PR #6, opened 2018-03-16
Fix transport authentication errors on redirect

jonjohnsonjr:
Add a name.Registry as the target registry for a bearerTransport. The
bearerTransport will only attach bearer tokens to requests matching the
registry's domain.

This fixes an issue where the registry would redirect to object storage,
which would reject the request due to an invalid Authorization header.

### DIFF
--- a/v1/remote/transport/bearer.go
+++ b/v1/remote/transport/bearer.go
@@ -16,19 +16,26 @@ package transport
 
 import (
 	"encoding/json"
-	"fmt"
 	"io/ioutil"
 	"net/http"
 	"net/url"
 
 	"github.com/google/go-containerregistry/authn"
+	"github.com/google/go-containerregistry/name"
 )
 
 type bearerTransport struct {
-	inner   http.RoundTripper
-	basic   authn.Authenticator
-	bearer  *authn.Bearer
-	realm   string
+	// Wrapped by bearerTransport.
+	inner http.RoundTripper
+	// Basic credentials that we exchange for bearer tokens.
+	basic authn.Authenticator
+	// Holds the bearer response from the token service.
+	bearer *authn.Bearer
+	// Registry to which we send bearer tokens.
+	registry name.Registry
+	// See https://tools.ietf.org/html/rfc6750#section-3
+	realm string
+	// See https://docs.docker.com/registry/spec/auth/token/
 	service string
 	scope   string
 }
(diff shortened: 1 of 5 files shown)


v1/remote/transport/bearer.go
```diff
@@ -22,15 +22,17 @@ import (
...
-	realm   string
-	service string
-	scope   string
+	inner    http.RoundTripper
+	basic    authn.Authenticator
+	bearer   *authn.Bearer
+	registry name.Registry
+	realm    string
+	service  string
+	scope    string
```
mattmoor: Looking at this list, it's long.  Can we add comments to each of these things describing their purpose?

v1/remote/transport/bearer.go
```diff
@@ -41,7 +43,11 @@ func (bt *bearerTransport) RoundTrip(in *http.Request) (*http.Response, error) {
 	if err != nil {
 		return nil, err
 	}
-	in.Header.Set("Authorization", hdr)
+
+	// Avoid setting Authorization header on the wrong domain (e.g. on redirect).
```
mattmoor: I would lead with something like: `http.Client` handles redirects at a layer above the `http.RoundTripper` abstraction, so to avoid forwarding Authorization headers to places we are redirected, only set it when the authorization header matches the registry with which we are interacting.

v1/remote/transport/bearer.go
```diff
@@ -41,7 +43,11 @@ func (bt *bearerTransport) RoundTrip(in *http.Request) (*http.Response, error) {
...
+	if in.Host == bt.registry.RegistryStr() {
+		in.Header.Set("Authorization", hdr)
```
mattmoor: Can you add a test?

mattmoor: Add this for Basic as well please.

jonjohnsonjr:
Should be RFAL.

v1/remote/transport/transport.go
```diff
@@ -53,7 +54,11 @@ func New(ref name.Reference, auth authn.Authenticator, t http.RoundTripper, a Sc
 	case anonymous:
 		return t, nil
 	case basic:
-		return &basicTransport{inner: t, auth: auth}, nil
+		u, err := url.Parse(ref.Context().Registry.String())
+		if err != nil {
+			return nil, err
+		}
+		return &basicTransport{inner: t, auth: auth, target: u.Host}, nil
```
mattmoor: Why isn't this just `target: ref.Context().RegistryStr()`?

jonjohnsonjr: Mea culpa

PR #7, opened 2018-03-16
Initial tarball.Image commit.

dlorenc:
This implements the interface, with panic'ing methods.
Manifest() is actually implemented as a start.


v1/tarball/image.go
```diff
@@ -15,17 +15,116 @@
...
+		if err != nil {
+			return nil, err
+		}
+		var buf bytes.Buffer
+		if hdr.Name == "manifest.json" {
+			_, err := io.Copy(&buf, tf)
+			if err != nil {
+				return nil, err
+			}
+			return v1.ParseManifest(buf.Bytes())
```
mattmoor: Sadly, this is not the manifest you are looking for :(

v1/tarball/image_test.go
```diff
@@ -0,0 +1,111 @@
...
+		}
+		if err := w.WriteHeader(&hdr); err != nil {
+			t.Fatalf("Error writing header to test tarfile: %v", err)
+		}
+		if _, err := w.Write(contents); err != nil {
+			t.Fatalf("Error writing contents to test tarfile: %v", err)
+		}
+	}
+	return &buf
+}
```
mattmoor: I'd recommend using `container_image` to produce all of your test data.  It builds *exactly* what you are looking for :)

dlorenc: bazel's container_image?

v1/tarball/image.go
dlorenc: Ah whoops, I was hoping I could reuse the type for `Config` and `Layers`, but it looks like those `Descriptors` are too strongly-typed for that :)

dlorenc:
OK, this should actually be working now.

I started to use container_image to generate the tarballs automatically, but that gets gazelle a bit angry. I generated them and just coped the results into testdata, which gazelle seems to automatically support: https://github.com/bazelbuild/rules_go#how-do-i-access-testdata

I figure this isn't that much worse, since either way we'd check in tarballs. This just checks in the images instead of the layers.

v1/tarball/image.go
```diff
@@ -15,17 +15,203 @@
...
+	mfstBytes, err := extractFileFromTar(i.path, "manifest.json")
+	if err != nil {
+		return err
+	}
+
+	tarMfst := tarManifest{}
+	if err := json.Unmarshal(mfstBytes, &tarMfst); err != nil {
+		return err
+	}
+	if len(tarMfst) != 1 {
```
mattmoor: I believe the python client allows passing the tag of the image to load.  Fine to TODO but I think we will want this for working with bundles.

v1/tarball/image.go
```diff
@@ -15,17 +15,203 @@
...
+		return errors.New("tarball must contain only a single image to be used with tarball.Image")
+	}
+
+	cfgBytes, err := extractFileFromTar(i.path, tarMfst[0].Config)
+	if err != nil {
+		return err
+	}
+
+	cfg, err := v1.ParseConfigFile(cfgBytes)
```
mattmoor: If I changed these interfaces to take an io.Reader, how problematic would it be? :)

I guess there is always bytes.NewBuffer.

v1/tarball/image.go
mattmoor: Hmm thinking more on this, I wonder if the best way to handle this would be:
1. Creat a ~Bundle type that loads the manifest and lazily accesses images.
1. Image loads that and then either extracts the Singleton or a specifically named image.

WDYT?

I see bundles (tarball) and lists (remote) as weakly supported by the python library (they were after thoughts), and I think this gets us the strong tarball support, while starting to improve the bundle story.

v1/tarball/image.go
```diff
@@ -15,17 +15,203 @@
...
+func (i *image) BlobSet() (map[v1.Hash]struct{}, error) {
+	panic("not implemented")
+}
+
+func (i *image) Digest() (v1.Hash, error) {
+	panic("not implemented")
+}
+
+func (i *image) MediaType() (types.MediaType, error) {
+	panic("not implemented")
```
mattmoor: This is just `i.manifest.MediaType`, right?

v1/tarball/image.go
```diff
@@ -15,17 +15,203 @@
...
+		},
+	}
+
+	for _, l := range tarMfst[0].Layers {
+		// TODO(dlorenc): support compressed layers.
+		uncompressed, err := extractFileFromTar(i.path, l)
+		if err != nil {
+			return err
+		}
+		layer, err := compress(uncompressed)
```
mattmoor: We need a library that holds compress/decompress utilities.

v1/tarball/image.go
```diff
@@ -15,17 +15,203 @@
...
+			MediaType: types.DockerLayer,
+			Size:      int64(len(layer)),
+			Digest:    v1.SHA256(string(layer)),
+		})
+	}
+	i.manifest = &manifest
+	return nil
+}
+
+func compress(l []byte) ([]byte, error) {
```
mattmoor: I think we want to switch to io.Reader everywhere to reduce our in-memory requirements.  This is a significant limitation of the python client (born of non-streaming http interfaces).

v1/tarball/image.go
```diff
@@ -15,17 +15,203 @@
...
+	Config   string
+	RepoTags []string
+	Layers   []string
+}
+
+// tarManifest is the struct used to represent a single image inside a `docker save` tarball.
+type tarManifest []singleManifest
+
+func (i *image) loadManifest() error {
+	mfstBytes, err := extractFileFromTar(i.path, "manifest.json")
```
mattmoor: We need to worry about the thread safety of these types.

v1/tarball/image.go
```diff
@@ -15,17 +15,203 @@
...
+	"github.com/google/go-containerregistry/v1/types"
 
 	"github.com/google/go-containerregistry/v1"
 )
 
-// TODO(mattmoor): This.
-// // image accesses an image from a remote registry
-// type image struct{}
-// var _ v1.Image = (*image)(nil)
+type tarBlob struct {
```
mattmoor: What is this used for?

v1/tarball/image_test.go
```diff
@@ -0,0 +1,64 @@
...
+// limitations under the License.
+
+package tarball
+
+import (
+	"path/filepath"
+	"testing"
+)
+
+var testdata = "testdata"
```
mattmoor: os.Getenv("TEST_SRCDIR") is the base you want, I think.

v1/tarball/image.go
```diff
@@ -15,17 +15,203 @@
...
+	return nil, fmt.Errorf("file %s not found in tar", filePath)
+}
+
+func (i *image) ConfigFile() (*v1.ConfigFile, error) {
+	if i.config == nil {
+		if err := i.loadManifest(); err != nil {
+			return nil, err
+		}
+	}
+	return i.config, nil
```
mattmoor: Accessing this technically shouldn't require loading the manifest IIRC.  Be lazier @dlorenc!!! ;-)

v1/tarball/testdata/README.md
```diff
@@ -0,0 +1,48 @@
+# Test Data
+
+The files in this directory are test data for tarball construction.
+
```
mattmoor: Shouldn't be bad to get these generated.  I hate checking in binaries :p

mattmoor: I can live with checking them in for things like malformed data or to exhibit behaviors that appear in the wild but rules_docker can't produce.  That's generally the line I've lived by. :)

v1/tarball/image.go
dlorenc: Why not? You need to parse the manifest to figure out which blob is the config, right?

dlorenc: I guess we can just guess that it's the file that ends with ".json" in the single manifest case?

mattmoor: Ugh, sorry.  My turn to be confused by the naming of this file.

v1/tarball/image.go
```diff
@@ -104,11 +141,13 @@ func (i *image) loadManifest() error {
...
-	if len(tarMfst) != 1 {
-		return errors.New("tarball must contain only a single image to be used with tarball.Image")
+
+	mfst, err := findSpecifiedManifest(i.tag, tarMfst)
+	if err != nil {
+		return err
 	}
 
-	cfgBytes, err := extractFileFromTar(i.path, tarMfst[0].Config)
+	cfgBytes, err := extractFileFromTar(i.path, mfst.Config)
```
mattmoor: Ok, so here's the nuance of the laziness I was talking about :)
1. We need to read `manifest.json` to access *everything* (might as well be eager), however,
1. We do NOT need `i.manifest` computed eagerly (this requires compressing all of the layers, which is expensive).

v1/tarball/image_test.go
```diff
@@ -53,12 +52,36 @@ func TestNoManifest(t *testing.T) {
...
 
-func TestBundle(t *testing.T) {
-	img, err := Image(filepath.Join(testdata, "no_manifest.tar"))
+func TestBundleSingle(t *testing.T) {
+	img, err := Image("test_bundle.tar", nil)
 	if err != nil {
 		t.Fatalf("Error loading image: %v", err)
 	}
 	if _, err := img.Manifest(); err == nil {
 		t.Fatalf("Error expected loading manifest.")
```
mattmoor: nit: `Fatalf` without format string.  Include the result you got?

v1/tarball/image.go
```diff
@@ -94,7 +105,33 @@ type singleManifest struct {
 // tarManifest is the struct used to represent a single image inside a `docker save` tarball.
 type tarManifest []singleManifest
 
+func findSpecifiedManifest(tag *name.Tag, tarMfst tarManifest) (*singleManifest, error) {
```
mattmoor: Seems like it would read best with `tarMfst` (which I keeping reading as "Tar MSFT", so we should rename :D) as the receiver:
```golang
func (tm tarManifest) manifestForTag(tag *name.Tag) ...
```

WDYT?

v1/tarball/image.go
```diff
@@ -94,7 +105,33 @@ type singleManifest struct {
...
+	if tag == nil {
+		if len(tarMfst) != 1 {
+			return nil, errors.New("tarball must contain only a single image to be used with tarball.Image")
```
mattmoor: Or a name must be provided?

dlorenc:
A bunch of refactoring for more laziness and renaming so @mattmoor stops setting mfst everywhere :)

v1/tarball/image_test.go
```diff
@@ -58,7 +58,7 @@ func TestBundleSingle(t *testing.T) {
 		t.Fatalf("Error loading image: %v", err)
 	}
 	if _, err := img.Manifest(); err == nil {
-		t.Fatalf("Error expected loading manifest.")
+		t.Fatalf("Error expected loading manifest: %v", err)
```
mattmoor: LOL, I meant to print the manifest.  This will always print `nil` :)

v1/tarball/image.go
```diff
@@ -95,69 +100,83 @@ func (i *image) Manifest() (*v1.Manifest, error) {
...
 	}
 	return nil, fmt.Errorf("tag %s not found in tarball", tag)
 }
 
-func (i *image) loadManifest() error {
-	i.lock.Lock()
-	defer i.lock.Unlock()
-
-	mfstBytes, err := extractFileFromTar(i.path, "manifest.json")
+func (i *image) loadTarDescriptorAndConfig() error {
```
mattmoor: Why not do this eagerly and avoid the need for locking?

dlorenc: Do you mean load these during the construction of the "Image" type in func Image()...?

dlorenc:
Sorry, jumped the gun a bit there, had to do a bit of refactoring/rebasing to fix it up for the readers.

v1/tarball/image_test.go
dlorenc: D'oh. Shouldn't code before coffee...

v1/tarball/image.go
```diff
@@ -177,44 +177,62 @@ func (i *image) loadManifestAndBlobs() error {
...
 			Size:      int64(len(cfgBytes)),
-			Digest:    v1.SHA256(string(cfgBytes)),
+			Digest:    sha,
 		},
 	}
 
 	for _, l := range i.imgDescriptor.Layers {
 		// TODO(dlorenc): support compressed layers.
 		uncompressed, err := extractFileFromTar(i.path, l)
+		defer uncompressed.Close()
```
mattmoor: This is function scope, not loop scope.  The workaround is to wrap this in:
```golang
func(){
...
}()
```

v1/tarball/image.go
```diff
@@ -177,44 +177,62 @@ func (i *image) loadManifestAndBlobs() error {
...
 			return err
 		}
-		r, err := compress.Compress(bytes.NewReader(uncompressed))
+		r, err := compress.Compress(uncompressed)
 		if err != nil {
 			return err
 		}
+
+		// TODO: figure out how to get the length of the compressed layer and compress it
+		// in one pass without reading it into memory.
```
mattmoor: Yeah, I have this same problem in stuff I'm doing :(

v1/tarball/image.go
mattmoor: Yeah.

dlorenc:
OK, this is RFAL. I can wait until your zip utils get merged, or fix it up after. Up to you.

v1/tarball/image.go
```diff
@@ -15,17 +15,257 @@
...
+	sha, _, err := v1.SHA256(ioutil.NopCloser(bytes.NewReader(cfgBytes)))
+	if err != nil {
+		return err
+	}
+	manifest := v1.Manifest{
+		SchemaVersion: 2,
+		MediaType:     types.DockerManifestSchema2,
+		Config: v1.Descriptor{
+			MediaType: types.DockerConfigJSON,
+			Size:      int64(len(cfgBytes)),
```
mattmoor: This can use what you have as `_` above, right?

dlorenc:
Going to merge this now, the GzipWrite/ReadClosers don't seem like that large of a win yet.

PR #8, opened 2018-03-16
Add error to the return of MediaType and Digest.

dlorenc:


dekkagaijin:
Could we just add an implicit `error` as the last return value? And only have to explicitly handle it if we wanted to?

Think this might revolutionize PL design.

dlorenc:
And an implicit first parameter of context?

dekkagaijin:
No, more exceptional than that.

PR #9, opened 2018-03-17
Switch parsing utilities to favor io.ReadCloser over []byte.

mattmoor:
This is in anticipation of using streaming interfaces to access the contents of image parts to avoid loading them into memory.

PR #10, opened 2018-03-17
Create a package of v1 utility functions.

mattmoor:
Similar idea to `"io/ioutil"`.  This holds assorted utilities and helpers for the `"v1"` package.

The main thing added in this change is `{Read,Write}AndCloser`, which turns an `io.Reader` and a `Close()` method into an `io.ReadCloser`.  This is useful in conjunction with methods of type: `func(io.Reader) io.Reader` (or the `io.Writer` analog) where the input is actually an `io.ReadCloser` (or `io.WriteCloser`).

An example of this is `gzip.NewReader`, which returns an `io.Reader`, but we want a variant that closes the input stream on `io.EOF`.

Base: https://github.com/google/go-containerregistry/pull/9


v1/manifest.go
```diff
@@ -38,10 +39,12 @@ type Descriptor struct {
 	Annotations map[string]string `json:"annotations,omitempty"`
 }
 
-// ParseManifest parses the provided string into a Manifest.
-func ParseManifest(data []byte) (*Manifest, error) {
+// ParseManifest parses the io.ReadCloser's contents into a Manifest.
+func ParseManifest(r io.ReadCloser) (*Manifest, error) {
+	defer r.Close()
+	decoder := json.NewDecoder(r)
```
ImJasonH: Normally this is just unlined: `json.NewDecoder(r).Decode(&m)`

v1/v1util/also.go
```diff
@@ -0,0 +1,57 @@
...
+
+import (
+	"io"
+)
+
+// ReadAndCloser implements io.ReadCloser by reading from a particular io.Reader
+// and then calling the provided "Close()" method.
+type ReadAndCloser struct {
+	Reader io.Reader
+	Closer func() error
```
ImJasonH: `CloseFunc`

v1/v1util/also.go
```diff
@@ -0,0 +1,57 @@
...
+
+// Close implements io.ReadCloser
+func (rac *ReadAndCloser) Close() error {
+	return rac.Closer()
+}
+
+// WriteAndCloser implements io.WriteCloser by reading from a particular io.Writer
+// and then calling the provided "Close()" method.
+type WriteAndCloser struct {
+	Writer io.Writer
```
ImJasonH: You can embed the Writer and avoid having to write the Write method yourself. Same with Reader

v1/v1util/also.go
```diff
@@ -0,0 +1,57 @@
...
+
+package v1util
+
+import (
+	"io"
+)
+
+// ReadAndCloser implements io.ReadCloser by reading from a particular io.Reader
+// and then calling the provided "Close()" method.
+type ReadAndCloser struct {
```
ImJasonH: This might feel better as a method that takes a Reader and close func, and returns a ReadCloser. That avoids having to export the kind of ugly type.

v1/v1util/also_test.go
```diff
@@ -0,0 +1,85 @@
+// Copyright 2018 Google LLC All Rights Reserved.
```
ImJasonH: Also_test?

v1/manifest.go
mattmoor: Done in the baseline change.

mattmoor:
I've been messing around with this change a bit locally, so ignore it for now.

v1/v1util/also.go
mattmoor: Good call.

v1/v1util/also.go
mattmoor: Done.

v1/v1util/also_test.go
mattmoor: I was going to call this `AlsoCloser` (after `NopCloser`) at first, but things got awkward when I needed both read and write.

v1/v1util/also.go
mattmoor: I'm going to make these private to `v1util` for now, since the aim is to use them for things like zip utilities, not this lower-level thing.

mattmoor:
Alright, I'm happier with this version of things.  In particular, this provides methods for:
`v1util.{Gzip,Gunzip}{Read,Write}Closer`

There is still a TODO to figure out the best way to avoid buffering things into memory when using `Reader`s to implement `Writers` and vice versa.  If folks have pointers / suggestions, I'm all ears, but I wanted to get the interface in place at least.

v1/v1util/zip.go
```diff
@@ -0,0 +1,100 @@
...
+	"compress/gzip"
+	"io"
+)
+
+// GzipReadCloser reads uncompressed input data from the io.ReadCloser and
+// returns an io.ReadCloser from which compressed data may be read.
+func GzipReadCloser(r io.ReadCloser) (io.ReadCloser, error) {
+	defer r.Close()
+
+	// TODO(mattmoor): How to avoid buffering this whole thing into memory?
```
ImJasonH: Does a Pipe help? It might add more complexity than you need. https://medium.com/stupid-gopher-tricks/streaming-data-in-go-without-buffering-3285ddd2a1e5

v1/v1util/zip.go
```diff
@@ -0,0 +1,100 @@
...
+// gunzipWriteCloser implements io.WriteCloser
+// It is used to implement GunzipWriteClose.
+type gunzipWriteCloser struct {
+	*bytes.Buffer
+	writer io.WriteCloser
+}
+
+// Close implements io.WriteCloser
+func (gwc *gunzipWriteCloser) Close() error {
+	// TODO(mattmoor): How to avoid buffering this whole thing into memory?
```
ImJasonH: Why does all of this happen in Close?

v1/v1util/zip.go
mattmoor: I think it's ~exactly what I want, but I think I'll save that for a follow-up.

v1/v1util/zip.go
mattmoor: Simple and effective until we switch to `io.Pipe`.  This saved me from writing my own buffered form of that.

ImJasonH:
Fine with punting to use `io.Pipe`

PR #26, opened 2018-03-18
Add validation steps to travis

### LINKED ISSUE #23: Add verification for go build of this repo
So that this can be vendored without Bazel.

dekkagaijin:
* Verify that `go build` works
* Validate `gofmt`
* Make `bazel test` error on failure
* Add `buildifier` steps

Fixes #23 

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,10 +3,14 @@ sudo: required
 dist: trusty
 
 language:
-  - java
+  - go
 
-jdk:
-  - oraclejdk8  # Building Bazel requires JDK8.
+before_install:
+  # Bazel requires JDK8
+  - sudo apt-get install oracle-java8-installer
+  
+install:
+  - go get -u github.com/bazelbuild/buildifier/buildifier
 
 services:
   - docker
@@ -21,4 +25,14 @@ addons:
 
 script:
   - bazel clean && bazel build //...
-  - bazel clean && bazel test //...
+  - bazel clean && bazel test --test_output=errors //...
+  
+  # Check for issues with the format of our bazel config files.
+  - buildifier -mode=check $(find . -name BUILD.bazel -type f)
+  - buildifier -mode=check $(find . -name WORKSPACE -type f)
+
+  # Verify that all source files are correctly formatted.
+  - gofmt -d -e -l .
+
+  - go clean -i
+  - go build -a -v -race ./... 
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -22,3 +23,10 @@ addons:
 script:
   - bazel clean && bazel build //...
   - bazel clean && bazel test //...
+
+  # Verify that all source files are correctly formatted.
+  - gofmt -d -e -l .
+
+  - go clean -i
+  - go get -u -t ./...
+  - go build -a -v -race ./...
```
dlorenc: Given that we're using bazel for stuff, should this be a bazel build instead of a go build? I guess both should work for build, but it's unlikely "go test" is going to keep working by itself for long.

dekkagaijin: `go test` doesn't work right now. We want to ensure that `go build` works on its own so that this lib can be vendored into pure go projects

mattmoor: The point of `go build` working is so that projects can vendor this using traditional Go tooling.  Those also typically prune tests, so it's less critical that `go test` work, but that's the easiest way to get Go coverage, so it might be worth a little effort.

.travis.yml
```diff
@@ -22,3 +23,10 @@ addons:
 script:
   - bazel clean && bazel build //...
   - bazel clean && bazel test //...
+
+  # Verify that all source files are correctly formatted.
+  - gofmt -d -e -l .
+
+  - go clean -i
+  - go get -u -t ./...
```
mattmoor: You should never need to `go get`, we have everything vendored.

.travis.yml
```diff
@@ -22,3 +23,10 @@ addons:
 script:
   - bazel clean && bazel build //...
   - bazel clean && bazel test //...
+
+  # Verify that all source files are correctly formatted.
+  - gofmt -d -e -l .
```
mattmoor: If you are adding linting, do you wanna add `buildifier`?

https://github.com/bazelbuild/rules_docker/blob/master/.travis.yml#L23-L26

Can you also add `--test_output=error` to the `bazel test` above?

.travis.yml
dekkagaijin: Unit tests can achieve compatibility pretty easily, and tests which exercise the file system can do so with a little effort (e.g. by using `testdata` and by taking arguments in the test which take the result of a `gendep` to receive references to files when run under bazel).

dekkagaijin:
@mattmoor done

PR #28, opened 2018-03-19
Add flavor text, link to py version, to README.md

dekkagaijin:


README.md
```diff
@@ -6,4 +6,4 @@
 
 ## Introduction
 
-This is a golang library for working with container registries.
+This is a golang library for working with container registries. It's largely based on the [Python library of the same name](https://github.com/google/containerregistry), but uses GitHub as the source of truth and will be the focus of future development.
```
mattmoor: I feel like this implies we're deprecating the Python library, which isn't really true (that's a massive amount of work).

A Go client is simply more suitable for a variety of workloads than Python for a variety of reasons.

dekkagaijin: fair enough, done

PR #31, opened 2018-03-20
Implement the "partial" helper package.

mattmoor:
This also implements `v1/random` that I was using as a testbed, and hooks it into @dlorenc's recent `v1/tarball`.

### DIFF
--- a/v1/partial/image.go
+++ b/v1/partial/image.go
@@ -1,0 +1,32 @@
+// Copyright 2018 Google LLC All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package partial
+
+import (
+	"github.com/google/go-containerregistry/v1"
+	"github.com/google/go-containerregistry/v1/types"
+)
+
+// imageCore is the core set of properties without which we cannot build a v1.Image
+type imageCore interface {
+	// ConfigFile returns this image's config file.
+	ConfigFile() (*v1.ConfigFile, error)
+
+	// ConfigName returns the hash of the image's config file.
+	ConfigName() (v1.Hash, error)
+
+	// MediaType of this image's manifest.
+	MediaType() (types.MediaType, error)
+}
--- a/v1/partial/image.go
+++ b/v1/partial/image.go
@@ -24,9 +24,6 @@ type imageCore interface {
 	// ConfigFile returns this image's config file.
 	ConfigFile() (*v1.ConfigFile, error)
 
-	// ConfigName returns the hash of the image's config file.
-	ConfigName() (v1.Hash, error)
-
 	// MediaType of this image's manifest.
 	MediaType() (types.MediaType, error)
 }
--- a/v1/partial/uncompressed.go
+++ b/v1/partial/uncompressed.go
@@ -1,0 +1,152 @@
+// Copyright 2018 Google LLC All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package partial
+
+import (
+	"bytes"
+	"encoding/json"
+	"io"
+	"sync"
+
+	"github.com/google/go-containerregistry/v1"
+	"github.com/google/go-containerregistry/v1/types"
+	"github.com/google/go-containerregistry/v1/v1util"
+)
+
+// UncompressedImageCore represents the bare minimum interface a natively
+// uncompressed image must implement for us to produce a v1.Image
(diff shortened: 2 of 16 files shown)


v1/partial/uncompressed.go
```diff
@@ -0,0 +1,152 @@
...
+func UncompressedToImage(uic UncompressedImageCore) (v1.Image, error) {
+	return &uncompressedImageExtender{
+		UncompressedImageCore: uic,
+	}, nil
+}
+
+// uncompressedImageExtender implements v1.Image by extending UncompressedImageCore with the
+// appropriate methods computed from the minimal core.
+type uncompressedImageExtender struct {
+	UncompressedImageCore
```
dlorenc: If you move these properties up to the UncompressedImage struct itself, could we avoid the transformation functions and embed the UncompressedImage struct directly in the sub-image type?

Something like:

```golang
// tarImage
type Image struct {
// tar stuff here
UncompressedImage
}
```

v1/partial/with_configfile.go
```diff
@@ -0,0 +1,38 @@
...
+}
+
+// DiffIDs is a helper for implementing v1.Image
+func DiffIDs(i withConfigFile) ([]v1.Hash, error) {
+	cfg, err := i.ConfigFile()
+	if err != nil {
+		return nil, err
+	}
+	dids := make([]v1.Hash, len(cfg.RootFS.DiffIDs))
+	for i, did := range cfg.RootFS.DiffIDs {
```
dlorenc: So the order of this is actually reversed from the cfg.RootFS.DiffIDs field? We might want to make an explicit comment about this somewhere in the v1.Image interface.

Right now it says "matches FSLayers", so I guess that might be clear enough. I'm going to have to look this up everytime anyway...

v1/tarball/image.go
```diff
@@ -56,46 +54,20 @@ func Image(path string, tag *name.Tag) (v1.Image, error) {
...
-// Manifest returns the v1.Manifest for the specified image.
-// This method memoizes the result, avoiding repeated reads from the tarball.
-func (i *image) Manifest() (*v1.Manifest, error) {
-	if i.manifest == nil {
-		if err := i.loadManifestAndBlobs(); err != nil {
-			return nil, err
-		}
-	}
-	return i.manifest, nil
+func (i *image) MediaType() (types.MediaType, error) {
```
dlorenc: Do you need to implement this one here, or is it covered in the partial?

v1/partial/uncompressed.go
mattmoor: Discussed a bit over chat last night, and it isn't clear there's a better way to do this.  Resolving for now, but open to later refactorings if we find something cleaner.

v1/tarball/image.go
mattmoor: This is part of `imageCore` because it could be either `Docker` or `OCI`.  I think we're biasing towards `Docker`, but should probably be biasing towards `OCI` since we called this `v1`...

@jonjohnsonjr WDYT?

v1/partial/image.go
```diff
@@ -0,0 +1,32 @@
...
+	"github.com/google/go-containerregistry/v1/types"
+)
+
+// imageCore is the core set of properties without which we cannot build a v1.Image
+type imageCore interface {
+	// ConfigFile returns this image's config file.
+	ConfigFile() (*v1.ConfigFile, error)
+
+	// ConfigName returns the hash of the image's config file.
+	ConfigName() (v1.Hash, error)
```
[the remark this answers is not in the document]
mattmoor: It may belong in `CompressedImageCore` because if you BYO `Manifest`, it must know a fixed name for `ConfigFile`...  I'll just remove it for now.

v1/partial/with_configfile.go
mattmoor: I think the comment belongs on the `DiffIDs` declaration on `v1.Image` specifying the ordering (added).

FWIW, I based the implementation (somewhat blindly) on [this](https://github.com/google/containerregistry/blob/master/client/v2_2/docker_image_.py#L48-L51).

v1/partial/with_configfile.go
```diff
@@ -0,0 +1,56 @@
...
+)
+
+// withConfigFile defines the subset of v1.Image used by these helper methods
+// TODO(mattmoor): Consider renaming this to withCore?
+type withConfigFile interface {
+	imageCore
+}
+
+// DiffIDs is a helper for implementing v1.Image
+func DiffIDs(i withConfigFile) ([]v1.Hash, error) {
```
ImJasonH: It's weird to have an exported method that takes an unexported type. Should `WithConfigFile` be exported?

v1/partial/with_configfile.go
```diff
@@ -0,0 +1,56 @@
...
+	"bytes"
+	"encoding/json"
+
+	"github.com/google/go-containerregistry/v1"
+	"github.com/google/go-containerregistry/v1/v1util"
+)
+
+// withConfigFile defines the subset of v1.Image used by these helper methods
+// TODO(mattmoor): Consider renaming this to withCore?
+type withConfigFile interface {
```
ImJasonH: What does this interface provide, since it seems to only be wrapping/renaming the `imageCore` interface?

v1/random/image.go
```diff
@@ -0,0 +1,83 @@
...
+	"github.com/google/go-containerregistry/v1"
+	"github.com/google/go-containerregistry/v1/partial"
+	"github.com/google/go-containerregistry/v1/types"
+	"github.com/google/go-containerregistry/v1/v1util"
+)
+
+// Image returns a pseudo-randomly generated Image.
+func Image(byteSize, layers int64) (v1.Image, error) {
+	layerz := make(map[v1.Hash][]byte)
+	for i := int64(0); i < layers; i = i + 1 {
```
ImJasonH: Unlike "lesser" languages, Go supports `i++` :)

v1/partial/with_configfile.go
mattmoor: Yeah, I privatized them, but you are right.

v1/partial/with_configfile.go
mattmoor: Now `WithConfigFile` is public, but I'm also conflicted (see TODO).

v1/random/image.go
mattmoor: Done

PR #34, opened 2018-03-20
Add support for compressed layers.

dlorenc:
This is missing an actual test because I don't have a way to generate these yet. Any suggestions on how to do that?


compress/compress.go
```diff
@@ -29,3 +32,12 @@ func Compress(r io.Reader) (io.Reader, error) {
 	}
 	return &buf, nil
 }
+
+// IsCompressed detects whether the input stream is compressed.
+func IsCompressed(r io.Reader) (bool, error) {
```
ImJasonH: This consumes the first two bytes of the reader, so downstream readers can't get it. Is that okay?

ISTR some Go stdlib thing does something like this, but returns a new `io.Reader` with the buffered bytes prepended.

So instead of `c, err := IsCompressed(r)` you'd have to write `r, c, err = IsCompressed(r)` which is a pain in and of itself.

compress/compress_test.go
```diff
@@ -0,0 +1,47 @@
...
+	}
+
+	buf := bytes.Buffer{}
+	gz := gzip.NewWriter(&buf)
+	if _, err := gz.Write(data); err != nil {
+		t.Fatalf("Error zipping data: %v", err)
+	}
+	gz.Close()
+
+	compressed, err = IsCompressed(&buf)
```
ImJasonH: Add a test that calling `IsCompressed(&buf)` multiple times continues to return the same value.

compress/compress.go
dlorenc: Yeah, at first I had a version of this that returned a reader so we could continue to read the input after deciding if it was compressed or not.

But then I realized the way this code is structured now, we don't need to read the layer anymore after (until much later, in a different function).

Maybe I should make it take a ReadCloser, and close it in this function to be explicit about that?

ImJasonH: Yeah in that case being explicit about what modifications you make to the reader seems really important. I'm curious how that later far-away read works if the first two bytes of the data have been consumed?

compress/compress_test.go
dlorenc: Done.

compress/compress.go
dlorenc: I made it return an un-advanced reader.

In this case we're just throwing away the reader after determining if it's compressed, so we ignore the newly returned one.

The compression would have been performed in another function, after opening the layer again.

dlorenc:
This should be RFAL @mattmoor. I filled in the missing compressedImageExtender.

dlorenc:
There's a bit of a chicken-and egg problem where I need this to do tarball.Write, and I can't really test this without tarball.Write. So "the tests are coming in a follow-up PR"™ :)

compress/compress.go
```diff
@@ -20,6 +20,9 @@ import (
 	"io"
 )
 
+var gzipMagicHeader = []byte{'\x1f', '\x8b'}
+
+// Compress compresses the input stream.
```
mattmoor: I believe this is now unused (though I didn't delete it).  Can we move `IsCompressed` into `v1util` and delete this?

v1/partial/with_manifest.go
```diff
@@ -107,3 +107,26 @@ func BlobToDiffID(i WithManifest, h v1.Hash) (v1.Hash, error) {
...
+		return v1.Hash{}, err
+	}
+	diffIDs, err := DiffIDs(i)
+	if err != nil {
+		return v1.Hash{}, err
+	}
+	if len(blobs) != len(diffIDs) {
+		return v1.Hash{}, fmt.Errorf("mismatched fs layers (%d) and diff ids (%d)", len(blobs), len(diffIDs))
+	}
+	for i, diffId := range diffIDs {
```
jonjohnsonjr: I don't love that this shadows the WithManifest.

v1/tarball/image.go
```diff
@@ -150,15 +196,74 @@ func extractFileFromTar(tarPath string, filePath string) (io.ReadCloser, error)
...
 	for idx, diffID := range i.config.RootFS.DiffIDs {
 		if diffID == h {
 			return extractFileFromTar(i.path, i.imgDescriptor.Layers[idx])
 		}
 	}
 	return nil, fmt.Errorf("diff id %q not found", h)
 }
+
+func (c *compressedImage) Blob(h v1.Hash) (io.ReadCloser, error) {
+	return nil, nil
```
mattmoor: ?

v1/tarball/image.go
```diff
@@ -150,15 +196,74 @@ func extractFileFromTar(tarPath string, filePath string) (io.ReadCloser, error)
...
+		MediaType:     types.DockerManifestSchema2,
+		Config: v1.Descriptor{
+			MediaType: types.DockerConfigJSON,
+			Size:      cfgSize,
+			Digest:    cfgHash,
+		},
+	}
+
+	for _, p := range c.imgDescriptor.Layers {
+		l, err := extractFileFromTar(c.path, p)
```
mattmoor: check `err`

v1/tarball/image.go
```diff
@@ -35,24 +40,48 @@ type image struct {
...
+	// Peek at the first layer and see if it's compressed.
+	compressed, err := img.areLayersCompressed()
+	if err != nil {
+		return nil, err
+	}
+	if compressed {
+		c := compressedImage{
+			image: img,
+		}
+		return partial.CompressedToImage(&c)
```
mattmoor: Can we just `&compressedImage{...}` inline? (and below)

v1/partial/compressed.go
```diff
@@ -39,7 +40,58 @@ type CompressedImageCore interface {
 // Assert that Image is a superset of this partial interface.
 var _ CompressedImageCore = (v1.Image)(nil)
 
+// uncompressedImageExtender implements v1.Image by extending UncompressedImageCore with the
+// appropriate methods computed from the minimal core.
+type compressedImageExtender struct {
+	CompressedImageCore
+
+	lock     sync.Mutex
+	manifest *v1.Manifest
```
mattmoor: These belong in the `compressImage` implementation in `v1/tarball`.  I suspect most other implementations don't need this.

mattmoor: In fact, it turns out you have it :)

v1/partial/compressed.go
```diff
@@ -39,7 +40,58 @@ type CompressedImageCore interface {
...
+func (i *compressedImageExtender) UncompressedBlob(h v1.Hash) (io.ReadCloser, error) {
+	return UncompressedBlob(i, h)
+}
+
+func (i *compressedImageExtender) UncompressedLayer(h v1.Hash) (io.ReadCloser, error) {
+	h, err := DiffIDToBlob(i, h)
+	if err != nil {
+		return nil, err
+	}
+	rc, err := i.Blob(h)
```
mattmoor: check `err`.

v1/tarball/image.go
```diff
@@ -150,15 +196,74 @@ func extractFileFromTar(tarPath string, filePath string) (io.ReadCloser, error)
...
+		if err != nil {
+			return nil, err
+		}
+		c.manifest.Layers = append(c.manifest.Layers, v1.Descriptor{
+			MediaType: types.DockerLayer,
+			Size:      size,
+			Digest:    sha,
+		})
+	}
+	return nil, nil
```
mattmoor: `c.manifest` LOL Kudos @jonjohnsonjr

dlorenc: haha yeah I fixed this one in my other branch, which adds tar.Write and was depending on this one. Screwed it up somehow trying to detangle the branches...

jonjohnsonjr: Ooh that's a good one... TODO: implement `git detangle`

v1/v1util/zip.go
```diff
@@ -98,3 +100,12 @@ func GunzipWriteCloser(w io.WriteCloser) (io.WriteCloser, error) {
 		writer: w,
 	}, nil
 }
+
+// IsCompressed detects whether the input stream is compressed.
+func IsCompressed(r io.Reader) (bool, error) {
```
mattmoor: IsGzipped?

PR #36, opened 2018-03-21
Tarball write support

dlorenc:


v1/tarball/write.go
```diff
@@ -29,5 +32,89 @@ type WriteOptions struct {
 
 // Write saves the image as the given tag in a tarball at the given path.
 func Write(p string, tag name.Tag, img v1.Image, wo WriteOptions) error {
```
ImJasonH: nit: Should `Write` take a `*WriteOptions` instead, so callers can pass `nil` for "default options"?

v1/tarball/write.go
```diff
@@ -29,5 +32,89 @@ type WriteOptions struct {
...
+	for _, layer := range layers {
+		r, err := img.Blob(layer)
+		if err != nil {
+			return err
+		}
+		blob, err := ioutil.ReadAll(r)
+		if err != nil {
+			return err
+		}
+		if err := writeFile(tf, layer.String(), blob); err != nil {
```
ImJasonH: `writeFile` should take an `io.Reader` to avoid having to buffer the entire layer contents. For the JSON blobs it's not so bad, those can be wrapped in a `bytes.NewReader`, but layers can be pretty big.

dlorenc: I started with that, but we need to write the Header first which needs to know the size, so it has to get buffered once.

But I think that we can get the BlobSize first and also pass that in. It has to get buffered once during manifest creation, but you're right that we shouldn't have to do it a second time.

v1/tarball/write.go
dlorenc: Probably. @mattmoor did you have any reason for making this a struct instead of a pointer initially?

v1/tarball/write.go
```diff
@@ -28,6 +32,91 @@ type WriteOptions struct {
...
 
 // Write saves the image as the given tag in a tarball at the given path.
-func Write(p string, tag name.Tag, img v1.Image, wo WriteOptions) error {
-	return fmt.Errorf("NYI: tarball.Write(%v)", p)
+func Write(p string, tag name.Tag, img v1.Image, wo *WriteOptions) error {
+	// Write in the compressed format.
+	// This is a tarball, on-disk, with:
+	// One manifest.json file at the top level containing information about several images.
+	// One file for each layer, named after the layer's SHA.
+	// One file for the config blob, named after it's sha.
```
jonjohnsonjr: its SHA

v1/tarball/write_test.go
```diff
@@ -0,0 +1,81 @@
...
+	tag, err := name.NewTag("gcr.io/foo/bar:latest", name.StrictValidation)
+	if err != nil {
+		t.Fatalf("Error creating test tag.")
+	}
+	if err := Write(fp.Name(), tag, randImage, nil); err != nil {
+		t.Fatalf("Unexpected error writing tarball: %v", err)
+	}
+
+	// Make sure the image is valid and can be loaded.
+	// Load it both by nil and by it's name.
```
jonjohnsonjr: its

v1/tarball/write.go
```diff
@@ -28,6 +32,91 @@ type WriteOptions struct {
...
+	}
+
+	// Generate the tar descriptor and write it.
+	layerPaths := []string{}
+	layers, err := img.FSLayers()
+	if err != nil {
+		return err
+	}
+	for _, l := range layers {
+		layerPaths = append(layerPaths, l.String())
```
mattmoor: nit: can we do this as part of the loop below and write `manifest.json` last?

v1/tarball/write.go
```diff
@@ -28,6 +32,91 @@ type WriteOptions struct {
...
 
 // Write saves the image as the given tag in a tarball at the given path.
-func Write(p string, tag name.Tag, img v1.Image, wo WriteOptions) error {
-	return fmt.Errorf("NYI: tarball.Write(%v)", p)
+func Write(p string, tag name.Tag, img v1.Image, wo *WriteOptions) error {
+	// Write in the compressed format.
+	// This is a tarball, on-disk, with:
+	// One manifest.json file at the top level containing information about several images.
+	// One file for each layer, named after the layer's SHA.
+	// One file for the config blob, named after its sha.
```
mattmoor: SHA :)

Issue #30, opened 2018-03-19
Consider adding a utility to validate Images


dlorenc:
This could be used in testing or by derived tools to validate the constraints imposed by Docker on an image. It could cover things like:

- relationship between config file history and layers
- layer sizes in the manifest match actual layer sizes
- config digest in the manifest matches the actual config digest

etc.

dekkagaijin:
* Validate that the images produced by `tarball` work with `docker load`

mattmoor:
See: https://github.com/google/go-containerregistry/pull/88

We should validate that the digest we have for the config in `Manifest` matches `ConfigName`.

PR #38, opened 2018-03-21
Created: `string` -> `time.Time`

### LINKED ISSUE #35: Docker can't parse our created times
Not sure where exactly the issue is, but Docker can't parse an empty creation time. Trying to docker load a saved tarball gives:

```shell
parsing time """" as ""2006-01-02T15:04:05Z07:00"": cannot parse """ as "2006"
```

dekkagaijin:
Fix #35 

### DIFF
--- a/v1/config.go
+++ b/v1/config.go
@@ -17,6 +17,7 @@ package v1
 import (
 	"encoding/json"
 	"io"
+	"time"
 )
 
 // ConfigFile is the configuration file that holds the metadata describing
@@ -25,7 +26,7 @@ import (
 type ConfigFile struct {
 	Architecture    string    `json:"architecture"`
 	Container       string    `json:"container"`
-	Created         string    `json:"created"`
+	Created         time.Time `json:"created"`
 	DockerVersion   string    `json:"docker_version"`
 	History         []History `json:"history"`
 	OS              string    `json:"os"`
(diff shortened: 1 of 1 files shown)


dlorenc:
Nice. We should add this to the list of things to validate in #30

dekkagaijin:
@dlorenc If we want to start taking dependencies on Docker's types, sure. They just use json.(Un)Marshal, so the only really interesting thing is the serialization roundtrips.

dlorenc:
Ah, it looks like this is the parsing code they use: https://github.com/moby/moby/blob/3a633a712c8bbb863fe7e57ec132dd87a9c4eff7/api/types/time/timestamp.go

I guess whatever we write has to be formatted in one of these four formats: https://github.com/moby/moby/blob/3a633a712c8bbb863fe7e57ec132dd87a9c4eff7/api/types/time/timestamp.go#L13

Up to you if you think it's worth testing.

dekkagaijin:
@dlorenc I'd say that that'd be a fairly superfluous reason to start going down the rabbit hole of vendoring  Docker. Might be better served with an e2e test in Travis. Luckily, these types are relatively static.

dekkagaijin:
FYI those 4 are extra, it also handles the standard RFC 3339- compliant formats:
https://github.com/moby/moby/blob/3a633a712c8bbb863fe7e57ec132dd87a9c4eff7/api/types/time/timestamp.go#L31

dlorenc:
Yeah definitely agree about not wanting to vendor docker just for this... I think it might be useful in #30 or somewhere else though. There are a bunch of things that can make docker unhappy even though something looks like an image, so having a utility to validate something that implements an Image interface would be really useful.

PR #40, opened 2018-03-25
Add a couple simple push/pull clients.

mattmoor:
These will fail on NYIs for now, but we have enough of the interface present that we can write them.

### DIFF
--- a/cmd/puller/main.go
+++ b/cmd/puller/main.go
@@ -43,6 +43,7 @@ func main() {
 	if err != nil {
 		log.Fatalln(err)
 	}
+	log.Printf("Pulling %v", t)
 
 	auth, err := authn.DefaultKeychain.Resolve(t.Registry)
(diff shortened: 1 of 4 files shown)


cmd/puller/main.go
```diff
@@ -0,0 +1,60 @@
...
+func main() {
+	flag.Parse()
+	if *path == "" {
+		log.Fatalln("-tarball must be specified.")
+	}
+	if *tag == "" {
+		log.Fatalln("-tag must be specified.")
+	}
+
+	t, err := name.NewTag(*tag, name.WeakValidation)
```
dlorenc: Maybe add a quick log line or fmt.Println with the t value we're about to push to/pull from, to avoid surprises from WeakValidation?

mattmoor: Done

PR #42, opened 2018-03-25
Introduce Raw{ConfigFile,Manifest} methods.

mattmoor:
These are needed to avoid inconsistencies in reserializing our structured representation.

This reorients the v1/partial interfaces around these as the new core that needs to be implemented, which leads to some slight simplification.


mattmoor:
This builds on https://github.com/google/go-containerregistry/pull/41, and takes it a bit further.

Issue #46, opened 2018-03-27
Surface structured errors from remote package


jonjohnsonjr:
The registry API has a [structured error response](https://docs.docker.com/registry/spec/api/#errors). The remote.* packages should support returning a custom error type with this information.

dekkagaijin:
+1

Guess I'll copy-paste more internal code for I M P A C T :3

PR #51, opened 2018-03-30
Add support for structured error handling.

mattmoor:
This changes adds support for parsing structured 4XX errors from a Docker registry.

Sample output against GCR:
```
2018/03/30 10:12:14 DENIED: "Please enable or contact project owners to enable the Google Container Registry API in Cloud Console at https://console.cloud.google.com/apis/api/containerregistry.googleapis.com/overview?project=convoy-adaptor before performing this operation."
```

Fixes: https://github.com/google/go-containerregistry/issues/46

### DIFF
--- a/v1/remote/error.go
+++ b/v1/remote/error.go
@@ -1,0 +1,107 @@
+// Copyright 2018 Google LLC All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package remote
+
+import (
+	"encoding/json"
+	"fmt"
+	"io/ioutil"
+	"net/http"
+	"strings"
+)
+
+// Error implements error to support the following error specification:
+// https://github.com/docker/distribution/blob/master/docs/spec/api.md#errors
+type Error struct {
+	Errors []Diagnostic `json:"errors,omitempty"`
+}
+
+// Check that Error implements error
+var _ error = (*Error)(nil)
+
+// Error implements error
+func (e *Error) Error() string {
+	switch len(e.Errors) {
+	case 0:
+		return "<empty remote.Error response>"
+	case 1:
+		return e.Errors[0].String()
+	default:
+		var errors []string
+		for _, d := range e.Errors {
+			errors = append(errors, d.String())
+		}
+		return fmt.Sprintf("multiple errors returned: %s",
+			strings.Join(errors, ";"))
+	}
+}
+
+// Diagnostic represents a single error returned by a Docker registry interaction.
+type Diagnostic struct {
+	Code    ErrorCode   `json:"code"`
+	Message string      `json:"message,omitempty"`
+	Detail  interface{} `json:"detail,omitempty"`
+}
+
+// String stringifies the Diagnostic
+func (d Diagnostic) String() string {
+	return fmt.Sprintf("%s: %q", d.Code, d.Message)
+}
+
+// ErrorCode is an enumeration of supported error codes.
+type ErrorCode string
+
+// The set of error conditions a registry may return:
+// https://github.com/docker/distribution/blob/master/docs/spec/api.md#errors-2
+const (
+	BlobUnknownErrorCode         ErrorCode = "BLOB_UNKNOWN"
+	BlobUploadInvalidErrorCode   ErrorCode = "BLOB_UPLOAD_INVALID"
+	BlobUploadUnknownErrorCode   ErrorCode = "BLOB_UPLOAD_UNKNOWN"
+	DigestInvalidErrorCode       ErrorCode = "DIGEST_INVALID"
+	ManifestBlobUnknownErrorCode ErrorCode = "MANIFEST_BLOB_UNKNOWN"
+	ManifestInvalidErrorCode     ErrorCode = "MANIFEST_INVALID"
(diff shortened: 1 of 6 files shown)


v1/remote/error.go
```diff
@@ -0,0 +1,107 @@
...
+	// so check the StatusCode before reading the body.
+	// https://github.com/docker/distribution/blob/master/docs/spec/api.md#errors
+	if resp.StatusCode < http.StatusBadRequest {
+		return nil
+	}
+	if resp.StatusCode >= http.StatusInternalServerError {
+		return nil
+	}
+
+	b, err := ioutil.ReadAll(resp.Body)
```
ImJasonH: `json.NewDecoder(resp.Body).Decode(&structuredError)`

ImJasonH: And `defer resp.Body.Close()`

v1/remote/error_test.go
```diff
@@ -0,0 +1,104 @@
...
+		body: "",
+	}, {
+		code: http.StatusUnauthorized,
+		body: "Not JSON",
+	}}
+
+	for _, test := range tests {
+		resp := &http.Response{
+			StatusCode: test.code,
+			Body:       v1util.NopReadCloser(bytes.NewBufferString(test.body)),
```
ImJasonH: `strings.NewReader` is shorter.

I forget what the benefit of `v1util.NopReadCloser` is over `ioutil.NopCloser`

v1/remote/error.go
```diff
@@ -0,0 +1,107 @@
...
+	NameUnknownErrorCode         ErrorCode = "NAME_UNKNOWN"
+	SizeInvalidErrorCode         ErrorCode = "SIZE_INVALID"
+	TagInvalidErrorCode          ErrorCode = "TAG_INVALID"
+	UnauthorizedErrorCode        ErrorCode = "UNAUTHORIZED"
+	DeniedErrorCode              ErrorCode = "DENIED"
+	UnsupportedErrorCode         ErrorCode = "UNSUPPORTED"
+)
+
+func checkError(resp *http.Response) error {
+	// According to the documentation, a structured error type is only return for 4XX errors,
```
jonjohnsonjr: I think you could interpret your link differently. 4XX should always return a structured error, but it doesn't say 4XX only.

v1/remote/error.go
```diff
@@ -0,0 +1,107 @@
...
+	// so check the StatusCode before reading the body.
+	// https://github.com/docker/distribution/blob/master/docs/spec/api.md#errors
+	if resp.StatusCode < http.StatusBadRequest {
+		return nil
+	}
+	if resp.StatusCode >= http.StatusInternalServerError {
+		return nil
```
jonjohnsonjr: Is 5xx ever not a terminal error? Can we handle them here, too?

v1/remote/error.go
```diff
@@ -0,0 +1,107 @@
...
+	NameInvalidErrorCode         ErrorCode = "NAME_INVALID"
+	NameUnknownErrorCode         ErrorCode = "NAME_UNKNOWN"
+	SizeInvalidErrorCode         ErrorCode = "SIZE_INVALID"
+	TagInvalidErrorCode          ErrorCode = "TAG_INVALID"
+	UnauthorizedErrorCode        ErrorCode = "UNAUTHORIZED"
+	DeniedErrorCode              ErrorCode = "DENIED"
+	UnsupportedErrorCode         ErrorCode = "UNSUPPORTED"
+)
+
+func checkError(resp *http.Response) error {
```
jonjohnsonjr: WDYT about something like this:
https://github.com/google/containerregistry/blob/1b7d4a482b1c239995f4a851010d16b2eefa176d/client/v2_2/docker_http_.py#L366

v1/remote/image.go
```diff
@@ -90,6 +90,10 @@ func (r *remoteImage) RawManifest() ([]byte, error) {
 	}
 	defer resp.Body.Close()
 
+	if err := checkError(resp); err != nil {
+		return nil, err
+	}
+
 	if resp.StatusCode != http.StatusOK {
 		// TODO(jonjohnsonjr): Parse this into a structured error.
```
jonjohnsonjr: We should be able to drop this now

v1/remote/image.go
```diff
@@ -151,8 +155,12 @@ func (r *remoteImage) Blob(h v1.Hash) (io.ReadCloser, error) {
 		return nil, err
 	}
 
+	if err := checkError(resp); err != nil {
+		resp.Body.Close()
+		return nil, err
+	}
+
 	if resp.StatusCode != http.StatusOK {
```
jonjohnsonjr: This feels redundant now

mattmoor: This checks other status codes :)

v1/remote/image.go
mattmoor: `checkError` doesn't handle 5xx or unsupported 2xx

v1/remote/error_test.go
mattmoor: I bias towards `bytes.Buffer` because it is just so much more capable, but I can change if you feel strongly.

I only added it for naming consistency, where we also have `NopWriteCloser`.

v1/remote/error.go
mattmoor: Done.

v1/remote/error.go
mattmoor: Done.

v1/remote/error.go
mattmoor: Done.

v1/remote/error.go
mattmoor: The client is responsible for closing `resp.Body` because we don't always read it.

I would have switched to the decoder pattern, but I need the raw body for new error modes in the refactoring Jon suggested.

v1/remote/image.go
mattmoor: Done

v1/remote/image.go
mattmoor: Done

mattmoor:
@ImJasonH LMK if you had other thoughts.  I may merge optimistically and follow up if you have any more comments.

Issue #60, opened 2018-04-14
Use cmp.Diff in place of reflect.DeepEqual


mattmoor:
From: github.com/google/go-cmp/cmp

PR #61, opened 2018-04-14
Switch to cmp.Diff in place of reflect.DeepEqual

mattmoor:
Fixes: https://github.com/google/go-containerregistry/issues/60


v1/hash.go
```diff
@@ -27,18 +27,18 @@ import (
 
 // Hash is an unqualified digest of some content, e.g. sha256:deadbeef
 type Hash struct {
-	algorithm string
-	hex       string
+	A string
+	H string
```
dekkagaijin: Alg, Hex? I know that lots of golang people prefer... _condensed_ labels, but you're triggering my front-end development PTSD

Is this meant to encourage use of the methods while preventing cmp.Diff from complaining about unexported fields? iirc, there's an argument you can pass to it to prevent that as well.

mattmoor: My preference would be `Algorithm` and `Hex`, but those are taken by methods.

Yes, this is for `cmp.Diff`.  This seemed the lesser evil than passing a "don't panic" flag everywhere.

jonjohnsonjr: Do they need to be methods?

mattmoor: Done

PR #64, opened 2018-04-16
Docker's config uses the field 'credsStore' instead of 'credStore'

bsnchan:
Pushing & pulling from DockerHub should now work

Signed-off-by: Brenda Chan <brchan@pivotal.io>
Signed-off-by: Dave Protasowski <davep@pivotal.io>


bsnchan:
dave will re-open to avoid cla/google bot

PR #67, opened 2018-04-16
Add ParseReference

ImJasonH:
This parses a string as either a tag or digest.

### DIFF
--- a/name/ref.go
+++ b/name/ref.go
@@ -14,7 +14,10 @@
 
 package name
 
-import "fmt"
+import (
+	"errors"
+	"fmt"
+)
 
 // Reference defines the interface that consumers use when they can
 // take either a tag or a digest.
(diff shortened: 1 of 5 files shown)


name/ref_test.go
```diff
@@ -0,0 +1,75 @@
+package name
```
mattmoor: license header.

name/ref.go
```diff
@@ -33,3 +36,14 @@ type Reference interface {
 	// Scope is the scope needed to access this reference.
 	Scope(string) string
 }
+
+// ParseReference parses the string as a reference, either by tag or digest.
+func ParseReference(s string, strict Strictness) (Reference, error) {
```
mattmoor: Can you update the places in `//cmd` to use this?

PR #71, opened 2018-04-17
Add unified CLI surface, d8s

### LINKED ISSUE #70: Unified CLI surface
The `cmd` directory currently has four separate binaries a user can use to interact with a remote registry: `puller`, `pusher`, `poke` (which gets only metadata), and `deleter`. It looks like we'll be adding `appender` soon in #63.

I think we should head off the proliferation of separate binaries and move them into a single command with subcommands for each operation. This will allow users to install a single binary into their path, and could allow some code reuse and help standardize conventions for the commands.

This would also make it easier to package as a [builder image](https://github.com/GoogleCloudPlatform/cloud-builders) for GCB.

## Subcommands

* `get` (instead of `poke`)
* `pull`
* `push`
* `delete`
* `append`
* `tag (list|add|remove)`
* `label (list|add|remove)`
* [`flatten`](https://github.com/google/containerregistry#flattenpar)
* [`rebase`](https://github.com/google/image-rebase)?

## Conventions

Here are some potentially controversial CLI conventions for us to bikeshed over:

* Use the default keychain for all authorization.

* If the image produces a new image, print that new image's reference by digest to stdout.

* **Positional args:** If a command operates on an image by reference, it should be a positional arg. Required flags should be positional args as much as makes sense. E.g., `the-command pull ubuntu ubuntu.tar.gz` instead of `the-command pull --tarball=ubuntu.tar.gz --tag=ubuntu`. If the command takes more than a couple args, or if both args are image references, they should be named flags to avoid confusion.

* **No `er` suffix:** Subcommands should be `pull`, `push`, `delete`, `append`, etc., and not `the-command puller foo bar.tar.gz`.

* **Use [`google/subcommands`](https://github.com/google/subcommands):** I don't have a strong preference, but that one seems good.

## What to name it

As with any Serious Engineering Endeavor, naming is the only hard part. Here are some naming suggestions for us to bikeshed over:

* [`reg` is taken](https://github.com/genuinetools/reg).
* `reggie`, like Clippy for container image registries.
* `imgctl` (pronounced "image cuttle", obviously).
* [`img` is taken](https://github.com/genuinetools/img).
* [`RegEdit.exe` is taken](https://en.wikipedia.org/wiki/Windows_Registry).

WDTY?

ImJasonH:
d8s is for "Dockerless", pronounced "dates". This can be considered a placeholder name, naming suggestions welcome.

This PR also modifies the `tarball.Write` method to be able to create output files if they don't already exist.

Fixes #70


cmd/d8s/delete.go
```diff
@@ -0,0 +1,42 @@
+package main
```
mattmoor: license headers, here and elsewhere.

v1/tarball/write.go
```diff
@@ -39,7 +39,7 @@ func Write(p string, tag name.Tag, img v1.Image, wo *WriteOptions) error {
 	// One file for each layer, named after the layer's SHA.
 	// One file for the config blob, named after its SHA.
 
-	w, err := os.OpenFile(p, os.O_WRONLY, 0644)
+	w, err := os.Create(p)
```
mattmoor: @dprotaso I think you have a similar change?

cmd/d8s/BUILD.bazel
```diff
@@ -2,19 +2,27 @@ load("@io_bazel_rules_go//go:def.bzl", "go_binary", "go_library")
...
+        "//vendor/github.com/google/subcommands:go_default_library",
     ],
 )
 
 go_binary(
-    name = "pusher",
+    name = "d8s",
     embed = [":go_default_library"],
     visibility = ["//visibility:public"],
 )
```
mattmoor: It would also be trivial to add a:
```python
go_image(
    name = "builder",
    binary = ":d8s",
)
```

cmd/d8s/delete.go
```diff
@@ -0,0 +1,42 @@
...
+	"github.com/google/go-containerregistry/v1/remote"
+	"github.com/google/subcommands"
+)
+
+type deleteCmd struct{}
+
+func (*deleteCmd) Name() string             { return "delete" }
+func (*deleteCmd) Synopsis() string         { return "Deletes a reference from its registry" }
+func (*deleteCmd) Usage() string            { return "delete <image>" }
+func (*deleteCmd) SetFlags(f *flag.FlagSet) {}
```
mattmoor: Shouldn't we parse `ref` here?

I think we should adopt standard terminology for `<tag>` vs. `<digest>` vs. `<reference>`.

mattmoor: We should also think about whether we can share parsing logic for these.

cmd/d8s/delete.go
```diff
@@ -0,0 +1,42 @@
...
+	}
+	ref := f.Args()[0]
+
+	r, err := name.ParseReference(ref, name.WeakValidation)
+	if err != nil {
+		log.Fatalln(err)
+	}
+
+	auth, err := authn.DefaultKeychain.Resolve(r.Context().Registry)
+	if err != nil {
```
mattmoor: I could see us wanting a mechanism for overriding this default behavior, e.g. to support `imagePullSecrets` or force a particular cred helper.

v1/tarball/write.go
dprotaso: Yup

mattmoor:
@jonjohnsonjr How do you think we should incorporate manifest lists into this?

cmd/d8s/delete.go
ImJasonH: I think `SetFlags` is only intended as a helper to parse flags, not args. I admit that I don't know though.

I've updated the usages to specify `<reference>` or `<tag>`

cmd/d8s/delete.go
ImJasonH: That makes sense. I can add a command-wide flag to override the auth method, so it's applicable to all commands.

Does `authn` support `imagePullSecrets` today? Could `DefaultKeychain` fallback to check `imagePullSecrets`? (These are questions for a future change, either way)

cmd/d8s/pull.go
```diff
@@ -20,7 +34,7 @@ func (*pullCmd) Name() string { return "pull" }
 func (*pullCmd) Synopsis() string {
 	return "Pulls an image by reference and stores its contents in a tarball"
 }
-func (*pullCmd) Usage() string            { return "pull <image> <tarball>" }
+func (*pullCmd) Usage() string            { return "pull <reference> <tarball>" }
```
mattmoor: I think it's fine for now, but we might want to rethink two positionals.

cmd/d8s/BUILD.bazel
```diff
@@ -1,20 +1,34 @@
...
+        "//v1:go_default_library",
         "//v1/remote:go_default_library",
         "//v1/tarball:go_default_library",
+        "//vendor/github.com/google/subcommands:go_default_library",
     ],
 )
 
 go_binary(
-    name = "pusher",
+    name = "d8s",
```
mattmoor: Can you add `pure = "on",` to this rule?

cmd/d8s/delete.go
mattmoor: Not yet.  I want to add a mode to these to support K8s-compatible auth resolution.

jonjohnsonjr:
@mattmoor docker has a [`docker manifest` command](https://docs.docker.com/edge/engine/reference/commandline/manifest/) that we could use for inspiration.

I'd err on being as lazy as possible here. We'll need to convert manifest lists to manifests for `pull` and `config`, at least.

PR #81, opened 2018-04-20
Accept access_token in token response

### LINKED ISSUE #54: Allow "access_token" in token exchange
We currently assume the response will contain "token". 

Azure does not: https://github.com/bazelbuild/rules_k8s/issues/128.

[The spec](https://docs.docker.com/registry/spec/auth/token/#requesting-a-token) allows for either.

jonjohnsonjr:
Fixes #54

### DIFF
--- a/v1/remote/transport/bearer.go
+++ b/v1/remote/transport/bearer.go
@@ -15,6 +15,8 @@
 package transport
 
 import (
+	"fmt"
+
 	"encoding/json"
 	"io/ioutil"
 	"net/http"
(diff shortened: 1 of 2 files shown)


v1/remote/transport/bearer.go
```diff
@@ -90,12 +92,28 @@ func (bt *bearerTransport) refresh() error {
...
 	}
+
+	// Find a token to turn into a Bearer authenticator
+	var bearer authn.Bearer
+	if response.Token != "" {
+		bearer = authn.Bearer{Token: response.Token}
+	} else if response.AccessToken != "" {
+		bearer = authn.Bearer{Token: response.AccessToken}
+	} else {
+		return fmt.Errorf("no token in bearer response")
```
dekkagaijin: unnecessary `Errorf`

PR #84, opened 2018-04-20
Add a new constructor in the mutate package for modifying an image co…

dlorenc:
…nfiguration.

Also rename End -> Env.

### DIFF
--- a/v1/mutate/mutate.go
+++ b/v1/mutate/mutate.go
@@ -111,6 +111,29 @@ func Append(base v1.Image, adds ...Addendum) (v1.Image, error) {
 	return image, nil
 }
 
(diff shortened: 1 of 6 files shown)


v1/mutate/mutate.go
```diff
@@ -111,6 +111,29 @@ func Append(base v1.Image, adds ...Addendum) (v1.Image, error) {
 	return image, nil
 }
 
+func NewFromConfig(base v1.Image, cfg v1.Config) (v1.Image, error) {
```
mattmoor: I'd probably just call this `mutate.Config`, and this needs a doc string.

I may just merge this and do that myself, if you don't get to it because I want to play with this change.

Issue #87, opened 2018-04-21
We should disallow accessing the config through LayerBy*


mattmoor:
Especially now that we have `v1.Layer`, it is awkward to reason about `ConfigFile` as a layer.

This has been the source of multiple bugs, we should just eliminate this and require that `LayerBy*` be used exclusively to access something from `Layers()`.

The history here is that since the config is stored as a blob in the registry, we had incorporated it into concepts like `blob_set()` (but not `fs_layers()`!)...

jonjohnsonjr:
Should we just close this? https://github.com/google/go-containerregistry/pull/285 😆

PR #88, opened 2018-04-21
Fix an issue accessing the Config when we mutate it.

mattmoor:
We were changing the `ConfigFile.Config`, but we weren't updating the `Manifest`'s digest for the `ConfigFile`.  This meant that things keying off of the `Manifest`'s `ConfigFile` digest field to access things through `LayerByDigest` would end up being delegated to the base image's implementation.  In my case, this was manifesting as us asking GCR for a non-existent blob (`remote.Image`).

Related: https://github.com/google/go-containerregistry/issues/87

### DIFF
--- a/v1/mutate/mutate.go
+++ b/v1/mutate/mutate.go
@@ -132,6 +132,10 @@ func Config(base v1.Image, cfg v1.Config) (v1.Image, error) {
 		diffIDMap:  make(map[v1.Hash]v1.Layer),
 		digestMap:  make(map[v1.Hash]v1.Layer),
 	}
+	image.manifest.Config.Digest, err = image.ConfigName()
+	if err != nil {
+		return nil, err
+	}
 	return image, nil
 }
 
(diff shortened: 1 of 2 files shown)


v1/mutate/mutate.go
```diff
@@ -132,6 +132,10 @@ func Config(base v1.Image, cfg v1.Config) (v1.Image, error) {
 		diffIDMap:  make(map[v1.Hash]v1.Layer),
 		digestMap:  make(map[v1.Hash]v1.Layer),
 	}
+	image.manifest.Config.Digest, err = image.ConfigName()
```
jonjohnsonjr: I haven't read through all of this file, but we do this above in `Append`, do we need to do it here too?

dlorenc:
if you rebase you'll pick up my patch that ignores vendor/ in gofmt.

v1/mutate/mutate.go
mattmoor: This doesn't call append :)

mattmoor:
I fixed the test

v1/mutate/mutate_test.go
```diff
@@ -117,8 +117,8 @@ func TestMutateConfig(t *testing.T) {
 		t.Fatalf("failed to mutate a config: %v", err)
 	}
 
-	if !manifestsAreEqual(t, source, result) {
-		t.Fatal("mutating the config mutated the manifest")
+	if manifestsAreEqual(t, source, result) {
+		t.Fatal("mutating the config MUST the manifest")
```
jonjohnsonjr: MUST mutate

Issue #91, opened 2018-04-22
name.NewTag bug


mattmoor:
`name.NewTag` has a `WeakValidation` bug.  When the host part contains a port (e.g. `foo.io:8383`), and the `:latest` tag is omitted, we incorrectly treat the port and repository portion as the tag, and then reject it due to illegal characters.

dekkagaijin:
lick

PR #93, opened 2018-04-23
fix bug when creating Tags with weak validation and  registries

dekkagaijin:
Fixes https://github.com/google/go-containerregistry/issues/91

### DIFF
--- a/name/tag.go
+++ b/name/tag.go
@@ -77,7 +77,7 @@ func NewTag(name string, strict Strictness) (Tag, error) {
 
 	// Split on ":"
 	parts := strings.Split(name, tagDelim)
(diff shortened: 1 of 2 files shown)


name/tag.go
```diff
@@ -76,7 +76,9 @@ func NewTag(name string, strict Strictness) (Tag, error) {
 	tag := ""
 
 	// Split on ":"
-	if parts := strings.Split(name, tagDelim); len(parts) >= 2 {
+	parts := strings.Split(name, tagDelim)
+	// Verify that we aren't confusing a tag for a hostname w/ port for the purposes of weak validaiton.
```
jonjohnsonjr: nit: validation

dekkagaijin: done

PR #94, opened 2018-04-23
panic instead of fatal, so that defers get executed (just in case)

dekkagaijin:
Probably not a huge biggie, but use Panic instead of Fatal, in case a `defer` closes resources that otherwise might leak in certain circumstances.

### DIFF
--- a/cmd/d8s/append.go
+++ b/cmd/d8s/append.go
@@ -59,33 +59,33 @@ func (a *appendCmd) Execute(
 
 	srcRef, err := name.ParseReference(src, name.WeakValidation)
 	if err != nil {
-		log.Fatalln(err)
+		log.Panicln(err)
 	}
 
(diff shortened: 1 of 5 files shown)


jonjohnsonjr:
I don't feel super strongly about this, but printing a stack trace on error seems a bit gross.

/shrug

mattmoor:
I feel strongly that we shouldn't print a massive stack trace on errors that should terminate the process.  This is useful to us, but not to anyone using the tool.  This was a pain point in the other library's tools for years, let's not go back there.

That said, we shouldn't be using `Fatal` inside of libraries of inner scopes where process termination might be highly undesirable anyways (e.g. this may be linked into a service).  I think we've kept this to the `//cmd` layer.  So it would be greatly preferable to refector things that run into this to return an error from the scope needing `defer` and `Fatal` at some final level.

Issue #80, opened 2018-04-20
Experiment: //cmd/ko


mattmoor:
**tl;dr** I want a more purpose built variation of what you get with `bazel` + `rules_go` + `rules_docker` + `rules_k8s` that simply wraps the Go toolchain directly.

### Background: State of Bazel

Today in Bazel, my build is described in `BUILD` files (largely generated by Gazelle) and extended to handle containerization and kubernetesification via:
```python
# BEGIN generated by Gazelle
go_library(
    name = "go_default_library",
    ...
)

go_test(
    name = "go_default_test",
    ...
)

go_binary(
    name = "name-of-directory",
    ...
)
# END generated by Gazelle

# Wrap the Go binary into a minimal container image.
go_image(
    name = "image",
    binary = ":name-of-directory",
)

# Helper for interacting with my K8s "Deployment"
k8s_object(
    name = "deployment",
    template = "deployment.yaml",
    images = {
        # Associate the image reference in deployment.yaml with the associated binary target.
        "gcr.io/foo/bar:baz": ":image",
    },
)
```

With this, I can quickly iterate and redeploy via `bazel run :deployment.apply`.

Unfortunately, it's painful to deal with Go in Bazel because it forks the source of truth for dependencies: the import paths in `*.go`.  This often leaves you running Gazelle before build commands, which breaks the promise of `rules_k8s`: a single command for rapid iteration.

You also end up checking in a lot of redundant configuration.

### Background: Go importpaths

As alluded to above, Go declares dependencies via a block like this at the top of each file:
```go
import (
    "github.com/google/go-containerregistry/authn"
    "github.com/google/go-containerregistry/name"
)
```

Binary targets are determined by the declaration of `package main` within a particular one of these paths.  For example, we have a binary that can be referenced via the Go importpath: `github.com/google/go-containerregistry/cmd/d8s`

#### Foreshadowing

Gee, Go importpaths look a *LOT* like Docker image references.

### Convention over Configuration

This idea builds around an extension of the Go importpath convention to tie Go binaries all the way through to one's Kubernetes configuration.  Instead of writing an image reference in your Kubernetes configuration, you would write a Go importpath:
```yaml
apiVersion: apps/v1beta1
kind: Deployment
metadata:
  name: foo
spec:
  replicas: 1
  template:
    spec:
      containers:
      - name: bar
        image: github.com/path/to/my/thing
```

[...]

PR #100, opened 2018-04-24
Implement ko/build

mattmoor:
This adds a package for handling builds for `ko`.  The `build.Interface` contains two methods:
1. `IsSupportedReference`: this is used to determine whether a particular string (in a K8s yaml) smells like a reference for which this `build.Interface` can produce a `v1.Image`.
1. `Build`: this is responsible for synthesizing a `v1.Image` for supported references.

At present, this contains two implementations:
1. `NewFixed`: this is an implementation intended to aid in testing packages that need a `build.Interface`.
1. `NewGo`: this is a real implementation that will perform a `go build` for import paths relative to the execution context and layer the resulting binary on top of a provided base image.

Progress towards: https://github.com/google/go-containerregistry/issues/80

### DIFF
--- a/ko/build/build.go
+++ b/ko/build/build.go
@@ -1,0 +1,31 @@
+// Copyright 2018 Google LLC All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package build
+
+import (
+	"github.com/google/go-containerregistry/v1"
+)
+
+// Interface abstracts different methods for turning a supported importpath
+// reference into a v1.Image.
+type Interface interface {
+	// IsSupportedReference determines whether the given reference is to an importpath reference
+	// that Ko supports building.
+	// TODO(mattmoor): Verify that some base repo: foo.io/bar can be suffixed with this reference and parsed.
+	IsSupportedReference(string) bool
+
+	// Build turns the given importpath reference into a v1.Image containing the Go binary.
+	Build(string) (v1.Image, error)
+}
--- a/ko/build/gobuild.go
+++ b/ko/build/gobuild.go
@@ -1,0 +1,182 @@
+// Copyright 2018 Google LLC All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package build
+
+import (
+	"archive/tar"
+	"bytes"
+	"fmt"
+	"io"
+	"io/ioutil"
+	"log"
+	"os"
+	"os/exec"
+	"path"
+	"strings"
+
+	"github.com/google/go-containerregistry/v1"
+	"github.com/google/go-containerregistry/v1/mutate"
+	"github.com/google/go-containerregistry/v1/tarball"
+	"github.com/google/go-containerregistry/v1/v1util"
+)
+
+var (
+	// getwd is a variable for testing.
+	getwd = os.Getwd
+)
+
+const (
+	appPath = "/app"
+)
+
+type Options struct {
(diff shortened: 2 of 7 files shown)


ko/build/build.go
```diff
@@ -0,0 +1,31 @@
...
+	"github.com/google/go-containerregistry/v1"
+)
+
+// Interface abstracts different methods for turning a supported importpath
+// reference into a v1.Image.
+type Interface interface {
+	// IsSupportedReference determines whether the given reference is to an importpath reference
+	// that Ko supports building.
+	// TODO(mattmoor): Verify that some base repo: foo.io/bar can be suffixed with this reference and parsed.
+	IsSupportedReference(string) bool
```
jonjohnsonjr: This is kinda long -- what about `CanBuild`?

ko/build/gobuild.go
```diff
@@ -0,0 +1,186 @@
...
+
+	// Start from a copy of the base image's config file, and set
+	// the entrypoint to our app.
+	cfg, err := withApp.ConfigFile()
+	if err != nil {
+		return nil, err
+	}
+	cfg = cfg.DeepCopy()
+	cfg.Config.Entrypoint = []string{appPath}
+	withEntrypoint, err := mutate.Config(withApp, cfg.Config)
```
jonjohnsonjr: you could just return this

ko/build/gobuild.go
```diff
@@ -0,0 +1,186 @@
...
+}
+
+// IsSupportedReference implements build.Interface
+func (gb *gobuild) IsSupportedReference(s string) bool {
+	// TODO(mattmoor): Consider supporting vendored things as well.
+	return strings.HasPrefix(s, gb.importpath)
+}
+
+func build(ip string) (string, error) {
+	file, err := ioutil.TempFile(os.TempDir(), "out")
```
jonjohnsonjr: Is there a way to just not write this to disk?

mattmoor: I don't want `go build` to stream be the output :P

`tarBinary` already avoids this for the tarball.

ko/build/build.go
mattmoor: Can is probably a superset of what this will opt to support.  e.g. go build only supports builds relative to the current thing on the gopath.

ko/build/gobuild.go
mattmoor: Done.

ko/build/gobuild.go
```diff
@@ -0,0 +1,182 @@
...
+}
+
+// Build implements build.Interface
+func (gb *gobuild) Build(s string) (v1.Image, error) {
+	// Do the build into a temporary file.
+	file, err := gb.build(s)
+	if err != nil {
+		return nil, err
+	}
+	defer os.Remove(file)
```
ImJasonH: Consider logging the error this might return.

ko/build/gobuild.go
```diff
@@ -0,0 +1,182 @@
...
+	// getwd is a variable for testing.
+	getwd = os.Getwd
+)
+
+const (
+	appPath = "/app"
+)
+
+type Options struct {
+	// TODO(mattmoor): Base Image?
```
ImJasonH: This is done now?

ko/build/gobuild.go
```diff
@@ -0,0 +1,182 @@
...
+	src := path.Join(os.Getenv("GOPATH"), "src")
+	if !strings.HasPrefix(wd, src) {
+		return "", fmt.Errorf("working directory %q must be on GOPATH %q", wd, src)
+	}
+	return strings.Trim(strings.TrimPrefix(wd, src), "/"), nil
+}
+
+// NewGo returns a build.Interface implementation that:
+//  1. builds go binaries named by importpath,
+//  2. containerizes the binary on a suitable base,
```
ImJasonH: "suitable" -> "specified"

The user might select a totally bonkersgonuts base image, which wouldn't be suitable at all.

ko/build/gobuild.go
```diff
@@ -0,0 +1,182 @@
...
+	// Create a layer from that tarball.
+	layer, err := tarball.LayerFromOpener(func() (io.ReadCloser, error) {
+		return v1util.NopReadCloser(bytes.NewBuffer(layerBytes)), nil
+	})
+	if err != nil {
+		return nil, err
+	}
+
+	// Augment the base image with our application layer.
+	withApp, err := mutate.AppendLayers(gb.opt.Base, layer)
```
ImJasonH: If a caller doesn't specify a valid `Base`, we won't find out until here, after we've built the image and assembled a layer from it. Can we sanity check earlier?

ko/build/gobuild_test.go
```diff
@@ -0,0 +1,226 @@
...
+			t.Fatalf("len(Layers()) = %v, want %v", got, want)
+		}
+	})
+
+	// While we have a randomized base image, the application layer should be completely deterministic.
+	// Check that when given fixed build outputs we get a fixed layer hash.
+	t.Run("check determinism", func(t *testing.T) {
+		expectedHash := v1.Hash{
+			Algorithm: "sha256",
+			Hex:       "f6c290fa444e8e94936132309a42764f8360fb88e38adb8b3077a21780128794",
```
ImJasonH: Where does this magic value come from? Can you derive it in the test itself? Or to make the point about determinism, build the thing twice and show that the digest matches, whatever it happens to be.

PR #102, opened 2018-04-24
Implement cmd/ko

mattmoor:
This uses the cobra library to implement the `ko` CLI, which wraps the `go` CLI and exposes a handful of commands for interacting with kubernetes configurations.  See `cmd/ko/README.md` for more details.

Fixes: #80

### DIFF
--- a/cmd/ko/README.md
+++ b/cmd/ko/README.md
@@ -52,31 +52,82 @@ spec:
         - containerPort: 8080
 ```
 
-**This extension of the Go convention enables us to create a tool with an
-extremely fast development cycle and effectively zero configuration.**
+### Determining supported import paths
+
+Similar to other tooling in the Go ecosystem, `ko` expects to execute in the
+context of your `$GOPATH`. This is used to determine what package(s) `ko`
+is expected to build.
+
+Suppose `GOPATH` is `~/gopath` and the current directory is
+`~/gopath/src/github.com/mattmoor/examples`. `ko` will deduce the base import
+path to be `github.com/mattmoor/examples`, and any references to subpackages
+of this will be built, containerized and published.
+
+For example, any of the following would be matched:
+* `github.com/mattmoor/examples`
+* `github.com/mattmoor/examples/cmd/foo`
+* `github.com/mattmoor/examples/bar`
+
+`ko` does not (currently) support `vendor`d binaries, or building other
+binaries from `GOPATH` (outside of the current scope).
+
+### Results
+
+Employing this convention enables `ko` to have effectively zero configuration
+and enable very fast development iteration. For
+[warm-image](https://github.com/mattmoor/warm-image), `ko` is able to
+build, containerize, and redeploy a non-trivial Kubernetes controller app in
+roughly 10 seconds (dominated by two `go build`s).
+
+```shell
+~/go/src/github.com/mattmoor/warm-image$ ko apply -f config/
+2018/04/25 17:11:28 Go building github.com/mattmoor/warm-image/cmd/sleeper
+2018/04/25 17:11:28 Go building github.com/mattmoor/warm-image/cmd/controller
+2018/04/25 17:11:29 Publishing us.gcr.io/convoy-adapter/github.com/mattmoor/warm-image/cmd/sleeper:latest
+2018/04/25 17:11:30 mounted sha256:eb05f3dbdb543cc610527248690575bacbbcebabe6ecf665b189cf18b541e3ca
+2018/04/25 17:11:30 mounted sha256:3ca7d60fa89dc8a1faea3046fd3516f23dab93489f3888ae539df4abc0973e52
+2018/04/25 17:11:30 mounted sha256:e2cc7c829942a768015dbcfdad7205c104cb85b84a79573555a8f4381b98110c
+2018/04/25 17:11:30 pushed us.gcr.io/convoy-adapter/github.com/mattmoor/warm-image/cmd/sleeper:latest
+2018/04/25 17:11:30 Published us.gcr.io/convoy-adapter/github.com/mattmoor/warm-image/cmd/sleeper@sha256:193acdbeff1ea9f105f49d97a6ceb7adbd30b3d64a8b9949382f4be9569cd06d
+2018/04/25 17:11:37 Publishing us.gcr.io/convoy-adapter/github.com/mattmoor/warm-image/cmd/controller:latest
+2018/04/25 17:11:37 mounted sha256:dc0dd55edef1443e976c835825479c3dc713bb689547f8a170f8a0d14f9ff734
+2018/04/25 17:11:37 mounted sha256:eb05f3dbdb543cc610527248690575bacbbcebabe6ecf665b189cf18b541e3ca
+2018/04/25 17:11:37 mounted sha256:fbc44e14a1d848ed485b5c3f03611c3e21aaa197fcba579255e5f8416a1b7172
+2018/04/25 17:11:38 pushed us.gcr.io/convoy-adapter/github.com/mattmoor/warm-image/cmd/controller:latest
+2018/04/25 17:11:38 Published us.gcr.io/convoy-adapter/github.com/mattmoor/warm-image/cmd/controller@sha256:78794915fca48d0c4b339dc1df91a72f1e4bc6a7b33beaeef8aecda0947d5d31
+clusterrolebinding "warmimage-controller-admin" configured
+deployment "warmimage-controller" unchanged
+namespace "warmimage-system" configured
+serviceaccount "warmimage-controller" unchanged
+customresourcedefinition "warmimages.mattmoor.io" configured
+```
 
 ## Usage
 
-`ko` also three commands that interact with Kubernetes resources.
+`ko` has three commands that interact with Kubernetes resources.
 
 ### `ko resolve`
 
-`ko resolve` builds on the [model](#the-ko-model) above to determine the set of
-Go import paths to build, containerize, and publish. It then outputs a
-multi-document yaml of the resources to `STDOUT`, replacing import paths with
-published image digests.
+`ko resolve` takes Kubernetes yaml files in the style of `kubectl apply`
+and (based on the [model above](#the-ko-model)) determines the set of
+Go import paths to build, containerize, and publish.
 
 To determine where to publish the images, `ko` currently requires the
 environment variable `KO_DOCKER_REPO` to be set to an acceptable docker
-repository (e.g. `gcr.io/your-project`). This will likely change in a
-future version.
+repository (e.g. `gcr.io/your-project`). **This will likely change in a
+future version.**
 
-Following the example above, this might result in:
+The output of `ko resolve` is the concatenated yaml with import paths
+replaced with published image digests.  Following the example above,
+this would be:
 
 ```shell
+# Command
 export PROJECT_ID=$(gcloud config get-value core/project)
 export KO_DOCKER_REPO="gcr.io/${PROJECT_ID}"
 ko resolve -f deployment.yaml
+
+# Output
 apiVersion: apps/v1beta1
 kind: Deployment
 metadata:
--- a/cmd/ko/README.md
+++ b/cmd/ko/README.md
@@ -1,0 +1,148 @@
+# `ko`
+
+`ko` is an **experimental** CLI to support rapid development of Go containers
+with Kubernetes and minimal configuration.
+
+## Installation
+
+`ko` can be installed via:
+
+```shell
+go install github.com/google/go-containerregistry/cmd/ko
+```
+
+## The `ko` Model
+
+`ko` is built around a very simple extension to Go's model for expressing
+dependencies using [import paths](https://golang.org/doc/code.html#ImportPaths).
+
+In Go, dependencies are expressed via blocks like:
+
+```go
+import (
+    "github.com/google/go-containerregistry/authn"
+    "github.com/google/go-containerregistry/name"
+)
+```
+
+Similarly (as you can see above), Go binaries can be referenced via import
+paths like `github.com/google/go-containerregistry/cmd/ko`.
+
+**One of the goals of `ko` is to make containers invisible infrastructure.**
+Simply replace image references in your Kubernetes yaml with the import path for
+your Go binary, and `ko` will handle containerizing and publishing that
+container image as needed.
+
+For example, you might use the following in a Kubernetes `Deployment` resource:
+
+```yaml
+apiVersion: apps/v1beta1
+kind: Deployment
+metadata:
+  name: hello-world
+spec:
+  replicas: 1
+  template:
+    spec:
+      containers:
+      - name: hello-world
+        # This is the import path for the Go binary to build and run.
+        image: github.com/mattmoor/examples/http/cmd/helloworld
+        ports:
+        - containerPort: 8080
+```
+
+**This extension of the Go convention enables us to create a tool with an
+extremely fast development cycle and effectively zero configuration.**
+
+## Usage
+
+`ko` also three commands that interact with Kubernetes resources.
+
+### `ko resolve`
+
+`ko resolve` builds on the [model](#the-ko-model) above to determine the set of
+Go import paths to build, containerize, and publish. It then outputs a
+multi-document yaml of the resources to `STDOUT`, replacing import paths with
+published image digests.
+
+To determine where to publish the images, `ko` currently requires the
+environment variable `KO_DOCKER_REPO` to be set to an acceptable docker
+repository (e.g. `gcr.io/your-project`). This will likely change in a
+future version.
+
+Following the example above, this might result in:
+
+```shell
+export PROJECT_ID=$(gcloud config get-value core/project)
+export KO_DOCKER_REPO="gcr.io/${PROJECT_ID}"
+ko resolve -f deployment.yaml
+apiVersion: apps/v1beta1
+kind: Deployment
+metadata:
+  name: hello-world
+spec:
+  replicas: 1
+  template:
+    spec:
+      containers:
+      - name: hello-world
+        # This is the digest of the published image containing the go binary
+        # at the embedded import path.
+        image: gcr.io/your-project/github.com/mattmoor/examples/http/cmd/helloworld@sha256:deadbeef
+        ports:
+        - containerPort: 8080
+```
+
+*Note that DockerHub does not currently support these multi-level names. We may
+employ alternate naming strategies in the future to broaden support, but this
+would sacrifice some amount of identifiability.*
+
+### `ko apply`
+
+`ko apply` is intended to parallel `kubectl apply`, but acts on the same
+resolved output as `ko resolve` emits. It is expected that `ko apply` will act
+as the vehicle for rapid iteration during development.  As changes are made to a
+particular application, you can run: `ko apply -f unit.yaml` to rapidly
+rebuild, repush, and redeploy their changes.
+
+`ko apply` will invoke `kubectl apply` under the covers, and therefore apply
+to whatever `kubectl` context is active.
+
+### `ko delete`
+
+`ko delete` simply passes through to `kubectl delete`, as with the `go`
+commands. It is exposed purely out of convenience for cleaning up resources
+created through `ko apply`.
+
+## Relevance to Release Management
+
+`ko` is also useful for helping manage releases. For example, if your project
+periodically releases a set of images and configuration to launch those images
+on a Kubernetes cluster, release binaries may be published and the configuration
+generated via:
+
+```shell
+export PROJECT_ID=<YOUR RELEASE PROJECT>
+export KO_DOCKER_REPO="gcr.io/${PROJECT_ID}"
+ko resolve -f config/ > release.yaml
+```
+
+This will publish all of the binary components as container images to
+`gcr.io/my-releases/...` and create a `release.yaml` file containing all of the
+configuration for your application with inlined image references.
+
+This resulting configuration may then be installed onto Kubernetes clusters via:
+
+```shell
+kubectl apply -f release.yaml
+```
+
+
+## Acknowledgements
+
+This work is based heavily on learnings from having built the
+[Docker](github.com/bazelbuild/rules_docker) and
+[Kubernetes](github.com/bazelbuild/rules_k8s) support for
+[Bazel](https://bazel.build). That work was presented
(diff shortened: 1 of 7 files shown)


cmd/ko/README.md
```diff
@@ -0,0 +1,146 @@
...
+import (
+    "github.com/google/go-containerregistry/authn"
+    "github.com/google/go-containerregistry/name"
+)
+```
+
+Similarly (as you can see above), Go binaries can be references via import
+paths like `github.com/google/go-containerregistry/cmd/ko`.
+
+With `ko`, we simply replace image references within Kubernetes yaml
```
jonjohnsonjr: This sentence is kind of hard to understand, and I think you've buried the lede a bit.

The Big Idea™ is hidden in here, i.e.:
> if I use `ko`, I don't have to worry about building and pushing containers anymore. I can just pretend kubernetes knows how to deploy my go binaries, and the container becomes an implementation detail.

I don't think that's immediately obvious unless you spell it out for attention-deficient folks like me. Splitting out the motivation and the implementation could make this sentence easier to swallow, too.

cmd/ko/README.md
```diff
@@ -0,0 +1,146 @@
...
+
+`ko` can act as a drop-in replacement for the `go` CLI, to which it passes
+through all of the top-level `go foo` commands.
+
+`ko` also exports three commands that interact with Kubernetes resources.
+
+### `ko delete`
+
+`ko delete` simply passes through to `kubectl delete`, as with the `go`
+commands, it is exposed purely out of convenience for cleaning up resources
```
jonjohnsonjr: "**. I**t is exposed..."

cmd/ko/README.md
```diff
@@ -0,0 +1,146 @@
...
+extremely fast development cycle and effectively zero configuration.**
+
+## Usage
+
+`ko` can act as a drop-in replacement for the `go` CLI, to which it passes
+through all of the top-level `go foo` commands.
+
+`ko` also exports three commands that interact with Kubernetes resources.
+
+### `ko delete`
```
jonjohnsonjr: This should go after `ko apply`

cmd/ko/README.md
```diff
@@ -0,0 +1,146 @@
...
+
+`ko delete` simply passes through to `kubectl delete`, as with the `go`
+commands, it is exposed purely out of convenience for cleaning up resources
+created through `ko apply`.
+
+### `ko resolve`
+
+`ko resolve` builds on the [model](#the-ko-model) above to determine the set of
+Go import paths to build, containerize and publish. It then outputs a
```
jonjohnsonjr: oxford comma pls

cmd/ko/README.md
```diff
@@ -0,0 +1,146 @@
...
+multi-document yaml of the resources to `STDOUT` replacing import paths with
```
jonjohnsonjr: to `STDOUT`**,** replacing...

cmd/ko/README.md
```diff
@@ -0,0 +1,146 @@
...
+    spec:
+      containers:
+      - name: hello-world
+        # This is the importpath for the Go binary to build and run.
+        image: github.com/examples/http/cmd/helloworld
+        ports:
+        - containerPort: 8080
+```
+
+**This extension of the Go convention enables us to create a tool with an
```
jonjohnsonjr: Any way you can easily "show, don't tell" this claim?

cmd/ko/README.md
```diff
@@ -0,0 +1,146 @@
...
+metadata:
+  name: hello-world
+spec:
+  replicas: 1
+  template:
+    spec:
+      containers:
+      - name: hello-world
+        # This is the importpath for the Go binary to build and run.
+        image: github.com/examples/http/cmd/helloworld
```
jonjohnsonjr: Given that https://github.com/examples is a real account... can you change this example? :)

cmd/ko/README.md
```diff
@@ -0,0 +1,146 @@
...
+
+To determine where to publish the images, `ko` currently requires the
+environment variable `KO_DOCKER_REPO` to be set to an acceptable docker
+repository (e.g. `gcr.io/your-project`). This will likely change in a
+future version.
+
+Following the example above, this might result in:
+
+```shell
+$ KO_DOCKER_REPO=gcr.io/your-project ko resolve -f deployment.yaml
```
jonjohnsonjr: The _first_ thing people will do is copy and paste these examples... this is definitely going to result in http requests to GCR.

Can you add something to make this fail to parse? Same comment for the rest of the gcr.io references.

Does `registry:latest` support mutl-level names?

cmd/ko/README.md
```diff
@@ -0,0 +1,146 @@
...
+kind: Deployment
+metadata:
+  name: hello-world
+spec:
+  replicas: 1
+  template:
+    spec:
+      containers:
+      - name: hello-world
+        # This is the importpath for the Go binary to build and run.
```
jonjohnsonjr: Weird to me that you started using "importpath" in a few places. It shows up in some comments and as a bazel-ism, not so much a go-ism, AFAICT.

cmd/ko/README.md
```diff
@@ -0,0 +1,146 @@
...
+(at least) DockerHub currently. We may employ alternate naming strategies in the
+future to broaden support, but this would sacrifice some amount of identifiability.*
+
+### `ko apply`
+
+`ko apply` is intended to parallel `kubectl apply`, but acts on the same
+resolved output as `ko resolve` emits. It is expected that `ko apply` will act
+as the vehicle for rapid iteration during development.  As changes are made to a
+particular application, users can run: `ko apply -f unit.yaml` to rapidly
+rebuild, repush and redeploy their changes.
```
jonjohnsonjr: oxford comma pls

cmd/ko/README.md
```diff
@@ -0,0 +1,146 @@
...
+periodically releases a set of images, and configuration to launch those images
+on a Kubernetes cluster, release binaries may be published and the configuration
+generated via:
+
+```shell
+KO_DOCKER_REPO=gcr.io/my-releases ko resolve -f config/ > release.yaml
+```
+
+This will publish all of the binary components as container images to
+`gcr.io/my-releases/...` and create a `release.yaml` file container all of the
```
jonjohnsonjr: s/container/containing/

cmd/ko/README.md
```diff
@@ -0,0 +1,146 @@
...
+as the vehicle for rapid iteration during development.  As changes are made to a
+particular application, users can run: `ko apply -f unit.yaml` to rapidly
+rebuild, repush and redeploy their changes.
+
+`ko apply` will invoke `kubectl apply` under the covers, and therefore apply
+to whatever `kubectl` context is active.
+
+## Relevance to Release Management
+
+`ko` is also useful for helping manage releases. For example, if your project
```
jonjohnsonjr: style nit: You have a mix of style throughout this doc with "we" at first, "users" above, "your" here, and the passive voice below (e.g. release binaries may be published).

cmd/ko/README.md
```diff
@@ -0,0 +1,146 @@
...
+periodically releases a set of images, and configuration to launch those images
```
jonjohnsonjr: extraneous comma

cmd/ko/README.md
```diff
@@ -0,0 +1,146 @@
...
+on a Kubernetes cluster, release binaries may be published and the configuration
+generated via:
+
+```shell
+KO_DOCKER_REPO=gcr.io/my-releases ko resolve -f config/ > release.yaml
+```
+
+This will publish all of the binary components as container images to
+`gcr.io/my-releases/...` and create a `release.yaml` file container all of the
+configuration for my application with inlined image references.
```
jonjohnsonjr: continuing the style nit from above: lol, and "my" here

cmd/ko/README.md
```diff
@@ -0,0 +1,146 @@
...
+
+## Installation
+
+`ko` can be installed via:
+
+```shell
+go install github.com/google/go-containerregistry/cmd/ko
+```
+
+## The `ko` Model
```
jonjohnsonjr: If this blows up, I think one of the first questions you'll get is "How does this compare to skaffold??" It might be worth dedicating a little section at the bottom to that.

cmd/ko/README.md
```diff
@@ -0,0 +1,146 @@
...
+      containers:
+      - name: hello-world
+        # This is the digest of the published image containing the go binary
+        # at the embedded importpath.
+        image: gcr.io/your-project/github.com/examples/http/cmd/helloworld@sha256:deadbeef
+        ports:
+        - containerPort: 8080
+```
+
+*It it notable that these multi-level names are known to not be supported against
```
jonjohnsonjr: > are known to not be supported against (at least) DockerHub currently

Wow. How about just:

> Note that DockerHub does not currently support these multi-level names.

cmd/ko/README.md
```diff
@@ -0,0 +1,146 @@
+# `ko`
+
+## What is `ko`?
```
ImJasonH: nit: I don't think this header adds much.

cmd/ko/README.md
ImJasonH: The term "importpath" (no spaces) might be a bazel-ism, but "import path" seems fine: https://golang.org/doc/code.html#ImportPaths

cmd/ko/README.md
ImJasonH: It feels weird to me that we're asking users to write a k8s-style YAML that won't be valid until you've `ko resolve`d it. Maybe we can signal that the YAML isn't really ready to use yet, by naming it `deployment.yaml.tmpl` or something in this example? Or otherwise making it clear that the `github.com/whatever/cmd/hello` is a placeholder which will be resolved? Possibly by surrounding it with `{{}}` ? Does Kubernetes have any kind of standard for this today?

cmd/ko/README.md
```diff
@@ -0,0 +1,146 @@
...
+        ports:
+        - containerPort: 8080
+```
+
+**This extension of the Go convention enables us to create a tool with an
+extremely fast development cycle and effectively zero configuration.**
+
+## Usage
+
+`ko` can act as a drop-in replacement for the `go` CLI, to which it passes
```
ImJasonH: It's unclear to me why I should ever use `ko build` over `go build`, even if it's just a passthrough. Maybe this means an example would be helpful, or maybe it means we can drop the passthrough altogether?

cmd/ko/README.md
```diff
@@ -0,0 +1,146 @@
...
+through all of the top-level `go foo` commands.
+
+`ko` also exports three commands that interact with Kubernetes resources.
+
+### `ko delete`
+
+`ko delete` simply passes through to `kubectl delete`, as with the `go`
```
ImJasonH: Likewise for `ko delete` -- what benefit do I get from running `ko delete foo.yaml` over `kubectl delete foo.yaml`?

ko/resolve/resolve.go
```diff
@@ -0,0 +1,163 @@
...
+	case map[interface{}]interface{}:
+		var sm sync.Map
+		var wg sync.WaitGroup
+		var mapError error
+		// Launch go routines for all of the key/value pairs.
+		// We could perhaps further parallelize by doing keys and values
+		// in parallel, but the likelihood of that materially improving
+		// things is vanishingly small since most keys are fixed in K8s
+		// anyways.
+		for k, v := range typed {
```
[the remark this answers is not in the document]
ImJasonH: Discussed offline, ignore this.

cmd/ko/README.md
dprotaso: To your point I wonder if instead of hijacking the `image` property you use `goSource` or something as explicit

also `deployment.yaml.tmpl` -> `deployment.yaml.ko` ?

cmd/ko/README.md
mattmoor: I've rephrased this to play up containers as an implementation detail that fades away.

cmd/ko/README.md
mattmoor: thanks for the link, I've added this to the section above where I talk about the model :)

cmd/ko/README.md
mattmoor: ugh, fine.

mattmoor: **tl;dr** In general, I have a bias towards adopting facets of `rules_k8s` that seem to be working well.

More and more in the context of Bazel, I've biased towards using non-sensical image names that are given meaning by `image_chroot` ([e.g.](https://github.com/mattmoor/warm-image/blob/master/controller.yaml#L16)), and this is effectively the same idea.  Since skaffold came up, I'll also point out that [this](https://github.com/GoogleContainerTools/skaffold/blob/master/examples/getting-started/skaffold.yaml#L5) never publishes a `:latest` tag, which is what is expected in all of the yamls :)

re: standard substitutions, no.  Part of my concern with `{{}}` is that it's invalid yaml, which introduces new fun.

re: `image:`, this isn't the only context for substitutions.  For example, it is an extremely useful pattern to be able to pass resolved container names as flags to controllers ([e.g.](https://github.com/mattmoor/warm-image/blob/master/controller.yaml#L21)).  Also, I find there is value in avoiding schema-sensitivity, since it means this model should work with most CRDs that deal with images as well ([e.g.](https://github.com/mattmoor/warm-image/blob/master/example-warmimage.yaml#L6)).

re: file extension, K8s doesn't have a monopoly on `yaml`.  :)

---
Ultimately, I am really trying to minimize the number of concepts folks need to grok to use this and everything we add increases the learning curve.

cmd/ko/README.md
mattmoor: Once this lands, I'll convert `WarmImage` to use it and include a sample invocation wrapped with `time`.

cmd/ko/README.md
mattmoor: I'll drop the Go passthrough for now.  Some of my thinking on this was that we might be able to piggyback on `go fmt` to also format your K8s yamls with `ko fmt`, but this is a low priority :)

cmd/ko/README.md
mattmoor: Done.

cmd/ko/README.md
mattmoor: I find the symmetry with `ko apply` useful, I have a stronger bias to keep this.

cmd/ko/README.md
mattmoor: Done.

cmd/ko/README.md
mattmoor: Done.

cmd/ko/README.md
mattmoor: Done.

cmd/ko/README.md
mattmoor: IDK about `registry:latest`.  Updated things to be copy/paste-able

cmd/ko/README.md
mattmoor: Done.

cmd/ko/README.md
mattmoor: Done.

cmd/ko/README.md
mattmoor: Done.

cmd/ko/README.md
mattmoor: Done.

cmd/ko/README.md
mattmoor: Done.

cmd/ko/README.md
mattmoor: The only instance of "We" that I could find is intended to refer to the ko maintainers.  I changed the reference below, and references to "users" to be directed at the reader ("you", "your").  LMK if I missed any.

cmd/ko/README.md
ImJasonH: I understand the motivation, and certainly introducing a new concept early on might confuse users.

My concern is that by piggybacking on Go import paths, users won't realize that their image is actually at `gcr.io/my-project/github.com/user/repo/cmd/foo/hello`, and not knowing that linkage will make it harder for them to debug issues in the future, for instance when they try to pull the image that's acting funny in the cluster to their local workstation.

A magical onboarding can be harmful if you don't understand the underlying magic. I think of it like the Sorcerer's Apprentice, you want to know at least a little about what's happening if it ever goes wrong. Maybe that's not the case for Bazel today either, I don't know.

It's probably more a documentation issue. We can spell it out extra clearly in the README that `github.com/foo/...` actually means `$KO_DOCKER_REPO/github.com/foo/...` and repeat it three times before you proceed. :grinning:

mattmoor: Perhaps it would help to share the output of a `ko apply`, since I think it helps clarify.

I'm modifying WarmImage to use this, and here's the output when I apply:
```shell
$ ko apply -f config/
2018/04/25 12:14:41 Go building github.com/mattmoor/warm-image/cmd/sleeper
2018/04/25 12:14:41 Go building github.com/mattmoor/warm-image/cmd/controller
2018/04/25 12:14:43 Publishing us.gcr.io/convoy-adapter/github.com/mattmoor/warm-image/cmd/sleeper:latest
2018/04/25 12:14:43 mounted sha256:eb05f3dbdb543cc610527248690575bacbbcebabe6ecf665b189cf18b541e3ca
2018/04/25 12:14:43 mounted sha256:3ca7d60fa89dc8a1faea3046fd3516f23dab93489f3888ae539df4abc0973e52
2018/04/25 12:14:43 mounted sha256:e2cc7c829942a768015dbcfdad7205c104cb85b84a79573555a8f4381b98110c
2018/04/25 12:14:44 pushed us.gcr.io/convoy-adapter/github.com/mattmoor/warm-image/cmd/sleeper:latest
2018/04/25 12:14:44 Published us.gcr.io/convoy-adapter/github.com/mattmoor/warm-image/cmd/sleeper@sha256:193acdbeff1ea9f105f49d97a6ceb7adbd30b3d64a8b9949382f4be9569cd06d
2018/04/25 12:14:50 Publishing us.gcr.io/convoy-adapter/github.com/mattmoor/warm-image/cmd/controller:latest
2018/04/25 12:14:51 mounted sha256:fbc44e14a1d848ed485b5c3f03611c3e21aaa197fcba579255e5f8416a1b7172
2018/04/25 12:14:51 mounted sha256:eb05f3dbdb543cc610527248690575bacbbcebabe6ecf665b189cf18b541e3ca
2018/04/25 12:14:51 mounted sha256:dc0dd55edef1443e976c835825479c3dc713bb689547f8a170f8a0d14f9ff734
2018/04/25 12:14:52 pushed us.gcr.io/convoy-adapter/github.com/mattmoor/warm-image/cmd/controller:latest
2018/04/25 12:14:52 Published us.gcr.io/convoy-adapter/github.com/mattmoor/warm-image/cmd/controller@sha256:78794915fca48d0c4b339dc1df91a72f1e4bc6a7b33beaeef8aecda0947d5d31
clusterrolebinding "warmimage-controller-admin" configured
deployment "warmimage-controller" created
namespace "warmimage-system" configured
serviceaccount "warmimage-controller" unchanged
customresourcedefinition "warmimages.mattmoor.io" configured
```

I think that this clearly indicates that the images are being published as `$KO_DOCKER_REPO/github.com/foo/...`, but I could be wrong?  WDYT?

cmd/ko/README.md
```diff
@@ -0,0 +1,199 @@
...
+```
+
+### Determining supported import paths
+
+Similar to other tooling in the Go ecosystem, `ko` expects to execute in the
+context of your `$GOPATH`. This is used to determine what package(s) `ko`
+is expected to build.
+
+Suppose `GOPATH` is `~/gopath` and the current directory is
+`~gopath/src/github.com/mattmoor/examples`. `ko` will deduce the base import
```
jonjohnsonjr: Missing a / ?

~/gopath/...

cmd/ko/README.md
```diff
@@ -0,0 +1,199 @@
...
+path to be `github.com/mattmoor/examples`, and any references to subpackages
+of this will be built, containerized and published.
+
+For examples, any of the following would be matched:
```
jonjohnsonjr: For example

cmd/ko/README.md
```diff
@@ -0,0 +1,199 @@
...
+```
+
+## Usage
+
+`ko` has three commands that interact with Kubernetes resources.
+
+### `ko resolve`
+
+`ko resolve` takes Kubernetes yaml files in the style of `kubectl apply`
+and based on the [model above](#the-ko-model) determines the set of
```
jonjohnsonjr: (based on the model above)

cmd/ko/README.md
jonjohnsonjr: Re: hijacking the `image` property. Using a stand-in image reference here feels dirty to me. I'm willing to be convinced that I'm wrong, but IMO this would be cleaner if you used another key, e.g. "import":

```yaml
    spec:
      containers:
      - name: hello-world
        # This is the import path for the Go binary to build and run.
        import: github.com/mattmoor/examples/http/cmd/helloworld
        ports:
        - containerPort: 8080
```

It's very clear that is *not* an image reference, and `import` would be replaced by an equivalent `image` by `ko`.

mattmoor: To reiterate:
> re: image:, this isn't the only context for substitutions. For example, it is an extremely useful pattern to be able to pass resolved container names as flags to controllers (e.g.). Also, I find there is value in avoiding schema-sensitivity, since it means this model should work with most CRDs that deal with images as well (e.g.).

At the end of the day, this is an experimental tool.  I see this as the heart of that experiment, and I'm pretty biased against changing it before even trying it.

cmd/ko/main.go
```diff
@@ -0,0 +1,266 @@
...
+			return nil
+		})
+		if err != nil {
+			return nil, err
+		}
+	}
+	return files, nil
+}
+
+func resolveFilesTo(fo *FilenameOptions, out io.Writer) {
```
jonjohnsonjr: It feels like you stopped naming this function halfway through

cmd/ko/README.md
mattmoor: Done

cmd/ko/README.md
mattmoor: Done

cmd/ko/README.md
mattmoor: Done.

cmd/ko/main.go
mattmoor: Done

PR #104, opened 2018-04-24
Add d8s copy

jonjohnsonjr:
This copies an image from a src reference to a dst reference, which
should generally be fast due to mounting.

### DIFF
--- a/cmd/d8s/copy.go
+++ b/cmd/d8s/copy.go
@@ -31,9 +31,9 @@ type copyCmd struct{}
 
 func (*copyCmd) Name() string { return "copy" }
 func (*copyCmd) Synopsis() string {
-	return "Lazily copies a remote src ref to a remote dst ref"
+	return "Efficiently copies a remote image from src reference to dst reference"
 }
-func (*copyCmd) Usage() string            { return "copy <src> <dst>" }
+func (*copyCmd) Usage() string            { return "copy <src reference> <dst reference>" }
 func (*copyCmd) SetFlags(f *flag.FlagSet) {}
 
 func (*copyCmd) Execute(ctx context.Context, f *flag.FlagSet, _ ...interface{}) subcommands.ExitStatus {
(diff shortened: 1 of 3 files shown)


cmd/d8s/copy.go
```diff
@@ -0,0 +1,80 @@
...
+	"github.com/google/go-containerregistry/v1/remote"
+)
+
+type copyCmd struct{}
+
+func (*copyCmd) Name() string { return "copy" }
+func (*copyCmd) Synopsis() string {
+	return "Lazily copies a remote src ref to a remote dst ref"
+}
+func (*copyCmd) Usage() string            { return "copy <src> <dst>" }
```
mattmoor: We are trying to use `<reference>` or `<tag>`

cmd/d8s/copy.go
```diff
@@ -0,0 +1,80 @@
...
+	"github.com/google/go-containerregistry/authn"
+	"github.com/google/go-containerregistry/name"
+	"github.com/google/go-containerregistry/v1/remote"
+)
+
+type copyCmd struct{}
+
+func (*copyCmd) Name() string { return "copy" }
+func (*copyCmd) Synopsis() string {
+	return "Lazily copies a remote src ref to a remote dst ref"
```
mattmoor: nit: "Lazily" is an implementation detail, how about "Efficiently"?

cmd/d8s/copy.go
```diff
@@ -0,0 +1,80 @@
...
+		log.Fatalln(err)
+	}
+	log.Printf("Pulling %v", src)
+
+	srcAuth, err := authn.DefaultKeychain.Resolve(src.Context().Registry)
+	if err != nil {
+		log.Fatalln(err)
+	}
+
+	i, err := remote.Image(src, srcAuth, http.DefaultTransport)
```
mattmoor: nit: img

cmd/d8s/copy.go
jonjohnsonjr: Done

cmd/d8s/copy.go
jonjohnsonjr: Done

cmd/d8s/copy.go
jonjohnsonjr: Done

PR #106, opened 2018-04-24
Migrate to spf13/cobra

### LINKED ISSUE #73: github.com/google/subcommands isn't very user friendly
Should consider using another library or patching it.

Biggest issue is if your subcommand returns `ExitUsageError` during `Execute` it doesn't actually print the usage. 

```bash
$ d8s append
$ # Nothing was printed!
```

Additionally, when performing `d8s [subcommand] -h` or `d8s [subcommand] -non-existent-flag` the output is confusing. See the weird `-o string ...` output below

```bash
$ d8s append -h
append [-o output-file] <src-reference> <dest-tag> <tarball>  -o string
    	output the resulting image to a new tarball
```

ImJasonH:
cobra

After this change, there are no positional args, everything is a flag. If we like single-positional-args but not double-positional-args I can undo this change for, e.g., `d8s manifest --ref foo` to go back to `d8s manifest foo`.

I'm not totally convinced I like the way I've crammed everything into `init`s. `cobra`'s documentation seems to expect flags to be top-level vars, but that got annoying since there were many `src`s and `dst`s all in scope at the same time. I wanted to avoid moving each command into its own package, since that seemed like overkill.

This change splits `get.go`'s commands into separate files, but keeps `getImage` in `get.go`, which I'm not sure I like.

Fixes #73 

/cc @dprotaso


Gopkg.lock
```diff
@@ -73,17 +73,23 @@
...
+  version = "v1.0"
 
 [[projects]]
   name = "github.com/pkg/errors"
   packages = ["."]
   revision = "645ef00459ed84a119197bfb8d8205042c6df63d"
   version = "v0.8.0"
 
+[[projects]]
+  name = "github.com/spf13/cobra"
```
mattmoor: It'll save me / you merge headaches if you copy the cobra bit from the Gopkg.toml in #102

mattmoor:
This is @jonjohnsonjr's pet area, so I'm gonna defer to him :)

cmd/d8s/append.go
```diff
@@ -15,106 +15,84 @@
...
-	if err := tarball.Write(file, tag, i, &tarball.WriteOptions{}); err != nil {
-		log.Fatalln(err)
+func init() {
+	var output, src, dst, tar string
+	appendCmd := &cobra.Command{
+		Use:   "append",
+		Short: "Append contents of a tarball to a remote image",
+		Run: func(*cobra.Command, []string) {
+			if src == "" || dst == "" || tar == "" {
+				log.Fatalln("Must provide -src, -dst and -tarball")
```
bsnchan: `tar` as per line 96 in this file

jonjohnsonjr:
Would you mind splitting out the cobra migration from the flags changes? I think the only commands that are ~ambiguous are pull, push, and append.

Is there any way to avoid the extra double indentation? I know caring about line length is rare in golang, but I'll personally find this annoying in my editor :grimacing:

ImJasonH:
I removed the redundant indentation, be careful what you wish for 😈

cmd/d8s/config.go
```diff
@@ -0,0 +1,44 @@
...
+	ref := args[0]
+	i, err := getImage(ref)
+	if err != nil {
+		log.Fatalln(err)
+	}
+	config, err := i.RawConfigFile()
+	if err != nil {
+		log.Fatalln(err)
+	}
+	fmt.Println(string(config))
```
jonjohnsonjr: Can you change this to `fmt.Print` so that `$ d8s config ubuntu | sha256sum` works?

cmd/d8s/manifest.go
```diff
@@ -0,0 +1,44 @@
...
+	ref := args[0]
+	i, err := getImage(ref)
+	if err != nil {
+		log.Fatalln(err)
+	}
+	manifest, err := i.RawManifest()
+	if err != nil {
+		log.Fatalln(err)
+	}
+	fmt.Println(string(manifest))
```
jonjohnsonjr: Same comment as config, `Print` instead of `Println`

ImJasonH: Excellent idea, done.

PR #109, opened 2018-04-25
Support $HOME/.docker if $DOCKER_CONFIG is unset

ImJasonH:
This is needed for image-rebase to work on Container Builder, where the
user's homedir is unset, but $HOME is explicitly set in each build step.

### DIFF
--- a/authn/keychain.go
+++ b/authn/keychain.go
@@ -16,12 +16,13 @@ package authn
 
 import (
 	"encoding/json"
+	"errors"
 	"fmt"
 	"io/ioutil"
 	"log"
 	"os"
-	"os/user"
 	"path"
+	"runtime"
 
 	"github.com/google/go-containerregistry/name"
 )
@@ -41,14 +42,23 @@ func configDir() (string, error) {
 	if dc := os.Getenv("DOCKER_CONFIG"); dc != "" {
 		return dc, nil
 	}
-	if h := os.Getenv("HOME"); h != "" {
-		return path.Join(h, ".docker"), nil
+	if h := homeDir(); h != "" {
+		return path.Join(homeDir(), ".docker"), nil
 	}
-	usr, err := user.Current()
-	if err != nil {
-		return "", err
+	return "", errNoHomeDir
+}
+
+var errNoHomeDir = errors.New("could not determine home directory")
+
+func homeDir() string {
+	if runtime.GOOS == "windows" {
+		home := os.Getenv("HOMEDRIVE") + os.Getenv("HOMEPATH")
+		if home == "" {
+			home = os.Getenv("USERPROFILE")
+		}
+		return home
 	}
-	return path.Join(usr.HomeDir, ".docker"), nil
+	return os.Getenv("HOME")
 }
 
 // authEntry is a helper for JSON parsing an "auth" entry of config.json
--- a/authn/keychain.go
+++ b/authn/keychain.go
@@ -22,6 +22,7 @@ import (
 	"log"
 	"os"
 	"path"
+	"runtime"
 
 	"github.com/google/go-containerregistry/name"
 )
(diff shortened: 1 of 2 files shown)


authn/keychain.go
```diff
@@ -41,6 +41,9 @@ func configDir() (string, error) {
 	if dc := os.Getenv("DOCKER_CONFIG"); dc != "" {
 		return dc, nil
 	}
+	if h := os.Getenv("HOME"); h != "" {
```
dekkagaijin: I think that the correct fix would be to remove the `user.Current().homedir` and port this code (in addition to supporting $DOCKER_CONFIG):
https://github.com/google/containerregistry/blob/master/client/docker_creds_.py#L211

The python version more closely duplicates the actual behavior of the Docker client, which uses `%USERPROFILE%` on windows and `~` otherwise. The Docker client code is kinda spaghetti, and I don't remember where it is off the top of my head, but you can trust that this is a detail that has been burned into my brain :p

dekkagaijin: Actually, it looks like the python lib has a bug here. DOCKER_CONFIG is supposed to be a directory, rather than the path including `config.json`

ImJasonH: I don't know anything about Windows, but I found https://stackoverflow.com/questions/7922270/obtain-users-home-directory which seems useful.

Added tests, some of which are `t.Skip`ped on non-Windows, but should work (famous last words).

OTOH, I can remove the `if runtime.GOOS == "windows"` check and just allow this weird behavior on non-Windows.

WDYT?

dekkagaijin: I couldn't find the actual code, but the docs mention %USERPROFILE% and $HOME:
https://github.com/docker/cli/blob/2f3b13742fb1de2eb3b89ed2350182b8e83cf9dc/docs/reference/commandline/login.md#privileged-user-requirement

authn/keychain.go
```diff
@@ -41,14 +42,23 @@ func configDir() (string, error) {
...
+	return "", errNoHomeDir
+}
+
+var errNoHomeDir = errors.New("could not determine home directory")
+
+func homeDir() string {
+	if runtime.GOOS == "windows" {
+		home := os.Getenv("HOMEDRIVE") + os.Getenv("HOMEPATH")
+		if home == "" {
+			home = os.Getenv("USERPROFILE")
```
dekkagaijin: Left comments in the previous thread. tl;dr Docker specifically uses %USERPROFILE% on Windows

ImJasonH: Okay, the new path is DOCKER_CONFIG -> USERPROFILE -> HOME, regardless of platform.

If this is terrible we can change it in the future.

authn/keychain.go
```diff
@@ -51,14 +50,10 @@ func configDir() (string, error) {
...
-		home := os.Getenv("HOMEDRIVE") + os.Getenv("HOMEPATH")
-		if home == "" {
-			home = os.Getenv("USERPROFILE")
-		}
-		return home
+	if h := os.Getenv("HOME"); h != "" {
+		return h
 	}
-	return os.Getenv("HOME")
+	return os.Getenv("USERPROFILE")
```
dekkagaijin: I'd keep the explicit branch for Windows since %HOME% is also set by default on newer versions, and is often the same value as %USERPROFILE%, but not necessarily (boy oh boy was that a fun week of debugging). Specifically:
```
// dockerUserHomeDir returns the current user's home directory, as interpreted by Docker.
func dockerUserHomeDir() string {
  if runtime.GOOS == "windows" {
    // Docker specifically expands "%USERPROFILE%" on Windows, 
    return os.Getenv("USERPROFILE")	
  }
  // Docker defaults to "$HOME" Linux and OSX. 
  return os.Getenv("HOME")
}
```

dekkagaijin: (sorry, should have just been explicit from the start)

PR #115, opened 2018-04-27
Generate docs via a new flag to crane

ImJasonH:
So far they're pretty weak ([e.g.](https://github.com/ImJasonH/go-containerregistry/blob/546329712311bfe314ca946e3b7bc8f4dbdb67c5/cmd/crane/doc/crane.md)), but that's just a matter of adding better docs to the subcommands and flags themselves. I'll work on those separately.

```
bazel run //cmd/crane:crane -- --gendoc=$PWD/cmd/crane/doc
```


cmd/crane/main.go
```diff
@@ -27,9 +28,18 @@ var rootCmd = &cobra.Command{
 	Run:   func(cmd *cobra.Command, _ []string) { cmd.Usage() },
 }
 
+var gendoc = flag.String("gendoc", "", "If set, path to generate docs and exit")
+
 func main() {
+	flag.Parse()
```
mattmoor: It's weird that we have two flag parsers.

ImJasonH: Yeah I agree. I thought about promoting it to a command, but then it's weird because users shouldn't have to run it or ever think about it. So I wanted to hide it from generated usage and docs. `flag.Parse` seemed like a good place to hide it.

Another option would be to have a _separate_ command that does nothing but generate docs for `crane`, which would require `rootCmd` to become exported, and bleh I just gave up and did the simplest dumbest thing.

jonjohnsonjr:
This is like the opposite of bespoke, hand-crafted READMEs 😢

jonjohnsonjr:
Is there a way to include examples in the output? this looks fine to me as a starting point.

ImJasonH:
Yeah there's an `Example` field on `Command`: https://godoc.org/github.com/spf13/cobra#Command

Not sure how it's rendered.

cmd/crane/main.go
jonjohnsonjr: Exporting rootCmd would actually be *really* nice for any hypothetical tool that might want to extend `crane` with additional features/commands... ;-)

jonjohnsonjr: Started towards this in #142.

Issue #123, opened 2018-05-06
crane append: cross repo mount causes "UNAUTHORIZED" error on Dockerhub


sclevine:
I've identified a potential bug with `crane append` on both macOS 10.13.4 and Linux (docker4mac).

Crane version: `48dcc10b3aad9a760316aba5c88138bcc60dbcd1`
Go version (mac): `go version go1.10 darwin/amd64`
Go version (linux): `go version go1.10 darwin/amd64` (`GOOS=linux`)

When the source is a Dockerhub repo and the destination is a different Dockerhub repo, I get the following error:
```
 opal:test stephen$ crane append packs/cflinuxfs2:run sclevine/some-app:latest ./layer.tgz
2018/05/06 16:58:34 UNAUTHORIZED: "authentication required"
```
I get the same error if I have write access to both repos:
```
 opal:test stephen$ crane append sclevine/cflinuxfs2:run sclevine/some-app:latest ./layer.tgz
2018/05/06 16:59:45 UNAUTHORIZED: "authentication required"
```
Using the same repo for both the image source and image destination succeeds.

The authorization error is coming from this call:
https://github.com/google/go-containerregistry/blob/48dcc10b3aad9a760316aba5c88138bcc60dbcd1/cmd/crane/append.go#L94

I was able to fix this by removing these lines (which create a cross-repo mount):
https://github.com/google/go-containerregistry/blob/48dcc10b3aad9a760316aba5c88138bcc60dbcd1/cmd/crane/append.go#L85-L87
This is very slow, presumably because the base layers are copied to the host and back to the registry.

sclevine:
More data from a failed attempt:

Request:
```
POST https://index.docker.io/v2/sclevine/some-app/blobs/uploads/?from=packs%2Fcflinuxfs2&mount=sha256%3A9535de6d506e0e353a9378d14ad3fe8f2d44bf7f1e8e7dd5b58c01a053878d37 application/json
(no body)
```

Response:
```
[...]

PR #124, opened 2018-05-07
Fix scope-related bug in remote.Write (#123)

sclevine:
This PR refactors `transport.New` to take fully-qualified scopes. This is necessary to fix `crane append` when the source and destination are two different Docker Hub repos .

See #123 for additional context.

### DIFF
--- a/v1/remote/transport/transport.go
+++ b/v1/remote/transport/transport.go
@@ -27,9 +27,9 @@ const (
 )
 
 // New returns a new RoundTripper based on the provided RoundTripper that has been
-// setup to authenticate with the remote registry hosting "ref", in the capacity
-// laid out by the specified Scope.
-func New(ref name.Reference, auth authn.Authenticator, t http.RoundTripper, scopes []string) (http.RoundTripper, error) {
+// setup to authenticate with the remote registry "reg", in the capacity
+// laid out by the specified scopes.
+func New(reg name.Registry, auth authn.Authenticator, t http.RoundTripper, scopes []string) (http.RoundTripper, error) {
 	// The handshake:
 	//  1. Use "t" to ping() the registry for the authentication challenge.
 	//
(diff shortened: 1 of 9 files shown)


v1/remote/transport/transport.go
```diff
@@ -29,7 +29,7 @@ const (
 // New returns a new RoundTripper based on the provided RoundTripper that has been
 // setup to authenticate with the remote registry hosting "ref", in the capacity
 // laid out by the specified Scope.
-func New(ref name.Reference, auth authn.Authenticator, t http.RoundTripper, a Scope) (http.RoundTripper, error) {
+func New(ref name.Reference, auth authn.Authenticator, t http.RoundTripper, scopes []string) (http.RoundTripper, error) {
```
mattmoor: I think this change would be a bit smaller if you used: `scopes ...string` instead

mattmoor: Though I guess not as much as I'd hoped at first because we're also hoisting the `.Scope` calls.

mattmoor:
I think @jonjohnsonjr should take this, he's our mounting expert.  This bug almost certainly exists in Python as well :(

v1/remote/transport/transport.go
ImJasonH: Making `scopes` variadic makes it harder to add more repeated args in the future, if we ever want to do that. I think I'd prefer to stick with `[]string`.

sclevine: I went back and forth about using an explicit slice vs. a variadic arg as well as whether we should keep the `Scope` type. I chose this implementation for the reason @ImJasonH suggests, and also because the data itself has changed from typed sentinel values to arbitrary data.

We should probably replace the `ref name.Reference` in `transport.New` with a `reg name.Registry`, given that only `ref.Context().Registry` is used. I think this would make the transport signature easier to understand. Any objections to that additional change?

jonjohnsonjr:
Thanks!

PR #125, opened 2018-05-08
Add option to force registry scheme

guilhermebr:
These changes make possible to use with `insecure registry` over HTTP.

Needed in https://github.com/GoogleContainerTools/kaniko/pull/169


name/registry.go
```diff
@@ -23,6 +23,7 @@ const (
 
 // Registry stores a docker registry name in a structured form.
 type Registry struct {
+	Scheme   string
```
mattmoor: I don't really like that this is public / mutable.

I wonder if we should simply move `transport.Scheme` to be a property of Registry, and add a `NewInsecureRegistry` constructor.

guilhermebr: Looks good! I will try this.

jonjohnsonjr:
Still planning to add this?

guilhermebr:
@jonjohnsonjr Sure,  I will back this week to solve this! sorry for the delay

jonjohnsonjr:
No worries! Just curious :)

guilhermebr:
@mattmoor Needs to create an option to set insecure using ref.ParseReference function. 
I'm not sure the best way to do this without broke compatibility... options I can see:

pkg/v1/remote/transport/scheme.go
```diff
@@ -15,28 +15,10 @@
...
 )
 
-// Detect more complex forms of local references.
-var reLocal = regexp.MustCompile(`.*\.local(?:host)?(?::\d{1,5})?$`)
-
-// Detect the loopback IP (127.0.0.1)
-var reLoopback = regexp.MustCompile(regexp.QuoteMeta("127.0.0.1"))
-
 // Scheme returns https scheme for all the endpoints except localhost.
 func Scheme(reg name.Registry) string {
```
ImJasonH: This func doesn't seem to add much anymore, can we just update callers to call `reg.Scheme` directly? (Could also be a separate PR)

pkg/v1/remote/transport/scheme.go
```diff
@@ -15,28 +15,10 @@
...
-	"strings"
-
 	"github.com/google/go-containerregistry/pkg/name"
 )
 
-// Detect more complex forms of local references.
-var reLocal = regexp.MustCompile(`.*\.local(?:host)?(?::\d{1,5})?$`)
-
-// Detect the loopback IP (127.0.0.1)
-var reLoopback = regexp.MustCompile(regexp.QuoteMeta("127.0.0.1"))
```
dekkagaijin: also `var reipv6Loopback = regexp.MustCompile(regexp.QuoteMeta("::1"))`?

pkg/v1/remote/transport/scheme.go
guilhermebr: done, but I'm afraid to break for external uses

pkg/v1/remote/transport/scheme.go
guilhermebr: done

bechhansen:
Hello @mattmoor 

I'm using https://github.com/getporter/porter and porter is using this library indirectly.
We currently have issues when Porter is pushing images to registries on internal networks.

Looks like this PR always forces the client to use unsecure http when using registries on the _local_ top-level domain.
Is this correctly understood? Is it in any way possible to make it use https on e.g. _registry.example.local_?

For modern zero-trust architectures we really always also want to use https on our internal networks.

@carolynvs this question might be relevant for you. Don't we want portal to be able to use https on local networks?

carolynvs:
@bechhansen I don't think there's anything porter can do to workaround this by calling the library differently. I suggest creating a new issue in this repo that proposes what you'd like to see (i.e. respecting the protocol used if specified) and see if they are up for a patch.

PR #134, opened 2018-05-12
Wrap cmd errors with some context

### LINKED ISSUE #74: errors returned via cli aren't user friendly
Currently errors in this library don't provide much context. Thus errors in the resulting cli suffer from poor UX

We should use [github.com/pkg/errors](https://github.com/pkg/errors) and wrap errors with more context.

jonjohnsonjr:
Fixes #74


jonjohnsonjr:
@ImJasonH I pushed another commit, WDYT about adding line numbers to the logged output?

Issue #133, opened 2018-05-12
Support registry-less minikube development in ko via daemon.Write


mattmoor:
Today ko's [`resolveFile`](https://github.com/google/go-containerregistry/blob/master/cmd/ko/resolve.go#L77) unconditionally uses `publish.NewDefault`.

`rules_k8s` [never cracked minikube development](https://github.com/bazelbuild/rules_k8s/issues/9), but the goal was to publish images to the exposed minikube daemon.  The python library never supported `daemon.Write`, but we do :)

I'd like us to expose an implementation of [`publish.Interface`](https://github.com/google/go-containerregistry/blob/master/ko/publish/publish.go#L23) that publishes to an exposed daemon via `daemon.Write`, and hook into this in `ko` via a flag: `--mk` or `--minikube`.

WDYT?

jonjohnsonjr:
Brilliant.

PR #135, opened 2018-05-12
Add a local mode to ko.

mattmoor:
This change adds a --local (or -L) flag to `ko`, which directs it to "publish" image references differently.  Instead of publishing the image to a registry, and substituting its digest, this change adds the option to publish the image to the docker daemon.

The scenario this is aimed at supporting is local development with minikube without a registry roundtrip.  With minikube as `kubectl config current-context`, this can load the images directly into the minikube docker daemon and then direct minikube to run the resulting templates.

Fixes: https://github.com/google/go-containerregistry/issues/133


ko/publish/daemon.go
```diff
@@ -0,0 +1,47 @@
...
+
+import (
+	"log"
+
+	"github.com/google/go-containerregistry/name"
+	"github.com/google/go-containerregistry/v1"
+	"github.com/google/go-containerregistry/v1/daemon"
+)
+
+// demon is intentionally misspelled to avoid name collision (and drive Jon nuts).
```
jonjohnsonjr: sigh

cmd/ko/local.go
```diff
@@ -0,0 +1,28 @@
...
+	"github.com/spf13/cobra"
+)
+
+type LocalOptions struct {
+	Local bool
+}
+
+func addLocalArg(cmd *cobra.Command, lo *LocalOptions) {
+	cmd.Flags().BoolVarP(&lo.Local, "local", "L", lo.Local,
+		"Whether to publish images to a local docker daemon vs. a registry.")
```
jonjohnsonjr: Would it ever make sense to want to do both? I guess you could just drop the flag and rerun it?

mattmoor: I don't think so?  You can use a registry with minikube too, but you lose the fully-local aspect.

ko/publish/daemon.go
```diff
@@ -0,0 +1,47 @@
...
+	return &demon{wo}
+}
+
+// Publish implements publish.Interface
+func (d *demon) Publish(img v1.Image, s string) (name.Reference, error) {
+	tag, err := name.NewTag(s, name.WeakValidation)
+	if err != nil {
+		return nil, err
+	}
+	log.Printf("Publishing (locally) %v", tag)
```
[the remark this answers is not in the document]
mattmoor: Done.

cmd/ko/commands.go
```diff
@@ -70,7 +71,7 @@ func addKubeCommands(topLevel *cobra.Command) {
 		Run: func(cmd *cobra.Command, args []string) {
 			// TODO(mattmoor): Use io.Pipe to avoid buffering the whole thing.
```
ImJasonH: You don't even need a pipe. Set up the exec.Command before here and pass `kubectlCmd.Stdin` to `resolveFilesToWriter`

ko/publish/daemon.go
```diff
@@ -0,0 +1,47 @@
...
+func (d *demon) Publish(img v1.Image, s string) (name.Reference, error) {
+	tag, err := name.NewTag(s, name.WeakValidation)
+	if err != nil {
+		return nil, err
+	}
+	log.Printf("Loading %v", tag)
+	if _, err := daemon.Write(tag, img, d.wo); err != nil {
+		return nil, err
+	}
+	log.Printf("Loaded %v", tag)
```
ImJasonH: Is this logging necessary?

ko/publish/daemon_test.go
```diff
@@ -0,0 +1,56 @@
...
+	"testing"
+
+	"github.com/docker/docker/api/types"
+	"github.com/google/go-containerregistry/v1/daemon"
+	"github.com/google/go-containerregistry/v1/random"
+)
+
+type MockImageLoader struct{}
+
+func (m *MockImageLoader) ImageLoad(_ context.Context, _ io.Reader, _ bool) (types.ImageLoadResponse, error) {
```
ImJasonH: Nit: if you don't use any args you don't have to name them.

PR #138, opened 2018-05-14
added healthcheck to config struct and healthcheck struct

chrisz100:
Needed to implement HEALTHCHECK commands to kaniko - see https://github.com/GoogleContainerTools/kaniko/issues/17#issuecomment-388868094

### DIFF
--- a/v1/config.go
+++ b/v1/config.go
@@ -64,6 +64,27 @@ type RootFS struct {
 	DiffIDs []Hash `json:"diff_ids"`
 }
 
+// HealthConfig holds configuration settings for the HEALTHCHECK feature.
+type HealthConfig struct {
+	// Test is the test to perform to check that the container is healthy.
+	// An empty slice means to inherit the default.
+	// The options are:
+	// {} : inherit healthcheck
+	// {"NONE"} : disable healthcheck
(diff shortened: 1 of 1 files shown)


chrisz100:
@dlorenc cc

v1/config.go
```diff
@@ -64,6 +64,27 @@ type RootFS struct {
...
+// HealthConfig holds configuration settings for the HEALTHCHECK feature.
+type HealthConfig struct {
+	// Test is the test to perform to check that the container is healthy.
+	// An empty slice means to inherit the default.
+	// The options are:
+	// {} : inherit healthcheck
+	// {"NONE"} : disable healthcheck
+	// {"CMD", args...} : exec arguments directly
+	// {"CMD-SHELL", command} : run command with system's default shell
+	Test []string `json:",omitempty"`
```
dlorenc: nit: should these be lowercased like `json:"test,omitempty"`

dlorenc: Ah, basically just to make it lowercase when marshalled, to match the way we marshal all the other values in this file

dlorenc:
Ah sorry for the back and forth, my coworkers point out that we actually should leave it the way you had it before to match exactly how Docker serializes. After that, this LGTM.

PR #142, opened 2018-05-16
Split cmd/crane implementation to pkg/crane

jonjohnsonjr:
Rewrites each "init()" to be "NewCmdFoo". This pattern was cribbed from
the kubectl command in kubernetes/kubernetes.

I'm doing this so I can reuse most of these for cmd/gcr and selectively
replace bits of it with gcr-specific stuff.

I can forsee us cleaning up the signatures of these crane cmd
implementations to be more friendly and then exposing them so that
folks don't have to rewrite these common functions over and over again.


jonjohnsonjr:
I didn't test that `rebase_cloudbuild.yaml` still works (mostly out of laziness), but since this was a mostly mechanical transformation, I don't expect to have broken anything (I tested most commands).

Someone may want to double check that I haven't broken anything, since we don't have any real tests for crane.

Issue #143, opened 2018-05-16
ko does not load images into minikube daemon


jszroberto:
I was testing `ko` to build/upload container images into Minikube's daemon but so far I found the following problems: 

1. It defaults to the docker env you have set up, in my case "Docker for Mac" not to Minikube (I think that should be said at least in the Readme)
2. I tried to run `eval $(minikube docker-env)`, so that the minikube daemon would be used as describe in [Reusing the docker daemon from Minikube](https://github.com/kubernetes/minikube/blob/master/docs/reusing_the_docker_daemon.md). This didn't help
3. I investigate this library and found that you use [NewEnvClient](https://godoc.org/github.com/docker/docker/client#NewEnvClient) for the daemon, which should use the variables provided by `minikube docker-env`. 
```
$ minikube docker-env                               
export DOCKER_TLS_VERIFY="1"
export DOCKER_HOST="tcp://192.168.64.12:2376"
export DOCKER_CERT_PATH="/Users/jszroberto/.minikube/certs"
export DOCKER_API_VERSION="1.23"
# Run this command to configure your shell:
# eval $(minikube docker-env)
```

4. I tracked it down and I saw that those variables are set when the client is created, so there should be used, but running `docker images` doesn't displayed the images that should be loaded. 
5. `ko` doesn't display much debugging information. It looks that the loading is done though. 

Do I miss something?

jszroberto:
I found a funny behavior. 

If I use `ko` build with the changes introduced in the pr https://github.com/google/go-containerregistry/pull/146, the images are loaded
```
REPOSITORY                                TAG                 IMAGE ID            CREATED             SIZE
ko/test1   latest              d5d98a64c534        3 minutes ago       52.5MB
ko/test2   latest              df61ed21ef34        3 minutes ago       2.45GB
ko/test3     latest              34f39829790e        3 minutes ago       2.44GB

```
but if I rebuild it based on your master, the image information gets messed up:

```
$  docker images                         
REPOSITORY                                TAG                 IMAGE ID            CREATED             SIZE
<none>                                    <none>              d5d98a64c534        About an hour ago   52.5MB
<none>                                    <none>              df61ed21ef34        About an hour ago   2.45GB
<none>                                    <none>              34f39829790e        About an hour ago   2.44GB
```

Going to trying this again with a fresh minikube

jszroberto:
That last comment looks to be a false alarm, not sure why loading to the minikube deamon images works today. 

[...]

PR #148, opened 2018-05-16
Don't use the latest tag

mattmoor:
Avoid the latest tag so that the K8s pull policy will default to IfNotPresent.

Also use a fake domain to avoid DoSing stuff if the kubelet starts pulling.

Fixes: https://github.com/google/go-containerregistry/issues/143


mattmoor:
cc @jszroberto 

If you have the time to check this out, I'd love to know if it fixes all your problems :)

jszroberto:
@mattmoor I've just tried it out and minikube is now able to use the loaded images from the daemon 👍

ko/publish/daemon.go
```diff
@@ -34,7 +35,11 @@ func NewDaemon(wo daemon.WriteOptions) Interface {
 
 // Publish implements publish.Interface
 func (d *demon) Publish(img v1.Image, s string) (name.Reference, error) {
-	tag, err := name.NewTag(s, name.WeakValidation)
+	h, err := img.Digest()
+	if err != nil {
+		return nil, err
+	}
+	tag, err := name.NewTag(fmt.Sprintf("local.dev/%s:%s", s, h.Hex), name.WeakValidation)
```
jonjohnsonjr: What about `ko.local` instead of `local.dev`?

https://tools.ietf.org/html/rfc6762

PR #157, opened 2018-05-19
Fallback to Anonymous if credentials are not found

### LINKED ISSUE #156: "credentials not found in native keychain" message should fallback to Anonymous
`authn.Authorization()` returns an error instead of `Anonymous.Authorization()` if the credential helper exits with code 1 and this message "credentials not found in native keychain".

See https://github.com/google/go-containerregistry/blob/aea8d2a7256c0e2216a15d9913d564332e3f1f8a/authn/helper.go#L76

dgageot:
Fixes #156

Signed-off-by: David Gageot <david@gageot.net>

### DIFF
--- a/authn/helper.go
+++ b/authn/helper.go
@@ -72,15 +72,17 @@ func (h *helper) Authorization() (string, error) {
 
 	var out bytes.Buffer
 	cmd.Stdout = &out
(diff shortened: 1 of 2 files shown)


ImJasonH:
Thanks!

PR #167, opened 2018-05-24
Fix mutate.Rebase DiffID calculation

jonjohnsonjr:
We were downloading all the layers just to calculate their DiffIDs. We
can get that from the ConfigFile.

### DIFF
--- a/v1/mutate/mutate.go
+++ b/v1/mutate/mutate.go
@@ -30,9 +30,9 @@ import (
 
 const whiteoutPrefix = ".wh."
 
-// Addendum contains layers and history to be appended
-// to a base image
+// Addendum contains the layer, history, and source image to be appended to a base image
 type Addendum struct {
+	Image   v1.Image
 	Layer   v1.Layer
 	History v1.History
 }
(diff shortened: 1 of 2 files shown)


v1/mutate/mutate.go
```diff
@@ -79,10 +79,22 @@ func Append(base v1.Image, adds ...Addendum) (v1.Image, error) {
...
+			diffID, err = partial.BlobToDiffID(add.Image, digest)
+			if err != nil {
+				return nil, err
+			}
+		} else {
+			diffID, err = add.Layer.DiffID()
+			if err != nil {
+				return nil, err
+			}
+		}
```
mattmoor: I'm fine with this as a band-aid, but I'd like to make Layer.DiffID elide the Blob fetch whenever possible.

mattmoor: Can you open an issue to track that?  It was something I'd meant to do, but fell off my radar.

jonjohnsonjr:
Closing this in favor of #187

PR #173, opened 2018-05-25
Add newline to `crane digest` output

jonjohnsonjr:
As (probably) the only user of this command, I find that I want a
newline more often than I don't want a newline.


ImJasonH:
Can you outline your use case some more? I admit I have never used this command in anger, but I had imagined it being useful in terms of

```sh
digest=$(crane digest foo)
echo getting manfiest of foo at $digest
crane manifest foo@$digest
```

ImJasonH:
Maybe an `-n` flag a la `echo -n`? (Though I hate to solve every tiny problem with a new flag)

PR #176, opened 2018-05-28
Log pushed digest to stderr

### LINKED ISSUE #169: ko resolve should not include output from the push

ImJasonH:
Fixes #169

### DIFF
--- a/v1/remote/write.go
+++ b/v1/remote/write.go
@@ -286,7 +286,7 @@ func (w *writer) commitImage() error {
 	}
 
 	// The image was successfully pushed!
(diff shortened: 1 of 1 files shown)


v1/remote/write.go
```diff
@@ -286,7 +286,7 @@ func (w *writer) commitImage() error {
 	}
 
 	// The image was successfully pushed!
-	fmt.Printf("%v: digest: %v size: %d\n", w.ref, digest, len(raw))
+	log.Printf("%v: digest: %v size: %d\n", w.ref, digest, len(raw))
```
jonjohnsonjr: I think there's an extra `\n` now :P

Issue #170, opened 2018-05-24
ko should have stricter flag parsing


jonjohnsonjr:
`ko apply L -f foo/` should probably error out, I meant `ko apply -L -f foo/`.

PR #183, opened 2018-05-31
Reject unrecognized arguments.

mattmoor:
Previously we silently ignored unrecognized flags, which led to non-intuitive behaviors when a typo was present.

Fixes: https://github.com/google/go-containerregistry/issues/170


jonjohnsonjr:
https://github.com/spf13/cobra/blob/2da4a54c5ceefcee7ca5dd0eea1e18a3b6366489/args.go#L27

mattmoor:
Much simpler, thanks.

Issue #165, opened 2018-05-24
Cross-repo blob mounts not working with crane rebase on Docker Hub


sclevine:
With the same bits for the original, old, new, and rebased image, I observe that one of the FS image layers is downloaded from the host machine and uploaded to the registry when all repositories are hosted on index.docker.io, but not when they are hosted on gcr.io.

When I configure the transport to log all outbound requests, I see an extra `PUT` request in the Docker Hub case. In the output of `crane rebase`, I see one less `mounted blob` line, and one more `pushed blob` line. In the GCR case, all FS layers are successfully mounted, and none appear to be copied through the host.

In the Docker Hub case, I observe network traffic that correlates with a layer download and upload.

Example command I've run to observe this behavior:
```bash
crane rebase --new_base packs/cflinuxfs2:build --old_base packs/cflinuxfs2@sha256:febf8638351de52e8f89922c4640731e7a2d01df5285ec5f6a13050e5e50c3f0 --original sclevine/export-test20 --rebased sclevine/export-test20
```

Example output (all index.docker.io):
```
2018/05/23 18:50:33 mounted blob: sha256:804bd5f007d2a47308203d462084378ed0991bf08a07d9567e131d72d18b6c83
2018/05/23 18:50:33 mounted blob: sha256:2bef03052cada8d5755b58a2fe0fbe8ef8bc6fe487064089961b36667745d637
2018/05/23 18:50:33 mounted blob: sha256:baff991d413d1e591236551c268d1895b8c4468fea9e8fcebe513a029092117b
2018/05/23 18:50:33 mounted blob: sha256:6503ba87cf54f220a454752e60c76f687f5eb7e53a4953fef07469c68502b1f3
2018/05/23 18:50:33 mounted blob: sha256:48519f0d2275d289d52e3f44d26a91833598622662e18ef5cfa1ed23942ac2a5
2018/05/23 18:50:34 pushed blob sha256:dc9ad858060e24bcf392befb2169c27afcccbb211a40d2d6c67bcee2ba4f11c4
2018/05/23 18:50:51 pushed blob sha256:485c6a10daa6280fc6ca431b8f89775f28fdeee388dac095f21ac350a353368c
index.docker.io/sclevine/export-test9:latest: digest: sha256:250576160c8041ac284b13b8796a9d9972464ec56035da30f811bce9242cc3c9 size: 1247
```

Example output (all gcr.io):
```
[...]

PR #186, opened 2018-06-01
Change the way mounting works in go-containerregistry.

mattmoor:
This completely changes the model for cross-repo mounting in the `go-containerregistry` library.  Rather than having clients of `remote.Write` pass in a set of potential mount sources, `remote.Write` is sensitive to a new `v1.Layer` wrapper type: `remote.MountableLayer`, which implements `v1.Layer` by embedding the layer it wraps and augmenting that with the name of the repository from which we should attempt to mount it, if written.  With this model, `remote.Image` now wraps all of the `v1.Layer`s it returns with the name of the repository from which it is pulling.

Fixes: https://github.com/google/go-containerregistry/issues/165

cc @sclevine

### DIFF
--- a/v1/remote/write.go
+++ b/v1/remote/write.go
@@ -31,8 +31,6 @@ import (
 // WriteOptions are used to expose optional information to guide or
 // control the image write.
 type WriteOptions struct {
-	// The set of paths from which to attempt to mount blobs.
-	MountPaths []name.Repository
 	// TODO(mattmoor): Expose "threads" to limit parallelism?
 }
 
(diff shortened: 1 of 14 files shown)


v1/remote/write.go
```diff
@@ -132,9 +136,13 @@ func (w *writer) initiateUpload(h v1.Hash) (location string, mounted bool, err e
...
 	}
+	l, err := w.img.LayerByDigest(h)
+	if err != nil {
+		return "", false, err
+	}
 	var from []string
-	for _, m := range w.options.MountPaths {
-		from = append(from, m.RepositoryStr())
+	if ml, ok := l.(*MountableLayer); ok {
+		from = append(from, ml.Repository.RepositoryStr())
```
jonjohnsonjr: I don't think it makes a lot of sense for `from` to be an array, now.

mattmoor: I combined this predicate with the one below.

Issue #166, opened 2018-05-24
Rebasing always downloads layers to the host


sclevine:
When rebasing with all repositories on GCR or all repositories on Docker Hub, I observe that layers are unexpectedly downloaded to the host, but not re-uploaded. I observe this with both `crane rebase` and via direct usage of `mutate.Rebase`.

The culprit appears to be the first execution of this line:
https://github.com/google/go-containerregistry/blob/7f1e66edb3a89174cb95bbc9ab5324449cd80d69/v1/mutate/mutate.go#L82

Coming from:
https://github.com/google/go-containerregistry/blob/7f1e66edb3a89174cb95bbc9ab5324449cd80d69/v1/mutate/rebase.go#L80

I'm guessing that `DiffID` requires an uncompressed image which must be downloaded from the registry before decompressing and calculating a SHA.

For me, this eats up more time when rebasing on Docker Hub than #165 does. On GCR, over 80% of  real time appears to be spent in the `DiffID` method.

jonjohnsonjr:
@sclevine can you try patching this in and see if it fixes you?

https://github.com/jonjohnsonjr/go-containerregistry/commit/d9e30031b9015b04effafe166823202611786839

This seems to work in my testing -- dropped rebase time from 17s to 6s.

sclevine:
Patch works for me! Also seeing ~6s on GCR.

jonjohnsonjr:
Improvement on dockerhub too? Or just GCR?

sclevine:
It's ~25s faster on Docker Hub for me, but #165 makes it take much longer than 6s. Still a great improvement. 😄

mattmoor:
@sclevine See the linked PR.  I've proposed an alternative to Jon's change that *should* avoid recomputing `DiffIDs` (and a model for augmenting the bare minimum we currently allow the base to provide to `partial`).  I've based this on the PR that fixes mounting against DockerHub, so I'd appreciate it if you could take it for a spin and confirm it fixes all the problems you were seeing.

sclevine:
Sounds great. I'll test both issues against 8eeff01c82cf64317009e4067f14791288ba2678 and report back.

sclevine:
@mattmoor Can confirm that this works. Docker Hub rebases are now ~6s for me. Thanks again!

PR #187, opened 2018-06-01
Avoid recomputing the DiffID of layers where we know it.

mattmoor:
Prior to this change, the `partial` package enabled folks to implement the bare minimum interface necessary to compute the remainder of the `v1.Image` or `v1.Layer` interface.  With this change, we start to adopt a pattern that allows implementors to go beyond this bare minimum and implement additional methods and have the `partial` wrappers respect and delegate to these additional methods.

The first use case of this pattern is in `remote.Layer` where we effectively know all of the layers' `DiffID`s, but because of the "bare minimum" constraint we had previously been unable to leverage this knowledge to elide recomputing the DiffID by downloading and uncompressing the layer.

Fixes: https://github.com/google/go-containerregistry/issues/166

cc @sclevine

### DIFF
--- a/v1/partial/compressed.go
+++ b/v1/partial/compressed.go
@@ -50,6 +50,11 @@ func (ule *compressedLayerExtender) Uncompressed() (io.ReadCloser, error) {
 
 // DiffID implements v1.Layer
 func (ule *compressedLayerExtender) DiffID() (v1.Hash, error) {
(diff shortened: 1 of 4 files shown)


mattmoor:
This is currently based on https://github.com/google/go-containerregistry/pull/186

jonjohnsonjr:
Fantastic.

PR #191, opened 2018-06-02
Fix BASIC challenge

### LINKED ISSUE #189: Auth issue in kaniko -- Unrecognized challenge: BASIC
This [issue](https://github.com/GoogleContainerTools/kaniko/issues/196) was opened in kaniko, and it seems like it's hitting an auth error from [here](https://github.com/google/go-containerregistry/blob/0ff1db03aaaabc9820e8a13353f61d8474814abd/v1/remote/transport/transport.go#L82) (I think it's happening when kaniko is trying to push the final image). It could have something to do with the docker config, just wondering if anyone knew why this might be happening?

jonjohnsonjr:
Java people like to yell, which `strings.Title` doesn't handle well.

Fixes #189


jonjohnsonjr:
RFAL

PR #203, opened 2018-06-04
Update crane append to accept flags, not args

### LINKED ISSUE #200: crane append should accept flags, not positional args
Cobra's generated docs for positional args are bad, and flags make it harder to mess up "src vs dst" ordering bugs anyway. 

Check with existing users whether they'll be broken by this change and whether we should support both for some period while they migrate.

Also document that crane is experimental and likely to change in breaking ways until an official release is cut.

ImJasonH:
Also: 
- mark required flags in append and rebase
- regenerate docs without "Auto generated" section

Fixes #200


pkg/crane/root.go
```diff
@@ -15,7 +15,8 @@ package crane
 import "github.com/spf13/cobra"
 
 var Root = &cobra.Command{
-	Use:   "crane",
-	Short: "Crane is a tool for managing container images",
-	Run:   func(cmd *cobra.Command, _ []string) { cmd.Usage() },
+	Use:               "crane",
+	Short:             "Crane is a tool for managing container images",
+	Run:               func(cmd *cobra.Command, _ []string) { cmd.Usage() },
+	DisableAutoGenTag: true,
```
jonjohnsonjr: !!! :D

pkg/crane/append.go
```diff
@@ -30,17 +30,23 @@ import (
...
 		Run: func(_ *cobra.Command, args []string) {
-			src, dst, tar := args[0], args[1], args[2]
-			doAppend(src, dst, tar, output)
+			doAppend(baseRef, newTag, newLayer, outFile)
 		},
 	}
-	appendCmd.Flags().StringVarP(&output, "output", "o", "", "Path to new tarball of resulting image")
+	appendCmd.Flags().StringVarP(&baseRef, "base", "b", "", "Name of base image to append to")
+	appendCmd.Flags().StringVarP(&newTag, "new_tag", "t", "", "Tag to apply to resulting image")
+	appendCmd.Flags().StringVarP(&newLayer, "new_layer", "f", "", "Path to tarball to append to image")
```
jonjohnsonjr: Why `-f`? As in file?

ImJasonH: Yeah. Open to suggestions if you're opposed.

PR #206, opened 2018-06-06
Adding mutate library methods for reproducible images.

### LINKED ISSUE #188: Best place to put new utility?
I'm working on adding a command to strip away timestamps from docker images to make them able to be built reproducibly, similar to the script we have [here](https://github.com/GoogleContainerTools/base-images-docker/blob/master/util/config_stripper.py) but in Go, and I was wondering where in the containerregistry this would fit best.

sharifelgamal:
Fixes #188

### DIFF
--- a/pkg/v1/mutate/mutate.go
+++ b/pkg/v1/mutate/mutate.go
@@ -387,11 +387,8 @@ func inWhiteoutDir(fileMap map[string]bool, file string) bool {
 	return false
 }
 
-//CREATED is a constant timestamp, set to the 0 time.
-var CREATED = time.Time{}
-
-// Reproducibilify strips out any timestamp from the layers and config of an image.
-func Reproducibilify(img v1.Image) (v1.Image, error) {
+// Time sets all timestamps in an image to the give timestamp.
+func Time(img v1.Image, t time.Time) (v1.Image, error) {
 	newImage := empty.Image
 
 	layers, err := img.Layers()
@@ -402,7 +399,7 @@ func Reproducibilify(img v1.Image) (v1.Image, error) {
 	// Strip away all timestamps from layers
 	var newLayers []v1.Layer
 	for _, layer := range layers {
-		newLayer, err := stripLayer(layer)
+		newLayer, err := layerTime(layer, t)
 		if err != nil {
 			return nil, err
 		}
@@ -421,15 +418,10 @@ func Reproducibilify(img v1.Image) (v1.Image, error) {
 	}
 
 	cfg := cf.DeepCopy()
-	cfg.Created = v1.Time{Time: CREATED}
-
-	cfg.Container = ""
-	cfg.Config.Hostname = ""
-	cfg.ContainerConfig.Hostname = ""
-	cfg.DockerVersion = ""
+	cfg.Created = v1.Time{Time: t}
 
 	for _, h := range cfg.History {
-		h.Created = v1.Time{Time: CREATED}
+		h.Created = v1.Time{Time: t}
 	}
 
 	newImage, err = ConfigFile(newImage, cfg)
@@ -441,7 +433,7 @@ func Reproducibilify(img v1.Image) (v1.Image, error) {
 
 }
 
-func stripLayer(layer v1.Layer) (v1.Layer, error) {
+func layerTime(layer v1.Layer, t time.Time) (v1.Layer, error) {
 	layerReader, err := layer.Uncompressed()
 	if err != nil {
 		return nil, err
@@ -460,7 +452,7 @@ func stripLayer(layer v1.Layer) (v1.Layer, error) {
 			return nil, err
 		}
 
-		header.ModTime = CREATED
+		header.ModTime = t
 		if err := tarWriter.WriteHeader(header); err != nil {
 			return nil, err
 		}
@@ -488,4 +480,29 @@ func stripLayer(layer v1.Layer) (v1.Layer, error) {
 	}
 
 	return layer, nil
+}
+
+//Canonical is a helper function to combine Time and ConfigFile to make a reproducible image
+func Canonical(img v1.Image) (v1.Image, error) {
+	// Set all timestamps to 0
+	created := time.Time{}
+	img, err := Time(img, created)
+	if err != nil {
+		return nil, err
+	}
+
+	cf, err := img.ConfigFile()
+	if err != nil {
+		return nil, err
+	}
+
+	// Get rid of useless random config
+	cfg := cf.DeepCopy()
+
+	cfg.Container = ""
+	cfg.Config.Hostname = ""
+	cfg.ContainerConfig.Hostname = ""
+	cfg.DockerVersion = ""
+
+	return ConfigFile(img, cfg)
 }
--- a/pkg/v1/mutate/mutate.go
+++ b/pkg/v1/mutate/mutate.go
@@ -17,6 +17,7 @@ package mutate
 import (
 	"archive/tar"
 	"bytes"
+	"compress/gzip"
 	"encoding/json"
 	"errors"
 	"fmt"
@@ -189,7 +190,7 @@ func ConfigFile(base v1.Image, cfg *v1.ConfigFile) (v1.Image, error) {
 	return image, nil
 }
 
-// Created mutates the provided v1.Image to have the provided v1.Time
+// CreatedAt mutates the provided v1.Image to have the provided v1.Time
 func CreatedAt(base v1.Image, created v1.Time) (v1.Image, error) {
 	m, err := base.Manifest()
 	if err != nil {
@@ -425,8 +426,8 @@ func inWhiteoutDir(fileMap map[string]bool, file string) bool {
 //CREATED is a constant timestamp, set to the 0 time.
 var CREATED = time.Time{}
 
-// StripConfig strips out any timestamp from the layers and config of an image.
-func StripConfig(img v1.Image) (v1.Image, error) {
+// Reproducibilify strips out any timestamp from the layers and config of an image.
+func Reproducibilify(img v1.Image) (v1.Image, error) {
 	newImage := empty.Image
 
 	layers, err := img.Layers()
@@ -478,7 +479,7 @@ func stripLayer(layer v1.Layer) (v1.Layer, error) {
 	if err != nil {
 		return nil, err
 	}
-	r, w := io.Pipe()
+	w := new(bytes.Buffer)
 	tarWriter := tar.NewWriter(w)
 	defer tarWriter.Close()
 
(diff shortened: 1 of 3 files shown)


pkg/v1/mutate/mutate.go
```diff
@@ -392,3 +393,73 @@ func inWhiteoutDir(fileMap map[string]bool, file string) bool {
...
+		}
+		newLayerName, err := newLayer.Digest()
+		if err != nil {
+			return nil, err
+		}
+
+		layerDescriptor := &v1.Descriptor{
+			MediaType: types.DockerLayer,
+		}
+		newLayers = append(newLayers, *layerDescriptor)
```
dlorenc: I think you may need to make a new image from scratch: https://github.com/google/go-containerregistry/blob/master/pkg/v1/empty/image.go

And then use mutate to append the new stripped layers into it.

Then use mutate.Config to copy the Config into the new image.

pkg/v1/mutate/mutate.go
```diff
@@ -161,6 +162,33 @@ func Config(base v1.Image, cfg v1.Config) (v1.Image, error) {
 	return image, nil
 }
 
+// ConfigFile mutates the provided v1.Image to have the provided v1.ConfigFile
+func ConfigFile(base v1.Image, cfg *v1.ConfigFile) (v1.Image, error) {
```
dlorenc: Do we actually need this - or can we use Config() above and set the ConfigFile.Config?

pkg/v1/mutate/mutate.go
```diff
@@ -392,3 +421,85 @@ func inWhiteoutDir(fileMap map[string]bool, file string) bool {
...
+
+	layers, err := img.Layers()
+	if err != nil {
+		return nil, err
+	}
+
+	// Strip away all timestamps from layers
+	var newLayers []v1.Layer
+	for _, layer := range layers {
+		newLayer, err := stripLayer(layer)
```
jonjohnsonjr: I think calling `stripLayer` here makes `StripConfig` a misleading name.

sharifelgamal: I agree that `StripConfig` is a bad name, I had it there more as a placeholder. Do you have an idea for a better name? `StripTimestamps`?

jonjohnsonjr: `StripTimestamps` is better, but this is also an opportunity to coin a new humorous term that is more fun to say, e.g. `Reproducibilify`, `TimeTravel`, or `Warp`.

pkg/v1/mutate/mutate.go
sharifelgamal: We change more than just ConfigFile.Config, so either we do this or we set the ConfigFile directly in the StripConfig method then call Config() to set everything else correctly. This felt better.

pkg/v1/mutate/mutate.go
dlorenc: The library function name should probably be boring, but I'm +1 on giving it a fun name in crane or some other CLI.

How about stripmetadata for now?

pkg/v1/mutate/mutate.go
```diff
@@ -392,3 +421,85 @@ func inWhiteoutDir(fileMap map[string]bool, file string) bool {
...
+		if err != nil {
+			return nil, err
+		}
+		newLayers = append(newLayers, newLayer)
+	}
+
+	newImage, err = AppendLayers(newImage, newLayers...)
```
dlorenc: Check error

pkg/v1/mutate/mutate.go
```diff
@@ -392,3 +421,85 @@ func inWhiteoutDir(fileMap map[string]bool, file string) bool {
...
+	return newImage, nil
+
+}
+
+func stripLayer(layer v1.Layer) (v1.Layer, error) {
+	layerReader, err := layer.Uncompressed()
+	if err != nil {
+		return nil, err
+	}
+	r, w := io.Pipe()
```
dlorenc: I don't think you want to use a pipe here, a byte buffer should be fine.

pkg/v1/mutate/mutate.go
jonjohnsonjr: It's not necessarily "stripping" the metadata/timestamps.

This reads in a nice way, I think:
```go
frozen, err := mutate.Time(img, time.Now())
```

dlorenc: I like that - but this actually does a bit more than remove times (there are some random numbers in there too we remove). Any thoughts on that?

jonjohnsonjr: I think I'd prefer two methods, then. One to set the time and one to remove various randomness. I would expect clients to compose these to achieve what they want. If I'm reading this correctly, the random numbers are just in the config file, so there shouldn't be any performance issues with splitting it up into two mutations:

```go
frozen, _ := mutate.Time(img, whenever)
happy, _ := mutate.ConfigFile(img, notRandom)
```

If we find a need to use this all over the place, we could have a helper do both of those, maybe:
```go
canon, _ := mutate.Canonicalize(img)
```

Which strips all the metadata and sets the time to something we hard-code. WDYT?

pkg/v1/mutate/mutate.go
```diff
@@ -160,7 +163,34 @@ func Config(base v1.Image, cfg v1.Config) (v1.Image, error) {
 	return image, nil
 }
 
-// Created mutates the provided v1.Image to have the provided v1.Time
+// ConfigFile mutates the provided v1.Image to have the provided v1.ConfigFile
+func ConfigFile(base v1.Image, cfg *v1.ConfigFile) (v1.Image, error) {
```
jonjohnsonjr: There is a lot of code duplication here, it seems. Is it possible to refactor `Config` and `CreatedAt` to use this? (genuine question, I have not given it much thought)

sharifelgamal: I suspect we can yes.

jonjohnsonjr: Looks great, thanks!

pkg/v1/mutate/mutate.go
```diff
@@ -392,3 +386,106 @@ func inWhiteoutDir(fileMap map[string]bool, file string) bool {
 	}
 	return false
 }
+
+//CREATED is a constant timestamp, set to the 0 time.
+var CREATED = time.Time{}
```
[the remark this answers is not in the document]
sharifelgamal: This constant is going away, so that won't be an issue.

pkg/v1/mutate/mutate.go
sharifelgamal: Makes sense, I'll make the switch.

pkg/v1/mutate/mutate.go
sharifelgamal: Done.

pkg/v1/mutate/mutate.go
```diff
@@ -392,3 +386,123 @@ func inWhiteoutDir(fileMap map[string]bool, file string) bool {
...
+	}
+	layer, err = tarball.LayerFromOpener(opener)
+	if err != nil {
+		return nil, err
+	}
+
+	return layer, nil
+}
+
+//Canonical is a helper function to combine Time and ConfigFile to make a reproducible image
```
jonjohnsonjr: Nit: missing space after `//`

pkg/v1/mutate/mutate.go
```diff
@@ -392,3 +386,123 @@ func inWhiteoutDir(fileMap map[string]bool, file string) bool {
...
+	if err != nil {
+		return nil, err
+	}
+
+	cf, err := img.ConfigFile()
+	if err != nil {
+		return nil, err
+	}
+
+	// Get rid of useless random config
```
jonjohnsonjr: Nit: "useless" is a bit harsh here, maybe "non-reproducible" or "host environment-based"

pkg/v1/mutate/mutate.go
```diff
@@ -392,3 +386,123 @@ func inWhiteoutDir(fileMap map[string]bool, file string) bool {
...
+	cfg := cf.DeepCopy()
+
+	cfg.Container = ""
```
jonjohnsonjr: I'm not familiar enough with docker to know for sure that these are useless... in particular I'm surprised that `ContainerConfig.Hostname` isn't needed, but I'll defer to you.

pkg/v1/mutate/mutate.go
```diff
@@ -392,3 +386,123 @@ func inWhiteoutDir(fileMap map[string]bool, file string) bool {
...
+		}
+		newLayers = append(newLayers, newLayer)
+	}
+
+	newImage, err = AppendLayers(newImage, newLayers...)
+	if err != nil {
+		return nil, err
+	}
+
+	// Strip away timestamps and base container info from the config file
```
jonjohnsonjr: Nit: update this comment

pkg/v1/mutate/mutate.go
```diff
@@ -392,3 +386,123 @@ func inWhiteoutDir(fileMap map[string]bool, file string) bool {
 	}
 	return false
 }
+
+// Time sets all timestamps in an image to the give timestamp.
```
jonjohnsonjr: Nit: `to the given timestamp`

pkg/v1/mutate/mutate.go
```diff
@@ -392,3 +386,117 @@ func inWhiteoutDir(fileMap map[string]bool, file string) bool {
...
+			return nil, err
+		}
+
+		header.ModTime = t
+		if err := tarWriter.WriteHeader(header); err != nil {
+			return nil, err
+		}
+
+		if header.Typeflag == tar.TypeReg {
+			_, err = io.Copy(tarWriter, tarReader)
```
ImJasonH: A supernit, but I tend to prefer `if _, err := io.Copy(...); err != nil { ... }` because it makes it clear that the error doesn't matter outside of this block. It's small, but it adds up.

pkg/v1/mutate/mutate.go
```diff
@@ -392,3 +386,117 @@ func inWhiteoutDir(fileMap map[string]bool, file string) bool {
...
+	}
+	layer, err = tarball.LayerFromOpener(opener)
+	if err != nil {
+		return nil, err
+	}
+
+	return layer, nil
+}
+
+// Canonical is a helper function to combine Time and ConfigFile to make a reproducible image
```
ImJasonH: Might help to define "reproducible" more clearly here. An image might not be reproducible just by stripping its timestamps etc out.

pkg/v1/mutate/mutate.go
```diff
@@ -128,22 +133,27 @@ func Append(base v1.Image, adds ...Addendum) (v1.Image, error) {
...
 	}
 
-	cf, err := base.ConfigFile()
+	cf.Config = cfg
+
+	return ConfigFile(base, cf)
+}
+
+// ConfigFile mutates the provided v1.Image to have the provided v1.ConfigFile
+func ConfigFile(base v1.Image, cfg *v1.ConfigFile) (v1.Image, error) {
```
mattmoor: We used Config in place of ConfigFile because it contained information the user might want to tweak, and getting it wrong made it a lot harder to screwup the image (e.g. it has diff ids).

Can you explain how you expect this to be (safely) used?  If it's for use within this package, let's make it private.

sharifelgamal: making it private is probably the safest move, especially since I added the Canonical wrapper, which is what kaniko is going to call.

sharifelgamal:
If this is ready, feel free to merge.

Issue #205, opened 2018-06-06
`daemon.Write` implementation is pretty naive


mattmoor:
[here](https://github.com/google/go-containerregistry/blob/master/pkg/v1/daemon/write.go#L49)

Today the `daemon.Write` interface just uses `tarball.Write` into the `docker load` interface.  While correct, this can be incredibly slow for scenarios like `ko -L` because of its lack of incrementality.

In particular, for a large base image we produce and stream a fat tarball to the daemon for every publish.  On top of this, since we don't have a local cache wrapping `remote.Image`, we download the base image every time.  `remote.Write` elides both upload *and* download through the careful use of existence checks.  We should be equally careful in `daemon.Write`.

One option to explore for this is what `rules_docker` did in its [incremental load script](https://github.com/bazelbuild/rules_docker/blob/master/container/incremental_load.sh.tpl).  However, we should be careful to measure how this performs on a full daemon (we've seen superlinear behavior in some of the daemon calls before).

mattmoor:
cc @dlorenc not sure if/what of your portfolio of tools might be using `daemon.Write`.

This would certainly be mitigated by the local caching we've been talking about, which feels relevant to kaniko + `FROM` caching, but ideally we'd just never access the content from layers present in the daemon.

mattmoor:
I think that this could take advantage of the recently added `l.(*remote.Layer)` trick to direct the daemon to pull the image itself (vs. having us side-load it).

Then we'd want to add the capacity to omit layers from the tarball via a callback [here](https://github.com/google/go-containerregistry/blob/master/pkg/v1/tarball/write.go#L76).

This should avoid potential pathological behavior in the daemon, which I believe would otherwise require us to:
1. Enumerate all images in the daemon
1. Enumerate all the diff-ids for each image

mattmoor:
Correction: `l.(*remote.MountableLayer)`.  We would also have to generalize this to capture the full `name.Reference` instead of just the `name.Repository`.

ekcasey:
@jonjohnsonjr We also create a daemon image using the `docker load` interface.

However, we know that we are extending a base that already exists in the daemon and therefore we can do a hack. We discovered we could load a tarball that omits layer blobs when a layer with the same chain ID already exists in the daemon. Therefore, when creating images we load tarballs with a `manifest.json` that look like this example:
```
[
  {
[...]

PR #209, opened 2018-06-10
Add the option to exclude certain layers from the tarball.

mattmoor:
This should be considered a relatively advanced option, but for folks that know
what they are doing you can reduce the amount of data that you need to encode in
the tarball for the daemon to load it.

The ultimate use case of this option will be from `daemon.Write`, which
currently uses the `docker load` interface to pull image into the daemon,
however, this currently reuploads (and redownloads) the base image on each write
in context like `ko`. If we can determine the set of layers that already exist
in the daemon we can elide these from the tarball to dramatically improve
performance.

Related: #205

### DIFF
--- a/pkg/v1/tarball/write.go
+++ b/pkg/v1/tarball/write.go
@@ -26,9 +26,15 @@ import (
 	"github.com/google/go-containerregistry/pkg/v1"
 )
 
+// LayerFilter defines a function for filtering layers.
+//  True - indicates the layer should be kept,
+//  False - indicates the layer should be excluded.
+type LayerFilter func(v1.Layer) (bool, error)
+
 // WriteOptions are used to expose optional information to guide or
 // control the image write.
 type WriteOptions struct {
+	LayerFilter LayerFilter
 	// TODO(mattmoor): Whether to store things compressed?
 }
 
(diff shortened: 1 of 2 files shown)


pkg/v1/tarball/write.go
```diff
@@ -89,6 +105,15 @@ func Write(tag name.Tag, img v1.Image, wo *WriteOptions, w io.Writer) error {
 		// https://www.gnu.org/software/gzip/manual/html_node/Overview.html
 		layerFiles[i] = fmt.Sprintf("%s.tar.gz", hex)
 
+		// We filter late because the lenght of layerFiles must match the diff_ids
```
jonjohnsonjr: nit: `length`

pkg/v1/tarball/write.go
```diff
@@ -66,6 +72,16 @@ func Write(tag name.Tag, img v1.Image, wo *WriteOptions, w io.Writer) error {
 		return err
 	}
 
+	// Default filter function
```
jonjohnsonjr: Maybe hoist this out into its own method?

pkg/v1/tarball/write.go
```diff
@@ -66,6 +72,16 @@ func Write(tag name.Tag, img v1.Image, wo *WriteOptions, w io.Writer) error {
...
+	filter := func(_ v1.Layer) (bool, error) {
+		return true, nil
+	}
+	if wo != nil {
+		if wo.LayerFilter != nil {
```
jonjohnsonjr: Collapse this into `if wo != nil && wo.LayerFilter != nil {` ?

Though I guess that makes it harder to add more options... up to you.

mattmoor:
@jonjohnsonjr I hear you, but at the same time, I'm not sure of a good way to do this closer to `daemon.Write` that doesn't also involve cloning a significant portion of `tarball.Write`.

dlorenc:
can you rebase this?

mattmoor:
At this point, it's more work to rebase than redo.  @jonjohnsonjr are you up for it?

PR #210, opened 2018-06-12
compressing layers for reproducible images

sharifelgamal:


jonjohnsonjr:
Maybe add a test for layertime?

Issue #213, opened 2018-06-14
ko needs a better way of generating temp file names


mattmoor:
*This is circumstantial evidence from a bug report, IDK that this is a real problem*

Here's the (redacted) output:
```
ko apply -f something.yaml
2018/06/14 13:59:22 Go building github.com/something/something
2018/06/14 13:59:22 Unexpected error running "go build": exit status 1
go build github.com/something/something: build output "/var/folders/.../T/out743897893" already exists and is not an object file
2018/06/14 13:59:22 error processing import paths in "something.yaml": exit status 1
```

We use [`ioutil.TempFile`](https://golang.org/pkg/io/ioutil/#TempFile) to produce the name of a file to which the `go build -o` should go [here](https://github.com/google/go-containerregistry/blob/d5ae41de3d3c22d36b735f5d5e3fb42d1ba32f25/pkg/ko/build/gobuild.go#L93).

This error is from a place in Go that is paranoid about overwriting files [here](https://go.googlesource.com/go/+/dev.ssa/src/cmd/go/build.go#1785).

I wonder if `ioutil.TempFile` is creating the empty file, and Go is seeing this and freaking out, but I'm puzzled by the fact that we don't hit this everywhere.

mattmoor:
cc @glyn

PR #214, opened 2018-06-14
Avoid creating build output file

glyn:
Create a new build directory for each file and use a fixed file name but don't
open the file (since that sometimes causes go build to fail).

Fixes https://github.com/google/go-containerregistry/issues/213

### DIFF
--- a/pkg/ko/build/gobuild.go
+++ b/pkg/ko/build/gobuild.go
@@ -90,12 +90,14 @@ func (gb *gobuild) IsSupportedReference(s string) bool {
 }
 
 func build(ip string) (string, error) {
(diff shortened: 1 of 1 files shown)


pkg/ko/build/gobuild.go
```diff
@@ -90,12 +90,14 @@ func (gb *gobuild) IsSupportedReference(s string) bool {
 }
 
 func build(ip string) (string, error) {
-	file, err := ioutil.TempFile(os.TempDir(), "out")
+	tmpDir, err := ioutil.TempDir("", "ko")
```
mattmoor: that's a neat trick to avoid the `Open` :D

PR #216, opened 2018-06-21
preserve config for reproducible images

sharifelgamal:
@dlorenc


pkg/v1/mutate/mutate.go
```diff
@@ -411,13 +411,23 @@ func Time(img v1.Image, t time.Time) (v1.Image, error) {
...
+	}
+
 	cf, err := newImage.ConfigFile()
 	if err != nil {
 		return nil, fmt.Errorf("Error setting config file: %v", err)
 	}
 
 	cfg := cf.DeepCopy()
+
+	cfg.Config = ocf.Config
```
jonjohnsonjr: I'd put `// Copy basic config over` here instead.

jonjohnsonjr:
Thanks!

PR #219, opened 2018-07-04
cross-repo blob mount should be used only when the layer comes from the same registry

everpeace:
Hi!  Thank you for maintaining great project!!

`go-containerregistry` uses [Cross Repository Blob Mount](https://docs.docker.com/registry/spec/api/#cross-repository-blob-mount) in [`writer.initiateUpload(h)`](https://github.com/google/go-containerregistry/blob/3f6471078a9661a9a439bd5e71a371aff429566a/pkg/v1/remote/write.go#L143-L149) even when the mountable layer comes from a different registry.  This would cause `401 Unauthorized` error.  It can easily happen when users try to push the layer to their own private registry.

Although I'm not so familiar with docker registry API,  I think cross repo blob mount should be used only when the mountable layer comes from the SAME registry with the target registry.

### DIFF
--- a/pkg/v1/remote/write.go
+++ b/pkg/v1/remote/write.go
@@ -145,7 +145,9 @@ func (w *writer) initiateUpload(h v1.Hash) (location string, mounted bool, err e
 	// if "mount" is specified, even if no "from" sources are specified.  If this turns out
 	// to not be broadly applicable then we should replace mounts without "from"s with a HEAD.
 	if ml, ok := l.(*MountableLayer); ok {
-		uv["from"] = []string{ml.Reference.Context().RepositoryStr()}
+		if w.ref.Context().RegistryStr() == ml.Reference.Context().RegistryStr() {
+			uv["from"] = []string{ml.Reference.Context().RepositoryStr()}
+		}
 	}
 	u.RawQuery = uv.Encode()
 
--- a/pkg/v1/remote/write_test.go
+++ b/pkg/v1/remote/write_test.go
@@ -129,6 +129,10 @@ func mustConfigName(t *testing.T, img v1.Image) v1.Hash {
 
 func setupWriter(repo string, img v1.Image, handler http.HandlerFunc) (*writer, closer, error) {
 	server := httptest.NewServer(handler)
+	return setupWriterWithServer(server, repo, img)
+}
+
+func setupWriterWithServer(server *httptest.Server, repo string, img v1.Image) (*writer, closer, error) {
 	u, err := url.Parse(server.URL)
 	if err != nil {
 		server.Close()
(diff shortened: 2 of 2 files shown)


everpeace:
@jonjohnsonjr @mattmoor would you mind reviewing this??  Thanks in advance 🙇

pkg/v1/remote/write.go
```diff
@@ -145,7 +145,9 @@ func (w *writer) initiateUpload(h v1.Hash) (location string, mounted bool, err e
 	// if "mount" is specified, even if no "from" sources are specified.  If this turns out
 	// to not be broadly applicable then we should replace mounts without "from"s with a HEAD.
 	if ml, ok := l.(*MountableLayer); ok {
-		uv["from"] = []string{ml.Reference.Context().RepositoryStr()}
+		if w.ref.Context().RegistryStr() == ml.Reference.Context().RegistryStr() {
+			uv["from"] = []string{ml.Reference.Context().RepositoryStr()}
```
everpeace: I think it's safe only when the layer comes from _the same_ registry with the target registry.

jonjohnsonjr:
Hi, thanks for your PR! I think there's a better approach to fixing this, but let me know if I'm misunderstanding anything 👍 

> go-containerregistry uses Cross Repository Blob Mount in writer.initiateUpload(h) even when the mountable layer comes from a different registry. This would cause 401 Unauthorized error. It can easily happen when users try to push the layer to their own private registry.

From the spec:
> If a mount fails due to invalid repository or digest arguments, the registry will fall back to the standard upload behavior and return a 202 Accepted

IMO, It sounds like your registry isn't following the spec. Are you using docker/distribution or something else?

> Although I'm not so familiar with docker registry API, I think cross repo blob mount should be used only when the mountable layer comes from the SAME registry with the target registry.

I think that's generally true, but by eagerly attempting to mount, a registry implementation can opportunistically dedupe. See this comment: https://github.com/opencontainers/distribution-spec/issues/18#issuecomment-400765775

I would consider this a registry implementation bug rather than a bug in this library. The registry should be robust to a source repository being deleted or changing ACLs. If it's impossible to fix this in your registry implementation, I'd rather see us improve our handling of fallback logic. What do you think?

mattmoor:
FWIW @jonjohnsonjr I'm fine with this, given where mounting landed in `docker/distribution`.

I'll spare you the history lesson and my sour grapes about where it landed :)

I am also curious what registry this is, for future reference.

everpeace:
> LGTM -- if you fix the typo I will merge :)

Thanks!  I pushed the fix.

jonjohnsonjr:
> I'm not sure this is the problem of harbor's implemetation or not.
> I would be very appreciated if you would help me to understand root cause clearer.

I'd say this is another case of the docker CLI becoming an implicit part of the registry spec :) In my opinion, harbor's handling of this is a bug, since [the spec indicates they should just fall back to normal upload behavior by returning a 202](https://docs.docker.com/registry/spec/api/#cross-repository-blob-mount).

With your change, I think we now have similar behavior to docker, but we probably want to be more robust in our error handling here regardless.

Thanks again for the fix, @everpeace 😄

PR #217, opened 2018-07-04
Rip bazel out

jonjohnsonjr:
I cannot understand why, but running the ko tests breaks my editor. When
it tries to run goimports on save, it breaks iff I run the ko unit
tests. Every time. I think the build package is *not* safe :)

Output:

```
$ go test ./pkg/... -count=1
ok  	github.com/google/go-containerregistry/pkg/authn	0.034s
?   	github.com/google/go-containerregistry/pkg/crane	[no test files]
?   	github.com/google/go-containerregistry/pkg/ko	[no test files]
ok  	github.com/google/go-containerregistry/pkg/ko/build	0.031s
open /tmp/go-build243642577/b184/vet.cfg: no such file or directory
go build github.com/google/go-containerregistry/pkg/v1 (testmain): open /tmp/go-build243642577/b185/importcfg: no such file or directory
go build github.com/google/go-containerregistry/pkg/v1/daemon (testmain): open /tmp/go-build243642577/b188/importcfg: no such file or directory
open /tmp/go-build243642577/b182/name.test: no such file or directory
go build github.com/google/go-containerregistry/pkg/v1/empty (testmain): open /tmp/go-build243642577/b191/importcfg: no such file or directory
go build github.com/google/go-containerregistry/pkg/v1/mutate (testmain): open /tmp/go-build243642577/b194/importcfg: no such file or directory
open /tmp/go-build243642577/b077/vet.cfg: no such file or directory
go build github.com/google/go-containerregistry/pkg/v1/random (testmain): open /tmp/go-build243642577/b197/importcfg: no such file or directory
go build github.com/google/go-containerregistry/pkg/v1/remote (testmain): open /tmp/go-build243642577/b200/importcfg: no such file or directory
open /tmp/go-build243642577/b137/publish.test: no such file or directory
FAIL	github.com/google/go-containerregistry/pkg/ko/publish [build failed]
go build github.com/google/go-containerregistry/pkg/v1/remote/transport (testmain): open /tmp/go-build243642577/b203/importcfg: no such file or directory
go build github.com/google/go-containerregistry/pkg/v1/tarball (testmain): open /tmp/go-build243642577/b206/importcfg: no such file or directory
open /tmp/go-build243642577/b062/vet.cfg: no such file or directory
go build github.com/google/go-containerregistry/pkg/v1/v1util (testmain): open /tmp/go-build243642577/b209/importcfg: no such file or directory
open /tmp/go-build243642577/b170/resolve.test: no such file or directory
[...]

PR #225, opened 2018-07-12
Make cmd/crane library target public

pcj:
This makes building cross-compiled binaries of crane easier

### DIFF
--- a/cmd/crane/BUILD.bazel
+++ b/cmd/crane/BUILD.bazel
@@ -6,7 +6,7 @@ go_library(
         "main.go",
     ],
     importpath = "github.com/google/go-containerregistry/cmd/crane",
-    visibility = ["//visibility:private"],
+    visibility = ["//visibility:public"],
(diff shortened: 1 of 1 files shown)


ImJasonH:
FWIW Bazel targets in general are on their way out: https://github.com/google/go-containerregistry/pull/217

mattmoor:
... which means vendoring this and Gazelle should just work after @jonjohnsonjr's change lands.

Gazelle looooooves making stuff public.

pcj:
Actually, no.  Gazelle didn't make the targets public.

mattmoor:
@pcj that seems strange, Gazelle makes like everything public.  Do you have a broken example published somewhere I could look at?

pcj:
I created an example:

https://github.com/pcj/go-containerregistry-gazelle-example

pcj:
Perhaps I can use the overlay feature of bazel-gazelle `git_repository` rule...

Issue #197, opened 2018-06-03
ko apply shouldn't need to be run inside GOPATH


ImJasonH:
It doesn't make sense to say:

```yaml
image: ./cmd/foo
```

It only makes sense IMO to include full importpaths in YAML, at which point we can run `ko apply` outside of GOPATH and have it assume importpaths are at `$GOPATH/src/$importpath`

For `ko publish` it probably makes sense to be less strict (#190), and if given a relative path, assume it's being run inside GOPATH.

This also allows `ko` (excluding `publish` with relative importpaths) to be run outside of `GOPATH`, which would enable us to easily produce a builder image for use in GCB.

[My fork's `kobin` branch](https://github.com/ImJasonH/go-containerregistry/tree/kobin?files=1) has this change, and also excludes building images from non-`package main` packages, but is broken by some interaction of the `go/build.Default.GOPATH` and Bazel. If we dropped Bazel for tests (#198) I believe we could merge it (relevant tests pass with `go test`).

jonjohnsonjr:
> but is broken by some interaction of the go/build.Default.GOPATH and Bazel

I wonder if this is fixed by #217

ImJasonH:
Closed in #229

PR #229, opened 2018-07-16
Use go/build.Import to resolve import paths

### LINKED ISSUE #215: ko should check IsCommand on import paths
See [here](https://golang.org/pkg/go/build/#Package.IsCommand).

Something like:
```go
import "go/build"

p, err := build.Import(importpath, srcDir, 0)
// check err
if !p.IsCommand() {
  // Return error.
}

```

ImJasonH:
Fixes #215 and #197 

With this change, only import paths that represent runnable commands (i.e., `package main`) are considered "supported". As a reminder, attempting to `ko publish` an unsupported path will fail, and unsupported paths are simply ignored when `ko resolve`ing.

This change also allows `ko` to be run outside of `GOPATH`, and requires import paths to be fully qualified -- `image: ./cmd/foo` will not `resolve`, it must be `image: github.com/my/repo/cmd/foo`

### DIFF
--- a/pkg/ko/build/gobuild_test.go
+++ b/pkg/ko/build/gobuild_test.go
@@ -35,6 +35,7 @@ func TestGoBuildIsSupportedRef(t *testing.T) {
 	for _, importpath := range []string{
 		filepath.FromSlash("github.com/google/go-containerregistry/cmd/crane"),
 		filepath.FromSlash("github.com/google/go-containerregistry/cmd/ko"),
+		filepath.FromSlash("github.com/google/go-containerregistry/vendor/k8s.io/code-generator/cmd/deepcopy-gen"), // vendored commands work too.
 	} {
 		t.Run(importpath, func(t *testing.T) {
 			if !ng.IsSupportedReference(importpath) {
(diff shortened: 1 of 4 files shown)


pkg/ko/build/gobuild_test.go
```diff
@@ -28,116 +25,32 @@ import (
...
-	}
-
-	for _, test := range supportedTests {
-		t.Run(test, func(t *testing.T) {
-			if !ng.IsSupportedReference(test) {
-				t.Errorf("IsSupportedReference(%v) = false, want true", test)
+	// Supported import paths.
+	for _, importpath := range []string{
+		filepath.FromSlash("github.com/google/go-containerregistry/cmd/crane"),
+		filepath.FromSlash("github.com/google/go-containerregistry/cmd/ko"),
```
mattmoor: Can you check something in the vendor directory, even if you have to force the inclusion of it in `Gopkg.toml`?  Maybe `deepcopy-gen`?

pkg/ko/build/gobuild_test.go
```diff
@@ -28,116 +25,33 @@ import (
...
+		filepath.FromSlash("github.com/google/go-containerregistry/vendor/k8s.io/code-generator/cmd/deepcopy-gen"), // vendored commands work too.
```
mattmoor: Can this work without the prefix thru `.../vendor/`?

ImJasonH: Maybe with a relative local path? (#231)

Issue #190, opened 2018-06-02
ko publish should support relative import paths


mattmoor:
I'd love to be able to say `ko publish ./cmd/foo` like I can say `go build ./cmd/foo`.

cc @ImJasonH

jonjohnsonjr:
What about `ko publish ./...` ?

ImJasonH:
If we support that we should probably only build `package main` packages, otherwise you're going to build a lot of useless images.

mattmoor:
I think this is done.

ImJasonH:
The only thing we don't have is `./...`, I'm not sure how important that is or if it's just nice-to-have. If somebody wants it we can open a new issue

PR #231, opened 2018-07-16
Allow relative paths in ko publish

ImJasonH:
Begins to address #190, based on work in #229 

With this change, if your working directory is within `$GOPATH/src`, you can run `ko publish ./some/local/import/path` (assuming that path is valid and points to a `package main`).

This does not yet support `ko publish ./...`, but this is a prerequisite.

### DIFF
--- a/cmd/ko/publish.go
+++ b/cmd/ko/publish.go
@@ -46,8 +46,8 @@ func publishImages(importpaths []string, lo *LocalOptions) {
 	}
 	for _, importpath := range importpaths {
 		if gb.IsLocalImport(importpath) {
-			// Interpret `ko publish ./cmd/foo` as `ko publish $GOPATH/cmd/foo`
-			// $PWD must be within $GOPATH/src
+			// Qualify relative imports to their fully-qualified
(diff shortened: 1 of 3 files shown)


cmd/ko/publish.go
```diff
@@ -15,22 +15,54 @@
...
+}
+
 func publishImages(importpaths []string, lo *LocalOptions) {
 	b, err := build.NewGo(gobuildOptions())
 	if err != nil {
 		log.Fatalf("error creating go builder: %v", err)
 	}
 	for _, importpath := range importpaths {
+		if gb.IsLocalImport(importpath) {
+			// Interpret `ko publish ./cmd/foo` as `ko publish $GOPATH/cmd/foo`
```
jonjohnsonjr: Is this comment correct? `$GOPATH/cmd/foo` seems unlikely.

ImJasonH: It was not. Updated.

cmd/ko/publish_test.go
```diff
@@ -0,0 +1,35 @@
+package main
```
mattmoor: license

PR #232, opened 2018-07-16
Add a mechanism for composing keychains.

mattmoor:
This enables us to change the fallback behavior of keychains from Anonymous to another keychain.

### DIFF
--- a/pkg/authn/multikeychain.go
+++ b/pkg/authn/multikeychain.go
@@ -1,0 +1,45 @@
+// Copyright 2018 Google LLC All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package authn
+
+import (
+	"github.com/google/go-containerregistry/pkg/name"
+)
+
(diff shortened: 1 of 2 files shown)


pkg/authn/multikeychain.go
```diff
@@ -0,0 +1,45 @@
...
+}
+
+// Resolve implements Keychain.
+func (mk *multiKeychain) Resolve(reg name.Registry) (Authenticator, error) {
+	for _, kc := range mk.keychains {
+		auth, err := kc.Resolve(reg)
+		if err != nil {
+			return nil, err
+		}
+		if auth != Anonymous {
```
ImJasonH: This may lead to confusing log messages saying "Some error happened, falling back on anonymous" that aren't actually true. Does it make sense to move that fallback message to here instead?

jonjohnsonjr: Or just remove it entirely :)

ImJasonH: I dunno, I think it's pretty helpful when debugging to see when you're about to do something anonymously. For instance, if you're trying to push an image and it fails with "You don't have access", you'd want to know that's because you fell back to anonymous credentials.

Maybe the best place for the warning is in [`WithAuthFromKeychain`](https://github.com/google/go-containerregistry/blob/f02501e1d74bff68d6e2d8962ea61277ea628b5d/pkg/v1/remote/options.go#L41), where it'll be authoritative about whether you're anonymous or not.

It's probably _not_ a good idea to log in the library itself, only in our commands.

mattmoor: @ImJasonH I agree, but saw fixing the logging as an orthogonal PR.  This logging is clearly presumptive about the context in which it is being used, and probably doesn't belong in a library.

`WithAuthFromKeychain` is one place it might go, but perhaps having a variant of that with logging is most appropriate.  WDYT?

mattmoor: Also +1 on usefulness

PR #235, opened 2018-07-17
This migrates the `publish.NewDefault` method to functional options.

mattmoor:


pkg/ko/publish/options.go
```diff
@@ -0,0 +1,61 @@
...
+// Set client on image using provided transport, and the default authenticator
+func (i *defaultOpener) setTransport(t http.RoundTripper) error {
+	i.t = t
+	return nil
+}
+
+// Set client on image using provided authenticator, and the default transport
+func (i *defaultOpener) setAuth(auth authn.Authenticator) error {
+	i.auth = auth
+	return nil
```
ImJasonH: These don't need to `return nil`, I think it's simpler to just do that above in the returned func.

ImJasonH: You probably don't even need `setAuth`? Just `i.auth = auth; return nil` above in `WithAuth`

mattmoor: Should be RFAL

pkg/ko/publish/options.go
```diff
@@ -0,0 +1,62 @@
...
+		if err != nil {
+			return err
+		}
+		i.setAuth(auth)
+		return nil
+	}
+}
+
+// Set client on image using provided transport, and the default authenticator
+func (i *defaultOpener) setTransport(t http.RoundTripper) {
```
ImJasonH: I think even `setTransport` and `setAuth` are pretty useless, and could just be removed, but I'll leave that up to you.

PR #237, opened 2018-07-18
remove duplicate scope string from token handshake

### LINKED ISSUE #221: Remote should dedupe scopes for token handshake
@everpeace pointed this out in #219 

We append a scope for each layer, which might be a bit too much. Many registries handle multiple scopes poorly, but others (chiefly dockerhub) require multiple scopes for cross-repo mounting to work.

See here: https://github.com/google/go-containerregistry/blob/3f6471078a9661a9a439bd5e71a371aff429566a/pkg/v1/remote/write.go#L45-L49

everpeace:
fixes #221

### DIFF
--- a/pkg/v1/remote/write.go
+++ b/pkg/v1/remote/write.go
@@ -42,13 +42,8 @@ func Write(ref name.Reference, img v1.Image, auth authn.Authenticator, t http.Ro
 	if err != nil {
 		return err
 	}
-	scopes := []string{ref.Scope(transport.PushScope)}
-	for _, l := range ls {
-		if ml, ok := l.(*MountableLayer); ok {
-			scopes = append(scopes, ml.Reference.Context().Scope(transport.PullScope))
-		}
-	}
 
+	scopes := scopesForUploadingImage(ref, ls)
 	tr, err := transport.New(ref.Context().Registry, auth, t, scopes)
 	if err != nil {
 		return err
@@ -294,6 +289,24 @@ func (w *writer) commitImage() error {
 	// The image was successfully pushed!
 	log.Printf("%v: digest: %v size: %d", w.ref, digest, len(raw))
 	return nil
+}
+
+func scopesForUploadingImage(ref name.Reference, layers []v1.Layer) []string {
+	// use a map as set to remove duplicates scope strings
+	scopeSet := map[string]struct{}{ref.Scope(transport.PushScope): {}}
+
+	for _, l := range layers {
+		if ml, ok := l.(*MountableLayer); ok {
+			scopeSet[ml.Reference.Context().Scope(transport.PullScope)] = struct{}{}
+		}
+	}
+
+	scopes := make([]string, 0)
+	for scope, _ := range scopeSet {
+		scopes = append(scopes, scope)
+	}
+
+	return scopes
 }
 
 // TODO(mattmoor): WriteIndex
--- a/pkg/v1/remote/write.go
+++ b/pkg/v1/remote/write.go
@@ -293,15 +293,21 @@ func (w *writer) commitImage() error {
 
 func scopesForUploadingImage(ref name.Reference, layers []v1.Layer) []string {
 	// use a map as set to remove duplicates scope strings
-	scopeSet := map[string]struct{}{ref.Scope(transport.PushScope): {}}
+	scopeSet := map[string]struct{}{}
 
 	for _, l := range layers {
 		if ml, ok := l.(*MountableLayer); ok {
-			scopeSet[ml.Reference.Context().Scope(transport.PullScope)] = struct{}{}
+			// we add push scope for ref.Context() after the loop
+			if ml.Reference.Context() != ref.Context() {
+				scopeSet[ml.Reference.Context().Scope(transport.PullScope)] = struct{}{}
+			}
 		}
 	}
 
 	scopes := make([]string, 0)
+	// Push scope should be the first element because a few registries just look at the first scope to determine access.
+	scopes = append(scopes, ref.Scope(transport.PushScope))
+
 	for scope, _ := range scopeSet {
 		scopes = append(scopes, scope)
 	}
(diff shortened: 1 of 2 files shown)


everpeace:
@jonjohnsonjr I'm working on the issue but I'm wondering how to test this.  Because I'm not so familiar with go-containerregistry data structures.  Could you give me some advice??

I extract the function [`scopesForUploadingImage`](https://github.com/everpeace/go-containerregistry/blob/dedup-scopes-for-token-handshake/pkg/v1/remote/write.go#L293-L313) to return scope strings.  This would be easier to test.

jonjohnsonjr:
@everpeace awesome, thanks for the PR!

For testing, I'd just change `scopesForUploadingImage` to take an array of layers instead of a `v1.Image`, e.g.:

```go
func scopesForUploadingImage(ref name.Reference, layers []v1.Layer) []string {
    // your code here
}
```

(note that you shouldn't need to return an error anymore, either 😄)

Then you can just pass in an array of various `MountableLayer`s. This would definitely lend itself to some [table driven tests](https://github.com/golang/go/wiki/TableDrivenTests), since it's a simple transformation between data structures.

@mattmoor any better ideas?

everpeace:
@jonjohnsonjr Thanks for your great advice!  I changed the interface of `scopesForUploadingImage` and wrote tests.  Would you mind if I asked you to review??

pkg/v1/remote/write_test.go
```diff
@@ -606,3 +608,139 @@ func TestWriteWithErrors(t *testing.T) {
 		t.Errorf("Write(); (-want +got) = %s", diff)
 	}
 }
+
+func TestScopesForUploadingImage(t *testing.T) {
+	serverHandler := http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
```
jonjohnsonjr: It looks like you're spinning up a fake server just to get a valid URL. I'd remove these lines and just hard-code the references to be, e.g.:
`example.com/sample/sample:latest`
`example.com/sample/another_repo1:latest`
`example.com/sample/another_repo2:latest`

I think that will be easier to follow, since the string concatenation obfuscates what's going on a bit.

pkg/v1/remote/write.go
```diff
@@ -296,4 +291,22 @@ func (w *writer) commitImage() error {
 	return nil
 }
 
+func scopesForUploadingImage(ref name.Reference, layers []v1.Layer) []string {
+	// use a map as set to remove duplicates scope strings
+	scopeSet := map[string]struct{}{ref.Scope(transport.PushScope): {}}
```
jonjohnsonjr: So this is kind of annoying (and not your fault), but I believe we want the `PushScope` to be the first scope due to how some registries are implemented. I believe a few registries just look at the first scope to determine access (this is a bug), and since PushScope is the broadest (it has both `push,pull`), we need it to be first in order to have write access.

So tl;dr, we want the push scope to be first, but we don't care about the order of any of the other scopes (just read access).

pkg/v1/remote/write_test.go
```diff
@@ -606,3 +608,139 @@ func TestWriteWithErrors(t *testing.T) {
...
+				referenceToUpload.Scope(transport.PushScope),
+				anotherRepo1.Scope(transport.PullScope),
+				anotherRepo2.Scope(transport.PullScope),
+			},
+		},
+	}
+
+	for _, testCase := range testCases {
+		actual := scopesForUploadingImage(testCase.reference, testCase.layers)
+		if !reflect.DeepEqual(sort.StringsAreSorted(actual), sort.StringsAreSorted(testCase.expected)) {
```
jonjohnsonjr: I think you want `sort.Strings` instead of `sort.StringsAreSorted`, but I'd probably use the [`cmpopts.SortSlices`](https://godoc.org/github.com/google/go-cmp/cmp/cmpopts#SortSlices) instead.

Per my other comment, I think you want to assert that the first element is what you expect (the push scope)... something like this:
```go
if want, got := testCase.expected[0], actual[0]; want != got {
  t.Errorf("Wrong first scope; want %v, got %v", want, got)
}
if diff := cmp.Diff(testCase.expected[1:], actual[1:], cmpopts.SortSlices()); diff != "" {
  t.Errorf("Wrong scopes (-want +got) = %v", diff)
}
```

I haven't tested that that even compiles, but it should be approximately right :P

pkg/v1/remote/write_test.go
```diff
@@ -606,3 +608,139 @@ func TestWriteWithErrors(t *testing.T) {
...
+			},
+			expected: []string{
+				referenceToUpload.Scope(transport.PushScope),
+				anotherRepo1.Scope(transport.PullScope),
+				anotherRepo2.Scope(transport.PullScope),
+			},
+		},
+	}
+
+	for _, testCase := range testCases {
```
jonjohnsonjr: I'd probably shorten `testCase` to `tc` just to reduce line length a bit.

everpeace:
@jonjohnsonjr Hi, sorry for my absense.  I updated my branch by following your comments.
But travis failed..  Test passed on my local as below.  I would be appriciated if you see some problems 🙇 

```
$ git rev-parse HEAD
ddd7874984fddf740f6bf1a5abcd0d55cfabf0e0

$ find . -name "*.go" | grep -v vendor/ | xargs gofmt -d -e -l

$ echo $?
0

$ go clean -i

$ go test -v -race -coverprofile=coverage.txt -covermode=atomic ./...
...
--- PASS: TestWriter (0.01s)
PASS
coverage: 74.0% of statements
ok      github.com/google/go-containerregistry/pkg/v1/v1util    1.086s  coverage: 74.0% of statements

$ echo $?
0
```

jonjohnsonjr:
@everpeace looks like this change removed the import of `transport`: https://github.com/google/go-containerregistry/commit/b0936d62e7942acf7b1b5d4c377caa5f0b7566a5#diff-b9d51c3f473efa8b6f285db706147bf0

Failure is here on travis: https://travis-ci.org/google/go-containerregistry/builds/409687465#L608

If you rebase onto master and add that import back, it should pass 👍

everpeace:
@jonjohnsonjr Thanks for pointing out!  I fixed my code and rebased to latest master.  Please take a look again 🙇

Issue #160, opened 2018-05-23
Support flattening importpaths in ko


mattmoor:
`ko` doesn't currently support DockerHub (and possibly other registries) because they lack support for multi-level names.

Consider adding support for flattening import paths to form image names under `KO_DOCKER_REPO`, which would plug in (somehow) [here](https://github.com/google/go-containerregistry/blob/master/cmd/ko/publish.go#L48)

Either `NewDefault` would construct a non-default publisher, or we'd pass some options through to direct the publisher to construct flat names.

mattmoor:
I think I want to switch `ko` over to functional options before tackling this.

PR #240, opened 2018-07-19
Change the default behavior to use flat image names.

mattmoor:
Add a functional option to `pkg/ko/publish` to control image names.

To enable the broadest possible support out-of-the-box, this changes our default naming behavior to use shortened names: `{package}-{hash of full import path}`.

The current behavior, which yields better discoverability, can be restored by passing `--preserve-import-paths` or `-P`.

I have verified with this flag that I can publish to DockerHub via `ko`.

Fixes: https://github.com/google/go-containerregistry/issues/160

### DIFF
--- a/cmd/ko/README.md
+++ b/cmd/ko/README.md
@@ -169,10 +169,6 @@ spec:
         - containerPort: 8080
 ```
 
-*Note that DockerHub does not currently support these multi-level names. We may
-employ alternate naming strategies in the future to broaden support, but this
-would sacrifice some amount of identifiability.*
-
 ### `ko apply`
 
 `ko apply` is intended to parallel `kubectl apply`, but acts on the same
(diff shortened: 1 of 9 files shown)


mattmoor:
cc @sebgoa FYI

pkg/ko/publish/options.go
```diff
@@ -42,21 +43,24 @@ func WithAuth(auth authn.Authenticator) Option {
 // authenticator on a default publisher using an authn.Keychain
 func WithAuthFromKeychain(keys authn.Keychain) Option {
 	return func(i *defaultOpener) error {
-		auth, err := keys.Resolve(i.base.Registry)
+		repo, err := name.NewRepository(i.base, name.WeakValidation)
```
ImJasonH: Should we also make this change in [`pkg/v1/remote/options.go`](https://github.com/google/go-containerregistry/blob/master/pkg/v1/remote/options.go#L41)?

mattmoor: No, that still takes a parsed image name.  We can't here because we parse DockerHub base references of the form: `docker.io/mattmoor` as `docker.io/library/mattmoor` :(

ImJasonH: Can you add a comment to that effect? I'm trying to keep some idiot (me) from coming by in a few months and trying to collapse both `WithAuthFromKeychain` implementations into one, and causing pain.

mattmoor: Done

cmd/ko/README.md
```diff
@@ -190,6 +186,31 @@ to whatever `kubectl` context is active.
 of convenience for cleaning up resources created through `ko apply`.
 
 
+## With DockerHub
```
jonjohnsonjr: I think it's canonically `Docker Hub` (here and elsewhere).

cmd/ko/README.md
```diff
@@ -190,6 +186,31 @@ to whatever `kubectl` context is active.
...
+Unfortunately, DockerHub does not support this sort of multi-level names, so you may see an error like:
+
+```shell
+$ KO_DOCKER_REPO=docker.io/mattmoor ko publish ./cmd/crane
+2018/07/19 03:25:56 Using base gcr.io/distroless/base:latest for github.com/google/go-containerregistry/cmd/crane
+2018/07/19 03:25:56 Publishing index.docker.io/mattmoor/github.com/google/go-containerregistry/cmd/crane:latest
+2018/07/19 03:25:57 error publishing github.com/google/go-containerregistry/cmd/crane: UNAUTHORIZED: "authentication required"
+```
+
+To support this, we have a flag that will flatten the import path to `{package}-{hash of import path}`:
```
jonjohnsonjr: What do you think about hard-coding the default value for this based on the registry? It feels impure but I think it's not that unreasonable to do for docker hub and gcr. Using the flag should override it and it wouldn't be hard to add other registries as needed.

pkg/ko/publish/options.go
```diff
@@ -43,7 +44,15 @@ func WithAuth(auth authn.Authenticator) Option {
 // authenticator on a default publisher using an authn.Keychain
 func WithAuthFromKeychain(keys authn.Keychain) Option {
 	return func(i *defaultOpener) error {
-		auth, err := keys.Resolve(i.base.Registry)
+		// We parse this lazily because it is a repository prefix, which
+		// means that docker.io/mattmoor actually gets interpreted as
+		// dockerio/library/mattmoor, which gets tricky when we start
```
jonjohnsonjr: `docker.io`

cmd/ko/README.md
```diff
@@ -190,6 +186,31 @@ to whatever `kubectl` context is active.
...
+
+```shell
+$ KO_DOCKER_REPO=docker.io/mattmoor ko publish --flat ./cmd/crane
```
sebgoa: `--flat` is not very intuitive for users even though technically telling.

Also if you set `KO_DOCKER_REPO` couldn't the `--flat` be implicit ?

jonjohnsonjr:
I think changing the default is the right thing to do.

cmd/ko/README.md
```diff
@@ -126,13 +123,26 @@ an argument. It prints the images' published digests after each image is publish
...
+2018/07/19 14:57:34 Using base gcr.io/distroless/base:latest for github.com/mattmoor/warm-image/cmd/sleeper
+2018/07/19 14:57:35 Publishing us.gcr.io/my-project/sleeper-ebdb8b8b13d4bbe1d3592de055016d37:latest
+2018/07/19 14:57:35 mounted blob: sha256:739b3deec2edb17c512f507894c55c2681f9724191d820cdc01f668330724ca7
+2018/07/19 14:57:35 mounted blob: sha256:57752e7f9593cbfb7101af994b136a369ecc8174332866622db32a264f3fbefd
+2018/07/19 14:57:35 mounted blob: sha256:59df9d5b488aea2753ab7774ae41a9a3e96903f87ac699f3505960e744f36f7d
+2018/07/19 14:57:36 us.gcr.io/my-project/sleeper-ebdb8b8b13d4bbe1d3592de055016d37:latest: digest: sha256:6c7b96a294cad3ce613aac23c8aca5f9dd12a894354ab276c157fb5c1c2e3326 size: 592
+2018/07/19 14:57:36 Published us.gcr.io/my-project/sleeper-ebdb8b8b13d4bbe1d3592de055016d37@sha256:6c7b96a294cad3ce613aac23c8aca5f9dd12a894354ab276c157fb5c1c2e3326
+```
+
+`ko publish` also supports relative import paths, when in the context of a particular repo.
```
ImJasonH: Mention that this only works if you're inside `$GOPATH`?

mattmoor:
I signed it!

PR #243, opened 2018-07-20
Add a gcrane tool

jonjohnsonjr:
This is a variant of crane that takes advantage of some gcr-specific
features to implement richer listing and garbage collection.

### DIFF
--- a/cmd/gcr/main.go
+++ b/cmd/gcr/main.go
@@ -1,0 +1,52 @@
+// Copyright 2018 Google LLC All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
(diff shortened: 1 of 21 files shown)


cmd/gcr/main.go
```diff
@@ -0,0 +1,52 @@
...
+	Run:   func(cmd *cobra.Command, _ []string) { cmd.Usage() },
+}
+
+func main() {
+	// Use google-specific implementations for listing and gc.
+	cmds.AddCommand(gcr.NewCmdGc())
+	cmds.AddCommand(gcr.NewCmdList())
+
+	// Use crane for everthing else so that this can be a drop-in replacement.
+	cmds.AddCommand(crane.NewCmdAppend())
```
ImJasonH: This will need to be kept in sync with `cmd/crane/main.go`, which isn't great. Can we somehow convince cobra to let us passthrough unknown commands to another tool? Or grab them and pass them through ourselves?

jonjohnsonjr: Done.

pkg/v1/google/list.go
```diff
@@ -0,0 +1,78 @@
...
+	"net/url"
+
+	"github.com/google/go-containerregistry/pkg/authn"
+	"github.com/google/go-containerregistry/pkg/name"
+	"github.com/google/go-containerregistry/pkg/v1/remote"
+	"github.com/google/go-containerregistry/pkg/v1/remote/transport"
+)
+
+type ManifestInfo struct {
+	// TODO: int64?
```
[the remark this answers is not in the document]
jonjohnsonjr: Done.

PR #245, opened 2018-07-26
Implement Walk for gcrane listing

jonjohnsonjr:
Add -r flag to `gcrane {gc,ls}`

### DIFF
--- a/pkg/gcrane/list.go
+++ b/pkg/gcrane/list.go
@@ -15,8 +15,6 @@
 package gcrane
 
 import (
-	"net/http"
-
 	"fmt"
 	"log"
 
@@ -49,18 +47,16 @@ func ls(root string, recursive bool) {
 	if err != nil {
 		log.Fatalln(err)
 	}
-	auth, err := authn.DefaultKeychain.Resolve(repo.Registry)
-	if err != nil {
-		log.Fatalln(err)
-	}
+
+	auth := google.WithAuthFromKeychain(authn.DefaultKeychain)
 
 	if recursive {
-		if err := google.Walk(repo, auth, http.DefaultTransport, printImages); err != nil {
+		if err := google.Walk(repo, printImages, auth); err != nil {
 			log.Fatalln(err)
 		}
 	}
 
-	tags, err := google.List(repo, auth, http.DefaultTransport)
+	tags, err := google.List(repo, auth)
 	if err != nil {
 		log.Fatalln(err)
 	}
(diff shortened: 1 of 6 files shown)


jonjohnsonjr:
@ImJasonH WDYT before I add tests? Generally reasonable?

pkg/v1/google/list.go
```diff
@@ -125,3 +125,58 @@ func List(repo name.Repository, auth authn.Authenticator, t http.RoundTripper) (
...
+		}
+	}
+
+	// We made it!
+	return nil
+}
+
+// Walk recursively descends repositories, calling walkFn.
+// TODO: Do we need a keychain since this can take a while?
+func Walk(root name.Repository, auth authn.Authenticator, t http.RoundTripper, walkFn WalkFunc) error {
```
ImJasonH: I'd love to start standardizing on funcitonal options for new methods especially. In particular `WithAuthFromKeychain` might help solve the `TODO` you have above, and options would let users avoid having to specify `http.DefaultTransport` all the time.

mattmoor: +1

pkg/gcrane/list.go
```diff
@@ -29,50 +27,67 @@ import (
...
+			fmt.Printf("%s:%s\n", repo, tag)
+		}
+		return
+	}
 
 	for _, child := range tags.Children {
 		fmt.Printf("%s/%s\n", repo, child)
 	}
 
+	_ = printImages(repo, tags, err)
```
ImJasonH: Can you add a comment about why this error is unhandled?

pkg/v1/google/options.go
```diff
@@ -0,0 +1,56 @@
...
+import (
+	"log"
+	"net/http"
+
+	"github.com/google/go-containerregistry/pkg/authn"
+)
+
+// WithTransport is a functional option for overriding the default transport
+// on a remote image
+func WithTransport(t http.RoundTripper) ListerOption {
```
ImJasonH: Not a comment on this PR per se, but I think we're going to end up with a bunch of copies of these the way we're going. I'd love to figure out how to reuse them while making it clear which reusable options are usable in which scenarios. I can think of a way of abusing interfaces with private methods, but that's pretty gross. 😷

pkg/gcrane/list.go
jonjohnsonjr: Handled it instead.

pkg/v1/google/options.go
jonjohnsonjr: Yeah I wasn't sure how to reuse these in a nice way 👎

I figure it's not a ton of code, so some duplication here isn't terrible, but I agree this is less than ideal :)

PR #250, opened 2018-08-06
Perform HEAD for blob before initiating upload

### LINKED ISSUE #249: HEAD checks for pushing remote images
In my experiments, I am reasonably certain that my ECR registry does not support mounted blobs, with or without the from param.

It will be nice to have HEAD checks when uploading blobs to remote registries. They are fairly cheap, and almost everyone supports them.

I can send a pull request if you think this is a good idea.

siddharthab:
Fixes #249.

### DIFF
--- a/pkg/v1/remote/write.go
+++ b/pkg/v1/remote/write.go
@@ -120,6 +120,26 @@ func (w *writer) nextLocation(resp *http.Response) (string, error) {
 	return resp.Request.URL.ResolveReference(u).String(), nil
 }
 
+// checkExisting checks if a blob exists already in the repository by making a
+// HEAD request to the blob store API.  GCR performs an existence check on the
+// initiation if "mount" is specified, even if no "from" sources are specified.
+// However, this is not broadly applicable to all registries, e.g. ECR.
+func (w *writer) checkExisting(h v1.Hash) (bool, error) {
+	u := w.url(fmt.Sprintf("/v2/%s/blobs/%s", w.ref.Context().RepositoryStr(), h.String()))
+
+	resp, err := w.client.Head(u.String())
+	if err != nil {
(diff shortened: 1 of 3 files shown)


pkg/v1/remote/write.go
```diff
@@ -120,52 +120,79 @@ func (w *writer) nextLocation(resp *http.Response) (string, error) {
...
-	if ml, ok := l.(*MountableLayer); ok {
-		if w.ref.Context().RegistryStr() == ml.Reference.Context().RegistryStr() {
-			uv["from"] = []string{ml.Reference.Context().RepositoryStr()}
+	// GCR will perform an existence check on the initiation if "mount" is
+	// specified, even if no "from" sources are specified.  However, this is not
+	// broadly applicable to all registries, e.g. ECR. So we use a simple HEAD
+	// check when the "from" source is not specified.
+	if ml, ok := l.(*MountableLayer); ok && w.ref.Context().RegistryStr() == ml.Reference.Context().RegistryStr() {
+		uv["from"] = []string{ml.Reference.Context().RepositoryStr()}
+	} else {
```
jonjohnsonjr: It seems like this misses the case where you're doing an ECR -> ECR upload, though maybe I'm misreading it. (e.g. `crane copy <ecr-src> <ecr-dst>` to retag something would not end up calling `checkExisting`)

siddharthab: Yes, you are right. I just went off the comment there and implemented what it said. But just checked, it did not work for `crane copy`. Now fixed; crane copy to an ECR destination will now reuse blobs in the destination image repository, if they exist.

jonjohnsonjr:
@siddharthab  thanks for the PR 👍

PR #252, opened 2018-08-09
support writing multiple images to a tarball

sharifelgamal:


mattmoor:
(drive-by thought without looking at the change)
Do we currently support writing multiple images to a tarball?  If we did I'd think this would be a degenerate case of that, e.g.
```go
tarball.Write(map[tag]v1.Image{
  "foo:bar": img,
  "baz:blah": img,
})
```

sharifelgamal:
Far as I can tell, we don't currently support writing multiple images to a tarball. Should I go ahead and do that instead?

jonjohnsonjr:
@sharifelgamal I agree with Matt, I think that's probably what we should do. In the python library we had [`multi_image_tarball`](https://github.com/google/containerregistry/blob/bf22fb4dc41f000ff20e6bc02c3649243096703d/client/v2_2/save_.py#L48).

Since that's more general, we can kill two birds with one stone here:
1. allowing multiple tags per image and
1. allowing multiple images per tarball.

sharifelgamal:
I'm going avoid this merge hell by opening a new PR.

PR #253, opened 2018-08-10
Fix bug where ko layers didn't include tar trailers

ImJasonH:
Because `buf.Bytes()` was called before `defer tw.Close()` in `tarKoData` and `tarBinary`, the returned `[]byte` didn't include tar trailer data added by `tw.Close()`.

Instead, return the `bytes.Buffer` and get its `Bytes()` after the func has returned and `tw.Close()` has been called.

Go's tar package seems to handle trailer-less tar data just fine, as evidenced by the fact that our tests and Docker itself had no problem with these images, but the Python implementation failed to extract layers without the trailer data.

### DIFF
--- a/pkg/ko/build/gobuild.go
+++ b/pkg/ko/build/gobuild.go
@@ -115,7 +115,7 @@ func build(ip string) (string, error) {
 	return file, nil
 }
 
-func tarBinary(binary string) (io.Reader, error) {
+func tarBinary(binary string) (*bytes.Buffer, error) {
 	buf := bytes.NewBuffer(nil)
 	tw := tar.NewWriter(buf)
 	defer tw.Close()
@@ -160,7 +160,7 @@ func kodataPath(s string) (string, error) {
 // Where kodata lives in the image.
 const kodataRoot = "/var/run/ko"
 
-func tarKoData(importpath string) (io.Reader, error) {
+func tarKoData(importpath string) (*bytes.Buffer, error) {
 	buf := bytes.NewBuffer(nil)
 	tw := tar.NewWriter(buf)
 	defer tw.Close()
@@ -239,8 +239,9 @@ func (gb *gobuild) Build(s string) (v1.Image, error) {
 	if err != nil {
 		return nil, err
 	}
+	dataLayerBytes := dataLayerBuf.Bytes()
 	dataLayer, err := tarball.LayerFromOpener(func() (io.ReadCloser, error) {
-		return v1util.NopReadCloser(dataLayerBuf), nil
+		return v1util.NopReadCloser(bytes.NewBuffer(dataLayerBytes)), nil
 	})
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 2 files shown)


ImJasonH:
Looks like our current `check_determinism` checks against constant known-ahead-of-time digests fail. Something in betwen Go 1.10 and 1.11beta3 changes the digest of the layers.

From HEAD:
```
$ docker run -e GOPATH=/gopath -v $GOPATH:/gopath --workdir=/gopath/src/github.com/google/go-containerregistry golang:1.10 go test ./pkg/ko/build/...
ok      github.com/google/go-containerregistry/pkg/ko/build     0.057s

$ docker run -e GOPATH=/gopath -v $GOPATH:/gopath --workdir=/gopath/src/github.com/google/go-containerregistry golang:1.11beta3 go test ./pkg/ko/build/...
--- FAIL: TestGoBuildNoKoData (0.01s)
    --- FAIL: TestGoBuildNoKoData/check_determinism (0.00s)
        gobuild_test.go:130: Digest() = sha256:ca5c6ceb4a7f054335e8ff65f1d42a1c43cd44054d26de8ad4ad4b0014d4f4ec, want sha256:a7be3daf6084d1ec81ca903599d23a514903c9e875d60414ff8009994615bd70
--- FAIL: TestGoBuild (0.01s)
    --- FAIL: TestGoBuild/check_determinism (0.00s)
        gobuild_test.go:208: Digest() = sha256:c049f84314c675467780475efe3cba5fabb13c4d49f7ee7137c057997efcba41, want sha256:9b5d822a4d4c0d1afab5dbe85b38e922240a7bf3b2bbbeff53189fe46f3a7b84
    --- FAIL: TestGoBuild/check_kodata (0.00s)
        gobuild_test.go:224: Digest() = sha256:5d6e7a5bb03c6c5fac80e4beb7cf6d514be943724e8b4573255bdbeafe77b40f, want sha256:7a7bafbc2ae1bf844c47b33025dd459913a3fece0a94b1f3ced860675be2b79c
FAIL
FAIL    github.com/google/go-containerregistry/pkg/ko/build     0.049s
```

Changing the digests to those produced by Go 1.11 (as I had done) means they fail against 1.10 (as [Travis did](https://travis-ci.org/google/go-containerregistry/builds/414564152?utm_source=github_status&utm_medium=notification)).

So, instead of checking determinism against a known-ahead-of-time digest, I've updated to check determinism against the same image built again, with the same version of Go. This passes whether that version is 1.10 or 1.11beta3.

This does mean that `ko`-produced images will be different if the user has 1.10 or 1.11beta3 installed on their machine. AFAICT this would have been a breaking change even before this PR.

ImJasonH:
The only real change to `archive/tar` between 1.10 and 1.11 is the fix for https://github.com/golang/go/issues/22768, where callers of `WriteHeader` who don't set `Typeflag` would get the incorrect type. By setting this flag ourselves, like we should have done from the beginning, we can retain determinism regardless of which version of Go's `archive/tar` is used.

PR #257, opened 2018-08-13
Add Image Index types

jonjohnsonjr:
This adds minimal structs and interfaces we need to represent an image index.

Please nit names :)

We'll probably want to add some more methods to the `ImageIndex` interface for convenience in the future, but this should be a sufficient and non-controversial starting point.

Also updated the codegen.


pkg/v1/manifest.go
```diff
@@ -37,6 +37,15 @@ type Descriptor struct {
...
 	Annotations map[string]string `json:"annotations,omitempty"`
+	Platform    *Platform         `json:"platform,omitempty"`
+}
+
+// Index represents an OCI image index in a structured way.
+type Index struct {
+	SchemaVersion int64             `json:"schemaVersion"`
+	MediaType     types.MediaType   `json:"mediaType"`
+	Manifests     []Descriptor      `json:"manifests"`
+	Annotations   map[string]string `json:"annotations,omitempty"`
```
mattmoor: nit: we have `config.go` and `manifest.go`...

pkg/v1/index.go
```diff
@@ -0,0 +1,33 @@
...
+
+type ImageIndex interface {
+	// MediaType of this image's manifest.
+	MediaType() (types.MediaType, error)
+
+	// Digest returns the sha256 of this index's manifest.
+	Digest() (Hash, error)
+
+	// Index returns this image index's manifest object.
+	Index() (*Index, error)
```
mattmoor: `ImageIndex` vs. `Index` is confusing.

pkg/v1/index.go
```diff
@@ -0,0 +1,33 @@
...
+
+	// RawManifest returns the serialized bytes of Index().
+	RawManifest() ([]byte, error)
```
mattmoor: RawIndex?

jonjohnsonjr: My thinking here was to get `Digest` for a `WithRawManifest`, but perhaps not worth the coupling?

pkg/v1/index.go
jonjohnsonjr: I think I like `ImageIndex` for the interface, WDYT about `IndexManifest` for the struct?

pkg/v1/manifest.go
jonjohnsonjr: Moved some things around. I think it kind of makes sense to keep the index in `manifest.go`, since it's a _kind_ of manifest, but it's not a strong opinion.

pkg/v1/manifest.go
```diff
@@ -37,6 +37,7 @@ type Descriptor struct {
 	Digest      Hash              `json:"digest"`
 	URLs        []string          `json:"urls,omitempty"`
 	Annotations map[string]string `json:"annotations,omitempty"`
+	Platform    *Platform         `json:"platform,omitempty"`
```
mattmoor: Did you run codegen?

jonjohnsonjr: Done now :)

PR #260, opened 2018-08-22
support writing multiple images to a tarball

sharifelgamal:


pkg/v1/tarball/write.go
```diff
@@ -44,78 +44,106 @@ func WriteToFile(p string, tag name.Tag, img v1.Image, wo *WriteOptions) error {
...
-	// Generate the tar descriptor and write it.
-	td := tarDescriptor{
-		singleImageTarDescriptor{
+			if err := writeTarEntry(tf, layerFiles[i], r, blobSize); err != nil {
+				return err
+			}
+		}
+
+		// Generate the tar descriptor and write it.
+
```
jonjohnsonjr: extra newline

pkg/v1/tarball/write.go
```diff
@@ -44,78 +44,106 @@ func WriteToFile(p string, tag name.Tag, img v1.Image, wo *WriteOptions) error {
 	return Write(tag, img, wo, w)
 }
 
+// MultiWrite writes the contents each image in the tagToImage map to its
```
jonjohnsonjr: This comment seems misplaced

pkg/v1/tarball/write.go
```diff
@@ -44,78 +44,106 @@ func WriteToFile(p string, tag name.Tag, img v1.Image, wo *WriteOptions) error {
...
+	return MultiWrite(map[name.Tag]v1.Image{tag: img}, wo, w)
+}
+
 // Write the contents of the image to the provided reader, in the compressed format.
 // The contents are written in the following format:
 // One manifest.json file at the top level containing information about several images.
 // One file for each layer, named after the layer's SHA.
 // One file for the config blob, named after its SHA.
-func Write(tag name.Tag, img v1.Image, wo *WriteOptions, w io.Writer) error {
+func MultiWrite(tagToImage map[name.Tag]v1.Image, wo *WriteOptions, w io.Writer) error {
```
jonjohnsonjr: This comment seems misplaced

sharifelgamal:
I've fixed up all the comments.

pkg/v1/tarball/write.go
```diff
@@ -44,78 +44,105 @@ func WriteToFile(p string, tag name.Tag, img v1.Image, wo *WriteOptions) error {
...
+
+		td = append(td, sitd)
 	}
+
 	tdBytes, err := json.Marshal(td)
 	if err != nil {
 		return err
 	}
-	return writeTarEntry(tf, "manifest.json", bytes.NewReader(tdBytes), int64(len(tdBytes)))
+	err = writeTarEntry(tf, "manifest.json", bytes.NewReader(tdBytes), int64(len(tdBytes)))
```
jonjohnsonjr: why not just `return writeTarEntry(...)`?

sharifelgamal: I was testing something out and forgot to switch this back. It's fixed.

jonjohnsonjr:
This looks fine to me 👍 

I'd like to see a couple tests of:
1. Multiple images
2. Multiple tags on one image

I played with this a bit to test the second case:

`$ cat test.go`
```golang
package main

import (
	"log"

	"github.com/google/go-containerregistry/pkg/v1"
)

func main() {
	t1, _ := name.NewTag("ubuntu", name.WeakValidation)
	t2, _ := name.NewTag("also-ubuntu", name.WeakValidation)

	i, err := remote.Image(t1, remote.WithAuthFromKeychain(authn.DefaultKeychain))
	if err != nil {
		log.Fatalf("reading image %q: %v", t1, err)
	}

	tags := make(map[name.Tag]v1.Image)

	tags[t1] = i
	tags[t2] = i

	if err := tarball.MultiWriteToFile("output.tar", tags, &tarball.WriteOptions{}); err != nil {
		log.Fatalf("writing image: %v", err)
	}
}
```
`$ cat output/manifest.json | jq .`
```json
[
  {
    "Config": "sha256:16508e5c265dcb5c05017a2a8a8228ae12b7b56b2cda0197ed5411bda200a961",
    "RepoTags": [
      "index.docker.io/library/also-ubuntu:latest"
    ],
    "Layers": [
      "124c757242f88002a858c23fc79f8262f9587fa30fd92507e586ad074afb42b6.tar.gz",
      "2ebc019eb4e2bbd192e61bce91038048924216d72dfe6ac3255322caaeb70144.tar.gz",
      "dac0825f7ffbea2ddf119026b0d6c4c453dfa38edb5c2abbe59bdba6ffdb3b9f.tar.gz",
      "82b0bb65d1bfb978cd646dcd653164f74b7e8656aa1b3a87657b8d244b56d324.tar.gz",
      "ef3b655c7f887451d42e45a04c2d13478171e4d451419bc8cf76c635f509532f.tar.gz"
    ]
  },
  {
    "Config": "sha256:16508e5c265dcb5c05017a2a8a8228ae12b7b56b2cda0197ed5411bda200a961",
    "RepoTags": [
      "index.docker.io/library/ubuntu:latest"
    ],
    "Layers": [
      "124c757242f88002a858c23fc79f8262f9587fa30fd92507e586ad074afb42b6.tar.gz",
      "2ebc019eb4e2bbd192e61bce91038048924216d72dfe6ac3255322caaeb70144.tar.gz",
      "dac0825f7ffbea2ddf119026b0d6c4c453dfa38edb5c2abbe59bdba6ffdb3b9f.tar.gz",
      "82b0bb65d1bfb978cd646dcd653164f74b7e8656aa1b3a87657b8d244b56d324.tar.gz",
      "ef3b655c7f887451d42e45a04c2d13478171e4d451419bc8cf76c635f509532f.tar.gz"
    ]
  }
]
```

I'm a bit surprised that this has two entries. I'd expect one entry with two elements in `RepoTags`, but it seems like this is the same as the behavior of the python library, so maybe that's fine: https://github.com/google/containerregistry/blob/03ec6a36903075ae6e905dd411c65e0efeec4f9f/client/v2_2/save_.py#L106

I would expect we can just dedupe based on the config digest, yeah? I'm not 100% sure about docker behavior w.r.t. tarballs, though.

@dlorenc wdyt? How does `RepoTags` work?

It also seems to take twice as long to run this with two tags vs one tag, which makes sense, but if they're pointing at the same image, we shouldn't have to download the layers twice. Can we check if we've already written a layer into the tarball before downloading them? This isn't a huge deal, but it would be nice to avoid any performance issues we can.

sharifelgamal:
RepoTags definitely works how we anticipate it, but I wrote this code in the simplest way possible, which creates a new entry for each tag, regardless of which image it points to. I can add a pretty simple dedup to prune multiple tags pointing to the same image.

sharifelgamal:
This should be ready

Issue #255, opened 2018-08-13
Support OCI image layout


jonjohnsonjr:
This is a quite a bit more sane than the docker tarball format, we should probably support it. It works a lot better as an intermediate format: https://github.com/knative/build/issues/215#issuecomment-412550394

https://github.com/opencontainers/image-spec/blob/master/image-layout.md

PR #261, opened 2018-08-23
Implement OCI Image Layout support

jonjohnsonjr:
Ref #255 

Feedback welcome :)

# what changes

`layout.Write` takes an ImageIndex and writes it to disk as an OCI Image Layout. This writes the top-level manifest as index.json.

`layout.Read` returns a `LayoutPath` for the given path. You use this as the entrypoint into reading from a layout.

`LayoutPath.Append{Image,Index}` writes an image or index to the image layout and updates the index.json.

`LayoutPath.WriteBlob` writes a blob to the OCI Image Layout, but doesn't update the index.json.

`LayoutPath.AppendDescriptor` adds a descriptor to the index.json. This + WriteBlob lets you store arbitrary stuff in the image layout.

`LayoutPath.Image` reads a specific image by digest from the image layout.

`LayoutPath.ImageIndex` reads the whole image layout as and image index.

# how do you use it?

You can construct arbitrary layouts via `WriteBlob` and `AppendDescriptor`. I think we'll want to make this a bit more ergonomic, e.g. `Append{Image,Index}` are going to be common operations, probably, since they just write an artifact to the blobs/ directory and update the `index.json`. There are some options, e.g. `WithURLs` and `WithAnnotations` to make it easier to Append things with the right descriptor.


pkg/v1/layout/index.go
```diff
@@ -0,0 +1,112 @@
...
+	img := &layoutImage{
+		path: i.path,
+		desc: *desc,
+	}
+	return partial.CompressedToImage(img)
+}
+
+func (i *layoutIndex) ImageIndex(h v1.Hash) (v1.ImageIndex, error) {
+	// Look up the digest in our manifest first to return a better error.
+	if _, err := i.findDescriptor(h); err != nil {
```
yolo3301: Do you need to check if the provided hash is really of a ImageIndex? It could be a hash of an Image right?

pkg/v1/layout/write.go
```diff
@@ -0,0 +1,240 @@
...
+	if err := writeImage(path, img); err != nil {
+		return err
+	}
+
+	mt, err := img.MediaType()
+	if err != nil {
+		return err
+	}
+
+	// TODO: This is to work arounda a validation bug in GCR.
```
yolo3301: What GCR bug?

cmd/compass/main.go
```diff
@@ -0,0 +1,191 @@
...
+	"github.com/google/go-containerregistry/pkg/name"
+	"github.com/google/go-containerregistry/pkg/v1/layout"
+	"github.com/google/go-containerregistry/pkg/v1/remote"
+	"github.com/spf13/cobra"
+)
+
+func main() {
+	// Parent command to which all subcommands are added.
+	cmds := &cobra.Command{
+		Use:   "compass",
```
ImJasonH: Why `compass` ? Can this just be `crane oci ...` instead?

cmd/compass/main.go
```diff
@@ -0,0 +1,191 @@
...
+		},
+	}
+
+	// TODO: rm
+	//  remove an entry from index.json, optionally garbage collecting unreferenced
+	//  blobs
+	// TODO: gc
+	//  garbage collect unreferenced blobs
+	for _, cmd := range []cobra.Command{{
+		Use:   "freeze",
```
ImJasonH: Why not `pull` and `push`?

cmd/compass/main.go
```diff
@@ -0,0 +1,191 @@
...
+	}, {
+		Use:   "ls",
+		Short: "List the manifests in a layout",
+		Args:  cobra.ExactArgs(1),
+		Run: func(_ *cobra.Command, args []string) {
+			ls(args[0])
+		},
+	}} {
+		cmd := cmd
+		cmds.AddCommand(&cmd)
```
ImJasonH: I think this is slightly more readable:

```
cmds.AddCommand(&cobra.Command{
  ...
})
cmds.AddCommand(&cobra.Command{
  ...
})
```

vs

```
for _, cmd := range []cobra.Command{
  ...
}, {
  ...
}} {
  cmd := cmd
  cmds.AddCommand(&cmd)
}
```

(And it's shorter!)

cmd/compass/main.go
```diff
@@ -0,0 +1,191 @@
...
+		log.Fatalf("reading image layout %q: %v", path, err)
+	}
+
+	manifest, err := ii.IndexManifest()
+	if err != nil {
+		log.Fatalf("reading image layout manifest %q: %v", path, err)
+	}
+
+	for _, desc := range manifest.Manifests {
+		// This logic is janky, we'd want smarter rewriting rules.
```
ImJasonH: Can you describe some of the smarts in case someone else is responsible for adding them?

pkg/v1/layout/testdata/test_image/555b1001d54ed229f56990845856f080b0707348b26b3aa85aaf58d5570cdee0/VERSION
```diff
@@ -0,0 +1 @@
+1.0
```
ImJasonH: Can you document in `testdata/test_image/README` how to regenerate these in the future?

pkg/v1/remote/image_test.go
```diff
@@ -250,7 +251,7 @@ func TestAcceptHeaders(t *testing.T) {
 			if r.Method != http.MethodGet {
 				t.Errorf("Method; got %v, want %v", r.Method, http.MethodGet)
 			}
-			if got, want := r.Header.Get("Accept"), string(types.DockerManifestSchema2); got != want {
+			if got, want := r.Header.Get("Accept"), string(types.DockerManifestSchema2); strings.Contains(want, got) {
```
ImJasonH: Can you just update `want` instead?

pkg/v1/layout/testdata/test_image/555b1001d54ed229f56990845856f080b0707348b26b3aa85aaf58d5570cdee0/VERSION
jonjohnsonjr: If only I knew.

cmd/compass/main.go
jonjohnsonjr: Chatted about this: https://github.com/google/go-containerregistry/issues/384 captures the ideas.

pkg/v1/layout/write.go
```diff
@@ -0,0 +1,318 @@
...
+			return nil, err
+		}
+	}
+
+	return AppendDescriptor(path, desc)
+}
+
+// AppendDescriptor adds a descriptor to the index.json of an ImageIndex located
+// at path.
+func AppendDescriptor(path string, desc v1.Descriptor) (v1.ImageIndex, error) {
```
glyn: I'd like to understand the relationship between the image layout and the returned `v1.ImageIndex` on this and other similar functions. Reading the [spec](https://github.com/opencontainers/image-spec/blob/master/image-layout.md) didn't help particularly because it seems that image indices can refer to other image indices. So is the intention that the returned image index always represents the whole enchilada or for some functions, does it represent one of the constituent image indices?

jonjohnsonjr: I'll attempt to explain but let me know if I can make this clearer in the comments or in general:

An [image index](https://github.com/opencontainers/image-spec/blob/master/image-index.md) is just a JSON data structure (AKA a manifest) that can refer to other manifests. Those manifests *usually* represent either an image or another index.

An [image layout](https://github.com/opencontainers/image-spec/blob/master/image-layout.md) is a way to store images on disk comprised of a "blobs/" directory with content-addressable contents and an "index.json" that is an index those contents and acts as an entrypoint. That index.json file's format is an [image index](https://github.com/opencontainers/image-spec/blob/master/image-index.md).

These top-level functions interact with either the "blobs/" directory or the "index.json" or both.

`WriteBlob`, `WriteImage`, and `WriteIndex` add files to the "blobs/" directory, but don't update index.json. This is useful for doing batches or writes without having to update the index.json in-betweeen.

`AppendImage` and `AppendIndex` add files to the "blobs" directory *and* update the index.json. I expect these would be the most common write operations.

`Write` takes an ImageIndex and essentially converts it to an image layout, where the image index manifest becomes the top-level index.json, and everything it references is written to the "blobs/" directory. This is useful for serializing an index to disk without having an extraneous wrapper around it.

The `layout.Image` and `layout.Index` functions for reading images/indexes mirrors what we're doing with `remote.Image` and `remote.Index` -- they just read from the data source and produce an image or index.

glyn: That's clear. I guess a clarification which would help me is to document the return values of the public functions. For example, `AppendDescriptor` might document its return values as follows:
```
// AppendDescriptor adds a descriptor to the index.json of an ImageIndex located
// at path. If successful, it returns a nil error and an ImageIndex representing the updated
// index.json (and equivalent to that which would be returned by calling Index(path) after 
// AppendDescriptor).
```

glyn: Alternatively, you might wish to remove the `ImageIndex` return values. The API provides the `Index` function to get hold of an `ImageIndex` if necessary.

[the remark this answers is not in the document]
glyn: @jonjohnsonjr What do you think about this? I wonder if something as simple as having these functions be methods on an `ImageLayoutPath` type (with underlying type `string`) and with a simple constructor function would actually make the whole API more readable. A constructor function is a more obvious entry point for the user.

If you adopted that approach, the `Write` function could then become a constructor function returning an `ImageLayoutPath`.

pkg/v1/layout/write.go
```diff
@@ -0,0 +1,318 @@
...
+	for _, opt := range options {
+		if err := opt(&desc); err != nil {
+			return nil, err
+		}
+	}
+
+	return AppendDescriptor(path, desc)
+}
+
+// AppendDescriptor adds a descriptor to the index.json of an ImageIndex located
```
glyn: I think this doc comment could be improved by talking in terms of an image layout. That might be simplest by turning it into a method as I mention below. But even if you decide against making this a method, I think it would help to raise the level of the description a bit. I know that an image layout is represented as an `index.json` of an `ImageIndex`, but calling it an image layout makes the intended usage clearer.

pkg/v1/layout/write.go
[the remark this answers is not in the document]
jonjohnsonjr: I think this is a good idea. I'd want to keep `Write` and `Index` functions, both of which would return an `ImageLayoutPath` (to mirror the approach of similar packages), but everything else would make sense as a method.

glyn:
@jonjohnsonjr I confirm I'm ok with my commits being contributed to this project.

jonjohnsonjr:
@ImJasonH I think this if RFAL.

pkg/v1/layout/image_test.go
```diff
@@ -0,0 +1,110 @@
...
+	lp, err := Read(testPath)
+	if err != nil {
+		t.Fatalf("Read() = %v", err)
+	}
+	img, err := lp.Image(manifestDigest)
+	if err != nil {
+		t.Fatalf("Image() = %v", err)
+	}
+
+	if err := validate.Image(img); err != nil {
```
ImJasonH: Can some of these be `t.Errorf`?

pkg/v1/layout/image_test.go
```diff
@@ -0,0 +1,110 @@
...
+	if err != nil {
+		t.Fatalf("Read() = %v", err)
+	}
+	img, err := lp.Image(manifestDigest)
+	if err != nil {
+		t.Fatalf("Image() = %v", err)
+	}
+
+	if _, err := img.LayerByDigest(bogusDigest); err == nil {
+		t.Fatalf("LayerByDigest(%s) = nil, expected err", bogusDigest)
```
ImJasonH: `t.Errorf`s here too

pkg/v1/layout/index.go
```diff
@@ -0,0 +1,146 @@
...
+
+var _ v1.ImageIndex = (*layoutIndex)(nil)
+
+type layoutIndex struct {
+	path     LayoutPath
+	rawIndex []byte
+}
+
+// ImageIndex is a convenience function which constructs a LayoutPath and returns its v1.ImageIndex.
+func ImageIndex(path string) (v1.ImageIndex, error) {
```
ImJasonH: How do you feel about `ImageIndexFromPath`? Is there any chance we'll want an `ImageIndex` constructor from anything else?

pkg/v1/layout/image_test.go
jonjohnsonjr: Done.

pkg/v1/layout/image_test.go
jonjohnsonjr: Done.

pkg/v1/layout/index.go
jonjohnsonjr: Done.

jonjohnsonjr:
@glyn I renamed a few things to satisfy the go linter, hope that's alright with you.

PR #262, opened 2018-08-23
Remove Docker-Content-Digest check

jonjohnsonjr:
See https://github.com/GoogleContainerTools/kaniko/issues/298

Too many registries don't work correctly, and we don't want to hard-code
them all.


mattmoor:
(sigh) wtf...

PR #267, opened 2018-09-04
change var name ref to cpyRef

zrss:
Optimize readability


zrss:
a tiny change @mattmoor could u take a look

PR #269, opened 2018-09-05
Add go module support

dgodd:
I am using go-containerregistry inside a project which uses go modules and it would be convenient if go-containerregistry included go.mod and go.sum files


ImJasonH:
Could you clarify how these files were generated? I assume it was just a `go mod -init -module github.com/google/go-containerregistry` which generated from `Gopkg.lock`?

Should we update `scripts/update-deps.sh` to generate both files and keep them up to date and consistent?

dgodd:
Yes. I just ran `go mod init` after creating a `go.mod` with the module name. (the gopkg.lock was used)

Keeping the two synced seems like a good idea. I'm happy to change my commit if that is desired.

PR #270, opened 2018-09-06
Memoize size/hash for uncompressedLayerExtender

dgodd:
Memoizing the hash and size take the time to write on image (for one of my images) from 13s down to 8s.

### DIFF
--- a/pkg/v1/partial/uncompressed.go
+++ b/pkg/v1/partial/uncompressed.go
@@ -37,8 +37,11 @@ type UncompressedLayer interface {
 // uncompressedLayerExtender implements v1.Image using the uncompressed base properties.
 type uncompressedLayerExtender struct {
 	UncompressedLayer
-	// TODO(mattmoor): Memoize size/hash so that the methods aren't twice as
+	// Memoize size/hash so that the methods aren't twice as
 	// expensive as doing this manually.
+	hash v1.Hash
+	size int64
+	lock sync.Mutex
 }
 
 // Compressed implements v1.Layer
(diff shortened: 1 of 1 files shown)


pkg/v1/partial/uncompressed.go
```diff
@@ -37,8 +37,11 @@ type UncompressedLayer interface {
 // uncompressedLayerExtender implements v1.Image using the uncompressed base properties.
 type uncompressedLayerExtender struct {
 	UncompressedLayer
-	// TODO(mattmoor): Memoize size/hash so that the methods aren't twice as
+	// Memoize size/hash so that the methods aren't twice as
 	// expensive as doing this manually.
+	hash v1.Hash
+	size int64
+	lock sync.Mutex
```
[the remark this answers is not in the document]
dgodd: I have just made this change in 611aad063148a69435dccd3cf8475262c11814f6, however due to the need to deal with errors I need to store an error as well. 

I'm on the fence as to which option is better, but I'm interested in any feedback either way, or for a better option.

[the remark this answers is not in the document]
dprotaso: You could store the error in `uncompressedLayerExtender` type. Or create a new union type ie. `HashOrError` as a helper

dgodd:
I believe that there are no outstanding issues with this PR. Could someone please review it again.

pkg/v1/partial/uncompressed.go
```diff
@@ -52,29 +56,31 @@ func (ule *uncompressedLayerExtender) Compressed() (io.ReadCloser, error) {
...
-		return -1, err
-	}
-	defer r.Close()
-	_, i, err := v1.SHA256(r)
-	return i, err
+	ule.calcSizeHash()
+	return ule.size, ule.hashSizeError
+}
+
+func (ule *uncompressedLayerExtender) calcSizeHash() {
```
ImJasonH: I think we should still have this return an error, and just return `nil` if the `Once` didn't fire:

```
func calcSizeHash() error {
  var err error
  s.once.Do(func() {
    var r io.ReadCloser
    r, err = ule.Compressed()
    ...
  })
  return err
}
```

This seems cleaner to me than storing the `hashSizeError` on the struct.

dgodd: @ImJasonH I may well be missing something, but won't that be a problem on the second call?  If `ule.Uncompressed` returns an error inside the once then we will have the situation that

upon calling `ule.Digest()` it returns the error that was received from `calcSizeHash`.
But upon the second call, to`url.Size()` we will return nil as the error because the once will not refire, and calcSizeHash() will thus return nil for error.

If you would prefer then I would happily have `calcSizeHash()` store the error in `hashSizeError` and return it so that it could be used, or even return all three values which might be cleaner than what I have now.

I'm open to suggestions, including that I have somehow incorrectly identied the problem

ImJasonH: Yeah good point. I guess we should keep the same behavior we have today and have `calcSizeHash` return the error every time, regardless of how many times the `Once` is called. I'm convinced.

PR #277, opened 2018-09-20
[WIP] Allow to publish multiple tags with --tags.

### LINKED ISSUE #181: ko should allow images to be tagged with other tags (not just :latest)
`ko apply`, `resolve`, `publish` all publish to `$KO_DOCKER_REPO/$importpath:latest` due to code [here](https://github.com/google/go-containerregistry/blob/36637b11473b5be14cc7271b034e098a9b572529/ko/publish/default.go#L47-L49).

Some users might want to tag these images with something other than `:latest`, possibly with multiple tags. Where that is configured is up for debate (in `.ko.yaml` or in another env var? How to specify multiple tags?).

markusthoemmes:
Fixes #181 

Implements the `--tags` argument as outlined in the issue. Opening early to get clearance on the approach taken.

TODO: Adjust `daemon_test.go` and `default_test.go` with additional tests to verify this actually works.

### DIFF
--- a/cmd/ko/publish.go
+++ b/cmd/ko/publish.go
@@ -83,18 +83,18 @@ func publishImages(importpaths []string, no *NameOptions, lo *LocalOptions, ta *
 		}
 
 		if lo.Local || repoName == publish.LocalDomain {
-			pub = publish.NewDaemon(daemon.WriteOptions{}, namer)
+			pub = publish.NewDaemon(daemon.WriteOptions{}, namer, ta.Tags)
 		} else {
 			if _, err := name.NewRepository(repoName, name.WeakValidation); err != nil {
 				log.Fatalf("the environment variable KO_DOCKER_REPO must be set to a valid docker repository, got %v", err)
 			}
-			opts := []publish.Option{publish.WithAuthFromKeychain(authn.DefaultKeychain), publish.WithNamer(namer)}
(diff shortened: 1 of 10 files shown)


cmd/ko/publish.go
```diff
@@ -93,7 +94,7 @@ func publishImages(importpaths []string, no *NameOptions, lo *LocalOptions) {
 				log.Fatalf("error setting up default image publisher: %v", err)
 			}
 		}
-		if _, err := pub.Publish(img, importpath); err != nil {
+		if _, err := pub.Publish(img, importpath, ta.Tags); err != nil {
```
mattmoor: I think we should pass this to the constructor instead of the `Publish` call(s).

Reviewing your other PR, my secret hope was that we could eventually get this down to one constructor with different functional options, but I need to play around sometime...

cmd/ko/tags.go
```diff
@@ -0,0 +1,28 @@
...
+	"github.com/spf13/cobra"
+)
+
+type TagsOptions struct {
+	Tags []string
+}
+
+func addTagsArg(cmd *cobra.Command, ta *TagsOptions) {
+	cmd.Flags().StringSliceVar(&ta.Tags, "tags", ta.Tags,
+		"Which tags to use for the produced image in addition to the digest.")
```
mattmoor: digest => default.

Also, I'd call this `extra_tags` because of this.

markusthoemmes:
Closing this for now, 'til another day!

PR #283, opened 2018-10-07
Refresh Bearer token

ianberinger:
Currently, a `bearerTransport` requests only one token during it's lifetime. After that token expires, all further requests will fail (`401`).

This seems to be the root cause of https://github.com/GoogleContainerTools/kaniko/issues/245

This PR enables `bearerTransport.RoundTrip` to refresh an expired token and retry the request.

### DIFF
--- a/pkg/v1/remote/transport/bearer.go
+++ b/pkg/v1/remote/transport/bearer.go
@@ -46,38 +46,34 @@ var _ http.RoundTripper = (*bearerTransport)(nil)
 
 // RoundTrip implements http.RoundTripper
 func (bt *bearerTransport) RoundTrip(in *http.Request) (*http.Response, error) {
-	hdr, err := bt.bearer.Authorization()
-	if err != nil {
-		return nil, err
-	}
-
-	// http.Client handles redirects at a layer above the http.RoundTripper
-	// abstraction, so to avoid forwarding Authorization headers to places
-	// we are redirected, only set it when the authorization header matches
-	// the registry with which we are interacting.
-	if in.Host == bt.registry.RegistryStr() {
-		in.Header.Set("Authorization", hdr)
-	}
-	in.Header.Set("User-Agent", transportName)
-
-	res, err := bt.inner.RoundTrip(in)
-
-	// Perform a token refresh() and retry the request in case the token has expired
-	if err == nil && res.StatusCode == http.StatusUnauthorized {
-		if err = bt.refresh(); err != nil {
-			return nil, err
-		}
-		hdr, err = bt.bearer.Authorization()
+	sendRequest := func() (*http.Response, error) {
+		hdr, err := bt.bearer.Authorization()
 		if err != nil {
 			return nil, err
 		}
 
+		// http.Client handles redirects at a layer above the http.RoundTripper
+		// abstraction, so to avoid forwarding Authorization headers to places
+		// we are redirected, only set it when the authorization header matches
+		// the registry with which we are interacting.
 		if in.Host == bt.registry.RegistryStr() {
 			in.Header.Set("Authorization", hdr)
 		}
 		in.Header.Set("User-Agent", transportName)
+		return bt.inner.RoundTrip(in)
+	}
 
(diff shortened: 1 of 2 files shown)


pkg/v1/remote/transport/bearer.go
```diff
@@ -60,8 +60,27 @@ func (bt *bearerTransport) RoundTrip(in *http.Request) (*http.Response, error) {
 	}
 	in.Header.Set("User-Agent", transportName)
 
-	// TODO(mattmoor): On 401s perform a single refresh() and retry.
-	return bt.inner.RoundTrip(in)
+	res, err := bt.inner.RoundTrip(in)
+
+	// Perform a token refresh() and retry the request in case the token has exipred
```
jonjohnsonjr: typo nit: expired

ianberinger: fixed

pkg/v1/remote/transport/bearer.go
```diff
@@ -60,8 +60,27 @@ func (bt *bearerTransport) RoundTrip(in *http.Request) (*http.Response, error) {
 	}
 	in.Header.Set("User-Agent", transportName)
 
-	// TODO(mattmoor): On 401s perform a single refresh() and retry.
-	return bt.inner.RoundTrip(in)
+	res, err := bt.inner.RoundTrip(in)
+
+	// Perform a token refresh() and retry the request in case the token has expired
+	if err == nil && res.StatusCode == http.StatusUnauthorized {
```
ImJasonH: I don't think we want to refresh and retry if `err == nil`, only if it was 401.

I'd also like to see some kind of reusable method we can call recursively to retry instead of setting up and calling `inner.RoundTrip` again.

ianberinger: Makes sense, does an inline function that does prepares and sends the request sound reasonable to you?

something like this:
```
func (bt *bearerTransport) RoundTrip(in *http.Request) (*http.Response, error) {
  sendRequest := func() (*http.Response, error) {...}
  res, err := sendRequest()

  if err == nil && res.StatusCode == http.StatusUnauthorized {...}

  return res, err
}
```

ImJasonH: Yeah that's perfect. I might just short-circuit the error case:

```
res, err := sendRequest()
if err != nil {
  return nil, err
}
if res.StatusCode == http.StatusUnauthorized {
  bt.refresh()
  return sendRequest()
}
```

pkg/v1/remote/transport/bearer.go
```diff
@@ -46,22 +46,37 @@ var _ http.RoundTripper = (*bearerTransport)(nil)
...
+		hdr, err := bt.bearer.Authorization()
+		if err != nil {
+			return nil, err
+		}
+
+		// http.Client handles redirects at a layer above the http.RoundTripper
+		// abstraction, so to avoid forwarding Authorization headers to places
+		// we are redirected, only set it when the authorization header matches
+		// the registry with which we are interacting.
+		if in.Host == bt.registry.RegistryStr() {
```
ImJasonH: I haven't thought to do this in other situations where I'm injecting a `RoundTripper` -- maybe I should? Is there a specific case where you've seen this as a problem? Or is there some semi-official guidance that you should do this?

Maybe we should this change in a separate PR, just to be sure? And with tests, definitely.

ianberinger: This was introduced in https://github.com/google/go-containerregistry/commit/29f7682fa6959d1c388ccac0f4261eb3e8b36d10.
The commit message mentions that some registries redirect to object storage that don't like the Authorization header.
`transport.basicTransport` has this as well: https://github.com/google/go-containerregistry/blob/master/pkg/v1/remote/transport/basic.go#L42

It looks like all `RoundTripper`'s in `pkg/v1/remote/transport` are covered.

ImJasonH: My mistake, I misread the diff and it looked like this was added in this PR.

PR #285, opened 2018-10-10
 compressedImageExtender.LayerByDigest properly finds config layer 

### LINKED ISSUE #284: crane cannot push compressed image tarballs
`crane` pushes all blobs to the remote registry but fails on the config blob. Manually decompressing the layers in the image tarball (and changing `manifest.json`) allows crane to successfully push.

The problem seems to boil down to a difference in implementation of `LayerByDigest`. In the `uncompressedImageExtender`, the digest is checked against the config name and a config layer is returned. In `compressedImageExtender`, no such branch exists and the code iterates on the true image layers instead.

aduong:
This behaviour mirrors `uncompressedImageExtender.LayerByDigest`.

Fixes #284

### DIFF
--- a/pkg/v1/partial/configlayer_test.go
+++ b/pkg/v1/partial/configlayer_test.go
@@ -1,0 +1,97 @@
+package partial
+
+import (
+	"fmt"
+	"io/ioutil"
+	"testing"
+
+	"github.com/google/go-containerregistry/pkg/v1"
+	"github.com/google/go-containerregistry/pkg/v1/types"
+)
+
+type testUIC struct {
+	configFile []byte
+}
+
+func (t testUIC) RawConfigFile() ([]byte, error) {
+	return t.configFile, nil
+}
+
+func (t testUIC) MediaType() (types.MediaType, error) {
+	return "", nil
+}
+
+func (t testUIC) LayerByDiffID(h v1.Hash) (UncompressedLayer, error) {
+	return nil, fmt.Errorf("no layer by diff ID %v", h)
+}
+
+func TestUncompressedLayerByDigest(t *testing.T) {
+	uic := testUIC{
+		configFile: []byte("{}"),
+	}
+	uie := &uncompressedImageExtender{
+		UncompressedImageCore: uic,
+	}
+	testLayerByDigestForConfigHash(t, uie)
+}
+
+type testCIC struct {
+	configFile []byte
+}
+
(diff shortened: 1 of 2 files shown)


pkg/v1/partial/configlayer_test.go
```diff
@@ -0,0 +1,97 @@
...
+
+func (testCIC) RawManifest() ([]byte, error) {
+	return nil, nil
+}
+
+func (t testCIC) LayerByDigest(h v1.Hash) (CompressedLayer, error) {
+	return nil, fmt.Errorf("no layer by diff ID %v", h)
+}
+
+func TestCompressedLayersByDigest(t *testing.T) {
```
ImJasonH: Can these be condensed into a table-driven test?

```
for _, image := range []v1.Image {
  &compressedImageExtender{
    CompressedImageCore: testCIC{
      configFile: []byte("{}"),
    },
}, {
  &uncompressedImageExtender{
    UncompressedImageCore: testUIC{
      configFile: []byte("{}"),
    },
  },
}} {
  hash, err := image.ConfigName()
  ... and the rest of testLayerByDigestForConfigHash ...
}
```

pkg/v1/partial/configlayer_test.go
```diff
@@ -0,0 +1,97 @@
...
+
+type testCIC struct {
+	configFile []byte
+}
+
+func (t testCIC) RawConfigFile() ([]byte, error) {
+	return t.configFile, nil
+}
+
+func (t testCIC) MediaType() (types.MediaType, error) {
```
ImJasonH: If these methods are not called during the test, you can simply embed the `v1.Image` interface into `testCIC`/`testUIC` to have the test structs implement the interface. If the methods are ever called, they'll panic, but it helps reduce no-op code and prevent future breakages if you can get away with it.

```
type testCIC struct {
  v1.Image
  configFile []byte
}

func (t testCIC) RawConfigFile() ([]byte, error) { return t.configFile, nil }
```

aduong: Nice. I'm keeping `testCIC.LayerByDigest` around because otherwise it panics. I'll remove the rest.

pkg/v1/partial/configlayer_test.go
aduong: Absolutely! I'm keeping the definition of the cases separate from the loop, if that's okay. If it's idiomatic otherwise, I'm happy to embed the table into the `for`.

PR #287, opened 2018-10-15
Accept a digest reference in crane pull

ImJasonH:
Before this change:

```
$ crane pull "gcr.io/knative-releases/github.com/knative/build/cmd/controller@sha256:04fa3a155337c65db154083e3240ea3b309093f377536466a222b3e8d774ae08" output.tar
2018/10/15 13:15:51 parsing tag "gcr.io/knative-releases/github.com/knative/build/cmd/controller@sha256:04fa3a155337c65db154083e3240ea3b309093f377536466a222b3e8d774ae08": repository can only contain the runes `abcdefghijklmnopqrstuvwxyz0123456789_-./`: knative-releases/github.com/knative/build/cmd/controller@sha256
```

After this change:

```
$ go run ./cmd/crane/main.go pull "gcr.io/knative-releases/github.com/knative/build/cmd/controller@sha256:04fa3a155337c65db154083e3240ea3b309093f377536466a222b3e8d774ae08" output.tar
2018/10/15 13:53:36 Pulling gcr.io/knative-releases/github.com/knative/build/cmd/controller@sha256:04fa3a155337c65db154083e3240ea3b309093f377536466a222b3e8d774ae08
$ ls output.tar
output.tar
```

The only tricky part is that `tarball.WriteToFile` (and all of `tarball.Write`) requires that it be given a tag to write to the output tarball. If `crane pull` is given a tag, we'll use that as we always have. If it's given a digest, we'll chop off the digest string and pass only the image name (e.g., `gcr.io/knative-releases/github.com/knative/build/cmd/controller` in the example above)

cc @dprotaso

### DIFF
--- a/pkg/crane/pull.go
+++ b/pkg/crane/pull.go
@@ -38,19 +38,36 @@ func NewCmdPull() *cobra.Command {
 
 func pull(_ *cobra.Command, args []string) {
 	src, dst := args[0], args[1]
-	// TODO: Why is only tag allowed?
-	t, err := name.NewTag(src, name.WeakValidation)
+
+	ref, err := name.ParseReference(src, name.WeakValidation)
 	if err != nil {
 		log.Fatalf("parsing tag %q: %v", src, err)
 	}
-	log.Printf("Pulling %v", t)
+	log.Printf("Pulling %v", ref)
 
-	i, err := remote.Image(t, remote.WithAuthFromKeychain(authn.DefaultKeychain))
+	i, err := remote.Image(ref, remote.WithAuthFromKeychain(authn.DefaultKeychain))
 	if err != nil {
-		log.Fatalf("reading image %q: %v", t, err)
+		log.Fatalf("reading image %q: %v", ref, err)
 	}
 
-	if err := tarball.WriteToFile(dst, t, i); err != nil {
+	// WriteToFile wants a tag to write to the tarball, but we might have
+	// been given a digest.
+	// If the original ref was a tag, use that. Otherwise, if it was a
+	// digest, remove the digest from the ref instead.
+	tag, ok := ref.(name.Tag)
+	if !ok {
+		d, ok := ref.(name.Digest)
+		if !ok {
+			log.Fatal("ref wasn't a tag or digest")
+		}
+		s := d.Repository.Name()
+		tag, err = name.NewTag(s, name.WeakValidation)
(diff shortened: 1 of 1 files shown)


pkg/crane/pull.go
```diff
@@ -38,19 +38,36 @@ func NewCmdPull() *cobra.Command {
...
+	i, err := remote.Image(ref, remote.WithAuthFromKeychain(authn.DefaultKeychain))
 	if err != nil {
-		log.Fatalf("reading image %q: %v", t, err)
+		log.Fatalf("reading image %q: %v", ref, err)
+	}
+
+	// WriteToFile wants a tag to write to the tarball, but we might have
+	// been given a digest.
+	// If the original ref was a tag, use that. Otherwise, if it was a
+	// digest, remove the digest from the ref instead.
```
jonjohnsonjr: FWIW in the python lib this became "i-was-a-digest": https://github.com/google/containerregistry/blob/63ff239fa771e81237804b24ae5ff237cedbf42a/tools/docker_puller_.py#L58-L74

Removing the digest just makes this `latest` in my head... not sure what the actual behavior is, but a sentinel to denote that it was pulled by digest seems better.

ImJasonH: That's a really good idea. Done.

pkg/crane/pull.go
```diff
@@ -38,19 +44,36 @@ func NewCmdPull() *cobra.Command {
...
```
jonjohnsonjr: nit: update this comment

ImJasonH:
I don't have a problem with `crane push --no-tag` or whatever, but I think we should wait until someone actually asks for it. I'm not sure what the use case would be for it?

jonjohnsonjr:
What I have in mind is staging images in a repo without making them easily discoverable yet. So you'd push an image by digest, test that it works, then tag it with something meaningful to expose it. Another case is garbage collection -- usually my images are pinned by a tag to prevent accidentally deleting something that's being used. If I'm producing a lot of images in a dev loop (i.e. using ko), I often want to delete them immediately afterwards. Using `latest` nearly achieves this, but it leaves the last image I  pushed tagged (arguably this is the right behavior).

I don't know if many registries even support pushing by digest, so probably not worth doing.

dprotaso:
woo thanks for this fix!

Issue #289, opened 2018-10-18
Treat 10.x.x.x registry addresses as local/insecure


berndtj:
10.x.x.x registries should be considered insecure in the same manor registries suffixed with .local are.  This is needed to pull from in cluster insecure registries:

Needed for: https://github.com/GoogleContainerTools/kaniko/issues/405

jonjohnsonjr:
Fixing this actually exacerbates another issue: https://github.com/google/go-containerregistry/issues/279

We probably want the behavior to first attempt https and fallback to http only for things that match the heuristic... (really, we need a flag or env var or something).

berndtj:
Yeah, it does.  I think the real problem is that we don't have a distinction between source and destination within Kaniko, so the insecure flag only applies to destination.  Fallback actually sounds like a pretty good solution that would "just work".

berndtj:
I've got another fix which should fix both my issue and #279.  If we just do a fallback in the ping, it seems to work.  I'm still evaluating this.

jonjohnsonjr:
Essentially yes, but I don't like it :(

I think what I'd really want is for our behavior to just be the same as this: https://docs.docker.com/registry/insecure/

I don't love how we implicitly assume certain endpoints are safe to be insecure, but it's harder to plumb these flags/configs around, so I'm not sure what to do 🙃

/cc @mattmoor what do you think?

jonjohnsonjr:
@berndtj I haven't forgotten about this, just pondering still what the best path forward would be.

I think I'm fine with this PR going in as-is if we followup with the fallback.

re: the fallback proposal, I think it would be cleaner to pull the retry into the bearer transport's [Roundtrip](https://github.com/google/go-containerregistry/blob/0a6b857c44c4f1f33da0a8c8286bfaf830013d86/pkg/v1/remote/transport/bearer.go#L48).

I'm not sure if it's easy (or even possible) to discriminate TLS-related errors from other connection-related errors, but perhaps retrying anything is a good idea.

In the bearer transport we could add something like:
```go
	res, err := sendRequest()
	if err != nil {
                if bt.registry.Scheme() == "http" {
                        in.URL.Scheme = "http"
                        return sendRequest()
                }
		return nil, err
	}
```

[...]

PR #290, opened 2018-10-18
Assume RFC1918 addresses are insecure registries

berndtj:
This change assumes RFC1918 addresses are insecure and use the
http scheme.  This is the same as we assume for .local addresses.

With this change images can be built from local insecure registries,
not just pushed.

Fix: #289

Signed-off-by: Berndt Jung <bjung@vmware.com>

### DIFF
--- a/pkg/name/registry.go
+++ b/pkg/name/registry.go
@@ -15,6 +15,7 @@
 package name
 
 import (
+	"net"
 	"net/url"
 	"regexp"
 	"strings"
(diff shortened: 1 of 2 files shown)


jonjohnsonjr:
Thanks! I think the push/pull distinction bug for `--insecure` is actually in kaniko, whereas this gets around that bug by allowing http for all RFC1918 addresses.

Tiny nit: there's a typo in the PR and commit: `s/1916/1918/g`.

berndtj:
Doh! I'll amend that.

PR #291, opened 2018-10-20
Add remote.StreamableLayer which enables streaming layer uploads

ImJasonH:
Today, pushing an image requires every layer to be at least buffered into memory before pushing, because `remote.Write` uploads each layer by its digest. The registry API doesn't actually require this, and allows blob writes without specifying digest.

This buffering negatively impacts kaniko in cases where kaniko's layer snapshot diff exceeds available memory, resulting in OOMs. It walks a file tree accumulating files in a layer, but has to buffer it all before writing. 👎 

While reading the layer contents (e.g., during remote upload), we'll capture the digest, diff ID and size of the layer for later use in constructing the config and manifest.

This is only the first step in enabling streaming layer writes for kaniko; the next step is teaching `mutate.Append` not to assume each layer is buffered and to only calculate the image's config and manifest on-demand after layers have already been streamily read.

cc @dlorenc

### DIFF
--- a/pkg/v1/remote/stream.go
+++ b/pkg/v1/remote/stream.go
@@ -15,32 +15,102 @@
 package remote
 
 import (
-	"errors"
+	"compress/gzip"
+	"crypto/sha256"
+	"encoding/hex"
+	"hash"
 	"io"
 
 	"github.com/google/go-containerregistry/pkg/v1"
+	"github.com/google/go-containerregistry/pkg/v1/v1util"
 )
 
 type StreamableLayer struct {
-	Blob io.ReadCloser
+	blob io.ReadCloser
+
+	digest, diffID v1.Hash
+	size           int64
 }
 
 var _ v1.Layer = (*StreamableLayer)(nil)
 
-func (s *StreamableLayer) Digest() (v1.Hash, error) {
-	return v1.Hash{}, errors.New("don't know digest of streamed layer")
-}
-
-func (s *StreamableLayer) DiffID() (v1.Hash, error) {
-	return v1.Hash{}, errors.New("don't know digest of streamed layer")
-}
+func (s *StreamableLayer) Digest() (v1.Hash, error) { return s.digest, nil }
+func (s *StreamableLayer) DiffID() (v1.Hash, error) { return s.diffID, nil }
+func (s *StreamableLayer) Size() (int64, error)     { return s.size, nil }
 
 func (s *StreamableLayer) Uncompressed() (io.ReadCloser, error) {
-	return nil, errors.New("refusing to decompress streamed layer")
+	return newCollectReader(s, s.blob), nil
 }
 
-func (s *StreamableLayer) Compressed() (io.ReadCloser, error) { return s.Blob, nil }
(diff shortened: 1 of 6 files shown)


pkg/v1/remote/write_test.go
```diff
@@ -564,11 +571,78 @@ func TestUploadOne(t *testing.T) {
...
+		t.Errorf("Digest got %q, want %q", dig, wantDigest)
+	}
+	if diffID, err := sl.DiffID(); err != nil {
+		t.Errorf("DiffID: %v", err)
+	} else if diffID.String() != wantDiffID {
+		t.Errorf("DiffID got %q, want %q", diffID, wantDiffID)
+	}
+	if size, err := sl.Size(); err != nil {
+		t.Errorf("Size: %v", err)
+	} else if size != int64(n) {
```
jonjohnsonjr: (for all of these tests) I'm pretty sure the result of `Size` should be the size of the compressed layer -- this being == 10000 makes me think there's a bug in the implementation.

pkg/v1/remote/write_test.go
```diff
@@ -564,11 +571,78 @@ func TestUploadOne(t *testing.T) {
...
+	}
+	defer closer.Close()
+
+	n := 10000
+	sl := NewStreamableLayer(ioutil.NopCloser(bytes.NewBufferString(strings.Repeat("a", n))))
+	if err := w.uploadOne(sl); err != nil {
+		t.Fatalf("uploadOne: %v", err)
+	}
+
+	// Check sl.Digest, sl.DiffID, sl.Size
```
jonjohnsonjr: Left this behind, I think.

pkg/v1/remote/write_test.go
```diff
@@ -564,11 +571,78 @@ func TestUploadOne(t *testing.T) {
...
+			if r.Method != http.MethodPost {
+				t.Errorf("Method; got %v, want %v", r.Method, http.MethodPost)
+			}
+			w.Header().Set("Location", streamPath)
+			http.Error(w, "Initiated", http.StatusAccepted)
+		case streamPath:
+			if r.Method != http.MethodPatch {
+				t.Errorf("Method; got %v, want %v", r.Method, http.MethodPatch)
+			}
+			// TODO(jasonhall): What should we check here?
```
jonjohnsonjr: You could compute the sha256 to make sure it actually uploaded the right content?

pkg/v1/remote/stream.go
```diff
@@ -0,0 +1,138 @@
...
+// NewStreamableLayer returns a StreamableLayer backed by the ReadCloser, which
+// is assumed to contain the layer's uncompressed data.
+//
+// If called, the streamable layer's Compressed method will return the
+// compressed view of the data.
+//
+// Whether through Uncompressed or Compressed, closing the returned ReadCloser
+// will close the underlying original ReadCloser, and will populate the layer's
+// Digest, DiffID and Size; until the stream is consumed, these will return
+// zero values.
```
jonjohnsonjr: This is true but it's probably more useful to mention the error.

ImJasonH:
This is currently broken, do not merge. 👎

PR #293, opened 2018-10-21
Fix transport Host checking on redirect

sheeprine:
Transport roundTripper ensure that we are not sending sensitive
authentication information to incorrect URLs by checking it against the
Host value of the request. Sometimes it's emptied if it's not a relative
redirect for example in case of an http to https redirect.

See https://github.com/golang/go/blob/release-branch.go1.10/src/net/http/client.go#L543
for details, in case of an http to https redirect req.Host and
req.URL.Host are equals and so the Host value empty.


sheeprine:
Sure, I'll try to have them by tomorrow.

sheeprine:
@ImJasonH Don't know if you saw the update due to yesterday's issue with github.
Tell me if the test are sufficient.
I'm currently debugging kaniko against an Harbor registry, might have some more patching to do.

PR #297, opened 2018-11-02
Allow kubectl global flags in ko apply

### LINKED ISSUE #288: Allow extra kubectl options to pass through `ko apply`
Instead of this:
    `ko resolve -f config/ | kubectl apply -f - -n myns`
I'd like to do this:
    `ko apply -f config/ -n myns`

Essentially, pass any unrecognized options through to `kubectl apply`, excluding those that might be relevant to both, e.g. `-f`

### LINKED ISSUE #295: Support targeting custom namespaces with ko
As a developer on a project I may want to define the resources for that project in a namespace of my choosing. This may be because I have a restricted account which forces me to use a specific namespace. Currently, a project can only provide static namespace definitions and as such a user has no easy way to chose where namespace-scoped resources are defined.

jonjohnsonjr:
Fixes #288 
Fixes #295


jonjohnsonjr:
@mattmoor WDYT

PR #300, opened 2018-11-07
Strawman: better gc

jonjohnsonjr:
This has been sitting around for a while so I thought I'd just open it as-is for opinions re: UX and naming and additional functionality.

This PR adds support for filtering `gcrane gc` based on the image's upload time (`--before` and `--after`)  _and_ the ability to actually delete images (`--delete`) _and_ the ability to delete tags for a matching image (`--untag`).

The default behavior stays the same (just print untagged images). I'm not convinced that `--untag` is a very clear flag name, so I'm open to suggestions.

### DIFF
--- a/pkg/gcrane/gc.go
+++ b/pkg/gcrane/gc.go
@@ -47,7 +47,10 @@ func NewCmdGc() *cobra.Command {
 			if err != nil {
 				log.Fatalln(err)
 			}
-			GarbageCollect(args[0], opts)
+
+			if err := GarbageCollect(args[0], opts); err != nil {
+				log.Fatalln(err)
+			}
 		},
 	}
 
@@ -68,20 +71,22 @@ type GCOptions struct {
 	Untagged bool
 	Tags     []string
 
-	// Set version of Tags, for convenience.
-	tagSet map[string]struct{}
-
 	Before *time.Time
 	After  *time.Time
 
 	Recursive bool
 	Delete    bool
+
+	// Private set version of Tags, for convenience.
+	tagSet map[string]struct{}
 }
 
-func GarbageCollect(root string, opts GCOptions) {
+// GarbageCollect iterates over a repository, printing and optionally deleting
+// images that match the criteria in opts.
+func GarbageCollect(root string, opts GCOptions) error {
 	repo, err := name.NewRepository(root, name.WeakValidation)
 	if err != nil {
-		log.Fatalln(err)
+		return err
 	}
 
 	auth := google.WithAuthFromKeychain(authn.DefaultKeychain)
@@ -93,16 +98,11 @@ func GarbageCollect(root string, opts GCOptions) {
 	}
 
 	if opts.Recursive {
-		if err := google.Walk(repo, opts.walkFn, auth); err != nil {
-			log.Fatalln(err)
-		}
-		return
+		return google.Walk(repo, opts.walkFn, auth)
 	}
 
 	tags, err := google.List(repo, auth)
-	if err := opts.walkFn(repo, tags, err); err != nil {
-		log.Fatalln(err)
-	}
+	return opts.walkFn(repo, tags, err)
 }
 
 func (o *GCOptions) getMatchingRefs(repo name.Repository, digest string, manifest google.ManifestInfo) ([]name.Reference, error) {
--- a/pkg/gcrane/gc.go
+++ b/pkg/gcrane/gc.go
@@ -17,32 +17,68 @@ package gcrane
 import (
 	"fmt"
 	"log"
+	"net/http"
+	"time"
 
 	"github.com/google/go-containerregistry/pkg/authn"
 	"github.com/google/go-containerregistry/pkg/name"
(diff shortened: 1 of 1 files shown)


pkg/gcrane/gc.go
```diff
@@ -17,62 +17,182 @@ package gcrane
...
+	After  *time.Time
+
+	Recursive bool
+	Delete    bool
+	Untag     bool
+}
+
+// GarbageCollect iterates over a repository, printing and optionally deleting
+// images that match the criteria in opts.
+func GarbageCollect(root string, opts GCOptions) error {
```
ImJasonH: How about `pkg/v1/google.GarbageCollect`? This effectively just calls methods available in other packages and in `google` but provides a higher-level all-in-one method.

pkg/gcrane/gc.go
```diff
@@ -17,62 +17,182 @@ package gcrane
...
 }
 
-func gc(root string, recursive bool) {
+type GCOptions struct {
+	Before *time.Time
+	After  *time.Time
+
+	Recursive bool
+	Delete    bool
+	Untag     bool
```
ImJasonH: Doc comments pls

pkg/gcrane/gc.go
```diff
@@ -17,62 +17,182 @@ package gcrane
...
+
+			if err := GarbageCollect(args[0], opts); err != nil {
+				log.Fatalln(err)
+			}
 		},
 	}
 
+	// Filters
+	cmd.Flags().StringVarP(&before, "before", "b", "", "Match images uploaded before this time (RFC3339)")
+	cmd.Flags().StringVarP(&after, "after", "a", "", "Match images uploaded after this time (RFC3339)")
```
ImJasonH: I'm having trouble thinking of a use case for `--after`, maybe just drop this for now until we find one?

pkg/gcrane/gc.go
```diff
@@ -17,62 +17,182 @@ package gcrane
...
+
+	// Behaviors
 	cmd.Flags().BoolVarP(&recursive, "recursive", "r", false, "Whether to recurse through repos")
+	cmd.Flags().BoolVarP(&remove, "delete", "D", false, "Delete images instead of just printing them")
```
ImJasonH: What do you think about calling this `--dry_run` where true (the default) prints images it _would_ delete, and false would list and then delete them. `-D` could still work for the short name.

pkg/gcrane/gc.go
```diff
@@ -17,62 +17,182 @@ package gcrane
...
+	cmd.Flags().BoolVarP(&untag, "untag", "U", false, "Also untag matching images")
 
 	return cmd
 }
 
-func gc(root string, recursive bool) {
+type GCOptions struct {
```
ImJasonH: No functional options?

When we figure out how to reuse functional opts it would be cool to be able to have callers pass the keychain, transport, etc., too.

pkg/gcrane/gc.go
jonjohnsonjr: The scenario I had in mind is something like:

A bug in my Dockerfile means my automation system is producing bad images. While putting out that fire, we fix the bug in the Dockerfile but don't clean up the images immediately. We *know* that there are a few days of garbage images between X and Y. We don't want to delete any of the older good images (so we can rollback) and we don't want to delete any of the new images (because they're in pod).

pkg/gcrane/gc.go
jonjohnsonjr: If it's defaulting to true, then I think you have to do `-D=false` or something for the short name, which is gross. I made these all default to false so that an empty struct for `GCOptions` would work, but adding functional options could fix that... I'll think on it.

pkg/gcrane/gc.go
jonjohnsonjr: I'm not totally convinced re: functional options for this, mostly because I'm not sure how to massage the type system into doing what I want without duplicating the world again :(

pkg/gcrane/gc.go
jonjohnsonjr: That's kind of why I like it in `pkg/gcrane`, since that's exactly what gcrane does. `GarbageCollect` is a lot higher level thing than the primitives in `pkg/v1/google` and `pkg/v1/remote`, so I don't _really_ think it belongs there.

pkg/gcrane/gc.go
wdhnl: I like --dry_run as an option name (or --dry-run)

wdhnl:
My use case is deleting images matching a tag with a regular expression.  The tooling that publishes the images is broken and doesn't set the timestamp correctly but the images are tagged with a datestamp, so I'd like to be able to delete images based on that.  Even if timestamps were fixed, I still couldn't GC based on that alone because I'd want to leave intact any images tagged as :latest or similar even if they were old.

Example tagged image: gcr.io/my-repo/services/rickroll/server:release_20180502.16_RC00

jonjohnsonjr:
Closing in favor of https://github.com/google/go-containerregistry/pull/820

PR #301, opened 2018-11-12
Add pkg/v1/stream.NewLayer to stream layer contents

ImJasonH:
This will enable scenarios where layer contents might not fit into available memory before being written to a registry in `remote.Write`.

This PR augments `mutate` and `remote` to lazily compute digests, etc., only after layer contents have been consumed and presumably uploaded to the registry.

### DIFF
--- a/pkg/v1/mutate/mutate_test.go
+++ b/pkg/v1/mutate/mutate_test.go
@@ -30,6 +30,7 @@ import (
 
 	"github.com/google/go-cmp/cmp"
 	"github.com/google/go-containerregistry/pkg/v1"
+	"github.com/google/go-containerregistry/pkg/v1/stream"
 	"github.com/google/go-containerregistry/pkg/v1/tarball"
 )
 
@@ -233,19 +234,19 @@ func TestMutateConfig(t *testing.T) {
 	}
 
 	if manifestsAreEqual(t, source, result) {
-		t.Fatal("mutating the config MUST mutate the manifest")
+		t.Error("mutating the config MUST mutate the manifest")
 	}
 
 	if configFilesAreEqual(t, source, result) {
-		t.Fatal("mutating the config did not mutate the config file")
+		t.Error("mutating the config did not mutate the config file")
 	}
 
 	if configSizesAreEqual(t, source, result) {
-		t.Fatal("adding an enviornment variable MUST change the config file size")
+		t.Error("adding an environment variable MUST change the config file size")
 	}
 
 	if !reflect.DeepEqual(cfg.Config.Env, newEnv) {
-		t.Fatalf("incorrect environment set %v!=%v", cfg.Config.Env, newEnv)
+		t.Errorf("incorrect environment set %v!=%v", cfg.Config.Env, newEnv)
 	}
 }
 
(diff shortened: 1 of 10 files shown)


pkg/v1/mutate/mutate_test.go
```diff
@@ -233,19 +234,19 @@ func TestMutateConfig(t *testing.T) {
...
 	}
 
 	if configFilesAreEqual(t, source, result) {
-		t.Fatal("mutating the config did not mutate the config file")
+		t.Error("mutating the config did not mutate the config file")
 	}
 
 	if configSizesAreEqual(t, source, result) {
-		t.Fatal("adding an enviornment variable MUST change the config file size")
+		t.Error("adding an environment variable MUST change the config file size")
```
jonjohnsonjr: Nice

pkg/v1/stream/layer.go
```diff
@@ -0,0 +1,176 @@
...
+func newCompressedReader(l *Layer) (*compressedReader, error) {
+	h := sha256.New()
+	zh := sha256.New()
+	count := &countWriter{}
+
+	// gzip.Writer writes to the output stream via pipe, a hasher to
+	// capture compressed digest, and a countWriter to capture compressed
+	// size.
+	pr, pw := io.Pipe()
+	zw, err := gzip.NewWriterLevel(io.MultiWriter(pw, zh, count), gzip.BestSpeed)
```
jonjohnsonjr: This being hardcoded is frustrating, since I think our primary user of this will want to manipulate the compression level, see: https://github.com/google/go-containerregistry/pull/155

pkg/v1/stream/layer.go
```diff
@@ -0,0 +1,176 @@
...
+	pr    io.Reader
+	count *countWriter
+
+	l *Layer // stream.Layer to update upon Close.
+}
+
+func newCompressedReader(l *Layer) (*compressedReader, error) {
+	h := sha256.New()
+	zh := sha256.New()
+	count := &countWriter{}
```
jonjohnsonjr: Is writing to a `countWriter` materially slower than incrementing the count in one of the other writer's `Write`s? I suspect not because countWriter isn't actually doing anything with those bytes, but I'm not sure how clever go's compiler is. I do like this though because it's simple.

ImJasonH: Not that I know of. At least it's likely not slower enough to be worth complicating this code to optimize it. If we find out it's a problem in the future we can look into it.

pkg/v1/stream/layer.go
ImJasonH: Ooh thanks for pointing that out.

I only hard-coded `BestSpeed` so that it would pass the test comparing against the `tarball` implementation, since it [uses `BestSpeed`](https://github.com/google/go-containerregistry/blob/d54baf9aa28edb9b985a6b35b57e26e3410c2443/pkg/v1/v1util/zip.go#L29) when compressing previously uncompressed layer contents.

I don't have a problem adding `stream.NewLayerLevel`, or maybe a functional arg to specify the level. Is it okay if we do it in a followup PR? I suspect we'll find more of these sorts of things once Kaniko actually starts using it.

jonjohnsonjr: I'm fine with this as is.

PR #302, opened 2018-11-13
Don't return config 'layer' in ConfigByDigest

### LINKED ISSUE #87: We should disallow accessing the config through LayerBy*

ImJasonH:
Fixes #87

### DIFF
--- a/pkg/v1/remote/write_test.go
+++ b/pkg/v1/remote/write_test.go
@@ -33,6 +33,7 @@ import (
 	"github.com/google/go-containerregistry/pkg/authn"
 	"github.com/google/go-containerregistry/pkg/name"
 	"github.com/google/go-containerregistry/pkg/v1"
+	"github.com/google/go-containerregistry/pkg/v1/partial"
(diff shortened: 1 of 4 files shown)


pkg/v1/remote/write_test.go
```diff
@@ -440,7 +441,7 @@ func TestInitiateUploadMountsWithMountFromTheSameRegistry(t *testing.T) {
 
 func TestStreamBlob(t *testing.T) {
 	img := setupImage(t)
-	h := mustConfigName(t, img)
+	_ = mustConfigName(t, img) // TODO
```
jonjohnsonjr: todo?

ImJasonH: If you insist!

PR #303, opened 2018-11-20
Delete unused v1util functions

### LINKED ISSUE #32: Use io.Pipe for v1util
See: https://medium.com/stupid-gopher-tricks/streaming-data-in-go-without-buffering-3285ddd2a1e5

jonjohnsonjr:
Fixes #32

### DIFF
--- a/pkg/v1/v1util/zip.go
+++ b/pkg/v1/v1util/zip.go
@@ -70,52 +70,6 @@ func GunzipReadCloser(r io.ReadCloser) (io.ReadCloser, error) {
 	}, nil
 }
 
-// GzipWriteCloser returns an io.WriteCloser to which uncompressed data may be
-// written, and the compressed data is then written to the provided
-// io.WriteCloser.
-func GzipWriteCloser(w io.WriteCloser) io.WriteCloser {
-	gw := gzip.NewWriter(w)
-	return &writeAndCloser{
-		Writer: gw,
-		CloseFunc: func() error {
-			if err := gw.Close(); err != nil {
-				return err
-			}
-			return w.Close()
-		},
-	}
-}
-
-// gunzipWriteCloser implements io.WriteCloser
-// It is used to implement GunzipWriteClose.
-type gunzipWriteCloser struct {
-	*bytes.Buffer
-	writer io.WriteCloser
-}
-
-// Close implements io.WriteCloser
-func (gwc *gunzipWriteCloser) Close() error {
-	// TODO(mattmoor): How to avoid buffering this whole thing into memory?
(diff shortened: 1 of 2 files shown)


jonjohnsonjr:
@dprotaso is there a more idiomatic way of doing this?

pkg/v1/v1util/zip.go
```diff
@@ -86,33 +86,45 @@ func GzipWriteCloser(w io.WriteCloser) io.WriteCloser {
...
+		}
+		defer gr.Close()
+
+		_, err = io.Copy(w, gr)
+		if err != nil {
+			pw.CloseWithError(err)
+			return
+		}
+
+		flushed <- true
```
ImJasonH: You shouldn't need this signal. Instead, closing the pipe writer will signal the stream is done and the reader will only return EOF afterwards.

pkg/v1/v1util/zip.go
```diff
@@ -86,33 +86,45 @@ func GzipWriteCloser(w io.WriteCloser) io.WriteCloser {
...
+
+		gr, err := gzip.NewReader(pr)
+		if err != nil {
+			pw.CloseWithError(err)
+			return
+		}
+		defer gr.Close()
+
+		_, err = io.Copy(w, gr)
+		if err != nil {
```
ImJasonH: You can just call `CloseWithError(err)` here, if `err == nil` it's equivalent to `Close` (which you should call instead of using the flushed chan).

jonjohnsonjr:
@ImJasonH PTAL.

pkg/v1/v1util/zip.go
```diff
@@ -70,52 +70,6 @@ func GunzipReadCloser(r io.ReadCloser) (io.ReadCloser, error) {
...
-// io.WriteCloser.
-func GunzipWriteCloser(w io.WriteCloser) (io.WriteCloser, error) {
-	return &gunzipWriteCloser{
-		Buffer: bytes.NewBuffer(nil),
-		writer: w,
-	}, nil
-}
-
 // IsGzipped detects whether the input stream is compressed.
 func IsGzipped(r io.Reader) (bool, error) {
```
ImJasonH: We should try to get rid of this too, it consumes the stream it checks. Otherwise, we should make it return a reconstructed Reader with those consumed bytes prepended.

jonjohnsonjr:
Moot now since we're deleting them 😄 but I was referring to blocking on a channel in `Close` to allow the pipe's goroutine's `io.Copy` to catch up to the caller's `Write`. It felt weird so maybe this is just a bad API.

PR #309, opened 2018-11-27
Surface credential helper output in errors

jonjohnsonjr:
If the helper returns an error, add the command output and binary name
to the error if the output doesn't contain credentials.

This makes things a lot easier to debug than the current state of
things, which looks like an HTTP request somehow returned an error code
of 1.


jonjohnsonjr:
@mattmoor any concerns with merging this?

mattmoor:
Nope

PR #315, opened 2018-12-02
Implement googley auth methods

jonjohnsonjr:
This caches tokens until they're expired, so it's much faster than using
the default keychain.

### DIFF
--- a/pkg/v1/google/auth_test.go
+++ b/pkg/v1/google/auth_test.go
@@ -1,0 +1,261 @@
+// Copyright 2018 Google LLC All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package google
+
+import (
+	"fmt"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"strings"
+	"testing"
+
+	"github.com/google/go-containerregistry/pkg/authn"
+	"github.com/google/go-containerregistry/pkg/name"
+	"golang.org/x/oauth2"
+)
+
+const (
+	// Fails to parse as JSON at all.
+	badoutput = ""
+
+	// Fails to parse token_expiry format.
+	badexpiry = `
+{
+  "credential": {
+    "access_token": "mytoken",
+    "token_expiry": "2018/12/02"
+  }
+}`
+
+	// Expires in 6,000 years. Hopefully nobody is using my software then.
+	success = `
+{
+  "credential": {
+    "access_token": "mytoken",
+    "token_expiry": "8018-12-02T04:08:13Z"
+  }
+}`
+)
+
+// We'll invoke ourselves with a special environment variable in order to mock
+// out the gcloud dependency of gcloudSource. The exec package does this, too.
+//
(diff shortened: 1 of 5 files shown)


jonjohnsonjr:
@dekkagaijin opening here just to show how I'd like to use it, but (after kubecon probably) it would be really great if I could consume the cred helper logic via an `oauth2.TokenSource` so we can automagically get the caching stuff for free. It wasn't too hard to massage gcloud's output into an `oauth2.Token` in here. I'd rather this be upstream than copying and pasting all the cred helper logic :)

pkg/v1/google/auth/env.go
```diff
@@ -0,0 +1,43 @@
...
+	"golang.org/x/oauth2"
+	googauth "golang.org/x/oauth2/google"
+)
+
+// NewEnvAuthenticator returns an authn.Authenticator that generates access
+// tokens from the environment we're running in.
+//
+// See: https://godoc.org/golang.org/x/oauth2/google#FindDefaultCredentials
+func NewEnvAuthenticator() (authn.Authenticator, error) {
+	scopes := []string{"https://www.googleapis.com/auth/cloud-platform"}
```
dekkagaijin: nit: make this a const

pkg/v1/google/auth/gcloud.go
```diff
@@ -0,0 +1,102 @@
...
+	"time"
+
+	"github.com/google/go-containerregistry/pkg/authn"
+	"golang.org/x/oauth2"
+)
+
+// NewGcloudAuthenticator returns an oauth2.TokenSource that generates access
+// tokens by shelling out to the gcloud sdk.
+func NewGcloudAuthenticator() (authn.Authenticator, error) {
+	// This is odd, but it's about what docker-credential-gcr does to test things.
```
dekkagaijin: I was directed toward `gcloud config config-helper` as the only supported way to obtain `gcloud`'s access_token. --force-auth-refresh ensures that we're getting a 1h token, and the --format makes it so we don't have to do any post-processing on the returned string.

pkg/v1/google/auth/gcloud.go
```diff
@@ -0,0 +1,102 @@
...
+type gcloudOutput struct {
+	Credential struct {
+		AccessToken string `json:"access_token"`
+		TokenExpiry string `json:"token_expiry"`
+	} `json:"credential"`
+}
+
+// expiryLayout matches the layout gcloud uses for token_expiry, see:
+// https://golang.org/pkg/time/#Parse.
+const expiryLayout = "2006-01-02T15:04:05Z"
```
dekkagaijin: `time.RFC3339`

dekkagaijin:
@jonjohnsonjr yeah I think I can look into it. like I said, there are some considerations that need to be made that make blanket caching impractical, but there's a solution that'll make 99% of people happier

pkg/v1/google/auth/gcloud.go
jonjohnsonjr: Thanks!

jonjohnsonjr:
I put a little bit more love into this, so I think it's ready for an actual review instead of a PoC.

jonjohnsonjr:
I think using TestMain and execing itself is breaking Codecov, but FWIW this is the only line that's not covered:
https://github.com/google/go-containerregistry/blob/d7328bffbccbc6f081d3ed509592653eca6cfb75/pkg/v1/google/auth.go#L42

pkg/v1/google/auth_test.go
```diff
@@ -0,0 +1,261 @@
...
+	}, {
+		env:        "badoutput",
+		wantPrefix: "failed to parse `gcloud config config-helper` output:",
+	}, {
+		env:        "badexpiry",
+		wantPrefix: "failed to parse gcloud token expiry:",
+	}}
+
+	for i, tc := range cases {
+		t.Run(fmt.Sprintf("cases[%d]", i), func(t *testing.T) {
```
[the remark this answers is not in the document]
jonjohnsonjr: Good idea, thanks.

PR #322, opened 2018-12-14
ko: log the import paths we are building

jonjohnsonjr:
When you run a ko command, it outputs nothing for a few seconds while it
builds go binaries. This change logs the import paths before we invoke
`go build` so that you know what it's doing.


jonjohnsonjr:
I think this is a bit more user friendly.

mattmoor:
Make sure that this doesn't break `ko resolve > release.yaml` pattern :)

jonjohnsonjr:
Works for me, `log` prints to stderr not stdout.

PR #326, opened 2018-12-18
ko: do builds in serial

jonjohnsonjr:
Attempt to fix a race in go build.

This makes ko take about 50% longer to execute, so we probably want to
revert it whenever the underlying issue is fixed.


mattmoor:
Bad.

jonjohnsonjr:
I believe https://github.com/google/go-containerregistry/pull/330 fixes this instead.

Issue #298, opened 2018-11-07
ko does not work with 1.11 modules


jasonkeene:
With go1.11 modules the source code for your module can live outside the `GOPATH`. This appears to break `ko`. When I run `apply` or `resolve` it does not modify `image: goImportPath`.

jasonkeene:
As a workaround I symlinked my source code into the default `GOPATH` (`~/go/src/...`).

mattmoor:
cc @jonjohnsonjr

PR #329, opened 2018-12-21
PoC: Prototype go module support

jonjohnsonjr:
Playing with a solution to https://github.com/google/go-containerregistry/issues/298

Really what we need is for `gobuild` to be smart enough to just find these import paths, but I'm not sure how to do make it do that.

Also, ideally, we could use the modload package, but it's internal :(

### DIFF
--- a/pkg/ko/build/gobuild.go
+++ b/pkg/ko/build/gobuild.go
@@ -16,8 +16,10 @@ package build
 
 import (
 	"archive/tar"
+	"bufio"
 	"bytes"
 	"errors"
+	"fmt"
 	gb "go/build"
 	"io"
 	"io/ioutil"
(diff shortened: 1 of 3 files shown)


jonjohnsonjr:
Oh man... I finally read [this page](https://blog.golang.org/modules2019) I've had open for a while. I should have read it before playing around with this:

> In the eight years that we’ve had GOPATH, an incredible amount of tooling has been created that assumes Go source code is stored in GOPATH. Moving to modules requires updating all code that makes that assumption. We’ve designed a new package, golang.org/x/tools/go/packages, that abstracts the operation of finding and loading information about the Go source code for a given target. This new package adapts automatically to both GOPATH and modules mode and is also extensible to tool-specific code layouts, such as the one used by Bazel. We’ve been working with tool authors throughout the Go community to help them adopt golang.org/x/tools/go/packages in their tools.

https://godoc.org/golang.org/x/tools/go/packages

Sigh. I'll look at using this.

PR #330, opened 2018-12-21
Set accurate manifest config digest when mutating config

ekcasey:
Signed-off-by: Emily Casey <ecasey@pivotal.io>
Signed-off-by: Matthew McNew <mmcnew@pivotal.io>

After upgrading `go-containerregistry` we were getting `MANIFEST_BLOB_UNKNOWN` errors when pushing images after the config had been mutated. It seems that the config digest in the manifest was not matching the actual digest of the config blob. We believe this PR fixes the error.

### DIFF
--- a/pkg/v1/mutate/mutate.go
+++ b/pkg/v1/mutate/mutate.go
@@ -88,21 +88,11 @@ func configFile(base v1.Image, cfg *v1.ConfigFile) (v1.Image, error) {
 	}
 
 	image := &image{
-		base:     base,
-		manifest: m.DeepCopy(),
+		base:       base,
+		manifest:   m.DeepCopy(),
+		configFile: cfg,
 	}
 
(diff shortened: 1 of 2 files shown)


pkg/v1/mutate/mutate.go
```diff
@@ -88,21 +88,12 @@ func configFile(base v1.Image, cfg *v1.ConfigFile) (v1.Image, error) {
...
-		return nil, err
-	}
-	d, sz, err := v1.SHA256(bytes.NewBuffer(rcfg))
-	if err != nil {
-		return nil, err
-	}
-	image.manifest.Config.Digest = d
-	image.manifest.Config.Size = sz
-	image.configFile = cfg
+
```
jonjohnsonjr: Remove extra newline?

ekcasey: removed

PR #336, opened 2019-01-03
ko apply --watch

mattmoor:
This adds experimental support for fsnotify integration for `ko apply`.

With this change you can run: `ko apply --watch -f config/` and it will perform the typical deployment step, and then begin to monitor the referenced go paths from each of the yaml files for changes using github.com/mattmoor/dep-notify (a Go dependency graph library built around fsnotify).

When a change is found, only the yaml files that reference the image in question are re-applied.  In watch mode, compilation failures are ignored (so they may be remedied and picked up by the watch).

### DIFF
--- a/cmd/ko/filestuff.go
+++ b/cmd/ko/filestuff.go
@@ -44,7 +44,11 @@ func addFileArg(cmd *cobra.Command, fo *FilenameOptions) {
 func enumerateFiles(fo *FilenameOptions) chan string {
 	files := make(chan string)
 	go func() {
+		// When we're done enumerating files, close the channel
 		defer close(files)
+		// When we are in --watch mode, we set up watches on the filesystem locations
+		// that we are supplied and continuously stream files, until we are sent an
+		// interrupt.
 		var watcher *fsnotify.Watcher
 		if fo.Watch {
 			var err error
@@ -55,15 +59,22 @@ func enumerateFiles(fo *FilenameOptions) chan string {
 			defer watcher.Close()
 		}
 		for _, paths := range fo.Filenames {
+			// Just pass through '-' as it is indicative of stdin.
 			if paths == "-" {
 				files <- paths
 				continue
 			}
+			// For each of the "filenames" we are passed (file or directory) start a
+			// "Walk" to enumerate all of the contained files recursively.
 			err := filepath.Walk(paths, func(path string, fi os.FileInfo, err error) error {
 				if err != nil {
 					return err
 				}
 
+				// If this is a directory, skip it if it isn't the current directory we are
+				// processing (unless we are in recursive mode).  If we decide to process
+				// the directory, and we're in watch mode, then we set up a watch on the
+				// directory.
 				if fi.IsDir() {
 					if path != paths && !fo.Recursive {
 						return filepath.SkipDir
@@ -71,8 +82,10 @@ func enumerateFiles(fo *FilenameOptions) chan string {
 					if watcher != nil {
 						watcher.Add(path)
 					}
+					// We don't stream back directories, we just decide to skip them, or not.
 					return nil
 				}
+
 				// Don't check extension if the filepath was passed explicitly
 				if path != paths {
 					switch filepath.Ext(path) {
@@ -81,7 +94,11 @@ func enumerateFiles(fo *FilenameOptions) chan string {
 					default:
 						return nil
 					}
+					// We weren't passed this explicitly, so elide the watch as we
+					// are already watching the directory.
 				} else {
+					// We were passed this directly, and so we may not be watching the
+					// directory, so watch this file explicitly.
 					if watcher != nil {
 						watcher.Add(path)
 					}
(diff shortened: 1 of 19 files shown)


mattmoor:
Oh, I probably have to `dep ensure` to pull in `dep-notify` or CI's not going to like this when it's not on `GOPATH`...

jonjohnsonjr:
I think you'll need to eventually run `./scripts/update-deps.sh` to remove all the symlinks (that break gcloud).

mattmoor:
Oh, scripts...  I looked in hack...  🙄

mattmoor:
This now incorporates a caching implementation of `build.Interface` and `publish.Interface` based around simply "future" abstractions for the return values for their respective abstractions.  This will improve non-watch apply performance when the same import path is referenced by multiple yaml files (e.g. [here](https://github.com/knative/serving/blob/7553a2341dddd52ae8691d72a4d2b17a2f03fefa/config/config-controller.yaml#L25) and [here](https://github.com/knative/serving/blob/7553a2341dddd52ae8691d72a4d2b17a2f03fefa/config/999-cache.yaml#L23)).

However, the main use case for this is to minimize the work done on a dep-notify when a yaml file contains N binary references, and only one of them changes (prior results may be used for unaffected targets).  [Here's](https://github.com/knative/build/blob/14db3d4a0d80ab4964dca7d4d5f0c0b0877651d7/config/controller.yaml#L32-L38) an example where that happens today.

In the latter case, when I edit the binary for `creds-init`, I only see the single binary being rebuilt; likewise for `git-init`.

pkg/ko/build/shared.go
```diff
@@ -0,0 +1,79 @@
...
+
+	m       sync.Mutex
+	results map[string]*future
+}
+
+// Caching implements Interface
+var _ Interface = (*Caching)(nil)
+
+// NewCaching wraps the provided build.Interface in an implementation that
+// shared build results for a given path until the result has been invalidated.
```
jonjohnsonjr: shares

pkg/ko/publish/future.go
```diff
@@ -0,0 +1,75 @@
...
+import (
+	"sync"
+
+	"github.com/google/go-containerregistry/pkg/name"
+)
+
+func newFuture(work func() (name.Reference, error)) *future {
+	// Create a channel on which to send the result.
+	ch := make(chan *result)
+	// Initiate the actual work, sending it's result
```
jonjohnsonjr: its

cmd/ko/filestuff.go
```diff
@@ -34,13 +36,24 @@ func addFileArg(cmd *cobra.Command, fo *FilenameOptions) {
 		"Filename, directory, or URL to files to use to create the resource")
 	cmd.Flags().BoolVarP(&fo.Recursive, "recursive", "R", fo.Recursive,
 		"Process the directory used in -f, --filename recursively. Useful when you want to manage related manifests organized within the same directory.")
+	cmd.Flags().BoolVarP(&fo.Watch, "watch", "W", fo.Watch,
+		"Continuously monitor the transitive dependencies of the passed yaml files, and redeploy whenever anything changes.")
 }
 
 // Based heavily on pkg/kubectl
 func enumerateFiles(fo *FilenameOptions) chan string {
```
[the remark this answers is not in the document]
mattmoor: This function was undercommented.  I went through and documented it, so that hopefully it is easier to follow.

pkg/ko/build/future.go
```diff
@@ -0,0 +1,75 @@
...
+import (
+	"sync"
+
+	v1 "github.com/google/go-containerregistry/pkg/v1"
+)
+
+func newFuture(work func() (v1.Image, error)) *future {
+	// Create a channel on which to send the result.
+	ch := make(chan *result)
+	// Initiate the actual work, sending it's result
```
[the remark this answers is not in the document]
mattmoor: done

pkg/ko/build/shared.go
mattmoor: Done.

pkg/ko/publish/future.go
mattmoor: Done

pkg/ko/publish/shared.go
```diff
@@ -0,0 +1,76 @@
...
+import (
+	"sync"
+
+	"github.com/google/go-containerregistry/pkg/name"
+	v1 "github.com/google/go-containerregistry/pkg/v1"
+)
+
+// Caching wraps a publisher implementation in a layer that shares publish results
+// for the same inputs using a simple "future" implementation.
+type Caching struct {
```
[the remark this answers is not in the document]
mattmoor: it does not, and I'm not.  this was a copy/paste vestige.

cmd/ko/filestuff.go
```diff
@@ -74,6 +111,23 @@ func enumerateFiles(fo *FilenameOptions) chan string {
 				log.Fatalf("Error enumerating files: %v", err)
 			}
 		}
+
+		// We're done watching the files we were passed and setting up watches.
+		// Now watch for changes in the watches we were set up and stream back
```
jonjohnsonjr: I can't parse this, should it be: "in the watches we set up"?

mattmoor: WHO WATCHES THE WATCHERS!?!?!

I rephrased

PR #342, opened 2019-01-08
Use google-specific authenticator for gcrane ls

jonjohnsonjr:
This cuts down on invocations of gcloud, which results in a dramatic
speedup for `gcrane ls -r`.

### DIFF
--- a/pkg/gcrane/list.go
+++ b/pkg/gcrane/list.go
@@ -18,7 +18,6 @@ import (
 	"fmt"
 	"log"
 
-	"github.com/google/go-containerregistry/pkg/authn"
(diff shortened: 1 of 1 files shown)


jonjohnsonjr:
```
$ time gcrane ls -r gcr.io/google-containers > /dev/null

real	8m3.650s
user	1m56.845s
sys	0m51.236s

$ git checkout google-auth-for-listing 
Switched to branch 'google-auth-for-listing'

$ go install

$ time gcrane ls -r gcr.io/google-containers > /dev/null

real	1m42.613s
user	0m5.002s
sys	0m3.850s
```

Much faster 😄

PR #345, opened 2019-01-11
Allow to override tags on publish

### LINKED ISSUE #181: ko should allow images to be tagged with other tags (not just :latest)

paivagustavo:
Sequence from #277. 

I have changed a little the context of the PR. The previous implementation only let you define additional tags, but when pushing an Image, you should have total control over the tags will be pushed.

There are moments when you don't want to publish a `latest` tag, for example when you need to generate a image from a hotfix from a old version of an application.

So with this in mind, when you pass a `--tag=` flag you override the `latest` default tag.

This closes #181 

Note: this requires some more testing.

### DIFF
--- a/cmd/ko/resolve.go
+++ b/cmd/ko/resolve.go
@@ -87,7 +87,7 @@ func makePublisher(no *NameOptions, lo *LocalOptions) (publish.Interface, error)
 
 		repoName := os.Getenv("KO_DOCKER_REPO")
 		if lo.Local || repoName == publish.LocalDomain {
-			return publish.NewDaemon(namer), nil
+			return publish.NewDaemon(namer, []string{}), nil
 		}
 		_, err := name.NewRepository(repoName, name.WeakValidation)
 		if err != nil {
--- a/cmd/ko/resolve.go
+++ b/cmd/ko/resolve.go
@@ -74,7 +74,7 @@ func makeBuilder() (*build.Caching, error) {
 	return build.NewCaching(innerBuilder)
 }
 
-func makePublisher(no *NameOptions, lo *LocalOptions) (publish.Interface, error) {
+func makePublisher(no *NameOptions, lo *LocalOptions, ta *TagsOptions) (publish.Interface, error) {
 	// Create the publish.Interface that we will use to publish image references
 	// to either a docker daemon or a container image registry.
 	innerPublisher, err := func() (publish.Interface, error) {
@@ -87,7 +87,7 @@ func makePublisher(no *NameOptions, lo *LocalOptions) (publish.Interface, error)
 
 		repoName := os.Getenv("KO_DOCKER_REPO")
 		if lo.Local || repoName == publish.LocalDomain {
-			return publish.NewDaemon(namer, []string{}), nil
+			return publish.NewDaemon(namer, ta.Tags), nil
 		}
 		_, err := name.NewRepository(repoName, name.WeakValidation)
 		if err != nil {
@@ -96,7 +96,8 @@ func makePublisher(no *NameOptions, lo *LocalOptions) (publish.Interface, error)
 
 		return publish.NewDefault(repoName,
 			publish.WithAuthFromKeychain(authn.DefaultKeychain),
-			publish.WithNamer(namer))
+			publish.WithNamer(namer),
+			publish.WithTags(ta.Tags))
 	}()
 	if err != nil {
 		return nil, err
@@ -109,7 +110,7 @@ func makePublisher(no *NameOptions, lo *LocalOptions) (publish.Interface, error)
 // resolvedFuture represents a "future" for the bytes of a resolved file.
 type resolvedFuture chan []byte
 
-func resolveFilesToWriter(fo *FilenameOptions, no *NameOptions, lo *LocalOptions, out io.WriteCloser) {
+func resolveFilesToWriter(fo *FilenameOptions, no *NameOptions, lo *LocalOptions, ta *TagsOptions, out io.WriteCloser) {
 	defer out.Close()
 	builder, err := makeBuilder()
 	if err != nil {
@@ -117,7 +118,7 @@ func resolveFilesToWriter(fo *FilenameOptions, no *NameOptions, lo *LocalOptions
 	}
 
 	// Wrap publisher in a memoizing publisher implementation.
-	publisher, err := makePublisher(no, lo)
+	publisher, err := makePublisher(no, lo, ta)
 	if err != nil {
 		log.Fatalf("error creating publisher: %v", err)
 	}
--- a/cmd/ko/tags.go
+++ b/cmd/ko/tags.go
@@ -1,0 +1,28 @@
+// Copyright 2018 Google LLC All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package main
+
+import (
+	"github.com/spf13/cobra"
+)
+
+type TagsOptions struct {
+	Tags []string
+}
+
+func addTagsArg(cmd *cobra.Command, ta *TagsOptions) {
+	cmd.Flags().StringSliceVar(&ta.Tags, "tags", ta.Tags,
+		"Which tags to use for the produced image instead of the default 'latest' tag.")
+}
--- a/cmd/ko/tags.go
+++ b/cmd/ko/tags.go
@@ -18,11 +18,12 @@ import (
 	"github.com/spf13/cobra"
 )
 
+// TagsOptions holds the list of tags to tag the built image
 type TagsOptions struct {
 	Tags []string
 }
 
 func addTagsArg(cmd *cobra.Command, ta *TagsOptions) {
-	cmd.Flags().StringSliceVar(&ta.Tags, "tags", ta.Tags,
+	cmd.Flags().StringSliceVarP(&ta.Tags, "tags", "t", []string{"latest"},
 		"Which tags to use for the produced image instead of the default 'latest' tag.")
 }
--- a/pkg/ko/publish/daemon.go
+++ b/pkg/ko/publish/daemon.go
@@ -49,13 +49,16 @@ func (d *demon) Publish(img v1.Image, s string) (name.Reference, error) {
 		return nil, err
 	}
 
-	for _, tagName := range d.tags {
+	tagsAndDigest := append(d.tags, h.Hex)
+
+	for _, tagName := range tagsAndDigest {
 		tag, err := name.NewTag(fmt.Sprintf("%s/%s:%s", LocalDomain, d.namer(s), tagName), name.WeakValidation)
 		if err != nil {
 			return nil, err
 		}
 		log.Printf("Loading %v", tag)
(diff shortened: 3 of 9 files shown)


paivagustavo:
As the person who made the initial implementation and one who reviewed it, I would appreciate the review from both, @markusthoemmes @mattmoor 

This is actually my first contribution to a Go project, any consideration will be very welcome.

pkg/ko/publish/daemon.go
```diff
@@ -47,14 +48,27 @@ func (d *demon) Publish(img v1.Image, s string) (name.Reference, error) {
 	if err != nil {
 		return nil, err
 	}
+
+	if len(d.tags) == 0 {
```
jonjohnsonjr: You can avoid this special casing if you just make the default value for `--tags` be `[]string{"latest"}`.

(The third parameter in `StringSliceVar` allows you to specify a default.)

cmd/ko/resolve.go
```diff
@@ -87,7 +87,7 @@ func makePublisher(no *NameOptions, lo *LocalOptions) (publish.Interface, error)
 
 		repoName := os.Getenv("KO_DOCKER_REPO")
 		if lo.Local || repoName == publish.LocalDomain {
-			return publish.NewDaemon(namer), nil
+			return publish.NewDaemon(namer, []string{}), nil
```
jonjohnsonjr: Shouldn't we also pass through the tags options for a local `resolve`?

pkg/ko/publish/options.go
```diff
@@ -72,3 +72,11 @@ func WithNamer(n Namer) Option {
 		return nil
 	}
 }
+
+// WithTags is a functional option for overriding the image tags
+func WithTags(tags []string) Option {
```
jonjohnsonjr: I think we want to use this in `makePublisher` as well. We just need to pass the TagsOptions through to it.

pkg/ko/publish/default.go
```diff
@@ -53,12 +55,17 @@ type Namer func(string) string
 //   ^--base--^ ^-------import path-------^
 func identity(in string) string { return in }
 
+// As some registries do not support push an image by digest, the default tag for pushing
```
jonjohnsonjr: small grammar nit: "... pushing an image by digest, ..."

cmd/ko/tags.go
```diff
@@ -0,0 +1,28 @@
...
+import (
+	"github.com/spf13/cobra"
+)
+
+type TagsOptions struct {
+	Tags []string
+}
+
+func addTagsArg(cmd *cobra.Command, ta *TagsOptions) {
+	cmd.Flags().StringSliceVar(&ta.Tags, "tags", ta.Tags,
```
jonjohnsonjr: I think i'd like this to be `StringSliceVarP` with `-t` as the shorthand so we can use it similarly to `docker build -t foo -t bar`. @ImJasonH @mattmoor thoughts?

paivagustavo:
I'll try to do these changes tonight. 

I've changed only the behaviour of `publish`, because of your [comment](https://github.com/google/go-containerregistry/issues/181#issuecomment-393299462) where you state:

>  I'm -1 on adding this to config anywhere.

But I can sure add this apply and resolve as well.

jonjohnsonjr:
Ah, sorry for the confusion, by that comment I meant I didn't want to add anything to [the config file](https://github.com/google/go-containerregistry/tree/master/cmd/ko#configuration-via-koyaml), since it's not project-specific, but user-specific.

pkg/ko/publish/daemon.go
paivagustavo: thanks, changed this.

pkg/ko/publish/options.go
paivagustavo: added this on the `makePublisher` function.

cmd/ko/tags.go
paivagustavo: that actually seems cool and very useful. I've added this too.

cmd/ko/resolve.go
paivagustavo: yup, add this here and on the `makePublisher`

pkg/ko/publish/daemon.go
```diff
@@ -47,14 +48,23 @@ func (d *demon) Publish(img v1.Image, s string) (name.Reference, error) {
...
 		return nil, err
 	}
+
+	for _, tagName := range d.tags {
+		tag, err := name.NewTag(fmt.Sprintf("%s/%s:%s", LocalDomain, d.namer(s), tagName), name.WeakValidation)
+		if err != nil {
+			return nil, err
+		}
+		log.Printf("Loading %v", tag)
+		if _, err := daemon.Write(tag, img); err != nil {
```
jonjohnsonjr: This is going to effectively "docker load" the tarball multiple times, which is slow. For this, we probably want to expose a new method in the `daemon` package to add a tag that just calls [`ImageTag`](https://github.com/moby/moby/blob/beef00cb2612754d73024fd89d417b3a7114f87f/client/image_tag.go#L12). I think that's out of scope for this PR, but can you just add a TODO? Something like:

`TODO: This is slow because we have to load the image multiple times. We should use client.ImageTag instead`

cc @nkubala does that sound right to you?

We could also possibly hide this by caching the result of `ImageLoad` via `map[v1.Image]<something>` and calling `ImageTag` instead of `ImageLoad` if the `v1.Image` matches... :man_shrugging:

pkg/ko/publish/daemon.go
```diff
@@ -47,14 +48,23 @@ func (d *demon) Publish(img v1.Image, s string) (name.Reference, error) {
...
+			return nil, err
+		}
+		log.Printf("Loaded %v", tag)
+	}
+
 	tag, err := name.NewTag(fmt.Sprintf("%s/%s:%s", LocalDomain, d.namer(s), h.Hex), name.WeakValidation)
```
jonjohnsonjr: I think we need to still load this one, since we'll be referencing it later. You could create a new list of tags that is `d.tags` + this tag, and use that for the above loop.

paivagustavo:
I've done everything you asked @jonjohnsonjr, with you could review this again. I guess we need to wait others to review this too.

In the process doing this, I found out that `ko run` is not documented on github and it has a bug when we pass a flag like `--preserve-import-paths` and any other, since it forwards any argument from the `ko` command to the `kubectl run` command, I think we need to address this too later in another issue, if you agree I'll open an issue for these.

pkg/ko/publish/daemon.go
paivagustavo: This is something that i thought while doing it, but I had the conclusion that it should be an improvement for another time. 
The same is valid to the `default.go` when doing `remote.Write`

jonjohnsonjr:
Yes that's definitely a bug in `ko run`, thanks for spotting it. Please open an issue :smile: cc @mattmoor

pkg/ko/publish/daemon.go
paivagustavo: Actually we don't. Any image that is published with a tag can be referenced by it's digest, so we don't need to upload it via the digest.

pkg/ko/publish/daemon.go
paivagustavo: @jonjohnsonjr  on the `default.go` we should be able to do this with `crane`? Just want to know to add a more specific TODO.

paivagustavo:
> Yes that's definitely a bug in `ko run`, thanks for spotting it. Please open an issue  cc @mattmoor

Opened issue #346, glad to help :)

pkg/ko/publish/daemon.go
paivagustavo: I guess there is not a way to tag an image, but we can generalize the commit function to do multiple pushes of the manifest for each tag.

pkg/ko/publish/daemon.go
jonjohnsonjr: Omg. I just typed up a huge response but unresolving the conversation made it go away. One second, I'll try to reproduce it.

jonjohnsonjr: I'm pretty sure we do still need this tag.

When using `--local`, we don't actually publish the images to a registry, we just sideload them into the docker daemon. Docker doesn't calculate these image digests until you push the image (it's expensive to calculate the digest because you have to gzip all the layers), so we can't reference them by their digests. Instead, we _tag_ these images with the digest, e.g.:

```
ko.local/webhook-37fa64b5b216d563332fd3210a121e52:5d887242fa90e49a31bb1f7caf9f2194de3890251acb65610f98fe60dd601cb1
```

instead of:

```
ko.local/webhook-37fa64b5b216d563332fd3210a121e52@sha256:5d887242fa90e49a31bb1f7caf9f2194de3890251acb65610f98fe60dd601cb1
```

It's a bit of a mess, but it's the only way to get minikube to work without a registry (as far as I know, we can't coerce docker into calculating a digest for us).

I don't have minikube installed, but you can see here that after running `ko resolve --tags "foo" -L -f config/` with your changes, docker doesn't know the digests for those images:
```
$ docker images --digests
REPOSITORY                                             TAG                 DIGEST                                                                    IMAGE ID            CREATED             SIZE
ko.local/webhook-37fa64b5b216d563332fd3210a121e52      foo                 <none>                                                                    4a98e85efff9        49 years ago        53.4MB
ko.local/activator-2d008c4c36689d255523e5fa210c0bec    foo                 <none>                                                                    9e529f7fe80a        49 years ago        61.2MB
ko.local/autoscaler-b41733a3e5694e8fc64777bdfd8e791a   foo                 <none>                                                                    f2315e73e8a3        49 years ago        62.1MB
ko.local/controller-43f0364ab2f6dab17267e80a1f6a4adc   foo                 <none>                                                                    48f90164b572        49 years ago        64.3MB
ko.local/queue-7204c16e44715cd30f78443fb99e0f58        foo                 <none>                                                                    dd1a04b833ea        49 years ago        50.9MB
```

pkg/ko/publish/daemon.go
jonjohnsonjr: I think we can optimize the `remote.Write` path a little bit to make it cheaper to add a tag to an existing image. We could also just expose a cheap "AddTag" function to do what is expected. Either way, I'm fine with just this TODO for now, we can revisit `remote.Write` later. It's much cheaper than the daemon side-loading thing (I think).

pkg/ko/publish/daemon.go
paivagustavo: Wow, that is a shame. Didn't know about this about the docker daemon, found some github issues from moby/docker that exposes this problem. I tested this on gcr, so i always end up getting an image with digest.

I'll add this tomorrow. So we actually only need to do this on the local daemon. Remote writes doens't need, right?

Thanks for clarifying this.

pkg/ko/publish/daemon.go
```diff
@@ -47,14 +48,27 @@ func (d *demon) Publish(img v1.Image, s string) (name.Reference, error) {
 	if err != nil {
 		return nil, err
 	}
+
+	tagsAndDigest := append(d.tags, h.Hex)
```
jonjohnsonjr: This is _probably_ fine in this case, but normally you'd want to use `copy`, then `append` for this.

It's a stupid go thing, but `append` manipulates the underlying array in strange ways that can lead to unexpected behavior: https://allegro.tech/2017/07/golang-slices-gotcha.html

paivagustavo: Hmm, didn't know about this, as you can see I am actually really new to go. I'll read about.

jonjohnsonjr:
Let's open a new PR for that since this one is ready to go.

PR #348, opened 2019-01-15
Work around kubectl buffering

mattmoor:
This works around buffering in kubectl that led me to workaround this issuewhen experimenting with demo material for ko [here]()https://github.com/mattmoor/kodemo/blob/90b630836e0237e67bfadb73c65bb5ad882c415e/config/service.yaml#L1-L936).

### DIFF
--- a/cmd/ko/commands.go
+++ b/cmd/ko/commands.go
@@ -131,7 +131,21 @@ func addKubeCommands(topLevel *cobra.Command) {
 			if err != nil {
 				log.Fatalf("error piping to 'kubectl apply': %v", err)
 			}
-			go resolveFilesToWriter(fo, no, lo, ta, stdin)
+
+			go func() {
+				// kubectl buffers data before starting to apply it, which
+				// can lead to resources being created more slowly than desired.
+				// In the case of --watch, it can lead to resources not being
+				// applied at all until enough iteration has occurred.  To work
+				// around this, we prime the stream with a bunch of empty objects
+				// which kubectl will discard.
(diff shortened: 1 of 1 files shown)


cmd/ko/commands.go
```diff
@@ -130,7 +130,20 @@ func addKubeCommands(topLevel *cobra.Command) {
...
 			}
-			go resolveFilesToWriter(fo, no, lo, stdin)
+
+			go func() {
+				// kubectl buffers data before starting to apply it, which
+				// can lead to resources being created more slowly than desired.
+				// In the case of --watch, it can lead to resources not being
+				// applied at all until enough iteration has occurred.  To work
+				// around this, we prime the stream with a bunch of empty objects
+				// which kubectl will discard.
```
jonjohnsonjr: Maybe add this before and/or after the loop:

```go
stdin.Write([]byte("# See https://github.com/google/go-containerregistry/pull/348\n"))
```

For anyone who somehow stumbles across this?

cmd/ko/commands.go
```diff
@@ -130,7 +130,20 @@ func addKubeCommands(topLevel *cobra.Command) {
...
+				for i := 0; i < 1000; i++ {
```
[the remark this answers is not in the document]
mattmoor: It was a round number that was big enough to get the job done :)

cmd/ko/commands.go
mattmoor: We only do this on the apply path, so opting for a comment in the code instead.

PR #350, opened 2019-01-15
Load image to daemon only once when writing multiples tags

paivagustavo:
follow up from #345 

With this, the image is written to the docker daemon only once with the first tag and the other tags are added via the `client.ImageTag`.

### DIFF
--- a/pkg/ko/publish/daemon.go
+++ b/pkg/ko/publish/daemon.go
@@ -72,7 +72,7 @@ func (d *demon) Publish(img v1.Image, s string) (name.Reference, error) {
 		if err != nil {
 			return nil, err
 		}
-		log.Printf("Added tagged %v", tagName)
+		log.Printf("Added tag %v", tagName)
 	}
 
 	return &digestTag, nil
--- a/pkg/ko/publish/daemon.go
+++ b/pkg/ko/publish/daemon.go
@@ -49,26 +49,27 @@ func (d *demon) Publish(img v1.Image, s string) (name.Reference, error) {
 		return nil, err
 	}
 
-	tagsAndDigest := append(d.tags, h.Hex)
-
-	for _, tagName := range tagsAndDigest {
+	var tags []name.Tag
+	for _, tagName := range d.tags {
 		tag, err := name.NewTag(fmt.Sprintf("%s/%s:%s", LocalDomain, d.namer(s), tagName), name.WeakValidation)
 		if err != nil {
 			return nil, err
 		}
-		log.Printf("Loading %v", tag)
-		// TODO: This is slow because we have to load the image multiple times.
-		// We should use client.ImageTag instead
-		if _, err := daemon.Write(tag, img); err != nil {
-			return nil, err
-		}
-		log.Printf("Loaded %v", tag)
+
+		tags = append(tags, tag)
 	}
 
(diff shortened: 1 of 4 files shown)


pkg/v1/daemon/write.go
```diff
@@ -44,6 +44,36 @@ var GetImageLoader = func() (ImageLoader, error) {
 	return cli, nil
 }
 
+// WriteAndTag writes the image with the first tag and add the others tags with the
+// client.ImageTag function in order to save unnecesary subsequents Writes
+func WriteAndTag(tags []name.Tag, img v1.Image) (string, error) {
```
jonjohnsonjr: I think I'd rather see this as a simpler:
```go
func Tag(src, dst name.Tag) error {
}
```

And move this logic of `Write` + `Tag` up into pkg/ko/publish/daemon.go, since it has a similar structure.

pkg/ko/publish/daemon.go
```diff
@@ -49,26 +49,27 @@ func (d *demon) Publish(img v1.Image, s string) (name.Reference, error) {
...
 
-	tag, err := name.NewTag(fmt.Sprintf("%s/%s:%s", LocalDomain, d.namer(s), h.Hex), name.WeakValidation)
+	digestTag, err := name.NewTag(fmt.Sprintf("%s/%s:%s", LocalDomain, d.namer(s), h.Hex), name.WeakValidation)
 	if err != nil {
 		return nil, err
 	}
+	tags = append(tags, digestTag)
+
+	log.Printf("Loading %v", tags)
+	if _, err := daemon.WriteAndTag(tags, img); err != nil {
```
jonjohnsonjr: I'd call `daemon.Write` with the `digestTag`, then loop over `d.tags` and call `daemon.Tag` for each of them. Then we can keep the `daemon` API simpler.

pkg/v1/daemon/write.go
paivagustavo: done.

pkg/ko/publish/daemon.go
paivagustavo: Indeed, I personally liked the result a lot, way simpler and more readable, thanks for the suggestion  :smile:

pkg/ko/publish/daemon_test.go
```diff
@@ -29,12 +29,19 @@ import (
 
 type MockImageLoader struct{}
 
+var Tags []string
```
paivagustavo: @jonjohnsonjr  this is the only thing that I felt that was a little hacky, do you have any suggestion here or is this ok?

pkg/v1/daemon/write.go
```diff
@@ -44,6 +44,17 @@ var GetImageLoader = func() (ImageLoader, error) {
 	return cli, nil
 }
 
+// Tag adds a tag to an already existent image.
+func Tag(src, dest name.Tag) error {
+	cli, err := GetImageLoader()
+
```
jonjohnsonjr: style nit: drop this newline

pkg/ko/publish/daemon_test.go
jonjohnsonjr: I think this approach is fine for now.

PR #352, opened 2019-01-27
Remove msg-s about using anonymous authenticator

distorhead:
These messages may mess up output of some program, that uses go-containerregistry as a library:

```
No matching credentials were found, falling back on anonymous
```

Maybe it should be optional? In this PR I've removed these messages, but it is not a good solution I guess. Don't know how to do it right way.

### DIFF
--- a/pkg/ko/publish/options.go
+++ b/pkg/ko/publish/options.go
@@ -15,7 +15,6 @@
 package publish
 
 import (
-	"log"
 	"net/http"
 
 	"github.com/google/go-containerregistry/pkg/authn"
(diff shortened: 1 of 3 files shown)


jonjohnsonjr:
How does it mess up the output? That goes to stderr, so you could just ` 2> /dev/null` to get rid of it...

I agree with you, though, we need some verbosity flags for the logs.

distorhead:
This is warning about using anonymous auth. But in our app there is a case with local development, where insecure docker registry with anonymous auth is normal.

Maybe at least configure `out` and `err` writers (default to `stdout` and `stderr`) in go-containerregistry initialization procedure. Then it will be possible to suppress some output.

jonjohnsonjr:
Addressed in https://github.com/google/go-containerregistry/pull/481

PR #359, opened 2019-02-05
Pull in the latest version of dep-notify to fix ko watch.

mattmoor:
With this `ko apply -W -f config/` on a directory outside of `GOPATH` seems to work properly.

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -341,11 +341,11 @@
 
 [[projects]]
   branch = "master"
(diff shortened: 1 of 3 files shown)


mattmoor:
Hmm, hold off on this for now.  I see rebuilds happening, but not re-applies.

mattmoor:
Ah, I've been using: https://github.com/google/go-containerregistry/pull/348  I didn't realize it handle merged.  I'll fix that up as well.

PR #360, opened 2019-02-05
Add http fallback

### LINKED ISSUE #279: Registry hosted at registry.<company>.local is always insecure
Hi!

We're hosting a private registry at registry.[company name].local. Removing the logic below for .local adresses we can host it as a secure registry.

https://github.com/google/go-containerregistry/blob/b0936d62e7942acf7b1b5d4c377caa5f0b7566a5/pkg/name/registry.go#L74

Since it already has the --insecure flag to force "http", I would like to remove the logic for .local adresses.

jonjohnsonjr:
This first attempts to use "https" for every request, falling back to
"http" if the registry matches our localhost heuristic or if it is
intentionally set to insecure via name.NewInsecureRegistry.

Fixes #279

### DIFF
--- a/pkg/v1/remote/transport/ping.go
+++ b/pkg/v1/remote/transport/ping.go
@@ -36,6 +36,9 @@ type pingResp struct {
 	// Following the challenge there are often key/value pairs
 	// e.g. Bearer service="gcr.io",realm="https://auth.gcr.io/v36/tokenz"
 	parameters map[string]string
+
+	// The registry's scheme to use. Communicates whether we fell back to http.
+	scheme string
 }
 
 func (c challenge) Canonical() challenge {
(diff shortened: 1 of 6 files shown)


jonjohnsonjr:
cc @berndtj

pkg/v1/remote/transport/ping.go
```diff
@@ -63,31 +63,43 @@ func parseChallenge(suffix string) map[string]string {
 func ping(reg name.Registry, t http.RoundTripper) (*pingResp, error) {
 	client := http.Client{Transport: t}
 
-	url := fmt.Sprintf("%s://%s/v2/", reg.Scheme(), reg.Name())
-	resp, err := client.Get(url)
-	if err != nil {
-		return nil, err
+	// Always try https first, falling back to http if we can.
```
mattmoor: I would do this here and propagate it to the other transports.  I don't think other transports should need to worry about fallback.

jonjohnsonjr: Done, PTAL.

PR #373, opened 2019-02-14
Make -B work with ko {resolve, apply}

### LINKED ISSUE #372: ko resolve and ko apply does not support base path option
It is documented that it should work in the commend, but doesn't look like it is implemented. `ko publish -B` works as expected, but `ko apply -B` and `ko resolve -B` does not actually use the base path.

See https://github.com/google/go-containerregistry/blob/master/cmd/ko/publish.go#L78 for the logic in `ko publish` versus https://github.com/google/go-containerregistry/blob/master/cmd/ko/resolve.go#L82 in `ko resolve`.

jonjohnsonjr:
Fixes #372

### DIFF
--- a/cmd/ko/flatname.go
+++ b/cmd/ko/flatname.go
@@ -19,6 +19,7 @@ import (
 	"encoding/hex"
 	"path/filepath"
 
+	"github.com/google/go-containerregistry/pkg/ko/publish"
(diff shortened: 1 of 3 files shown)


jonjohnsonjr:
cc @dgerd

PR #374, opened 2019-02-15
Add validate.Image and crane validate

### LINKED ISSUE #30: Consider adding a utility to validate Images

jonjohnsonjr:
Fixes #30

### DIFF
--- a/cmd/crane/doc/crane_validate.md
+++ b/cmd/crane/doc/crane_validate.md
@@ -1,0 +1,25 @@
+## crane validate
+
+Validate that an image is well-formed
+
+### Synopsis
+
+Validate that an image is well-formed
+
+```
+crane validate [flags]
+```
+
+### Options
+
+```
+      --daemon string    Name of image in daemon to validate
+  -h, --help             help for validate
+      --remote string    Name of remote image to validate
+      --tarball string   Path to tarball to validate
+```
+
+### SEE ALSO
+
+* [crane](crane.md)	 - Crane is a tool for managing container images
+
--- a/pkg/v1/validate/image.go
+++ b/pkg/v1/validate/image.go
@@ -1,0 +1,276 @@
+// Copyright 2018 Google LLC All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package validate
+
+import (
+	"archive/tar"
+	"bytes"
+	"compress/gzip"
+	"crypto/sha256"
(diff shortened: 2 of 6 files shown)


cmd/crane/doc/crane_validate.md
```diff
@@ -0,0 +1,25 @@
...
+Validate that an image is well-formed
+
+```
+crane validate [flags]
+```
+
+### Options
+
+```
+      --daemon string    Name of image in daemon to validate
```
ImJasonH: How about `crane validate [image]` (positional arg) and if it matches a remote image ref we assume it's a remote image ref, otherwise check for a tarball file. `-L` can tell us to look for an image in the daemon, as we do elsewhere in `crane` and `ko`.

pkg/v1/validate/image.go
```diff
@@ -0,0 +1,276 @@
...
+		return err
+	}
+
+	pcf, err := v1.ParseConfigFile(bytes.NewReader(rc))
+	if err != nil {
+		return err
+	}
+
+	if cn != hash {
+		return fmt.Errorf("mismatched config digest: ConfigName()=%s, SHA256(RawConfigFile())=%s", cn, hash)
```
ImJasonH: What do you think about collecting validation errors and reporting all of them at once at the end? This can help make debugging faster since you can see "the digest was wrong" (not that helpful on its own) alongside "the config content was wrong" ("oh _that's_ why the digest was wrong!").

pkg/v1/validate/image.go
```diff
@@ -0,0 +1,276 @@
...
+	layers, err := img.Layers()
+	if err != nil {
+		return err
+	}
+
+	digests := []v1.Hash{}
+	diffids := []v1.Hash{}
+	sizes := []int64{}
+	for _, layer := range layers {
+		// TODO: Test layer.Uncompressed.
```
ImJasonH: Should we do something to prevent some real smart person from consuming `stream.Layer` contents during validation, thus preventing it from being usable after that point?

It might just be a documentation issue. `stream.Layer` could make it clearer that you only get one chance to stream the layer, so it better be on its way to a `remote.Image` etc. if you want to do anything useful.

This makes me think we might want a "streaming-validating" image, which only validates things on-demand as the image is read. `validate.Image` could just return a `v1.Image` implementation that returns errors when it sees something that's invalid.

_That_ makes me wonder why validation is opt-in at all, and whether we should just always validate images for sanity all the time. We might break some clients, but only those that were producing invalid images, so 🤷‍♂️

cmd/crane/doc/crane_validate.md
jonjohnsonjr: I think there are some ambiguous cases and doing fallback means that we might match the wrong thing. I kind of like this pattern since it's easier to extend with new formats without having to thing about how that would interact with the others.

pkg/v1/validate/image.go
jonjohnsonjr: Sounds hard 🤷‍♂️

pkg/v1/validate/image.go
jonjohnsonjr: Done.

PR #378, opened 2019-02-20
PoC: ko py

jonjohnsonjr:


cmd/ko/commands.go
```diff
@@ -295,4 +320,113 @@ func addKubeCommands(topLevel *cobra.Command) {
 	addTagsArg(run, ta)
 
 	topLevel.AddCommand(run)
+
+	// TODO: Write docs
+	py := &cobra.Command{
+		Use:   "py -f FILENAME",
+		Short: "Print the input files with image references resolved to built/pushed image digests.",
+		Long:  `This sub-command finds import path references within the provided files, builds them into Go binaries, containerizes them, publishes them, and prints the resulting yaml.`,
```
dprotaso: It's not clear to me after reading this how it's different from `ko resolve`

dprotaso: Ah - I see the details here -  https://github.com/google/go-containerregistry/issues/379

cmd/ko/commands.go
```diff
@@ -295,4 +320,113 @@ func addKubeCommands(topLevel *cobra.Command) {
 	addTagsArg(run, ta)
 
 	topLevel.AddCommand(run)
+
+	// TODO: Write docs
+	py := &cobra.Command{
```
dprotaso: `py` - I originally thought we were supporting python.

Call it `copy` please lol

jonjohnsonjr: But the puns!

jonjohnsonjr: What about `ko migrate`?

comigrate (verb): (chiefly in biochemistry) To migrate together

PR #380, opened 2019-02-20
Name the app using the base of the importpath

### LINKED ISSUE #375: Name binary after importpath instead of 'ko-app'
The process name/entrypoint for all ko built containers is '/ko-app'. This makes it hard to decipher on the host which running '/ko-app' is from which container.

My proposal would be that we change it so that a container built from the importpath
`github.com/knative/serving/cmd/autoscaler`

has an entrypoint of
`/ko-app/autoscaler`

bradhoekstra:
Currently the entrypoint is always '/ko-app', which makes it difficult to tell which ko-built app is which from the host system.

This changes it so that a container built from the importpath
`github.com/knative/serving/cmd/autoscaler`

has an entrypoint of
`/ko-app/autoscaler`

Fixes #375

### DIFF
--- a/pkg/ko/build/gobuild.go
+++ b/pkg/ko/build/gobuild.go
@@ -31,7 +31,10 @@ import (
 	"github.com/google/go-containerregistry/pkg/v1/tarball"
 )
 
-const appDir = "/ko-app"
+const (
+	appDir             = "/ko-app"
+	defaultAppFilename = "ko-app"
+)
 
 // GetBase takes an importpath and returns a base v1.Image.
 type GetBase func(string) (v1.Image, error)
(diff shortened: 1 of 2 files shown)


pkg/ko/build/gobuild.go
```diff
@@ -249,8 +249,10 @@ func (gb *gobuild) Build(s string) (v1.Image, error) {
 	}
 	layers = append(layers, dataLayer)
 
+	appPath := filepath.Join(appDir, filepath.Base(s))
```
bradhoekstra: Are there any cases here where `filepath.Base(s)` would be invalid (ie. return "." or "/")?

As far as I can tell `ko` should only work on well-defined importpaths inside GOPATH so this shouldn't be an issue.

pkg/ko/build/gobuild.go
```diff
@@ -132,7 +132,7 @@ func tarBinary(binary string) (*bytes.Buffer, error) {
 		return nil, err
 	}
 	header := &tar.Header{
-		Name:     appPath,
+		Name:     name,
```
bradhoekstra: Any issues with just creating the file here without explicitly creating/setting permissions on the parent directory? I don't see any in my testing.

jonjohnsonjr: I think containerd gets mad if you don't, is that right? @dlorenc

dlorenc: I don't think it gets mad, it just creates missing directories with a different set of permissions than docker. If you dont care about the permissions of these parent directories, it should just work. If you care, it's best to create it explicitly.

pkg/ko/build/gobuild.go
jonjohnsonjr: We'd like to make it work outside of GOPATH, I think, for go modules. It might make sense to check for "." or "/" explicitly and return a better error just in case we hit that path for some reason. It would make future debugging a lot easier.

bradhoekstra: Alternatively, in that case we could just use the legacy app name and containers will continue to build.

bradhoekstra: In the case that this happens now the entrypoint will default to "/ko-app/ko-app".

pkg/ko/build/gobuild.go
bradhoekstra: Containerd was creating the directory with 0700 permissions, which seems like it would be an issue when running as anything but root.

I explicitly create the parent directory now with 0555 permissions.

Issue #119, opened 2018-04-29
Support Manifest Lists


mattmoor:
We should add support for manifest lists, if only to avoid needing this non-Fatal workaround for DockerHub:
```
2018/04/29 07:27:03 manifest digest: "sha256:90f24abe180424046a5d53f6fc6f9fdb8f79b835cb2fd7d1a782e4c30dfb5dcc" does not match Docker-Content-Digest: "sha256:d2518289e66fd3892c2dae5003218117abeeed2edbb470cba544aef480fb6b3a" for "index.docker.io/library/ubuntu:latest"
```

Jon, you're easily the best person for the job :)

jonjohnsonjr:
This will be a lot of work to make it compose well with everything.

As a starting point, we can just stash the manifest list and resolve it to the amd64/llinux linux it points to. Making this work transparently for e.g. `crane copy` will take some effort (i.e. we will need to teach each sink about manifest lists).

listx:
ping?

bbrowning:
Because Knative uses go-containerregistry, this bug prevents Knative from working with quay.io and likely several other registries. You'll see the manifest digest error message which is fatal for non-DockerHub repositories.

jonjohnsonjr:
@bbrowning I believe quay has a different issue, but your point is generally correct :) working on this now.

bbrowning:
@jonjohnsonjr Yes, the issue is probably a bit different after a conversation on the Knative Slack. It manifests in a similar error message around digests not matching, but it was pointed out to me that quay.io isn't on schema 2 yet.

PR #386, opened 2019-03-01
Implement remote.Index

jonjohnsonjr:
This adds support for reading an ImageIndex from a remote repository.

I refactored some of the shared code between `Index` and `Image` to a separate "fetcher" struct. We should probably pull all of this out into a separate remote client that is shared between the read and write path, since it's very similar, but that's out of scope for this PR.

See #119 

Write support incoming once this lands.

### DIFF
--- a/pkg/v1/remote/image.go
+++ b/pkg/v1/remote/image.go
@@ -35,7 +35,7 @@ import (
 
 // remoteImage accesses an image from a remote registry
 type remoteImage struct {
-	fetch        fetcher
+	fetcher
 	manifestLock sync.Mutex // Protects manifest
 	manifest     []byte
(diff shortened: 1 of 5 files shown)


pkg/v1/remote/image.go
```diff
@@ -34,12 +35,12 @@ import (
 
 // remoteImage accesses an image from a remote registry
 type remoteImage struct {
-	ref          name.Reference
-	client       *http.Client
+	fetch        fetcher
```
ImJasonH: Maybe consider embedding this instead?

jonjohnsonjr: Done.

Issue #388, opened 2019-03-01
Add an interface that represents a union of Image and ImageIndex


jonjohnsonjr:
When we do a `crane pull` from a registry, we don't know ahead of time if the artifact at that reference is an `Image` or an `ImageIndex`. For backwards compatibility, the registry will down-convert an ImageIndex to an Image for us, so we have to first try to pull as an `ImageIndex`, then try to pull as an `Image`.

It would be really nice if we had a method that returned a descriptor + the ability to turn that descriptor into an `Image` or an `ImageIndex`. Then we could know whether we're dealing with an image or an index and deal with it appropriately rather than having to do this fallback logic. As it is, we have to do an extra token exchange and re-fetch the manifest bytes when we do the fallback.

PR #389, opened 2019-03-01
crane/gcrane: Support copying Indexes

jonjohnsonjr:
This adds support to `crane cp` and `gcrane cp` for copying indexes.

### DIFF
--- a/pkg/crane/copy.go
+++ b/pkg/crane/copy.go
@@ -15,6 +15,7 @@
 package crane
 
 import (
+	"fmt"
 	"log"
 	"net/http"
 
@@ -51,6 +52,11 @@ func doCopy(_ *cobra.Command, args []string) {
 	}
 	log.Printf("Pushing %v", dstRef)
 
+	srcAuth, err := authn.DefaultKeychain.Resolve(srcRef.Context().Registry)
+	if err != nil {
+		log.Fatalf("getting creds for %q: %v", srcRef, err)
+	}
+
 	dstAuth, err := authn.DefaultKeychain.Resolve(dstRef.Context().Registry)
 	if err != nil {
 		log.Fatalf("getting creds for %q: %v", dstRef, err)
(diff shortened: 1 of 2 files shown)


jonjohnsonjr:
Rebased!

pkg/crane/copy.go
```diff
@@ -61,6 +56,24 @@ func doCopy(_ *cobra.Command, args []string) {
 		log.Fatalf("getting creds for %q: %v", dstRef, err)
 	}
 
+	// First, try to copy as an index.
+	// If that fails, try to copy as an image.
+	// We have to do things in this order because fallback logic exists in the
+	// registry to convert an index to an  image.
```
ImJasonH: Extra space

pkg/crane/copy.go
```diff
@@ -61,6 +56,24 @@ func doCopy(_ *cobra.Command, args []string) {
...
+	// TODO(#388): Figure out which artifact is returned at runtime.
+	idx, err := remote.Index(srcRef, remote.WithAuthFromKeychain(authn.DefaultKeychain))
+	if err == nil {
+		if err := remote.WriteIndex(dstRef, idx, dstAuth, http.DefaultTransport); err == nil {
+			// We succeeded, so return. If we encounter any errors just fall through.
+			// This is less than ideal.
```
ImJasonH: Yeah :(

This means `crane cp` gets slower in the common case where it's copying an image. Any ideas how to do this better?

pkg/gcrane/copy.go
```diff
@@ -63,8 +63,15 @@ func doCopy(args []string, recursive bool) {
...
 
-		if err := singleCopy(src, dst, srcAuth, dstAuth); err != nil {
-			log.Fatalf("failed to copy image: %v", err)
+		// First, try to copy as an index.
+		// If that fails, try to copy as an image.
+		// We have to try this second because fallback logic exists in the registry
+		// to convert an index to an  image.
+		// TODO(#388): Figure out which artifact is returned at runtime.
+		if err := copyIndex(src, dst, srcAuth, dstAuth); err != nil {
+			if err := copyImage(src, dst, srcAuth, dstAuth); err != nil {
```
ImJasonH: This version is a lot more readable, can we use this style in crane copy?

pkg/crane/copy.go
jonjohnsonjr: Yeah but not with our current interfaces:
https://github.com/google/go-containerregistry/issues/388

We need some return type for `remote.Get` or something that is a tagged union so we can collapse the remote image wave function down to either an image or an index, depending on what we get back from the registry.

As it is, this makes "crane cp" slower by a pretty small amount, since it only affects a single copy.

pkg/gcrane/copy.go
jonjohnsonjr: Done

PR #391, opened 2019-03-05
Allow tag and digest

### LINKED ISSUE #351: Parsing image references with both tag and sha fails
[name.ParseReference](https://github.com/google/go-containerregistry/blob/efb7e1b888e142e2c66af20fd44e76a939b2cc3e/pkg/name/ref.go#L41) does not handle Docker image references like `ubuntu:latest@sha256:868fd30a0e47b8d8ac485df174795b5e2fe8a6c8f056cc707b232d65b8a1ab68` (which Docker supports). This causes GoogleContainerTools/kaniko#535.

This could probably be fixed by adding some logic to `ParseReference` or implementing #78.

ViceIce:
fixes #351

### DIFF
--- a/pkg/name/digest.go
+++ b/pkg/name/digest.go
@@ -83,6 +83,11 @@ func NewDigest(name string, strict Strictness) (Digest, error) {
 		}
 	}
 
+	tag, err := NewTag(base, strict)
+	if err == nil {
+		base = tag.Repository.Name()
+	}
+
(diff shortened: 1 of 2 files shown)


jonjohnsonjr:
We might want to extend this to store an optional tag in the digest strcut so that `NewDigest(str).String() == str` but I think this is fine for now. Thanks!

PR #394, opened 2019-03-05
Add empty.Index

jonjohnsonjr:


pkg/v1/empty/index.go
```diff
@@ -0,0 +1,62 @@
...
+import (
+	"encoding/json"
+	"errors"
+
+	"github.com/google/go-containerregistry/pkg/v1"
+	"github.com/google/go-containerregistry/pkg/v1/partial"
+	"github.com/google/go-containerregistry/pkg/v1/types"
+)
+
+var Index = index()
```
ImJasonH: Why not just `var Index = emptyIndex{}` ?

ImJasonH: Doc comment?

jonjohnsonjr: I can't remember, I think golint complained because I was doing something wrong and this was an attempt to fix that.

PR #396, opened 2019-03-06
Dedupe layer uploads where possible

ImJasonH:
Fixes issue in https://github.com/GoogleContainerTools/kaniko/pull/599

Prior to #301 we deduped layers using `BlobSet`, but that was dropped when some layers might not be digestable ahead of time.

~~TODO: tests, I wanted to get this out in case there were comments about the overall direction.~~

### DIFF
--- a/pkg/v1/remote/write.go
+++ b/pkg/v1/remote/write.go
@@ -57,9 +57,21 @@ func Write(ref name.Reference, img v1.Image, auth authn.Authenticator, t http.Ro
 	}
 
 	// Upload individual layers in goroutines and collect any errors.
+	// If we can dedupe by the layer digest, try to do so. If the layer is
+	// a stream.Layer, we can't dedupe and might re-upload.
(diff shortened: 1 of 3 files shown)


pkg/v1/remote/write.go
```diff
@@ -57,9 +57,25 @@ func Write(ref name.Reference, img v1.Image, auth authn.Authenticator, t http.Ro
 	}
 
 	// Upload individual layers in goroutines and collect any errors.
+	// If we can dedupe by the layer digest, try to do so. If the layer is
+	// a stream.Layer, we can't dedupe and might re-upload.
 	var g errgroup.Group
+	seen := map[v1.Hash]struct{}{}
```
clrprod: If you call this uploaded - it'll probably be a little clearer.

Also if you use map[v1.Hash]bool, then you can do

if uploaded[h]

PR #398, opened 2019-03-06
Resolve manifest lists to manifests in remote.Image

jonjohnsonjr:
We currently rely on the registry to do fallback for us, but this is
unreliable:
* Some registries won't do it.
* This only works for amd64/linux

### DIFF
--- a/pkg/v1/remote/image.go
+++ b/pkg/v1/remote/image.go
@@ -21,6 +21,7 @@ import (
 	"io/ioutil"
 	"net/http"
 	"net/url"
+	"runtime"
 	"strings"
 	"sync"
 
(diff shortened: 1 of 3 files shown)


pkg/v1/remote/image.go
```diff
@@ -283,3 +294,38 @@ func (r *remoteImage) LayerByDigest(h v1.Hash) (partial.CompressedLayer, error)
...
+	}
+	for _, childDesc := range index.Manifests {
+		p := v1.Platform{
+			Architecture: "amd64",
+			OS:           "linux",
+		}
+		if childDesc.Platform != nil {
+			p = *childDesc.Platform
+		}
+		if runtime.GOARCH == p.Architecture && runtime.GOOS == p.OS {
```
dlorenc: Would it make sense to make runtime.GOARCH and runtime.GOOS parameters here? I've had cases where I want to resolve things for a different platform (think crane digest on a macbook).

jonjohnsonjr: Yeah we probably want to expose something like `WithPlatform` in our image options, good thinking.

dlorenc:
Looks like you're missing boilerplate in ././pkg/executor/push_test.go

pkg/v1/remote/image.go
jonjohnsonjr: Changed this to default to amd64/linux to get the previous behavior (kind of). We can override that with `WithPlatform` wherever it makes sense.

PR #400, opened 2019-03-07
Add tarball.LayerFromReader

clrprod:
This provides a more idiomatic way to create
layers from readers without having to understand Openers.

### DIFF
--- a/pkg/v1/tarball/layer.go
+++ b/pkg/v1/tarball/layer.go
@@ -15,6 +15,7 @@
 package tarball
 
 import (
+	"bytes"
 	"compress/gzip"
 	"io"
 	"io/ioutil"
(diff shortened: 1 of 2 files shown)


ImJasonH:
You may also be interested in [`stream.NewLayer`](https://godoc.org/github.com/google/go-containerregistry/pkg/v1/stream#NewLayer) which gives you a layer from an `io.ReadCloser`.

It was designed to enable uploading a large layer without having to buffer it all in memory or on disk. It has some downsides -- you can't get its digest until it's been consumed, and since it's a stream you can only consume it once -- but if all you want is to construct a layer from files on disk, there's a thing for that.

pkg/v1/tarball/layer.go
```diff
@@ -103,6 +103,18 @@ func LayerFromOpener(opener Opener) (v1.Layer, error) {
...
 
+// LayerFromReader returns a v1.Layer given a io.Reader.
+func LayerFromReader(reader io.Reader) (v1.Layer, error) {
+	// Buffering due to Opener requiring multiple calls.
+	a, err := ioutil.ReadAll(reader)
+	if err != nil {
+		return nil, err
+	}
+	return LayerFromOpener(func() (io.ReadCloser, error) {
+		bytes.NewReader(a), nil
```
jonjohnsonjr: Missing a return here.

jonjohnsonjr: Missing `return`

PR #401, opened 2019-03-07
Only attempt to mount for MountableLayer

jonjohnsonjr:
Uploading to Quay is failing with `INVALID_REQUEST: Invalid request`. My theory is this mounting quirk, but we still need to confirm this fixes things before merging.


pkg/v1/remote/write.go
```diff
@@ -200,10 +200,9 @@ func (w *writer) checkExistingManifest(h v1.Hash, mt types.MediaType) (bool, err
 func (w *writer) initiateUpload(from, mount string) (location string, mounted bool, err error) {
 	u := w.url(fmt.Sprintf("/v2/%s/blobs/uploads/", w.ref.Context().RepositoryStr()))
 	uv := url.Values{}
-	if mount != "" {
+	if mount != "" && from != "" {
+		// Quay will fail if we specify a "mount" without a "from".
```
mattmoor: Can you confirm that we started issuing `HEAD` requests?

Issue #390, opened 2019-03-04
Better logging


jonjohnsonjr:
We currently use the standard library's `log` package, but that has some issues:

https://github.com/google/go-containerregistry/issues/266
https://github.com/google/go-containerregistry/pull/352

There are some messages we'd like to get to users because traditionally they've been very useful in diagnosing problems. The way we're using `log` means that there's no way to turn that off.

One approach would be to log nothing, but expose issues through some diagnose command:
https://github.com/google/go-containerregistry/issues/246

We could also use a no-op logger by default that just discards messages, and register a real logger in our own binaries: https://golang.org/pkg/log/#New

We could also have a verbose logger that is useful for debugging that gets turned on via some flag.

ekcasey:
+1

For consumers of this library it would be nice to have control over the output without having to mess with the standard logger.

PR #405, opened 2019-03-14
authn/k8schain : fallthrough in case of `gcloud` command not available

vdemeester:
It's not the prettiest change ever, but it keeps the error in the general case, but allows to fallthrough in a `multikeychain`. This type of errors could be used for other chains too :angel: 

It fixes cases where using `google.Keychain` with `authn.NewMultiKeychain` where `gcloud` command is not available (see https://gist.github.com/vdemeester/c397ffb3fd19b4cc2ba3243dc4db9f83). It should use `Anonymous` instead of failing hard in that case.

Signed-off-by: Vincent Demeester <vdemeest@redhat.com>

### DIFF
--- a/pkg/v1/google/auth.go
+++ b/pkg/v1/google/auth.go
@@ -19,6 +19,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"log"
 	"os"
 	"os/exec"
 	"time"
@@ -58,6 +59,13 @@ func NewEnvAuthenticator() (authn.Authenticator, error) {
 // NewGcloudAuthenticator returns an oauth2.TokenSource that generates access
 // tokens by shelling out to the gcloud sdk.
 func NewGcloudAuthenticator() (authn.Authenticator, error) {
+	if _, err := exec.LookPath("gcloud"); err != nil {
+		// TODO(#390): Use better logger.
+		// gcloud is not available, fall back to anonymous
+		log.Println("gcloud binary not found")
+		return authn.Anonymous, nil
+	}
+
 	ts := gcloudSource{GetGcloudCmd()}
 
 	// Attempt to fetch a token to ensure gcloud is installed and we can run it.


jonjohnsonjr:
What does this fix?

vdemeester:
@jonjohnsonjr I've updated the description (should also update the commit message I think) : 

> It fixes cases where using `google.Keychain` with `authn.NewMultiKeychain` where `gcloud` command is not available (see https://gist.github.com/vdemeester/c397ffb3fd19b4cc2ba3243dc4db9f83)

pkg/v1/google/auth.go
```diff
@@ -115,6 +127,11 @@ func (gs gcloudSource) Token() (*oauth2.Token, error) {
 	// Don't attempt to interpret stderr, just pass it through.
 	cmd.Stderr = os.Stderr
 
+	if _, err := exec.LookPath("gcloud"); err != nil {
+		// gcloud is not available, fall back to anonymous
+		return nil, fallthroughErr{msg: "`gcloud` binary not found"}
```
jonjohnsonjr: What about just returning `authn.Anonymous, nil` here instead of a new error? Then I think you wouldn't need any other code changes... WDYT?

vdemeester: @jonjohnsonjr That was my first take, and the simplest :sweat_smile: But this means using `google.Keychain` alone will fail silently if `gcloud` is missing — I wasn't sure if that was what we want :angel: 

I can definitely update the PR to `return authn.Anonymous, nil` instead :wink:

jonjohnsonjr: I'd expect users of google.Keychain to have things set up appropriately. We don't have a great story for signaling non-fatal warnings (#390) unfortunately :/

I think the current behavior of returning Anonymous if the keychain failed to find a match (even due to error) is the best we can do for now. I wouldn't be opposed to adding something like:
```go
// TODO(#390): Use better logger.
log.Println("gcloud binary not found")
```

I don't think many people use `google.Keychain`, and those that do probably wouldn't mind that warning message for now 😄

vdemeester: alright, let's go for that then :wink:

vdemeester:
@jonjohnsonjr updated :wink:

pkg/v1/google/auth.go
```diff
@@ -115,6 +121,13 @@ func (gs gcloudSource) Token() (*oauth2.Token, error) {
 	// Don't attempt to interpret stderr, just pass it through.
 	cmd.Stderr = os.Stderr
 
+	if _, err := exec.LookPath("gcloud"); err != nil {
```
jonjohnsonjr: Can we check this earlier so we don't have to communicate with callers via `anonError`?

I'm thinking at the top of `NewGcloudAuthenticator` we do something like:
```go
cmd := GetGcloudCmd()
if _, err := exec.LookPath(cmd.Path); err != nil {
  // TODO(#390): Use better logger.
  // gcloud is not available, fall back to anonymous
  log.Println("gcloud binary not found")
  return authn.Anonymous, nil
}
ts := gcloudSource{cmd}
...
```

vdemeester: Fair, I’ll update ✌🏼

vdemeester:
@jonjohnsonjr updated :angel:

vdemeester:
@jonjohnsonjr yes sir :medal_military: I'll update that on pipeline :wink:

PR #406, opened 2019-03-14
Add --all-tags and --oci to crane push/pull

jonjohnsonjr:
This is a demo to work out any kinks with the OCI image layout library.
There are some TODOs we will probably want to address.

### DIFF
--- a/pkg/crane/pull.go
+++ b/pkg/crane/pull.go
@@ -144,7 +144,7 @@ func pullLayout(src, dst string) error {
 		// For now, we just do the fallback thing. try in
 		if err := appendIndex(path, ref, auth, annotation); err != nil {
 			if err := appendImage(path, ref, auth, annotation); err != nil {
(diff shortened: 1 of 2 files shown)


pkg/crane/pull.go
```diff
@@ -77,3 +96,73 @@ func pull(_ *cobra.Command, args []string) {
 		log.Fatalf("writing image %q: %v", dst, err)
 	}
 }
+
+func pullLayout(src, dst string) error {
+	// Open the layout.
+	path, err := layout.FromPath(dst)
+	if err != nil {
+		// TODO: This could be cleaner. We can have an entrypoint that inits the layout
```
glyn: Probably time to add such an entry point. E.g. `FromPathOrNew` could be a helper function encapsulating just the logic here.

jonjohnsonjr:
Rebasing this will be hard, just going to close it.

PR #409, opened 2019-03-16
Add remote.CheckPushPermission

ImJasonH:
Ideas welcome on a better name. I didn't want to add a more generic "can I (push | pull)?" method since it's already pretty trivial to pull an image, and not costly since you'll fail to pull the manifest first.

The immediate use is in Kaniko, which takes a `--destination` flag and could fail fast if the current creds don't have permission to push to that destination. `ko` could likewise fail fast before `go build`ing if the current creds don't have permission on `KO_DOCKER_REPO`.

Callers that already have an image constructed and just want to push it (e.g., `crane push`) should call `remote.Write` to avoid unnecessary calls to `initiateUpload`.

### DIFF
--- a/pkg/v1/remote/check.go
+++ b/pkg/v1/remote/check.go
@@ -34,6 +34,21 @@ func CheckPushPermission(ref name.Reference, kc authn.Keychain, t http.RoundTrip
 		ref:    ref,
 		client: &http.Client{Transport: tr},
 	}
-	_, _, err = w.initiateUpload("", "")
+	loc, _, err := w.initiateUpload("", "")
+	if loc != "" {
+		// Since we're only initiating the upload to check whether we
+		// can, we should attempt to cancel it, in case initiating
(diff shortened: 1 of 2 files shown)


ImJasonH:
@jonjohnsonjr correctly pointed out that we shouldn't need to initiate an upload to check whether the user has push permissions, `transport.New` should achieve that simply by requesting a token from the registry's token server.

However, Dockerhub's token server seems to be non-compliant with Docker's own spec, and always responds with a token even if the request is anonymous or the repo doesn't exist, or it exists and the credentialed user doesn't have push access. Filed https://github.com/docker/hub-feedback/issues/1771 against the registry. Until that issue is resolved, to ensure the token that Dockerhub responds with can actually push, we have to attempt to initiate a push.

For GCR and other spec-compliant registries, we could pass the check after `transport.New` succeeds, but we'd need to special-case Dockerhub (and any other non-spec-compliant registries) to proceed to initiate the upload.

PR #414, opened 2019-03-20
Replace LayerFromOpener with LayerFromReader where it makes sense.

clrprod:
This pretty much removes every reference to LayerFromOpener except for mutate.go


clrprod:
Weird semantics emerge.

PR #422, opened 2019-03-27
Fix test to not depend on current runtime

jonjohnsonjr:
In #398 we changed the manifest resolution to default to amd64/linux but
didn't update the test, which expected resolution to use the current
runtime. Travis didn't catch this because it runs on amd64/linux.

### DIFF
--- a/pkg/v1/remote/image_test.go
+++ b/pkg/v1/remote/image_test.go
@@ -21,7 +21,6 @@ import (
 	"net/http"
 	"net/http/httptest"
 	"net/url"
-	"runtime"
 	"strings"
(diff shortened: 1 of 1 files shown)


jonjohnsonjr:
cc @dprotaso

dprotaso:
wooo

dprotaso:
For context the following platform on Mac (darwin/amd64) didn't match the default platform (amd64/linux) 

```go
manifest.Manifests[1].Platform = &v1.Platform{
	Architecture: runtime.GOARCH,
	OS:           runtime.GOOS,
}
```

PR #423, opened 2019-03-27
expose the layer media type

dprotaso:
Signed-off-by: Dave Protasowski <dprotaso@gmail.com>

### DIFF
--- a/pkg/v1/layer.go
+++ b/pkg/v1/layer.go
@@ -16,6 +16,8 @@ package v1
 
 import (
 	"io"
+
+	"github.com/google/go-containerregistry/pkg/v1/types"
 )
 
 // Layer is an interface for accessing the properties of a particular layer of a v1.Image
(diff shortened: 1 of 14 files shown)


ImJasonH:
Can you motivate this change a bit with some context? Is there some feature that's blocked by not having this?

dprotaso:
Folks seem to pushing new 'archive' types to registries

ie. 
https://github.com/google/crfs#introducing-stargz
https://github.com/containerd/containerd/issues/2267#issuecomment-455406681

Hence why it'd be nice if the `v1.Layer` were to expose the original media type of the layer.

jonjohnsonjr:
I'm not completely convinced... I had some similar changes at one point when I was playing with the OCI image layout stuff, but then I didn't actually use it in any implementation. I think the `Layer` interface is kind of strange because of `Compressed`/`Uncompressed`. Perhaps what those really should be are something like `Distributable` and `Tarball` to distinguish between passing through the registry vs consuming as a filesystem.

dprotaso:
I think it’s less about the original of the stream but the content aka
media type. ie. docker layers are a tar stream

tar most likely will be supplemented in the future by some other thing so
knowing the media type would be valuable

the above examples demonstrate that

Issue #424, opened 2019-03-27
Gcrane: unrecognized HTTP status: 429 Too Many Requests


paulros:
I'm trying to migrate a large number of data (over 38000) to a new GCR. Gcrane fails with a following error:
"failed to copy image: reading image "XXXXX": unrecognized HTTP status: 429 Too Many Requests" that means a Cloud Storage JSON API usage limit was exceeded.

Command executed: 
gcrane cp -r [old_GCR] [new_GCR]

How can it be fixed?

jonjohnsonjr:
Oh wow that's a lot of images. My implementation was pretty naive, and for that number of images probably hits some quota for GCR/GCS.

I can hack together something that does throttling to stay under [GCR's posted limits](https://cloud.google.com/container-registry/quotas).

Sorry about that! I'll post back here when I've put something together.

jonjohnsonjr:
@paulros sorry this took so long, let me know if that doesn't help things.

PR #425, opened 2019-03-28
Add some throttling to gcrane cp -r

jonjohnsonjr:
Overall QPS is based on the rate limits documented on gcr.io. I've also
added a task queue to limit the number of concurrent image uploads so
that we can ensure we make some forward progress uploading the images.

Fixes https://github.com/google/go-containerregistry/issues/424

### DIFF
--- a/pkg/gcrane/copy.go
+++ b/pkg/gcrane/copy.go
@@ -28,6 +28,7 @@ import (
 	"github.com/google/go-containerregistry/pkg/v1/types"
 	"github.com/spf13/cobra"
 	"golang.org/x/sync/errgroup"
+	"golang.org/x/time/rate"
 )
 
 func init() { Root.AddCommand(NewCmdCopy()) }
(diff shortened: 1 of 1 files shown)


pkg/gcrane/copy.go
```diff
@@ -144,83 +147,131 @@ func copyIndex(src, dst string, srcAuth, dstAuth authn.Authenticator) error {
...
-//					for each tag in index {
-//						go func {
-//							copyIndex(tag, rename(tag, dst))
-//						}
-//					}
-//				}
-//			}
-//		}
-//	}
+type throttledTransport struct {
```
ImJasonH: It seems a bit weird to throttle at the transport level, is that just because it's the easiest/lowest place to inject the wait?

It might make more sense to have the `rate.Wait` [in the loop where goroutines are pulling tasks from the `tasks` chan](https://github.com/google/go-containerregistry/blob/4c6c546743b750542d3aa6fb224ad533e6aa404a/pkg/gcrane/copy.go#L240), but then it's not enforcing HTTP-requests-per-second, it's limiting copy-operations-per-second, where each might involve multiple HTTP requests.

A more robust solution would be to have the transport backoff and retry when it gets a 429, and then we don't have to guess what the rate limit will be ahead of time, we can just do as much as we can until the server tells us to back off.

jonjohnsonjr: > It seems a bit weird to throttle at the transport level, is that just because it's the easiest/lowest place to inject the wait?

I think it's the most correct. GCR will throttle at the request level, so this is pretty close to the truth. I think we may be overcounting some requests (e.g. clients get redirected to GCS, we'll count the GCS request and the GCR redirect twice).

> A more robust solution would be to have the transport backoff and retry when it gets a 429, and then we don't have to guess what the rate limit will be ahead of time, we can just do as much as we can until the server tells us to back off.

 I think this is fine but we will effectively do a ton of work at once, then wait ten minutes before continuing. For interactive use cases that's not much better than failing (IMO), but for non-interactive cases this is probably better than doing any throttling... good point. I will update to do this.

jonjohnsonjr:
Closing in favor of more recent https://github.com/google/go-containerregistry/pull/512

PR #426, opened 2019-03-29
Rewrite remote streaming layer handling

jonjohnsonjr:
Assume errors in v1.Layer.Digest are related to streaming.

Retry any errors that could be related to streaming layers after we've
uploaded all layers.

Remove any references to stream.Layer from remote.

### DIFF
--- a/pkg/v1/remote/write.go
+++ b/pkg/v1/remote/write.go
@@ -28,7 +28,6 @@ import (
 	v1 "github.com/google/go-containerregistry/pkg/v1"
 	"github.com/google/go-containerregistry/pkg/v1/partial"
 	"github.com/google/go-containerregistry/pkg/v1/remote/transport"
-	"github.com/google/go-containerregistry/pkg/v1/stream"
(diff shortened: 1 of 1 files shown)


dprotaso:
Curious is stream.Layer used anywhere else?

pkg/v1/remote/write.go
```diff
@@ -58,16 +57,16 @@ func Write(ref name.Reference, img v1.Image, auth authn.Authenticator, t http.Ro
 
 	// Upload individual layers in goroutines and collect any errors.
 	// If we can dedupe by the layer digest, try to do so. If the layer is
-	// a stream.Layer, we can't dedupe and might re-upload.
+	// a streaming layer, we can't dedupe and might re-upload.
```
ImJasonH: This comment could be even more general. "If we can't determine the digest to dedupe for whatever reason..."

ImJasonH:
> Curious is stream.Layer used anywhere else?

Doesn't appear to be. Aside from `remote.Write` there's a passing reference to it in a comment in `pkg/v1/validate/image.go`.

PR #432, opened 2019-04-12
Add functional options to the rest of remote

jonjohnsonjr:
Ref https://github.com/google/go-containerregistry/issues/161

### DIFF
--- a/pkg/crane/append.go
+++ b/pkg/crane/append.go
@@ -16,7 +16,6 @@ package crane
 
 import (
 	"log"
-	"net/http"
 
 	"github.com/google/go-containerregistry/pkg/authn"
 	"github.com/google/go-containerregistry/pkg/name"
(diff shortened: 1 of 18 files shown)


pkg/crane/append.go
```diff
@@ -82,12 +81,7 @@ func doAppend(src, dst, tar, output string) {
 		return
 	}
 
-	dstAuth, err := authn.DefaultKeychain.Resolve(dstTag.Context().Registry)
-	if err != nil {
-		log.Fatalf("getting creds for %q: %v", dstTag, err)
-	}
-
-	if err := remote.Write(dstTag, image, dstAuth, http.DefaultTransport); err != nil {
+	if err := remote.Write(dstTag, image, remote.WithAuthFromKeychain(authn.DefaultKeychain)); err != nil {
```
ImJasonH: Should `Write` assume the default keychain? Or anonymous keychain?

jonjohnsonjr: Some thoughts...

0. IMO, default for _read_ operations should be anonymous, so you don't accidentally leak private images by using this library and exposing it to the internet.

BUT

1. There are probably very few world-writable registries out there, so `remote.Write` will essentially always need some credentials, so anonymous by default doesn't make sense.
2. Making `remote.Write`'s default behavior read from a config file for credentials would violate the principle of least astonishment (for me).
3. Deviating from the behavior of all the other methods would be surprising from a consumer perspective and annoying from our code perspective.

So IDK.

ImJasonH: I think I'm convinced. I don't think users should be surprised that they need to explicitly authorize `remote.Write`, and the default is easy to specify.

PR #434, opened 2019-04-18
Updated the error messages

### LINKED ISSUE #433: crane: error messages are unhelpful
```
$ crane pull ubuntu
Error: accepts 2 arg(s), received 1
Usage:
  crane pull [flags]
```

What it's complaining about is the lack of an arg to specify output path, but you wouldn't be able to tell from the error message, only from reading generated md docs.

A better error message would be something like `Error: must specify output path` with `Usage: crane pull [image] [output-path]`

hpandeycodeit:
Fixes #433

### DIFF
--- a/pkg/crane/pull.go
+++ b/pkg/crane/pull.go
@@ -34,8 +34,8 @@ func init() { Root.AddCommand(NewCmdPull()) }
 
 // NewCmdPull creates a new cobra.Command for the pull subcommand.
 func NewCmdPull() *cobra.Command {
-	cmd := &cobra.Command{
-		Use:   "pull",
+	return &cobra.Command{
+		Use:   "crane pull [image] [output-path]",
 		Short: "Pull a remote image by reference and store its contents in a tarball",
 		Args: func(cmd *cobra.Command, args []string) error {
 			if len(args) < 2 {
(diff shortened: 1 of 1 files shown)


pkg/crane/pull.go
```diff
@@ -34,12 +34,25 @@ func init() { Root.AddCommand(NewCmdPull()) }
 
 // NewCmdPull creates a new cobra.Command for the pull subcommand.
 func NewCmdPull() *cobra.Command {
-	return &cobra.Command{
+	cmd := &cobra.Command{
 		Use:   "pull",
```
jonjohnsonjr: I think you can just update `Use` to achieve the same thing.

hpandeycodeit: right. I don't why I did that. Reverted it now.

pkg/crane/pull.go
```diff
@@ -35,10 +35,15 @@ func init() { Root.AddCommand(NewCmdPull()) }
 // NewCmdPull creates a new cobra.Command for the pull subcommand.
 func NewCmdPull() *cobra.Command {
 	return &cobra.Command{
-		Use:   "pull",
+		Use:   "crane pull [image] [output-path]",
 		Short: "Pull a remote image by reference and store its contents in a tarball",
-		Args:  cobra.ExactArgs(2),
-		Run:   pull,
+		Args: func(cmd *cobra.Command, args []string) error {
+			if len(args) < 2 {
```
u5surf: We  also should handle when length of args is 0. So we can use switch statement instead of ifs.

hpandeycodeit: Is there a different error message when args equal to 0?

pkg/crane/pull.go
```diff
@@ -35,10 +35,15 @@ func init() { Root.AddCommand(NewCmdPull()) }
...
+				return errors.New("must specify output path")
```
jonjohnsonjr: This doesn't compile (missing `errors` import).

jonjohnsonjr:
Closing in favor of https://github.com/google/go-containerregistry/pull/584

PR #436, opened 2019-04-22
registry: Implement basic blob handler.

clrprod:
This doesn't handle chunked uploads yet - although its design allows for it.

### DIFF
--- a/pkg/registry/blobs.go
+++ b/pkg/registry/blobs.go
@@ -1,0 +1,110 @@
+package registry
+
+import (
+	"bytes"
+	"crypto/sha256"
+	"encoding/hex"
+	"fmt"
+	"io"
+	"math/rand"
+	"net/http"
+	"strings"
+)
+
+// Returns whether this url should be handled by the blob handler
+// This is complicated because blob is indicated by the trailing path, not the leading path.
+// https://github.com/opencontainers/distribution-spec/blob/master/spec.md#pulling-a-layer
+// https://github.com/opencontainers/distribution-spec/blob/master/spec.md#pushing-a-layer
+func isBlob(req *http.Request) bool {
+	elem := strings.Split(req.URL.Path, "/")
+	elem = elem[1:]
+	if len(elem) < 3 {
+		return false
+	}
+	return elem[len(elem)-2] == "blobs" || (elem[len(elem)-3] == "blobs" &&
+		elem[len(elem)-2] == "uploads")
+}
+
+// blobs
+type blobs struct {
+	// Blobs are content addresses. we store them globally underneath their sha and make no distinctions per image.
+	contents map[string][]byte
+	// Each upload gets a unique id that writes occur to until finalized.
+	uploads map[string][]byte
+}
+
+// https://github.com/opencontainers/distribution-spec/blob/master/spec.md#pulling-a-layer
+// https://github.com/opencontainers/distribution-spec/blob/master/spec.md#pushing-a-layer
+func (b *blobs) handle(resp http.ResponseWriter, req *http.Request) {
+	elem := strings.Split(req.URL.Path, "/")
+	elem = elem[1:]
+	// Must have a path of form /v2/{container}/blobs/{upload,sha256:}
+	if len(elem) < 4 {
+		resp.WriteHeader(http.StatusBadRequest)
+		return
+	}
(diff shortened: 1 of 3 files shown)


pkg/registry/blobs.go
```diff
@@ -0,0 +1,110 @@
...
+	if req.Method == "POST" && target == "uploads" {
+		id := fmt.Sprint(rand.Int63())
+		resp.Header().Set("Location", fmt.Sprintf("/v2/%s/blobs/uploads/%s",
+			strings.Join(elem[1:len(elem)-2], ","),
+			id))
+		resp.Header().Set("Range", "0-0")
+		resp.WriteHeader(http.StatusAccepted)
+		return
+	}
+
```
jonjohnsonjr: No `PATCH`? We always do `HEAD`, `POST`, `PATCH`, `PUT`, so I think this will fail with our own uploads.

pkg/registry/registry.go
```diff
@@ -19,6 +27,13 @@ func version(resp http.ResponseWriter, req *http.Request) {
 // New returns a handler which implements the docker registry protocol. It should be registered at the site root.
 func New() http.Handler {
 	m := http.NewServeMux()
-	m.HandleFunc("/v2/", version)
+	v := v{
+		blobs: blobs{
+
```
jonjohnsonjr: nit: extra newline

pkg/registry/blobs.go
```diff
@@ -0,0 +1,110 @@
...
+	// Each upload gets a unique id that writes occur to until finalized.
+	uploads map[string][]byte
+}
+
+// https://github.com/opencontainers/distribution-spec/blob/master/spec.md#pulling-a-layer
+// https://github.com/opencontainers/distribution-spec/blob/master/spec.md#pushing-a-layer
+func (b *blobs) handle(resp http.ResponseWriter, req *http.Request) {
+	elem := strings.Split(req.URL.Path, "/")
+	elem = elem[1:]
+	// Must have a path of form /v2/{container}/blobs/{upload,sha256:}
```
jonjohnsonjr: `{container}` here is not really a standard term, I'd use either `{repo}` or `{name}`

pkg/registry/blobs.go
```diff
@@ -0,0 +1,110 @@
...
+	"fmt"
+	"io"
+	"math/rand"
+	"net/http"
+	"strings"
+)
+
+// Returns whether this url should be handled by the blob handler
+// This is complicated because blob is indicated by the trailing path, not the leading path.
+// https://github.com/opencontainers/distribution-spec/blob/master/spec.md#pulling-a-layer
```
jonjohnsonjr: These links seem redundant with the `handle` comment.

pkg/registry/blobs.go
```diff
@@ -0,0 +1,110 @@
...
+func (b *blobs) handle(resp http.ResponseWriter, req *http.Request) {
+	elem := strings.Split(req.URL.Path, "/")
+	elem = elem[1:]
+	// Must have a path of form /v2/{container}/blobs/{upload,sha256:}
+	if len(elem) < 4 {
+		resp.WriteHeader(http.StatusBadRequest)
+		return
+	}
+	target := elem[len(elem)-1]
+	service := elem[len(elem)-2]
```
jonjohnsonjr: Why `service`?

pkg/registry/blobs.go
clrprod: Per https://github.com/opencontainers/distribution-spec/blob/master/spec.md#chunked-upload I think that's technically a chunked upload. I can add it (although I'd prefer to do it in another PR).

pkg/registry/blobs.go
clrprod: Deleted

pkg/registry/blobs.go
clrprod: I needed a name and calling it upload service seemed to work? If you have a suggestion I'm all ears.

pkg/registry/blobs.go
clrprod: Done.

pkg/registry/blobs.go
jonjohnsonjr: SGTM.

PR #443, opened 2019-04-22
registry: Full implementation

clrprod:
- Add manifest support
- Increase Coverage
- Support monolithic, stream and chunked uploads.
- Add locks for concurrency.
- Add headers and path changes to get docker + crane working.
- Store manifest content type to get docker working with crane pushed container.
- Return spec compatible error codes everywhere.
- Log all requests (and failures).

### DIFF
--- a/pkg/registry/blobs.go
+++ b/pkg/registry/blobs.go
@@ -19,6 +19,9 @@ import (
 func isBlob(req *http.Request) bool {
 	elem := strings.Split(req.URL.Path, "/")
 	elem = elem[1:]
+	if elem[len(elem)-1] == "" {
+		elem = elem[:len(elem)-1]
+	}
 	if len(elem) < 3 {
 		return false
 	}
@@ -35,13 +38,19 @@ type blobs struct {
 	lock    sync.Mutex
 }
 
-func (b *blobs) handle(resp http.ResponseWriter, req *http.Request) {
+func (b *blobs) handle(resp http.ResponseWriter, req *http.Request) *regError {
 	elem := strings.Split(req.URL.Path, "/")
 	elem = elem[1:]
+	if elem[len(elem)-1] == "" {
+		elem = elem[:len(elem)-1]
+	}
 	// Must have a path of form /v2/{name}/blobs/{upload,sha256:}
 	if len(elem) < 4 {
-		resp.WriteHeader(http.StatusBadRequest)
-		return
+		return &regError{
+			Status:  http.StatusBadRequest,
+			Code:    "NAME_INVALID",
+			Message: "blobs must be attached to a container",
+		}
 	}
 	target := elem[len(elem)-1]
 	service := elem[len(elem)-2]
@@ -51,14 +60,17 @@ func (b *blobs) handle(resp http.ResponseWriter, req *http.Request) {
 		defer b.lock.Unlock()
 		b, ok := b.contents[target]
 		if !ok {
-			resp.WriteHeader(http.StatusNotFound)
-			return
+			return &regError{
+				Status:  http.StatusNotFound,
+				Code:    "BLOB_UNKNOWN",
+				Message: "Unknown blob",
+			}
 		}
 
 		resp.Header().Set("Content-Length", fmt.Sprint(len(b)))
 		resp.Header().Set("Docker-Content-Digest", target)
 		resp.WriteHeader(http.StatusOK)
-		return
+		return nil
 	}
 
 	if req.Method == "GET" {
@@ -66,15 +78,18 @@ func (b *blobs) handle(resp http.ResponseWriter, req *http.Request) {
 		defer b.lock.Unlock()
 		b, ok := b.contents[target]
 		if !ok {
-			resp.WriteHeader(http.StatusNotFound)
-			return
+			return &regError{
+				Status:  http.StatusNotFound,
+				Code:    "BLOB_UNKNOWN",
+				Message: "Unknown blob",
+			}
 		}
 
 		resp.Header().Set("Content-Length", fmt.Sprint(len(b)))
 		resp.Header().Set("Docker-Content-Digest", target)
 		resp.WriteHeader(http.StatusOK)
 		io.Copy(resp, bytes.NewReader(b))
-		return
+		return nil
 	}
 
 	if req.Method == "POST" && target == "uploads" {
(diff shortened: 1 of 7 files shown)


pkg/registry/blobs.go
```diff
@@ -31,6 +32,7 @@ type blobs struct {
 	contents map[string][]byte
 	// Each upload gets a unique id that writes occur to until finalized.
 	uploads map[string][]byte
+	lock    sync.Mutex
```
jonjohnsonjr: https://golang.org/pkg/sync/#RWMutex instead?

pkg/registry/manifest.go
```diff
@@ -0,0 +1,83 @@
...
+	"bytes"
+	"fmt"
+	"io"
+	"net/http"
+	"strings"
+	"sync"
+)
+
+type manifests struct {
+	// maps container -> manifest tag/ digest -> manifest
```
jonjohnsonjr: I would probably s/container/repo/ everywhere.

jonjohnsonjr: nit: drop the space after `tag/`

pkg/registry/manifest.go
```diff
@@ -0,0 +1,83 @@
...
+	container := strings.Join(elem[1:len(elem)-2], "/")
+
+	if req.Method == "GET" {
+		m.lock.Lock()
+		defer m.lock.Unlock()
+		if _, ok := m.manifests[container]; !ok {
+			resp.WriteHeader(http.StatusNotFound)
+			return
+		}
+		m, ok := m.manifests[container][target]
```
jonjohnsonjr: You could store the previous access of `m.manifests[container]` instead of doing it twice.

pkg/registry/manifest.go
```diff
@@ -0,0 +1,83 @@
...
+		}
+		resp.Header().Set("Content-Length", fmt.Sprint(len(m)))
+		resp.WriteHeader(http.StatusOK)
+		return
+	}
+
+	if req.Method == "PUT" {
+		m.lock.Lock()
+		defer m.lock.Unlock()
+		if _, ok := m.manifests[container]; !ok {
```
jonjohnsonjr: Maybe hoist this check to the top and populate it if `req.Method == "PUT"`, then we can know that `m.manifests[container]` exists for the rest of `handle()`.

pkg/registry/registry.go
```diff
@@ -6,7 +6,8 @@ import (
 )
 
 type v struct {
-	blobs blobs
+	blobs     blobs
+	manifests manifests
```
jonjohnsonjr: What about `/v2/.../tags/list`?

clrprod: Listing tags isn't required
https://github.com/opencontainers/distribution-spec/blob/master/spec.md#listing-image-tags

Manifest has all the information though to do it.

pkg/registry/blobs.go
clrprod: It's simpler to not use it - If people actually run into performance issues that we can flip over.

pkg/registry/manifest.go
clrprod: That would require a global lock which I'd like to avoid.

pkg/registry/registry.go
```diff
@@ -1,4 +1,16 @@
-// Package registry implements a docker V2 registry.
+// Package registry implements a docker V2 registry and the OCI distribution specification.
+// 
+// It is designed to be used anywhere a low dependency container registry is needed, with an 
+// initial focus on tests. 
+//
+// It's goal is to be standards compliant and it's strictness will increase over time.
+//
+// As an example
```
jonjohnsonjr: Pull this out into an [example test](https://blog.golang.org/examples)?

pkg/registry/registry.go
```diff
@@ -1,4 +1,16 @@
-// Package registry implements a docker V2 registry.
+// Package registry implements a docker V2 registry and the OCI distribution specification.
+// 
+// It is designed to be used anywhere a low dependency container registry is needed, with an 
+// initial focus on tests. 
+//
+// It's goal is to be standards compliant and it's strictness will increase over time.
```
jonjohnsonjr: `s/it's/its/`

pkg/registry/registry.go
clrprod: Done

pkg/registry/registry.go
clrprod: Done

pkg/registry/blobs.go
```diff
@@ -31,78 +35,181 @@ type blobs struct {
...
+		elem = elem[:len(elem)-1]
+	}
 	// Must have a path of form /v2/{name}/blobs/{upload,sha256:}
 	if len(elem) < 4 {
-		resp.WriteHeader(http.StatusBadRequest)
-		return
+		return &regError{
+			Status:  http.StatusBadRequest,
+			Code:    "NAME_INVALID",
+			Message: "blobs must be attached to a container",
```
ImJasonH: must be attached to repo?

pkg/registry/manifest.go
```diff
@@ -0,0 +1,121 @@
...
+		}
+		m, ok := m.manifests[repo][target]
+		if !ok {
+			return &regError{
+				Status:  http.StatusNotFound,
+				Code:    "MANIFEST_UNKNOWN",
+				Message: "Unknown manifest",
+			}
+		}
+		rd := sha256.Sum256(m.blob)
```
ImJasonH: You could maybe use https://godoc.org/github.com/google/go-containerregistry/pkg/v1#SHA256 if you wanted to, I don't feel strongly about it.

pkg/registry/registry.go
```diff
@@ -1,38 +1,62 @@
-// Package registry implements a docker V2 registry.
+// Package registry implements a docker V2 registry and the OCI distribution specification.
+//
+// It is designed to be used anywhere a low dependency container registry is needed, with an
+// initial focus on tests.
+//
+// Its goal is to be standards compliant and its strictness will increase over time.
+
 package registry
```
ImJasonH: Should this be `pkg/v1/registry`? 🤔

pkg/registry/manifest.go
jonjohnsonjr: I think this is kind of nice as-is because it has zero dependencies outside of the standard library

ImJasonH: Yeah that makes sense. Makes me wonder whether we need `v1.SHA256` at all tbh.

pkg/registry/blobs.go
clrprod: Done.

pkg/registry/registry.go
clrprod: I don't think so? The go package field is the leaf element generally (i.e. v1 uses package v1).

The language spec indicates only letters and unicode digits are allowed here.
https://golang.org/ref/spec#Package_clause

pkg/registry/manifest.go
clrprod: I like the code having zero dependencies. This could be cleaned up with a helper function - but I'll probably only do this once I find a bug and have to touch it again.

PR #447, opened 2019-04-25
registry: Add a travis command line test for docker and crane compatibility

clrprod:
There is a weird bug in crane where it has on off by 1 error with writing tar files, so some of the test is commented out.

### DIFF
--- a/cmd/registry/test.sh
+++ b/cmd/registry/test.sh
@@ -1,0 +1,24 @@
+#!/bin/bash
+set -ev
+
+go install ./cmd/registry
+go install ./cmd/crane
+
+registry &
+PID=$!
+
+crane pull debian:latest debianc.tar
+crane push debianc.tar localhost:1338/debianc:latest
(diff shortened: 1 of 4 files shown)


clrprod:
Fixed the weird bug - was an off by 1 error in chunked uploads which docker trusted. Validating blob hashes + lengths during manifest upload would have caught this.

cmd/registry/test.sh
```diff
@@ -0,0 +1,26 @@
+#!/bin/bash
+set -ev
+
+go install github.com/google/go-containerregistry/cmd/registry
+go install github.com/google/go-containerregistry/cmd/crane
```
ImJasonH: Can this be `go build ./cmd/crane -o crane`, to make it clear that it's building the local source, and not pulling and installing code from the remote GitHub repo?

clrprod: Done.

PR #453, opened 2019-05-02
Generate mocks for image and image index

glyn:


pkg/v1/image.go
```diff
@@ -19,6 +19,7 @@ import (
 )
 
 // Image defines the interface for interacting with an OCI v1 image.
+//go:generate counterfeiter -o fake/FakeImage.go . Image
```
jonjohnsonjr: I think snake_case is the standard convention for golang files names? Just: `image.go`

pkg/v1/index.go
```diff
@@ -19,6 +19,7 @@ import (
 )
 
 // ImageIndex defines the interface for interacting with an OCI image index.
+//go:generate counterfeiter -o fake/FakeImageIndex.go . ImageIndex
```
jonjohnsonjr: Same here: `image_index.go`

glyn:
> LGTM other than filename nits. I'm assuming you can control that, but I haven't dug into counterfeiter so maybe they need to be PascalCase?

No, my mistake. I'll fix that.

pkg/v1/image.go
glyn: Fixed.

pkg/v1/index.go
glyn: Fixed.

jonjohnsonjr:
Thanks!

PR #454, opened 2019-05-03
[g]crane: Add version command

jonjohnsonjr:
Inspired by https://github.com/google/ko/pull/25

### DIFF
--- a/pkg/crane/version.go
+++ b/pkg/crane/version.go
@@ -1,0 +1,36 @@
+package crane
+
+import (
+	"fmt"
+	"log"
+	"os/exec"
+	"strings"
+
+	"github.com/spf13/cobra"
+)
+
+// Version can be set via:
+// -ldflags="-X 'github.com/google/go-containerregistry/pkg/crane.Version=$TAG'"
+var Version string
+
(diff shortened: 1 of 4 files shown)


pkg/crane/version.go
```diff
@@ -0,0 +1,36 @@
...
+// NewCmdVersion creates a new cobra.Command for the version subcommand.
+func NewCmdVersion() *cobra.Command {
+	return &cobra.Command{
+		Use:   "version",
+		Short: "Print the version",
+		Args:  cobra.NoArgs,
+		Run: func(_ *cobra.Command, _ []string) {
+			if Version == "" {
+				// If Version is unset, use the current commit.
+				hash, err := exec.Command("git", "rev-parse", "HEAD").Output()
```
ImJasonH: As mentioned in https://github.com/google/ko/pull/25 (sorry I didn't notice it in this PR), we should try to run `git rev-parse HEAD` from within ko's directory, not within the current working directory, where it makes no sense.

jonjohnsonjr: Oh wow I really wasn't thinking 🤦‍♂

PR #456, opened 2019-05-17
Add fallback Authorization without protocol

abergmeier:
This way docker-credential-secretservice works, too.

### DIFF
--- a/pkg/authn/helper.go
+++ b/pkg/authn/helper.go
@@ -80,11 +80,9 @@ func (h *helper) Authorization() (string, error) {
 		// Since some keychains allow to omit the protocol we have to
 		// fallback to test without protocol
 		fallbackCmd := exec.Command(helperName, "get")
-
 		fallbackCmd.Stdin = strings.NewReader(h.domain.String())
 		out.Reset()
 		fallbackCmd.Stdout = &out
-
 		cmdErr = h.r.Run(fallbackCmd)
 
 		// Even fallback was not found - give up now and fall back on
--- a/pkg/authn/helper.go
+++ b/pkg/authn/helper.go
@@ -74,11 +74,25 @@ func (h *helper) Authorization() (string, error) {
 	cmd.Stdout = &out
 	cmdErr := h.r.Run(cmd)
 
-	// If we see this specific message, it means the domain wasn't found
-	// and we should fall back on anonymous auth.
 	output := strings.TrimSpace(out.String())
+	// If we see this specific message, it means the URL wasn't found
 	if output == magicNotFoundMessage {
-		return Anonymous.Authorization()
+		// Since some keychains allow to omit the protocol we have to
+		// fallback to test without protocol
+		fallbackCmd := exec.Command(helperName, "get")
+
+		fallbackCmd.Stdin = strings.NewReader(h.domain.String())
+		out.Reset()
+		fallbackCmd.Stdout = &out
+
+		cmdErr = h.r.Run(fallbackCmd)
+
+		// Even fallback was not found - give up now and fall back on
+		// anonymous auth.
+		output = strings.TrimSpace(out.String())
+		if output == magicNotFoundMessage {
+			return Anonymous.Authorization()
+		}
 	}
 
 	// Any other output should be parsed as JSON and the Username / Secret
--- a/pkg/authn/helper_test.go
+++ b/pkg/authn/helper_test.go
@@ -15,8 +15,10 @@
 package authn
 
 import (
+	"bytes"
 	"errors"
 	"os/exec"
+	"strings"
 	"testing"
 
 	"github.com/google/go-containerregistry/pkg/name"
@@ -52,6 +54,17 @@ func (pr *printRunner) Run(cmd *exec.Cmd) error {
 	_, err := cmd.Stdout.Write([]byte(pr.msg))
 	return err
 }
+
+// customRunner implements runner to delegate to f
+type customRunner struct {
+	f func(*exec.Cmd) error
+}
+
+// Run implements runner
+func (cr *customRunner) Run(cmd *exec.Cmd) error {
+	return cr.f(cmd)
+}
+
 
 // errorPrintRunner implements runner to write a fixed message to stdout
 // and exit with an error code.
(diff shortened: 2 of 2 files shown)


abergmeier:
Also added tests, to prevent this from regressing.

jonjohnsonjr:
Is this a bug in docker-credential-secretservice or intentional behavior? Is there an issue we can link to in the comments?

pkg/authn/helper_test.go
```diff
@@ -136,3 +149,61 @@ func TestBadOutput(t *testing.T) {
...
+	output := `{"ServerURL":"","Username":"mainuser","Secret":"FVLGCQr_a-FZJM2ON227-YXdAEJn3oZZnnWs6Jd6iVE"}`
+	f := func(cmd *exec.Cmd) error {
+		buf := new(bytes.Buffer)
+		buf.ReadFrom(cmd.Stdin)
+		s := buf.String()
+
+		var err error
+
+		if strings.HasPrefix(s, "foo.dev") {
+			_, err = cmd.Stdout.Write([]byte(output))
```
ImJasonH: In practice, I think writing to stdout is pretty reliable. Maybe this can just be `fmt.Fprint(...)` instead, and just ignore the error?

pkg/authn/helper_test.go
```diff
@@ -136,3 +149,61 @@ func TestBadOutput(t *testing.T) {
...
+		return err
+	}
+
+	h := &helper{name: "test", domain: testDomain, r: &customRunner{f: f}}
+	got, err := h.Authorization()
+	if err != nil {
+		t.Fatalf("Authorization() = %v", got)
+	}
+
+	if got != `Basic bWFpbnVzZXI6RlZMR0NRcl9hLUZaSk0yT04yMjctWVhkQUVKbjNvWlpubldzNkpkNmlWRQ==` {
```
ImJasonH: Can you make it more obvious where this auth token comes from? Even just having it generated from base64-encoding consts used elsewhere will make it clearer, with a bonus that it'll be easy to change the consts in the future if we ever need to.

pkg/authn/helper_test.go
```diff
@@ -136,3 +149,61 @@ func TestBadOutput(t *testing.T) {
 		t.Errorf("Authorization() = %v", got)
 	}
 }
+
+// TestHTTPSURL checks that helper saving https works
+func TestHTTPSURL(t *testing.T) {
+	output := `{"ServerURL":"","Username":"mainuser","Secret":"FVLGCQr_a-FZJM2ON227-YXdAEJn3oZZnnWs6Jd6iVE"}`
```
ImJasonH: Is `ServerURL` needed? Does it not default to `""` if unset?

pkg/authn/helper.go
```diff
@@ -74,11 +74,25 @@ func (h *helper) Authorization() (string, error) {
...
-	// If we see this specific message, it means the domain wasn't found
-	// and we should fall back on anonymous auth.
 	output := strings.TrimSpace(out.String())
+	// If we see this specific message, it means the URL wasn't found
 	if output == magicNotFoundMessage {
-		return Anonymous.Authorization()
+		// Since some keychains allow to omit the protocol we have to
+		// fallback to test without protocol
+		fallbackCmd := exec.Command(helperName, "get")
+
```
ImJasonH: Nit: unnecessary blank line here

pkg/authn/helper_test.go
abergmeier: I definitely do not want to ignore an error. Using `fmt` would not make the code any smaller, I am afraid.

pkg/authn/helper_test.go
abergmeier: Done.

pkg/authn/helper_test.go
abergmeier: It is not needed. It is what _docker-credential-secretservice_ returns. And I want to ensure, that it does not cause any problems.

pkg/authn/helper.go
abergmeier: Done

abergmeier:
> Is this a bug in docker-credential-secretservice or intentional behavior?
Not sure. Does it state anywhere, that URLs have to be stored with a protocol prefix?
Or is this only the implementation detail of Google and AWS impls?

> Is there an issue we can link to in the comments?
Not that I know of.

abergmeier:
Is this good to merge?

jonjohnsonjr:
It seems like this should be fixed in docker-credential-secretservice. I believe this was done in https://github.com/docker/docker-credential-helpers/pull/72 for osx and https://github.com/docker/docker-credential-helpers/pull/139 for wincred.

abergmeier:
> It seems like this should be fixed in docker-credential-secretservice. I believe this was done in docker/docker-credential-helpers#72 for osx and docker/docker-credential-helpers#139 for wincred.

Are you sure about this? From a quick scan, I think these 2 patches rather change behavior for paths.
Did you find any spec stating that a scheme is enforced?

> I don't mind merging this if it's too much trouble to fix in secretservice, but I'd rather fix the actual problem than adding a workaround :P

From my perspective this is not a theoretical problem (lets just fix dependencies) but a bunch of practical problems:
1. We have this behavior currently in the wild with at least _docker-credential-secretservice_.
  Not supporting this means that you force people to (be able to) upgrade to a certain version of _docker-credential-secretservice_. At least in Enterprise world, this is not necessarily possible/easy.
2. The problem actually surfaces in other binaries (e.g. _Kaniko_). And the way that `Authorization` currently works, it is really hard to analyze - report that problem to the user.
  So not supporting this behavior IMO would at least force a rework of `helper` so that it no longer quietly falls back to Anonymous. And as a result probably also break the exposed interface of go-containerregistry. And then change using binaries.

jonjohnsonjr:
We added the scheme to work around an issue with osxkeychain https://github.com/google/go-containerregistry/blob/1c6c7f61e8a5402b606c3c6db169fdcd1b0712b7/pkg/authn/helper.go#L70 links to https://github.com/bazelbuild/rules_docker/issues/111 which looks like it would have been fixed by https://github.com/docker/docker-credential-helpers/pull/72

Given that's been fixed for a while, I'm wondering if we can just drop the scheme instead of falling back. What do you think?

abergmeier:
> Given that's been fixed for a while, I'm wondering if we can just drop the scheme instead of falling back. What do you think?

I am not sure whether this may lead to problems in cases, where you have both _http_ and _https_ stored for a certain registry.
Especially for Enterprise there are a lot of unrooted CAs and then incentive to enforce `http` instead of `https`.

jonjohnsonjr:
> I am not sure whether this may lead to problems in cases, where you have both http and https stored for a certain registry.

In that case we probably want to derive the scheme from [h.domain.Scheme()](https://godoc.org/github.com/google/go-containerregistry/pkg/name#Registry.Scheme).

Do you happen to know how docker handles this? Does it also fall back like this?

abergmeier:
> Do you happen to know how docker handles this? Does it also fall back like this?

Not exactly sure which part of the Docker Suite you mean ;)

That said - I get the impression that like a lot of things coming from Docker the credential-helper interfaces may be underdefined.

jonjohnsonjr:
> Not exactly sure which part of the Docker Suite you mean ;)

Fair enough :smile: I was thinking the best we can really expect to do is be compatible with the community edition docker CLI.

> That said - I get the impression that like a lot of things coming from Docker the credential-helper interfaces may be underdefined.

That's my impression as well :) it seems very much an implementation-as-the-spec kind of thing.

I did some spelunking as to how docker/cli does this for `docker push`, here's what I've found:

1.  They first call [ResolveAuthConfig](https://github.com/docker/cli/blob/223656805372d19c6669954764aaea1db6aafa1e/cli/command/registry.go#L80-L88) with an [IndexInfo](https://github.com/moby/moby/blob/da823cf3a5a3437763ba12b2fa369c826d4a2bf4/api/types/registry/registry.go#L73).
2. If the index is `Official` (generally, dockerhub), they [consult the daemon via ElectAuthServer](https://github.com/docker/cli/blob/223656805372d19c6669954764aaea1db6aafa1e/cli/command/registry.go#L83). Otherwise, it just uses [IndexInfo.Name](https://github.com/moby/moby/blob/da823cf3a5a3437763ba12b2fa369c826d4a2bf4/api/types/registry/registry.go#L74-L75) which [appears to be just the hostname](https://github.com/moby/moby/blob/fc01c2b481097a6057bec3cd1ab2d7b4488c50c4/registry/config.go#L408) (without a scheme). I don't think we want to consult a daemon, but their [fallback path](https://github.com/docker/cli/blob/223656805372d19c6669954764aaea1db6aafa1e/cli/command/registry.go#L32-L38) for the `Official` index returns [a hard-coded URL](https://github.com/moby/moby/blob/fc01c2b481097a6057bec3cd1ab2d7b4488c50c4/registry/config.go#L39) (we should probably do the same for dockerhub).
3. They then pass either the hard-coded dockerhub key or the bare domain [all](https://github.com/docker/cli/blob/70f48f223157f5b567a87c563fa4997065fc6fde/cli/config/configfile/file.go#L288-L290) [the](https://github.com/docker/cli/blob/70f48f223157f5b567a87c563fa4997065fc6fde/cli/config/configfile/file.go#L276-L278) [way](https://github.com/docker/cli/blob/70f48f223157f5b567a87c563fa4997065fc6fde/cli/config/credentials/native_store.go#L22-L30) [down](https://github.com/docker/cli/blob/70f48f223157f5b567a87c563fa4997065fc6fde/cli/config/credentials/native_store.go#L43-L56) [to](https://github.com/docker/cli/blob/70f48f223157f5b567a87c563fa4997065fc6fde/cli/config/credentials/native_store.go#L118) [Get](https://github.com/docker/docker-credential-helpers/blob/master/client/client.go#L52-L54).

So it seems like for dockerhub (or some other official registry, which can happen for windows daemons?), serverURL is `"https://index.docker.io/v1/"`, but for everything else it's just the domain. How do you feel about matching that behavior?

abergmeier:
If we accept Docker's behavior as implementation-as-the-spec, then I think we are beyond discussing.
Then the only thing that helps is writing integration tests, to ensure that go-containerregistry behavior matches dockercli behavior.

jonjohnsonjr:
> If we accept Docker's behavior as implementation-as-the-spec, then I think we are beyond discussing.
Then the only thing that helps is writing integration tests, to ensure that go-containerregistry behavior matches dockercli behavior.

Fair enough, filed https://github.com/google/go-containerregistry/issues/479

I'm okay with merging this and revising it as part of #479.

jonjohnsonjr:
> Then the only thing that helps is writing integration tests, to ensure that go-containerregistry behavior matches dockercli behavior.

This was (finally, sorry) done in https://github.com/google/go-containerregistry/pull/462

I suspect you've moved on but if this fixes or doesn't fix the issue you were seeing, I'd love to know!

PR #457, opened 2019-05-17
Add error details in CheckPushPermission.

abergmeier:
Really nasty to find know exactly where things went wrong without.

### DIFF
--- a/pkg/v1/remote/check.go
+++ b/pkg/v1/remote/check.go
@@ -1,6 +1,7 @@
 package remote
 
 import (
+	"fmt"
 	"net/http"
 
 	"github.com/google/go-containerregistry/pkg/authn"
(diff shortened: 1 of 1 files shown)


pkg/v1/remote/check.go
```diff
@@ -18,13 +19,13 @@ import (
 func CheckPushPermission(ref name.Reference, kc authn.Keychain, t http.RoundTripper) error {
 	auth, err := kc.Resolve(ref.Context().Registry)
 	if err != nil {
-		return err
+		return fmt.Errorf("Resolving authorization for %v failed: %v", ref.Context().Registry, err)
```
jonjohnsonjr: Small nit, can you make these error strings start with lower case? See https://github.com/golang/go/wiki/CodeReviewComments#error-strings

abergmeier: Done.

PR #458, opened 2019-05-22
Composable layer cache package

### LINKED ISSUE #325: Caching support
I saw there was a PR https://github.com/google/go-containerregistry/pull/164 that tried to introduce caching, what is the status on this? It would be nice to get caching support as not just speed is a potential issue but also potentially cost when pulling from a private repo.

ImJasonH:
This change adds the `pkg/v1/cache` package, containing a `Cache` interface to describe a general-purpose layer cache, as well as `NewFilesystemCache` which caches layer contents on disk.

Unlike the previous attempt in #164, this `cache` package can be used with any `v1.Image` implementation, including those with `stream.Layer`s and `mutate.Append`ed layers.

The filesystem cache implementation lazily caches layer contents only when they're actually read by the caller, but the `Cache` interface leaves laziness up to the implementor -- an eager implementation could consume the layer contents immediately then return buffered or disk-backed `Layer` contents if it preferred.

A _very_ naive in-memory `Cache` is included in tests, don't even think about using it for anything real.

This also adds a flag to `crane pull` to specify a local disk cache for pulled layers.

Fixes #325 

cc @yiranwang52 and @codygibb

### DIFF
--- a/pkg/crane/pull.go
+++ b/pkg/crane/pull.go
@@ -60,7 +60,7 @@ func pull(src, dst, cachePath string) {
 		log.Fatalf("reading image %q: %v", ref, err)
 	}
 	if cachePath != "" {
-		i = cache.NewImage(i, cache.NewFilesystemCache(cachePath))
+		i = cache.Image(i, cache.NewFilesystemCache(cachePath))
 	}
 
 	// WriteToFile wants a tag to write to the tarball, but we might have
--- a/pkg/crane/pull.go
+++ b/pkg/crane/pull.go
@@ -55,14 +55,12 @@ func pull(src, dst, cachePath string) {
 	}
 	log.Printf("Pulling %v", ref)
 
-	var c cache.Cache
-	if cachePath != "" {
-		c = cache.NewFilesystemCache(cachePath)
-	}
-
-	i, err := remote.Image(ref, remote.WithAuthFromKeychain(authn.DefaultKeychain), remote.WithCache(c))
+	i, err := remote.Image(ref, remote.WithAuthFromKeychain(authn.DefaultKeychain))
 	if err != nil {
 		log.Fatalf("reading image %q: %v", ref, err)
+	}
+	if cachePath != "" {
+		i = cache.NewImage(i, cache.NewFilesystemCache(cachePath))
 	}
 
 	// WriteToFile wants a tag to write to the tarball, but we might have
--- a/pkg/crane/pull.go
+++ b/pkg/crane/pull.go
@@ -20,6 +20,7 @@ import (
 
 	"github.com/google/go-containerregistry/pkg/authn"
 	"github.com/google/go-containerregistry/pkg/name"
+	"github.com/google/go-containerregistry/pkg/v1/cache"
 	"github.com/google/go-containerregistry/pkg/v1/remote"
 	"github.com/google/go-containerregistry/pkg/v1/tarball"
 	"github.com/spf13/cobra"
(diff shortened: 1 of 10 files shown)


ImJasonH:
Added the functional option to `remote` and added a `-c` flag to `crane pull` to test things out:

```
$ time ./crane pull -c ~/.cache/ ubuntu out.tar
2019/05/22 10:22:59 Pulling index.docker.io/library/ubuntu:latest
2019/05/22 10:22:59 No matching credentials were found, falling back on anonymous
2019/05/22 10:22:59 No matching credentials were found, falling back on anonymous

real	0m4.758s
user	0m0.456s
sys	0m0.478s

$ time ./crane pull -c ~/.cache/ ubuntu out.tar
2019/05/22 10:23:06 Pulling index.docker.io/library/ubuntu:latest
2019/05/22 10:23:06 No matching credentials were found, falling back on anonymous
2019/05/22 10:23:06 No matching credentials were found, falling back on anonymous
2019/05/22 10:23:08 Layer sha256:6abc03819f3e00a67ed5adc1132cfec041d5f7ec3c29d5416ba0433877547b6f found in cache
2019/05/22 10:23:08 Layer sha256:05731e63f21105725a5c062a725b33a54ad8c697f9c810870c6aa3e3cd9fb6a2 found in cache
2019/05/22 10:23:08 Layer sha256:0bd67c50d6beeb55108476f72bea3b4b29a9f48832d6e045ec66b7ac4bf712a0 found in cache

real	0m2.799s
user	0m0.833s
sys	0m0.108s
```

pkg/v1/remote/options.go
```diff
@@ -104,3 +106,14 @@ func WithPlatform(p v1.Platform) Option {
 		return nil
 	}
 }
+
+// WithCache is a functional option to specify a Cache implementation that
+// should be used when pulling remote image layers.
+//
+// It is only used for remote.Image.
+func WithCache(c cache.Cache) Option {
```
jonjohnsonjr: This doesn't really enable anything that clients couldn't do by just wrapping the `remote.Image` themselves, right? I'm hesitant to add this as an option if we don't need it

pkg/v1/cache/fs.go
```diff
@@ -0,0 +1,80 @@
...
+
+func (rc *readcloser) Read(b []byte) (int, error) {
+	return rc.t.Read(b)
+}
+
+func (rc *readcloser) Close() error {
+	for _, c := range rc.closes {
+		if err := c(); err != nil {
+			return err
+		}
```
codygibb: Sorry for the drive-by review -- you might want to make a best effort to always call every close function, even if one of them returns an error. Most clients don't handle `Close()` errors (e.g. `defer x.Close()`), so if the first close fails, no one will notice that the second close (i.e. closing the file) was never called.

ImJasonH: Excellent point, thanks for the drive-by review! (I'm not sure it's a drive-by if I @ you on it though 😆 )

pkg/v1/remote/options.go
ImJasonH: I thought it would be nice for usability, but then I tried it the other way and it was basically the same. Done.

pkg/v1/remote/image.go
```diff
@@ -56,7 +56,11 @@ func Image(ref name.Reference, options ...Option) (v1.Image, error) {
 		return nil, err
 	}
 
-	return desc.Image()
+	img, err := desc.Image()
```
jonjohnsonjr: `return desc.Image()`

pkg/v1/cache/cache.go
```diff
@@ -0,0 +1,88 @@
...
+	// Delete removes the Layer with the given Hash from the Cache.
+	Delete(v1.Hash) error
+}
+
+// ErrNotFound is returned by Get when no layer with the given Hash is found.
+var ErrNotFound = errors.New("layer was not found")
+
+// NewImage returns a new Image whose layers will be pulled from the cache if
+// they are found, and written to the cache as they are read from the
+// underlyiing Image.
```
jonjohnsonjr: underlying

pkg/v1/cache/cache.go
```diff
@@ -0,0 +1,88 @@
...
+func NewImage(i v1.Image, c Cache) v1.Image {
```
jonjohnsonjr: `cache.Image`?

ImJasonH:
I'm going to call this not-WIP anymore, since I think most people who have looked at it agree it's a good (enough) API to start using and iterating on.

ImJasonH:
cc @sharifelgamal @priyawadhwa in case this looks useful to Kaniko, or in case something about this is a bad idea, based on your experience with caching layers in Kaniko.

pkg/crane/pull.go
```diff
@@ -55,6 +59,9 @@ func pull(_ *cobra.Command, args []string) {
 	if err != nil {
 		log.Fatalf("reading image %q: %v", ref, err)
 	}
+	if cachePath != "" {
+		i = cache.NewImage(i, cache.NewFilesystemCache(cachePath))
```
jonjohnsonjr: `cache.Image`

Issue #431, opened 2019-04-11
Handling BLOB_UPLOAD_INVALID?


tbarrella:
Hello,

This is to try to address https://github.com/GoogleContainerTools/kaniko/issues/607, in which BLOB_UPLOAD_INVALID is happening seemingly because of the following situation:

1. Kaniko tries to upload blob X
2. Somehow, Kaniko is told to retry blob X, but actually the upload succeeds
3. Kaniko tries to upload blob X again -> "BLOB_UPLOAD_INVALID" because blob X + 1 is expected, or more specifically
>BLOB_UPLOAD_INVALID: “First byte of the layer part is {position of first byte of X} instead of {position of first byte of X + 1}”

When this happens, the push fails completely

I know ideally step 2 wouldn't happen, but would it make sense to have some way to handle this from the client (uploader) side?

jonjohnsonjr:
Seems like a transient ECR error. On the client side I think the best thing we could do is retry, but since it's a 4xx error instead of a 5xx error, we'll need to parse the error, look for this error code, and retry the entire upload.

Alternatively we could just retry the layer upload if we encounter any errors. That seems less elegant but would require a lot less code 🤔

tbarrella:
Hello, thank you! I'm not very knowledgeable about what's best for here, or if other libraries have a standard way of handling 4xx errors (the [spec](https://docs.docker.com/registry/spec/api/#pushing-an-image) is a bit vague on that). From https://github.com/docker/distribution/blob/master/docs/spec/api.md#on-failure-bad-request-4, maybe it makes sense to parse the "code" because there are a few different 400 errors for blob upload

tbarrella:
I've been looking into this more because I'm still encountering this frequently (retrying the layer upload would be more preferable for me than retrying the whole build since some builds take a while). The registry API [lists possible failure responses](https://github.com/docker/distribution/blob/3226863cbcba6dbc2f6c83a37b28126c934af3f8/registry/api/v2/descriptors.go#L1254) for blob upload including specifying when they should be retried.

[...]

PR #459, opened 2019-05-22
Retry layer upload on BLOB_UPLOAD_INVALID error

tbarrella:
To close https://github.com/google/go-containerregistry/issues/431

I'm still new to Go so am especially receptive to changes. It seems like there should be tests for the retry, but I wasn't sure if there's a most idiomatic way

### DIFF
--- a/pkg/v1/remote/write.go
+++ b/pkg/v1/remote/write.go
@@ -20,8 +20,10 @@ import (
 	"fmt"
 	"io"
 	"log"
+	"math"
 	"net/http"
 	"net/url"
+	"time"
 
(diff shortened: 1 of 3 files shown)


pkg/v1/remote/write.go
```diff
@@ -37,6 +39,9 @@ type manifest interface {
 	Digest() (v1.Hash, error)
 }
 
+const maxRetries = 2
```
ImJasonH: Can you add a comment about what gets retried this many times? Alternatively, just move them into the method where they're used.

Issue #421, opened 2019-03-26
Problems authenticating to Azure ACR


tjk:
I've been trying to solve an issue with https://github.com/GoogleContainerTools/skaffold in which fetching a remote image digest from Azure Container Registry (ACR) returns a 401.

Turns out the `docker-credential-acr-*` returns a username of `<token>` and secret representing a refresh_token (rather than a username + password). 

https://github.com/google/go-containerregistry/blob/master/pkg/v1/remote/transport/transport.go determines it must use bearer auth strategy and when it tries to refresh the `bearerTransport`, it performs a GET request with `Basic` username+password auth. I have a patch that instead performs a POST with the refresh_token that the docker-credential helper returns which fixes the 401 but I need advice on how to more properly fix this.

Please let me know if seeing my patch would be more helpful.

jonjohnsonjr:
Thanks for filing an issue!

To make sure I understand, from looking at the registry docs, it looks like:
* we **only** support the [token authentication specification](https://docs.docker.com/registry/spec/auth/token/),
* whereas ACR **only** supports the [oauth authentication specification](https://docs.docker.com/registry/spec/auth/oauth/).

Reading through the oauth docs:
> Note: Not all token servers implement oauth2. If the request to the endpoint returns 404 using the HTTP POST method, refer to Token Documentation for using the HTTP GET method supported by all token servers.

So I'm thinking for an actual fix, we'd want to first attempt to do the oauth exchange, then do the token exchange if that fails. Ideally, ACR would just support token authentication, given (emphasis mine) this snipper:
> refer to Token Documentation for using the HTTP GET method supported by **all** token servers

It's probably more likely we could land a fix here than against ACR, but it might be worth pursuing that as well. I'm wondering if this is a bug in the cred helper, or how it's configured or something? cc @shhsu who seems to have written most of it. Is this expected?

Is there anything in the [challenge](https://docs.docker.com/registry/spec/auth/token/#how-to-authenticate) that indicates we should use oauth?

Seeing your patch would definitely be helpful, as well as the responses from ACR and from their credential helper (with any sensitive information or secrets redacted).

Do you happen to know if there are docs somewhere for how [the credential helper response](https://godoc.org/github.com/docker/docker-credential-helpers/credentials#Credentials) is supposed to work with the oauth exchange? I guess we just use the `Secret` as a `refresh_token`?

tjk:
Sorry got a busy, some responses --

> Reading through the oauth docs:

[...]

PR #462, opened 2019-05-29
Fix auth

jonjohnsonjr:
According to the credential helper protocol, secrets stored as identity
tokens have username `<token>`. However, it's very unclear what an identity token is!

This PR outsources a lot of our own credential handling to docker/cli/cli/config.

We also treat `<token>` as a bearer token and do a separate token handshake for oauth.

Fixes https://github.com/google/go-containerregistry/issues/421
Fixes https://github.com/google/go-containerregistry/issues/479
Fixes https://github.com/google/go-containerregistry/issues/493
Fixes https://github.com/google/ko/issues/90
Obsoletes https://github.com/google/go-containerregistry/pull/456
Related https://github.com/GoogleContainerTools/jib/issues/1490

https://github.com/docker/cli/blob/master/docs/reference/commandline/login.md#credential-helper-protocol

https://docs.docker.com/registry/spec/auth/oauth/

~~I think our authn package is actually kind of weird and I'm wondering how gross it would be to reuse some of the docker stuff for credential helpers.~~

This now uses docker's cli package for parsing config files!


balopat:
Hi @jonjohnsonjr - is there a way to help with this PR? We are really interested on this getting from the Skaffold side :)

jonjohnsonjr:
I'd love if someone could test this. I don't have immediate access to a registry or credential setup that this fixes.

pkg/authn/k8schain/k8schain_test.go
```diff
@@ -102,7 +102,7 @@ func TestAttachedServiceAccount(t *testing.T) {
 	if err != nil {
 		t.Errorf("Authorization() = %v", err)
 	}
-	if got != want {
+	if *got != *want {
```
ImJasonH: This will panic if either is nil, which might happen in the error case. Either fatal on error, or I think `reflecf.DeepEqual` might handle the nil case for you.

go.mod
```diff
@@ -9,8 +9,10 @@ require (
 	github.com/Microsoft/go-winio v0.4.14 // indirect
```
ImJasonH: Can we mark go mod files as generated for GitHub code reviews?

PR #464, opened 2019-06-05
implement SetDefaultKeychainConfigDir to allow for custom docker config

xwinxu:
Supplemented the configuration directory lookup procedure by adding an exported function SetDefaultKeychainConfigDir to set docker configuration file directory based on client import. Feature requested for rules-docker dependency on this repository for registry puller functionality.

### DIFF
--- a/pkg/authn/keychain.go
+++ b/pkg/authn/keychain.go
@@ -35,7 +35,19 @@ type Keychain interface {
 
 // defaultKeychain implements Keychain with the semantics of the standard Docker
 // credential keychain.
-type defaultKeychain struct{}
+type defaultKeychain struct {
+	configDir string
+}
+
+// SetDefaultKeychainConfigDir sets the client config directory to specified location.
+func SetDefaultKeychainConfigDir(p string) error {
+	dk, ok := DefaultKeychain.(*defaultKeychain)
+	if !ok {
+		return fmt.Errorf("Unable to set custom config dir")
+	}
+	dk.configDir = p
+	return nil
+}
 
 // configDir returns the directory containing Docker's config.json
 func configDir() (string, error) {


pkg/authn/keychain.go
```diff
@@ -35,7 +35,19 @@ type Keychain interface {
 
 // defaultKeychain implements Keychain with the semantics of the standard Docker
 // credential keychain.
-type defaultKeychain struct{}
+type defaultKeychain struct {
+	configDir string
```
smukherj1: Add comment saying this will store a custom docker config directory. The convention in Go is
`// configDir will store a custom docker client config directory. If this is blank, config directory resolution will mimic docker's default behavior`

pkg/authn/keychain.go
```diff
@@ -35,7 +35,19 @@ type Keychain interface {
...
+}
+
+// SetDefaultKeychainConfigDir sets the client config directory to specified location.
```
smukherj1: Sets the client config directory of the default keychain to the specified location.

pkg/authn/keychain.go
```diff
@@ -35,7 +35,19 @@ type Keychain interface {
...
+func SetDefaultKeychainConfigDir(p string) error {
+	dk, ok := DefaultKeychain.(*defaultKeychain)
+	if !ok {
+		return fmt.Errorf("Unable to set custom config dir")
```
smukherj1: errors begin with small letter

pkg/authn/keychain.go
```diff
@@ -35,7 +35,19 @@ type Keychain interface {
 
 // defaultKeychain implements Keychain with the semantics of the standard Docker
 // credential keychain.
-type defaultKeychain struct{}
+type defaultKeychain struct {
+	configDir string
+}
+
+// SetDefaultKeychainConfigDir sets the client config directory to specified location.
+func SetDefaultKeychainConfigDir(p string) error {
```
jonjohnsonjr: It's a bit gross, but you could just call:
```
os.Setenv("DOCKER_CONFIG", p)
```

in the caller's code instead of adding this method here. It's not great, but neither is a mutating package-level vars.

I wouldn't really expect a library to want to set this directly, unless you don't have control over the environment :/

pkg/authn/keychain.go
```diff
@@ -35,7 +35,19 @@ type Keychain interface {
...
+	dk, ok := DefaultKeychain.(*defaultKeychain)
```
xingao267: What about making this function called on the *defaultKeychain similar to the Resolve function below? You just set the configDir to p here. And when you call it, you call it on the exported DefaultKeychain instance auth.DefaultKeychain, which is more similar to how pull.go is currently consuming auth.DefaultKeychain when invoking Resolve(), or the python implementation, which seems making a little bit more sense to me.

If you are not planning to implement consuming the new configDir attribute in configDir() or Resolve() function, maybe add a TODO to keep track. I would probably do that together in this PR as it's small and making this feature complete.

smukherj1: This wouldn't work as is because "DefaultKeychain" is of type "Keychain" which is an interface that doesn't have the "SetConfigDir" function. If the function were added to the interface, every implementation of the interface would need to add this function. This would be an API breaking update to go-containerregistry as "Keychain" is part of the public API.

PR #466, opened 2019-06-07
Add Platform.Features attr and more platform matching checks

xiaohegong:
- Added "Feature" attr to the `Platform` struct for broader compatibility
- Added more platform matching validity checks

I've added two helper functions to accomplish these, please advise me if these should be moved to another util file.


pkg/v1/platform.go
```diff
@@ -21,4 +21,5 @@ type Platform struct {
 	OSVersion    string   `json:"os.version,omitempty"`
 	OSFeatures   []string `json:"os.features,omitempty"`
 	Variant      string   `json:"variant,omitempty"`
+	Features     []string `json:"CPU features,omitempty"`
```
xingao267: json:"features,omitempty"

pkg/v1/remote/index.go
```diff
@@ -182,3 +182,48 @@ func (r *remoteIndex) childDescriptor(child v1.Descriptor, platform v1.Platform)
 		platform:   platform,
 	}, nil
 }
+
+// Check if the given platform can run the required platform.
+// Matching platforms must have identical architecture, OS,
+// OS version (if provided), and variant (if provided)
+// Required features must be a subset of the features in the given platform.
+func (r *remoteIndex) matchesPlatform(given v1.Platform, required v1.Platform) bool {
```
xingao267: I think this function doesn't have to be a function called on *remoteIndex.

And for args you can do (given, required v1.Platform)

pkg/v1/remote/index.go
```diff
@@ -182,3 +182,48 @@ func (r *remoteIndex) childDescriptor(child v1.Descriptor, platform v1.Platform)
...
+	// required fields
+	if given.Architecture != required.Architecture || given.OS != required.OS {
+		return false
+	}
+
+	// optional fields
+	if given.OSVersion != "" && given.OSVersion != required.OSVersion {
+		return false
+	}
+	if given.Variant != "" && given.Variant != required.Variant {
```
xingao267: If I understand correctly, `required` is the one that user inputs? and `given` is from the manifest list, then we should check if `required.Variant` (and `required.OSVersion`, etc) is not empty instead of the given one?

pkg/v1/platform.go
xiaohegong: Got it, fixed

pkg/v1/remote/index.go
```diff
@@ -182,3 +182,48 @@ func (r *remoteIndex) childDescriptor(child v1.Descriptor, platform v1.Platform)
...
+		return false
+	}
+
+	// Verify any required features are a subset of this platform's features.
```
xingao267: //  Verify required platform's features are a subset of given platform's features.

xingao267: Maybe `is a subset`, I'm not sure.

pkg/v1/remote/index.go
```diff
@@ -182,3 +182,48 @@ func (r *remoteIndex) childDescriptor(child v1.Descriptor, platform v1.Platform)
...
+		return false
+	}
+	if !isSubset(given.Features, required.Features) {
+		return false
+	}
+
+	return true
+}
+
+// Check if the required array of strings is a subset of the given lst
```
xingao267: isSubset checks if ....

pkg/v1/remote/index.go
```diff
@@ -182,3 +182,48 @@ func (r *remoteIndex) childDescriptor(child v1.Descriptor, platform v1.Platform)
 		platform:   platform,
 	}, nil
 }
+
+// Check if the given platform can run the required platform.
```
xingao267: matchesPlatform checks if ...

pkg/v1/remote/index.go
```diff
@@ -182,3 +182,48 @@ func (r *remoteIndex) childDescriptor(child v1.Descriptor, platform v1.Platform)
 		platform:   platform,
 	}, nil
 }
+
+// matchesPlatform checks if the given platform can run the required platform.
```
xingao267: maybe say this

// matchesPlatform checks if the given platform matches the required platforms.
// The given platform matches the required platform if
//   - architecture and OS are identical
//   - OS version and variant are identical if provided
//   - features and OS features of the required platform is a subset of those of the given platform

pkg/v1/remote/index.go
```diff
@@ -182,3 +182,48 @@ func (r *remoteIndex) childDescriptor(child v1.Descriptor, platform v1.Platform)
...
+	}
+	if !isSubset(given.Features, required.Features) {
+		return false
+	}
+
+	return true
+}
+
+// isSubset checks if the required array of strings is a subset of the given lst
+func isSubset(lst []string, required []string) bool {
```
xingao267: the args can be combined as well, `lst, required []string`

pkg/v1/remote/index.go
xingao267: - features and OS features of the required platform are subsets of those of the given platform

pkg/v1/remote/index.go
```diff
@@ -182,3 +182,49 @@ func (r *remoteIndex) childDescriptor(child v1.Descriptor, platform v1.Platform)
 		platform:   platform,
 	}, nil
 }
+
+// matchesPlatform checks if the given platform matches the required platforms.
+// The given platform matches the required platform if
+// - architecture and OS are identical
+// - OS version and variant are identical if provided
+// - features and OS features of the required platform are subsets of those of the given platform
+func matchesPlatform(given, required v1.Platform) bool {
```
jonjohnsonjr: Want to write a unit test for `matchesPlatform`?

jonjohnsonjr: This is a good candidate for some simple [table driven tests](https://github.com/golang/go/wiki/TableDrivenTests) since it's a pure function.

We have some similar ones [here](https://github.com/google/go-containerregistry/blob/2687bd5ba651f247a89696f9dc8120100a12f763/pkg/name/registry_test.go#L139-L215).

jonjohnsonjr:
Travis is failing because of `gofmt` errors:
```
$ find . -name "*.go" | grep -v vendor/ | xargs gofmt -d -e -l
./pkg/v1/remote/index_test.go
diff -u ./pkg/v1/remote/index_test.go.orig ./pkg/v1/remote/index_test.go
--- ./pkg/v1/remote/index_test.go.orig	2019-06-10 19:18:00.241249087 +0000
+++ ./pkg/v1/remote/index_test.go	2019-06-10 19:18:00.245247225 +0000
@@ -254,9 +254,9 @@
 	tests := []struct {
 		// want is the expected return value from matchesPlatform
 		// when the given platform is 'given' and the required platform is 'required'.
-		given       v1.Platform
-		required    v1.Platform
-		want        bool
+		given    v1.Platform
+		required v1.Platform
+		want     bool
 	}{{ // The given & required platforms are identical. matchesPlatform expected to return true.
 		given: v1.Platform{
 			Architecture: "amd64",
```

I'd try to set up your editor to automatically run gofmt and goimports on save, if you can.

jonjohnsonjr:
Thanks!

Issue #468, opened 2019-06-13
Is it legit to send empty Authorization header for anonymous request?


yolo3301:
I noticed that when anon authn is used, requests to the registry will be sent with empty `Authorization` header (instead of dropping the Authorization header). However, it doesn't seem comply with http spec.

According to https://tools.ietf.org/html/rfc7235#appendix-C:

> credentials = auth-scheme [ 1*SP ( token68 / [ ( "," / auth-param )
    *( OWS "," [ OWS auth-param ] ) ] ) ]

Then

> auth-scheme = token

And

> token = <token, see [RFC7230], Section 3.2.6>

And 

> token          = 1*tchar

This essentially means the Authorization header cannot be empty. Any thoughts on this?

Related code: https://github.com/google/go-containerregistry/blob/master/pkg/v1/remote/transport/bearer.go#L62

jonjohnsonjr:
Yeah that's definitely a bug, this is actually causing https://github.com/knative/serving/issues/4355

PR #469, opened 2019-06-17
Skip sending Authorization header if it's empty

yolo3301:
To fix https://github.com/google/go-containerregistry/issues/468


jonjohnsonjr:
(Leaving this as a breadcrumb for the future in case things change.)

We might want a similar check here: https://github.com/google/go-containerregistry/blob/342edc910ac5a0efefb73b303040dfcc5569e91d/pkg/v1/remote/transport/bearer.go#L61

But I don't think we'll ever hit that path, so let's not worry about it yet :)

jonjohnsonjr:
Thanks!

yolo3301:
Yea, I looked at the bearer transport - it always uses the bearer authenticator, which always has `Bearer ` as the prefix of the Authorization header.

PR #471, opened 2019-06-19
Add retries for temporary network issues

### LINKED ISSUE #77: Don't use http.DefaultTransport in the CLI
It has no timeouts.

See: https://github.com/golang/go/issues/24138 and https://blog.cloudflare.com/the-complete-guide-to-golang-net-http-timeouts/

jonjohnsonjr:
We pretty commonly see this issue when communicating with a registry:
```
Get https://gcr.io/v2/: net/http: TLS handshake timeout
```

This causes a huge amount of pain if it's a failure within e.g. a large
CI job, since the entire thing would fail.

Fixes #77

This uses `k8s.io/apimachinery/pkg/util/wait` because we happen to already depend on it.

### DIFF
--- a/pkg/v1/remote/transport/retry.go
+++ b/pkg/v1/remote/transport/retry.go
@@ -1,0 +1,89 @@
+// Copyright 2018 Google LLC All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package transport
+
+import (
+	"net/http"
+	"time"
+
+	"github.com/google/go-containerregistry/pkg/internal/retry"
+	"k8s.io/apimachinery/pkg/util/wait"
+)
+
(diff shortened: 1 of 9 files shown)


pkg/v1/remote/transport/retry.go
```diff
@@ -0,0 +1,77 @@
...
+
+func (t *RetryTransport) RoundTrip(in *http.Request) (out *http.Response, err error) {
+	RetryTemporary(func() error {
+		out, err = t.Inner.RoundTrip(in)
+		return err
+	}, t.Backoff)
+	return
+}
+
+func RetryTemporary(f func() error, backoff wait.Backoff) (err error) {
```
ImJasonH: Should we check if `f` is `nil` here, and return an error? Otherwise, we'll panic on line 56.

pkg/v1/remote/transport/retry.go
```diff
@@ -0,0 +1,77 @@
...
+		}
+		return true, err
+	}
+	wait.ExponentialBackoff(backoff, condition)
+	return
+}
+
+func RetryErrors(f func() error, backoff wait.Backoff) (err error) {
+	condition := func() (bool, error) {
+		err = f()
```
ImJasonH: ditto

pkg/v1/remote/transport/retry.go
```diff
@@ -0,0 +1,89 @@
...
+
+// WithBackoff sets the backoff for retry operations.
+func WithBackoff(backoff wait.Backoff) Option {
+	return func(o *options) {
+		o.backoff = backoff
+	}
+}
+
+// WithPredicate sets the predicate for retry operations.
+func WithPredicate(predicate func(error) bool) Option {
```
ImJasonH: How about `WithRetryBackoff` and `WithRetryPredicate`?

PR #473, opened 2019-06-23
Mirror Config to ContainerConfig on mutations.

mattmoor:


mattmoor:
This works around a limitation in downstream tooling where when a container is launched from an image as an image `Config` is used, but when it's launched from an image like it's a snapshot of a container, then `ContainerConfig` is used.

This bad assumption works fine for imperatively built containers, but blows up for properly built containers.

Issue #472, opened 2019-06-21
Upload layer loses authentication on PATCH


lw346:
Set up:
 * Docker V2 registry with bearer token authentication required
 * Uploading an image via kaniko - revision 8621d738a07bc74b2adeafd175a3c738423577a0

What is happening:
 1. Ping sent to `https://registry.example.com/v2/`
    - Response with 401 directing to `registry-auth.example.com`
 1. Request made to `registry-auth.example.com` with appropriate scope
    - Response with 200 providing `<token>` in the JSON response body
 1. POST request made to `https://registry.example.com/v2/test-image/blobs/uploads/`
    - Request contains Authentication header set to `Bearer <token>`
    - Response is 202 Accepted with Location set to `https://registry.example.com/v2/test-image/blobs/uploads/some-hash`
 1. PATCH request made to `https://registry.example.com/v2/test-image/blobs/uploads/some-hash`
    - Request does not contain any Authentication header
    - Response is 401 directing to `registry-auth.example.com`
 1. Upload attempt terminates with error

What I expect to happen:
 * PATCH request in step 4 should contain Authentication header set to Bearer token

jonjohnsonjr:
Are you certain that the host for steps 4 and 5 are both `registry.example.com`? Is step 4 being redirected to another host?

lw346:
After taking a packet capture, it's the same host, but a different scheme - ie. calling POST on `https://registry.example.com` redirects to `http://registry.example.com`.

I'm happy to take this as user error, but it was still a bit confusing to debug.

jonjohnsonjr:
> https://registry.example.com redirects to http://registry.example.com

Looks like dropping the Authorization head in this case is the [golang http client](https://golang.org/pkg/net/http/#Client), which seems to be the right thing to do from a security perspective.

> registry.example.com:443

Ah. [This check](https://github.com/google/go-containerregistry/blob/ca8b66cb1b793f7fb0e3c191ad9a37eb3dc80ea0/pkg/v1/remote/transport/bearer.go#L56-L61) is probably failure because of that port. We could use [`net.SplitHostPort`](https://golang.org/pkg/net/#SplitHostPort) (or something else) to normalize the host when we match that...

PR #475, opened 2019-06-26
Resolve ports to well-known values when comparing Host values to decide whether or not to send Bearer Authorization header

lw346:
Previously the code considered 'https://registry.example.com/' functionally different to 'https://registry.example.com:443', and similarly for 'http' and port 80.  This change always resolves Host values (which in Go may or may not be a combination of host and port) to include the host and port based on the scheme being used such that these two examples are considered functionally equivalent.

This should resolve https://github.com/google/go-containerregistry/issues/472

### DIFF
--- a/pkg/v1/remote/transport/bearer.go
+++ b/pkg/v1/remote/transport/bearer.go
@@ -174,7 +174,7 @@ func (bt *bearerTransport) canonicalAddress(host string) (address string, err er
 	// - [ipv6]:port
 	// As net.SplitHostPort returns an error if the host does not contain a port, we should only attempt
 	// to call it when we know that the address contains a port
-	if strings.Count(host, ":") == 1 || (strings.Count(host, ":") >= 2 && strings.Contains(host, "]")) {
+	if strings.Count(host, ":") == 1 || (strings.Count(host, ":") >= 2 && strings.Contains(host, "]:")) {
 		hostname, port, err := net.SplitHostPort(host)
 		if err != nil {
 			return "", err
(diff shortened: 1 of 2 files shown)


pkg/v1/remote/transport/bearer.go
```diff
@@ -144,3 +163,28 @@ func (bt *bearerTransport) refresh() error {
...
+	// - ipv4
+	// - ipv4:port
+	// - ipv6
+	// - [ipv6]:port
+	// As net.SplitHostPort returns an error if the host does not contain a port, we should only attempt
+	// to call it when we know that the address contains a port
+	if strings.Count(host, ":") == 1 || (strings.Count(host, ":") >= 2 && strings.Contains(host, "]")) {
+		hostname, port, err := net.SplitHostPort(host)
+		if err != nil {
+			return "", err
```
jonjohnsonjr: I'm a bit worried that returning an error here might cause failures for things that used to work, e.g. if canonicalizing `in.Host` fails for some reason, but `in.URL.Host` would have succeeded, this would now fail.

What if we just use the original host in cases where we can't split it? Then we don't have to return an error (and don't have to handle it).

```go
hostname, port, err := net.SplitHostPort(host)
if err != nil {
  hostname = host
}
if port == "" {
  port = portMap[bt.scheme]
}
return net.JoinHostPort(hostname, port)
```

lw346: Seems like a fair concern - I'll update this today.

vindvaki: @jonjohnsonjr I believe @lw346's latest update addresses this concern. Can you take another look?

jonjohnsonjr: Ah sorry I missed that this got updated.

PR #476, opened 2019-06-29
Decoupled crane logic from crane command line code

### LINKED ISSUE #407: Decouple pkg/crane from command line flags
The original intention for `pkg/crane` was to be a library that `cmd/crane` consumes, but we ended up consolidating a lot of code because we were adding a lot of boilerplate for no reason.

I'd like to expose these high-level crane operations into easy to consume libraries, for example, someone who wants to copy an image basically has to re-implement all the code under `pkg/crane/copy.go`. Ideally, we'd expose a function:

```
package crane

func Copy(src, dst string, options ...Option) {
  // Do the copying.
}
```

And have sane defaults that can be overridden with options.

marekschwarz:
- Added commands package
  - Added api package
  - Extracted logic from commands into api package

This should resolve #407


jonjohnsonjr:
Oof, I wish I had seen your comment yesterday. I had been working on this recently, and it's slightly different from what you had here. I've rebased that and pushed it here: https://github.com/google/go-containerregistry/pull/478

Instead of doing a direct CLI -> package translation, I tried to tease apart what `crane` does into a set of functions that is usable without having to spend too much time learning the abstractions in this library, e.g.:

`$ crane pull` == `crane.Save(crane.Pull(...))`

`$ crane export` == `crane.Export(crane.Pull(...))`

`$ crane push` == `crane.Push(crane.Load(...))`

because this would allow someone who wanted to export the filesystem of a tarball image to easily discover they could do so via:

`crane.Export(crane.Load(...))`

Apologies for not responding -- I don't mean to waste your time. I'd love your feedback on my PR if you're interested 😄

marekschwarz:
No Problem, I just could have waited for your response before starting to work on it 😄 The time isn't wasted at all, I'm currently learning go and was searching for some practical examples.
 
I like your approach in solving this issue. I also considered coding it like that, but discarded it.
I discarded it because I saw that [ko](https://github.com/google/ko) kept its [ko/cmd/ko](https://github.com/google/ko/tree/master/cmd/ko) directory very clean by only putting the main-logic  into it and placed the commands themselves into the [pkg/commands](https://github.com/google/ko/tree/master/pkg/commands) directory.
With this approach, ko nicely decoupled the commands from the main logic.

Maybe we can try something like this:

- The "Business Logic" of crane goes into `pkg/crane`, so the user can still use `crane.foo()`, without having to rename `pkg/crane/api` to `crane` on importing it, as you designed it in your PR
- The command logic goes into `pkg/crane/commands`, to keep the `cmd/crane` directory clean

jonjohnsonjr:
I think what we have is pretty close to this now, going to close this PR but feel free to reopen if I missed something.

PR #477, opened 2019-06-29
Added command to generate bash_completion script

marekschwarz:


jonjohnsonjr:
I think travis is failing because you need to run `./hack/update-code-gen.sh` in order to re-generate the docs.

jonjohnsonjr:
@marekschwarz want to pick this back up?

PR #478, opened 2019-06-30
Turn crane into a library

### LINKED ISSUE #407: Decouple pkg/crane from command line flags

jonjohnsonjr:
Fixes #407

This change used to be a bit more compelling, before we started using functional options for everything, since it saves a bit of boilerplate. As is, it mostly just saves parsing strings via `name.ParseReference` and typing `remote.WithAuthFromKeychain(authn.DefaultKeychain)`.

I do think this strikes a nice balance for people who don't necessarily want to learn about our abstractions but do want to do something more sophisticated than what the `crane` CLI happens to expose.

### DIFF
--- a/cmd/crane/cmd/append.go
+++ b/cmd/crane/cmd/append.go
@@ -1,0 +1,64 @@
+// Copyright 2018 Google LLC All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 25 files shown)


wdhnl:
I like this approach.  
The CLI isn't quite what I'm looking for and being able to use the libraries easily would be perfect.

jonjohnsonjr:
@wdhnl going to merge this so you can play with it, I'd really value any feedback you have.

PR #481, opened 2019-07-17
Add logs package

### LINKED ISSUE #480: library prints annoying logs
When used in Skaffold, the library prints annoying logs to the console and we have no simple way to hide them.

For example:
 + https://github.com/google/go-containerregistry/blob/9c974872bc17b108ff910a5c97eaa676e3dfab2e/pkg/authn/keychain.go#L103
 + https://github.com/google/go-containerregistry/blob/9c974872bc17b108ff910a5c97eaa676e3dfab2e/pkg/authn/keychain.go#L109
 + https://github.com/google/go-containerregistry/blob/46a3528b82270ecea7c731c14e698ea24aa0bbca/pkg/v1/google/auth.go#L65

jonjohnsonjr:
Fixes #480

We log for two reasons:
1. Warning for something that might have gone wrong.
2. Progress for things that are going right.

This introduces logs.Warn and logs.Progress which just expose two
package level log.Loggers, which will allow consumers of this library to
override them.

cc @dgageot this will make our library not log anything by default, see `cmd/crane/main.go` for enabling progress and/or warning logs. We'll want to do the same for `ko` to keep the same behavior.

### DIFF
--- a/cmd/gcrane/main.go
+++ b/cmd/gcrane/main.go
@@ -20,7 +20,13 @@ import (
 
 	crane "github.com/google/go-containerregistry/cmd/crane/cmd"
 	"github.com/google/go-containerregistry/pkg/gcrane"
+	"github.com/google/go-containerregistry/pkg/logs"
 )
+
+func init() {
+	logs.Warn.SetOutput(os.Stderr)
+	logs.Progress.SetOutput(os.Stderr)
+}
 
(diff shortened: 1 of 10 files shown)


jonjohnsonjr:
I'm not sure if we love this but it's the least invasive thing I can imagine.

PR #484, opened 2019-07-23
Create a tls version of the registry.New()

clrprod:
This handles the complexity of generating certs and forcing transports
to the server with the correct root ca. This should be usable for
replacing any given container registry in a test.

### DIFF
--- a/pkg/registry/tls.go
+++ b/pkg/registry/tls.go
@@ -1,0 +1,91 @@
+package registry
+
+import (
+	"bytes"
+	"context"
+	"crypto/ecdsa"
+	"crypto/elliptic"
+	"crypto/rand"
+	"crypto/tls"
+	"crypto/x509"
+	"encoding/pem"
+	"math/big"
+	"net"
+	"net/http"
+	"net/http/httptest"
+	"time"
+)
+
+// TLS returns a httptest server, as well as a transport that has been modified to
+// send all requests to the given server. The TLS certs are generated for the given domain
+// which should correspond to the domain the container is stored in.
+func TLS(domain string) (*httptest.Server, *http.Transport, error) {
+
+	s := httptest.NewUnstartedServer(New())
+
+	template := x509.Certificate{
+		SerialNumber: big.NewInt(1),
+		NotBefore:    time.Now().Add(-1 * time.Hour),
+		NotAfter:     time.Now().Add(time.Hour),
+		IPAddresses: []net.IP{
+			net.IPv4(127, 0, 0, 1),
+			net.IPv6loopback,
+		},
+		DNSNames: []string{domain},
+
+		KeyUsage:              x509.KeyUsageKeyEncipherment | x509.KeyUsageDigitalSignature | x509.KeyUsageCertSign,
+		ExtKeyUsage:           []x509.ExtKeyUsage{x509.ExtKeyUsageServerAuth},
+		BasicConstraintsValid: true,
+		IsCA:                  true,
+	}
+
+	priv, err := ecdsa.GenerateKey(elliptic.P521(), rand.Reader)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	b, err := x509.CreateCertificate(rand.Reader, &template, &template, &priv.PublicKey, priv)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	pc := &bytes.Buffer{}
+	if err := pem.Encode(pc, &pem.Block{Type: "CERTIFICATE", Bytes: b}); err != nil {
+		return nil, nil, err
+	}
+
+	ek, err := x509.MarshalECPrivateKey(priv)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	pk := &bytes.Buffer{}
+	if err := pem.Encode(pk, &pem.Block{Type: "EC PRIVATE KEY", Bytes: ek}); err != nil {
+		return nil, nil, err
+	}
+
+	c, err := tls.X509KeyPair(pc.Bytes(), pk.Bytes())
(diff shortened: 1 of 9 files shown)


pkg/registry/tls.go
```diff
@@ -0,0 +1,92 @@
+package registry
```
jonjohnsonjr: missing copyright

Also I'm tempted to call this registrytest to match the httptest pattern, what do you think? Or is the `registry` package already used exclusively for testing?

pkg/registry/tls.go
```diff
@@ -0,0 +1,92 @@
...
+	"crypto/x509"
+	"encoding/pem"
+	"math/big"
+	"net"
+	"net/http"
+	"net/http/httptest"
+	"time"
+)
+
+// TLS returns a httptest server, with a http client that has been configured to
```
jonjohnsonjr: nit: "returns an httptest"
nit: "with an http client"

pkg/registry/tls.go
```diff
@@ -0,0 +1,92 @@
...
+// send all requests to the given server. The TLS certs are generated for the given domain
```
jonjohnsonjr: s/given server/returned server/

pkg/registry/tls.go
```diff
@@ -0,0 +1,92 @@
...
+// which should correspond to the domain the container is stored in.
```
jonjohnsonjr: s/container/image/

pkg/registry/tls.go
```diff
@@ -0,0 +1,92 @@
...
+			RootCAs: certpool,
+		},
+		DialContext: func(ctx context.Context, network, addr string) (net.Conn, error) {
+			return net.Dial(s.Listener.Addr().Network(), s.Listener.Addr().String())
+		},
+	}
+	s.Client().Transport = t
+
+	return s, nil
+
```
jonjohnsonjr: nit: drop newline

pkg/registry/tls_test.go
```diff
@@ -0,0 +1,35 @@
+package registry_test
```
jonjohnsonjr: looks like we're missing copyright headers on this entire package actually

pkg/registry/tls.go
```diff
@@ -0,0 +1,92 @@
...
+	"net/http/httptest"
+	"time"
+)
+
+// TLS returns a httptest server, with a http client that has been configured to
+// send all requests to the given server. The TLS certs are generated for the given domain
+// which should correspond to the domain the container is stored in.
+// If you need a transport, Client().Transport is correctly configured.
+func TLS(domain string) (*httptest.Server, error) {
+
```
jonjohnsonjr: nit: drop newline

pkg/registry/tls_test.go
```diff
@@ -0,0 +1,35 @@
...
+	if err != nil {
+		t.Fatalf("Unable to get image digest: %v", err)
+	}
+
+	d, err := name.NewDigest("test.com/foo@" + rd.String())
+	if err != nil {
+		t.Fatalf("Unable to parse digest: %v", err)
+	}
+	if err := remote.Write(d, i, remote.WithTransport(s.Client().Transport)); err != nil {
+		t.Fatalf("Unable to write image to remove: %s", err)
```
jonjohnsonjr: s/remove/remote/

pkg/registry/tls_test.go
```diff
@@ -0,0 +1,35 @@
...
+	"testing"
+
+	"github.com/google/go-containerregistry/pkg/name"
+	"github.com/google/go-containerregistry/pkg/registry"
+	"github.com/google/go-containerregistry/pkg/v1/random"
+	"github.com/google/go-containerregistry/pkg/v1/remote"
+)
+
+func TestTLS(t *testing.T) {
+	s, err := registry.TLS("test.com")
```
jonjohnsonjr: I'd usually say to use "example.com" for this, but I feel like that doesn't test that we can pick the domain (since httptest uses example.com itself). Does the hard-coded httptest cert work for "registry.example.com" as well? If not, we could use "registry.example.com" to both avoid using "test.com" and assert that it works for arbitrary domains.

pkg/registry/tls.go
clrprod: registry is pretty much used only for testing. I think if we move this to another package it may be a lot less discoverable. It's not like this is dangerous to use in prod if you want a in memory registry for some reason.

pkg/registry/tls.go
clrprod: Done.

pkg/registry/tls.go
clrprod: done

pkg/registry/tls.go
clrprod: Done.

pkg/registry/tls.go
clrprod: Done.

pkg/registry/tls_test.go
clrprod: Added them to the entire package.

pkg/registry/tls.go
clrprod: Done.

pkg/registry/tls_test.go
clrprod: Done.

pkg/registry/tls_test.go
clrprod: Done.

pkg/registry/tls.go
jonjohnsonjr: makes sense

Issue #413, opened 2019-03-19
Replace tarball.LayerFromOpener with stream.NewLayer


jonjohnsonjr:
The `tarball` package has methods for reading/writing images from/to the `docker save` tarball format. This aligns with how the other packages are organized, for example:
* `remote` is for reading/writing from/to registries.
* `daemon` is for reading/writing from/to container daemons.
* `layout` is for reading/writing from/to OCI image layouts.

However, the `tarball` package _also_ has methods for constructing `v1.Layer`s from a file or from an `Opener`. It _kind of_ makes sense to be in the `tarball` package because the layers happen to be compressed tarballs, but really that's an implementation detail. These functions aren't for tarball-specific layers, they're just for layers in general.

We should try to replace these with `stream.Layer` wherever we can. The `tarball.Layer` currently reads the whole tarball twice because it eagerly computes the digest/size/diffid. For uncompressed layers, `stream.Layer` is just better.

For compressed layers, we usually have some metadata that makes eagerly computing those things unnecessary (e.g. `remoteLayer`). For reading compressed layer from a `tarball.Image`, we should know the diffids from the config file.

tl;dr `LayerFromOpener` is bad, let's make `stream.Layer` work. I elaborated on this a bit [here](https://github.com/google/go-containerregistry/issues/412#issuecomment-474043461).

PR #485, opened 2019-07-24
Create a content package to create a image from a content map.

clrprod:
This is done so that there is a easy way to create images for testing,
or make simple images from within go programs.

### DIFF
--- a/pkg/v1/content/content.go
+++ b/pkg/v1/content/content.go
@@ -1,0 +1,57 @@
+// Copyright 2018 Google LLC All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Package content allows you to create images directly from contents.
+package content
+
+import (
+	"archive/tar"
+	"bytes"
+
+	v1 "github.com/google/go-containerregistry/pkg/v1"
+	"github.com/google/go-containerregistry/pkg/v1/empty"
+	"github.com/google/go-containerregistry/pkg/v1/mutate"
+	"github.com/google/go-containerregistry/pkg/v1/tarball"
+)
+
+// Image creates a image with the given contents. These images are reproducible and should be consistent.
+func Image(content ...map[string][]byte) (v1.Image, error) {
+
+	tl := []v1.Layer{}
+	for _, l := range content {
+		b := &bytes.Buffer{}
+		w := tar.NewWriter(b)
+		for f, c := range l {
+			if err := w.WriteHeader(&tar.Header{
+				Name: f,
+				Size: int64(len(c)),
+			}); err != nil {
+				return nil, err
+			}
+			if _, err := w.Write(c); err != nil {
+				return nil, err
+			}
+		}
+		if err := w.Close(); err != nil {
+			return nil, err
+		}
+		y, err := tarball.LayerFromReader(b)
+		if err != nil {
+			return nil, err
+		}
+		tl = append(tl, y)
+	}
+
+	return mutate.AppendLayers(empty.Image, tl...)
+}
--- a/pkg/v1/content/content.go
+++ b/pkg/v1/content/content.go
@@ -18,6 +18,7 @@ package content
 import (
 	"archive/tar"
 	"bytes"
+	"sort"
 
 	v1 "github.com/google/go-containerregistry/pkg/v1"
 	"github.com/google/go-containerregistry/pkg/v1/empty"
@@ -26,13 +27,21 @@ import (
 )
 
 // Image creates a image with the given contents. These images are reproducible and should be consistent.
+// The layers are applied from the begging to the end of the specified maps - the last map will be the topmost layer.
 func Image(content ...map[string][]byte) (v1.Image, error) {
-
 	tl := []v1.Layer{}
 	for _, l := range content {
 		b := &bytes.Buffer{}
 		w := tar.NewWriter(b)
-		for f, c := range l {
+
+		fn := []string{}
+		for f := range l {
+			fn = append(fn, f)
+		}
+		sort.Strings(fn)
+
+		for _, f := range fn {
+			c := l[f]
 			if err := w.WriteHeader(&tar.Header{
 				Name: f,
 				Size: int64(len(c)),
--- a/pkg/v1/content/content.go
+++ b/pkg/v1/content/content.go
@@ -26,36 +26,41 @@ import (
 	"github.com/google/go-containerregistry/pkg/v1/tarball"
 )
 
-// Image creates a image with the given contents. These images are reproducible and should be consistent.
+// Layer creates a layer from a single file map. These layers are reproducible and consistent.
+func Layer(content map[string][]byte) (v1.Layer, error) {
+	b := &bytes.Buffer{}
+	w := tar.NewWriter(b)
+
+	fn := []string{}
+	for f := range content {
+		fn = append(fn, f)
+	}
+	sort.Strings(fn)
+
+	for _, f := range fn {
+		c := content[f]
+		if err := w.WriteHeader(&tar.Header{
+			Name: f,
+			Size: int64(len(c)),
+		}); err != nil {
+			return nil, err
+		}
+		if _, err := w.Write(c); err != nil {
+			return nil, err
+		}
+	}
+	if err := w.Close(); err != nil {
+		return nil, err
+	}
+	return tarball.LayerFromReader(b)
+}
+
+// Image creates a image with the given filemaps as its contents. These images are reproducible and consistent.
 // The layers are applied from the begging to the end of the specified maps - the last map will be the topmost layer.
 func Image(content ...map[string][]byte) (v1.Image, error) {
 	tl := []v1.Layer{}
 	for _, l := range content {
-		b := &bytes.Buffer{}
-		w := tar.NewWriter(b)
-
-		fn := []string{}
-		for f := range l {
-			fn = append(fn, f)
-		}
-		sort.Strings(fn)
-
-		for _, f := range fn {
-			c := l[f]
-			if err := w.WriteHeader(&tar.Header{
-				Name: f,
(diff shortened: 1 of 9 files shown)


ImJasonH:
This definitely seems like something we want in a `testing` package, if we don't expect people to use this in a real production scenario.

clrprod:
I don't think this is really testing only - I'll probably use it in probers, and if I had to make a container with a go binary, I'd seriously think about using this over docker. It's a slightly better ux swiss army knife.

pkg/v1/content/content.go
```diff
@@ -0,0 +1,57 @@
...
+
+	v1 "github.com/google/go-containerregistry/pkg/v1"
+	"github.com/google/go-containerregistry/pkg/v1/empty"
+	"github.com/google/go-containerregistry/pkg/v1/mutate"
+	"github.com/google/go-containerregistry/pkg/v1/tarball"
+)
+
+// Image creates a image with the given contents. These images are reproducible and should be consistent.
+func Image(content ...map[string][]byte) (v1.Image, error) {
+
```
ImJasonH: nit: unnecessary newline

pkg/v1/content/content.go
```diff
@@ -0,0 +1,57 @@
...
+	"bytes"
+
+	v1 "github.com/google/go-containerregistry/pkg/v1"
+	"github.com/google/go-containerregistry/pkg/v1/empty"
+	"github.com/google/go-containerregistry/pkg/v1/mutate"
+	"github.com/google/go-containerregistry/pkg/v1/tarball"
+)
+
+// Image creates a image with the given contents. These images are reproducible and should be consistent.
+func Image(content ...map[string][]byte) (v1.Image, error) {
```
ImJasonH: What's the thinking behind making this variadic? It opens up the possibility that 2+ maps could define the same key and we'd have to document which wins in that case.

pkg/v1/content/content.go
```diff
@@ -0,0 +1,57 @@
...
+	"archive/tar"
+	"bytes"
+
+	v1 "github.com/google/go-containerregistry/pkg/v1"
+	"github.com/google/go-containerregistry/pkg/v1/empty"
+	"github.com/google/go-containerregistry/pkg/v1/mutate"
+	"github.com/google/go-containerregistry/pkg/v1/tarball"
+)
+
+// Image creates a image with the given contents. These images are reproducible and should be consistent.
```
ImJasonH: Is this true? Iterating over a map like we do on line 35 isn't guaranteed to be in the same order, so I think we could end up with different images from the same content.

pkg/v1/content/content.go
```diff
@@ -0,0 +1,57 @@
...
+				return nil, err
+			}
+			if _, err := w.Write(c); err != nil {
+				return nil, err
+			}
+		}
+		if err := w.Close(); err != nil {
+			return nil, err
+		}
+		y, err := tarball.LayerFromReader(b)
```
ImJasonH: I wonder if a `content.Layer(map[string][]byte)` would be just as helpful, and more powerful since you could `mutate.Append` content layers on top of an existing `v1.Image`.

pkg/v1/content/content_test.go
```diff
@@ -0,0 +1,81 @@
...
+
+	"github.com/google/go-containerregistry/pkg/v1/content"
+)
+
+func TestImageFromContents(t *testing.T) {
+	tcs := []struct {
+		Name     string
+		Contents []map[string][]byte
+		Digest   string
+	}{
```
ImJasonH: nit: collapse newlines and whitespace to conserve space:

```
tcs := []struct {
 ...fields...
}{{
 ...
}, {
 ...
}}
```

pkg/v1/content/content_test.go
```diff
@@ -0,0 +1,81 @@
...
+
+package content_test
+
+import (
+	"testing"
+
+	"github.com/google/go-containerregistry/pkg/v1/content"
+)
+
+func TestImageFromContents(t *testing.T) {
```
ImJasonH: Re: my earlier comment, it might be useful to have a test that asserts that two images from the same content results in the same digest etc.

jonjohnsonjr:
> if we don't expect people to use this in a real production scenario.

I'm torn on this. On one hand, this is basically how [rules_docker](https://github.com/bazelbuild/rules_docker) works, which IMO is the ideal _model_ for image construction. On the other hand, this doesn't compose that well with our other abstractions.

What if we just pulled out the middle bit and put it in `tarball`?:

```go
func LayerFromFileMap(files map[string][]byte) (v1.Layer, error) {
	b := &bytes.Buffer{}
	w := tar.NewWriter(b)
	for f, c := range l {
		if err := w.WriteHeader(&tar.Header{
			Name: f,
			Size: int64(len(c)),
		}); err != nil {
			return nil, err
		}
		if _, err := w.Write(c); err != nil {
			return nil, err
		}
	}
	if err := w.Close(); err != nil {
		return nil, err
	}
	return tarball.LayerFromReader(b)
}

```

Then this becomes a really small helper function:

```go
func Image(content ...map[string][]byte) (v1.Image, error) {
	ll := make([]v1.Layer, 0, len(content))
	for _, fm := range content {
		l, err := tarball.LayerFromFileMap(fm)
		if err != nil {
			return nil, err
		}
		ll = append(ll, l)
	}
	return mutate.AppendLayers(empty.Image, ll...)
}
```

It's really annoying that `LayerFromFileMap` would return an error (you could just inline everything otherwise), but I'm not sure under what circumstances `tar.Writer` can fail... it would be nice to do something like:

```go
func main() {
	img, _ := mutate.AppendLayers(
		tarball.LayerFromFileMap(map[string][]byte{
			"/test": []byte("testy"),
			"/bar":  []byte("not useful"),
		}),
		tarball.LayerFromFileMap(map[string][]byte{
			"/foo": []byte("fooey"),
		}),
	)
	d, _ := img.Digest()
	fmt.Println(d)
}
```

pkg/v1/content/content.go
jonjohnsonjr: > we'd have to document which wins in that case

IIUC each map is a layer, so we'd just need to document the order of these layers.

pkg/v1/content/content.go
jonjohnsonjr: We might want to make `mutate.Canonical` sort the files in each layer.

pkg/v1/content/content.go
jonjohnsonjr: You beat me to it.

pkg/v1/content/content.go
jonjohnsonjr: Though I'd err towards *also* doing [this](https://stackoverflow.com/a/18342865/6220920) for whatever code implements the
```go
func(map[string][]byte) v1.Layer
```
part.

ImJasonH:
+1 to `tarball.LayerFromFileMap`

We could have `v1util.Must(v1.Layer, error) v1.Layer` that `log.Exit`s if `err != nil` so the minimal helper becomes:

```go
img, _ := mutate.AppendLayers(empty.Image,
	v1util.Must(tarball.LayerFromFileMap(map[string][]byte{
		"/test": []byte("testy"),
		"/bar":  []byte("not useful"),
	})),
	v1util.Must(tarball.LayerFromFileMap(map[string][]byte{
		"/foo": []byte("fooey"),
	})),
)
d, _ := img.Digest()
fmt.Println(d)
```

Similar to https://golang.org/pkg/text/template/#Must

jonjohnsonjr:
> `v1util.Must(v1.Layer, error) v1.Layer`

I love this. Possibly:

`v1util.MustLayer(v1.Layer, error) v1.Layer`

So we can extend it to various common types we have to use. I'd use that all over the place for tests.

clrprod:
jon - I think you've invented monads.

pkg/v1/content/content.go
clrprod: I expect 99% of users will just use a single layer, but it's easier to get it included in the variadic form since if someone wants two layers they can do it. (it also doesn't make the code or function signature that much harder).

pkg/v1/content/content.go
clrprod: Done

pkg/v1/content/content.go
clrprod: Also updated the docs.

pkg/v1/content/content.go
clrprod: Yeah - this needs to be sorted. I thought this would be caught by the tests, but it wasn't. I added a new test which appears to fail :D.

clrprod: Done. Sorting is added and some more tests are added.

pkg/v1/content/content_test.go
clrprod: Tightened this up - I almost want a tool to do this.

pkg/v1/content/content_test.go
clrprod: Added

pkg/v1/content/content.go
clrprod: I made a content.Layer. It sounds like maybe that should live in tarball? (which may not be discoverable by the user, but if we move Layer, I'd also move Image). This also has nothing to do with tarballs other than how it is implemented. If there was a image or layer package it'd be obvious this should live there.

I also don't like content super well as a file name (so maybe filemap?).

pkg/v1/content/content_test.go
```diff
@@ -0,0 +1,164 @@
...
+			l1, _ := content.Layer(tc.Contents)
+			l2, _ := content.Layer(tc.Contents)
+
+			d1, _ := l1.Digest()
+			d2, _ := l2.Digest()
+
+			if d1 != d2 {
+				t.Fatalf("Non matching digests, want %q, got %q", d1, d2)
+			}
+
```
ImJasonH: nit: unnecessary empty lines all over this file

pkg/v1/content/content.go
clrprod: Renamed to filemap.

pkg/v1/content/content_test.go
clrprod: I hit enter a lot :P I deleted all the ones I saw hopefully.

jonjohnsonjr:
I'm weakly opposed to a new package. `filemap.Image` taking only one layer is odd. Given that `filemap.Layer` produces a tarball for the layer, I'm leaning towards keeping it in the `tarball` package.

I think adding `tarball.LayerFromFileMap` is mostly uncontroversial, even if `tarball` *feels* kind of wrong, that is captured here already: https://github.com/google/go-containerregistry/issues/413

clrprod:
Offline discussion with jonjohnsonjr - moving this to the crane package since crain is mostly sugar and this is a pretty high level abstraction.

PR #490, opened 2019-08-06
transport: Store statuscode and body within transport.Error

clrprod:
This allows for much better debugging of the returned error.

### DIFF
--- a/pkg/v1/remote/transport/error.go
+++ b/pkg/v1/remote/transport/error.go
@@ -26,6 +26,10 @@ import (
 // https://github.com/docker/distribution/blob/master/docs/spec/api.md#errors
 type Error struct {
 	Errors []Diagnostic `json:"errors,omitempty"`
+	// The http status code returned.
+	StatusCode int
+	// The raw body if we couldn't understand it.
+	rawBody string
 }
 
 // Check that Error implements error
@@ -35,7 +39,10 @@ var _ error = (*Error)(nil)
 func (e *Error) Error() string {
 	switch len(e.Errors) {
 	case 0:
-		return "<empty transport.Error response>"
+		if len(e.rawBody) == 0 {
+			return fmt.Sprintf("unsupported status code %d", e.StatusCode)
+		}
+		return fmt.Sprintf("unsupported status code %d; body: %s", e.StatusCode, e.rawBody)
 	case 1:
 		return e.Errors[0].String()
 	default:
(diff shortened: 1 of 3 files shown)


pkg/v1/remote/transport/error.go
```diff
@@ -26,6 +26,10 @@ import (
 // https://github.com/docker/distribution/blob/master/docs/spec/api.md#errors
 type Error struct {
 	Errors []Diagnostic `json:"errors,omitempty"`
+	// The http status code returned.
+	StatusCode int
+	// The raw body if we couldn't understand it.
+	RawBody string
```
jonjohnsonjr: I'd not expose `RawBody` until we actually have a use case for it.

clrprod: Done.

pkg/v1/remote/transport/error_test.go
```diff
@@ -94,10 +98,22 @@ func TestCheckErrorNotError(t *testing.T) {
...
-		} else if se, ok := err.(*Error); ok {
-			t.Errorf("CheckError(%d, %s) = %v, wanted another type", test.code, test.body, se)
+		err := CheckError(resp, http.StatusOK)
+		if err == nil {
+			t.Fatalf("CheckError(%d, %s) = nil, wanted error", test.code, test.body)
+		}
+		se, ok := err.(*Error)
+		if !ok {
+			t.Fatalf("CheckError(%d, %s) = %v, wanted error type", test.code, test.body, err)
+
```
jonjohnsonjr: nit: extra newline here

pkg/v1/remote/transport/error.go
```diff
@@ -35,7 +39,7 @@ var _ error = (*Error)(nil)
 func (e *Error) Error() string {
 	switch len(e.Errors) {
 	case 0:
-		return "<empty transport.Error response>"
+		return fmt.Sprintf("unsupported status code %d; body: %q", e.StatusCode, e.rawBody)
```
jonjohnsonjr: Not super important, but I would guess `%q` is going to look bad pretty often for multi-line bodies. It might also be nice to just omit the body if it's empty.

pkg/v1/remote/transport/error_test.go
clrprod: Done

pkg/v1/remote/transport/error.go
clrprod: Used %s and skipped body if empty.

PR #498, opened 2019-08-28
Add so much coverage

jonjohnsonjr:
Add test coverage for any low hanging fruit.
Add some more validation to pkg/v1/validate.
Add .codecov.yml to ignore generated files.

Real fixes this exposed:

* Drop unused method in partial.
* Make crane.Export take a Writer instead of WriteCloser.
* Fix log.Fatal instead of returning err in pkg/crane.
* Fix some invalid error messages in pkg/crane.
* Make GzipReadCloserLevel not return an error.
* Fix dropped pipe errors (https://golang.org/issue/24283).
* Make ParseReference return an ErrBadName.
* Remove redundant tarball.Write functions.

### DIFF
--- a/pkg/crane/crane_test.go
+++ b/pkg/crane/crane_test.go
@@ -1,0 +1,300 @@
+// Copyright 2019 Google LLC All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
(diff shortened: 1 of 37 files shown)


pkg/crane/crane_test.go
```diff
@@ -0,0 +1,300 @@
...
+	} {
+		if err == nil {
+			t.Error("expected err, got nil")
+		}
+	}
+}
+
+// e drops the first parameter so we can use the result of a function
+// that returns two values as an expression above. This is a bit of a go quirk.
+func e(_ interface{}, err error) error {
```
ImJasonH: Nit: define this in the method it's used? (Up to you)

PR #503, opened 2019-09-04
Add random.Layer

jonjohnsonjr:
This will make it easier to test foreign layers.

### DIFF
--- a/pkg/v1/random/image.go
+++ b/pkg/v1/random/image.go
@@ -18,9 +18,12 @@ import (
 	"archive/tar"
 	"bytes"
 	"crypto/rand"
+	"crypto/sha256"
+	"encoding/hex"
 	"fmt"
 	"io"
 	"io/ioutil"
+	mrand "math/rand"
 	"time"
 
 	v1 "github.com/google/go-containerregistry/pkg/v1"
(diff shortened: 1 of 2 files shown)


pkg/v1/random/image.go
```diff
@@ -141,3 +124,32 @@ func (i *image) LayerByDiffID(diffID v1.Hash) (partial.UncompressedLayer, error)
 	}
 	return l, nil
 }
+
+// Layer returns a layer with pseudo-randomly generated content.
+func Layer(byteSize int64, fileName string, mt types.MediaType) (v1.Layer, error) {
```
ImJasonH: Seems weird to have `random.Layer` take a filename. Can we just generate a random filename too?

If you think it should stay, at least document why it's needed in the doc comment.

pkg/v1/random/image.go
```diff
@@ -141,3 +124,32 @@ func (i *image) LayerByDiffID(diffID v1.Hash) (partial.UncompressedLayer, error)
...
+	var b bytes.Buffer
```
ImJasonH: We can get the digest using `io.MultiWriter` while we `CopyN` it, and avoid buffering.

Issue #207, opened 2018-06-06
Support foreign/restricted layers


dekkagaijin:
This is mostly an issue when naively attempting to pull layers. 

Docker V2 Schema2 media type: `application/vnd.docker.image.rootfs.foreign.diff.tar.gzip`

OCI NonDistributable: https://github.com/opencontainers/image-spec/blob/master/specs-go/v1/mediatype.go

Tagging @jonjohnsonjr for others that I missed

dekkagaijin:
I'll be doing this for the python lib, and will do the work for ggcr, as well, if it isn't picked up beforehand.

genevieve:
Hey @jonjohnsonjr! Are you looking for someone to make a PR for this functionality? Are you interested in this library supporting this functionality?

dekkagaijin:
@genevieve all the love this lib gets is a good thing, most likely.

I only updated the Python lib to support remote -> remote registry pushes before leaving the team, and don't know if it's been updated internally or externally to support local pulls or pushes.

jonjohnsonjr:
@genevieve yeah we're definitely interested in supporting this, I just haven't found the time to add it.

At one point I started working on it but got distracted: https://github.com/google/go-containerregistry/compare/master...jonjohnsonjr:foreign

I'm not sure how far behind that branch is -- rebasing might be harder than starting from scratch :)

genevieve:
@dekkagaijin @jonjohnsonjr Thank you both for your quick responses! I'm going to try working on it now and I'll get back to you if I find myself in over my head!

PR #506, opened 2019-09-04
Implement foreign layer support

jonjohnsonjr:
Fixes https://github.com/google/go-containerregistry/issues/207

* Support downloading layers from the `URLs` field in a `remote.Image` manifest.
* Skip uploading foreign layers.
* Populate `LayerSources` when writing a `tarball.Image`.
* Read `LayeSources` when reading a `tarball.Image`.

### DIFF
--- a/pkg/v1/hash_test.go
+++ b/pkg/v1/hash_test.go
@@ -83,3 +83,28 @@ func TestSHA256(t *testing.T) {
 		t.Errorf("n; got %v, want %v", got, want)
 	}
 }
+
+// This tests that you can use Hash as a key in a map (needs to implement both
+// MarshalText and UnmarshalText).
+func TestTextMarshalling(t *testing.T) {
+	foo := make(map[Hash]string)
+	b, err := json.Marshal(foo)
+	if err != nil {
+		t.Fatalf("could not marshal: %v", err)
+	}
+	if err := json.Unmarshal(b, &foo); err != nil {
+		t.Errorf("could not unmarshal: %v", err)
+	}
+
+	h := &Hash{
+		Algorithm: "sha256",
+		Hex:       strings.Repeat("a", 64),
+	}
+	text, err := h.MarshalText()
+	if err != nil {
+		t.Fatal(err)
+	}
+	if err := h.UnmarshalText(text); err != nil {
+		t.Fatal(err)
+	}
+}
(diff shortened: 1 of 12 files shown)


jonjohnsonjr:
cc @smukherj1

pkg/v1/hash_test.go
```diff
@@ -83,3 +83,28 @@ func TestSHA256(t *testing.T) {
 		t.Errorf("n; got %v, want %v", got, want)
 	}
 }
+
+// This tests that you can use Hash as a key in a map (needs to implement both
+// MarshalText and UnmarshalText).
+func TestTextMarshalling(t *testing.T) {
+	foo := make(map[Hash]string)
```
ImJasonH: Is this JSON-marshalling test more than an elaborate type assertion, checking that it satisfies the `TextMarshaler` interface?

pkg/v1/hash_test.go
```diff
@@ -83,3 +83,28 @@ func TestSHA256(t *testing.T) {
...
+
+	h := &Hash{
+		Algorithm: "sha256",
+		Hex:       strings.Repeat("a", 64),
+	}
+	text, err := h.MarshalText()
+	if err != nil {
+		t.Fatal(err)
+	}
+	if err := h.UnmarshalText(text); err != nil {
```
ImJasonH: Do you want to test that the hex value persists as expected through marshal/unmarshaling?

pkg/v1/remote/image.go
```diff
@@ -132,18 +132,42 @@ func (rl *remoteLayer) Digest() (v1.Hash, error) {
...
+	for _, s := range d.URLs {
+		u, err := url.Parse(s)
+		if err != nil {
+			return nil, err
+		}
+		urls = append(urls, *u)
+	}
+
+	// The lastErr for most pulls will be the same (the first error), but for
+	// foreign layers we'll want to surface the last one.
```
ImJasonH: Why do we want the last one?

pkg/v1/hash_test.go
jonjohnsonjr: This one is weird to me. Originally, I just had this json marshalling code in the test, which would fail unless I implemented the text marshalling. Unfortunately, the json marshalling doesn't actually provide coverage for the text marshalling implementation (I don't understand why), so I added the second half of this test to grab the coverage.

pkg/v1/remote/image.go
jonjohnsonjr: Updated the comment.

Issue #500, opened 2019-09-03
Support copying schema 1 images


jonjohnsonjr:
This has been bugging me for a while, and it seems like it wouldn't be too bad to fix.

Already discussed this to death why we don't want to fully support schema 1:
https://github.com/google/go-containerregistry/issues/296
https://github.com/google/go-containerregistry/issues/377

However, it wouldn't be too bad to just support copying schema 1 images from one registry to another. All we need to do is download/upload each blob then PUT the manifest. We wouldn't need to compute diffids or synthesize a config file.

This might also be a nice opportunity to decompose some of the remote functionality into smaller pieces, e.g. `remote.Blob` and `remote.WriteBlob`.

PR #509, opened 2019-09-05
Implement schema 1 image copying

jonjohnsonjr:
Adds a new package, pkg/internal/legacy, where we can keep this without worrying about the API being terrible.

Fixes https://github.com/google/go-containerregistry/issues/500


jonjohnsonjr:
cc @listx

PR #510, opened 2019-09-05
Update Keychain interface to accept repo

### LINKED ISSUE #495: k8schain: Support repo-level authentication
From https://github.com/knative/serving/issues/4435#issuecomment-506207586

Docker ([here](https://godoc.org/github.com/docker/cli/cli/config/configfile#ConfigFile.GetCredentialsStore) and [here](https://godoc.org/github.com/docker/cli/cli/config/credentials#Store)) and [ggcr](https://godoc.org/github.com/google/go-containerregistry/pkg/authn#Keychain) both expect credentials to be configured per registry, whereas [kubernetes](https://godoc.org/k8s.io/kubernetes/pkg/credentialprovider#DockerKeyring) allows them to be configured per-repository:
> DockerKeyring tracks a set of docker registry credentials, maintaining a reverse index across the registry endpoints. A registry endpoint is made up of a host (e.g. registry.example.com), but it may also contain a path (e.g. registry.example.com/foo) This index is important for two reasons:
> - registry endpoints may overlap, and when this happens we must find the most specific match for a given image
> - iterating a map does not yield predictable results

(I haven't actually played with docker's behavior to see if the `registryHostname` string accepts a path, but I suspect it does not.)

We need to figure out a reasonable way to support both these things.

Without thinking about it too much, we could have our Keychain interface accept a [`Repository`](https://godoc.org/github.com/google/go-containerregistry/pkg/name#Repository) (instead of a [`Registry`](https://godoc.org/github.com/google/go-containerregistry/pkg/name#Registry)) and have our [`DefaultKeychain`](https://godoc.org/github.com/google/go-containerregistry/pkg/authn#DefaultKeychain) just pull out the registry part, whereas the [`k8schain`](https://godoc.org/github.com/google/go-containerregistry/pkg/authn/k8schain) keychain would pass along the entire repository... open to better ideas though :smile:

jonjohnsonjr:
Fixes #495

Introduces `authn.Target` which both `name.Repository` and `name.Registry` implement, so this should be backwards compat.

### DIFF
--- a/pkg/authn/k8schain/k8schain_test.go
+++ b/pkg/authn/k8schain/k8schain_test.go
@@ -146,7 +146,7 @@ func TestImagePullSecrets(t *testing.T) {
 	for _, tc := range []struct {
 		name   string
 		auth   authn.Authenticator
-		target authn.Target
+		target authn.Resource
 	}{{
 		name:   "registry",
 		auth:   &authn.Basic{Username: username, Password: password},
--- a/pkg/authn/k8schain/k8schain_test.go
+++ b/pkg/authn/k8schain/k8schain_test.go
@@ -123,7 +123,7 @@ func TestImagePullSecrets(t *testing.T) {
 		Type: corev1.SecretTypeDockercfg,
 		Data: map[string][]byte{
 			corev1.DockerConfigKey: []byte(
-				fmt.Sprintf(`{"fake.registry.io": {"auth": "%s"}, "fake.registry.io/more/specific": {"auth": "%s"}}`,
+				fmt.Sprintf(`{"fake.registry.io": {"auth": %q}, "fake.registry.io/more/specific": {"auth": %q}}`,
 					base64.StdEncoding.EncodeToString([]byte(username+":"+password)),
 					base64.StdEncoding.EncodeToString([]byte(specificUser+":"+specificPass))),
 			),
--- a/pkg/authn/k8schain/k8schain_test.go
+++ b/pkg/authn/k8schain/k8schain_test.go
@@ -109,6 +109,7 @@ func TestAttachedServiceAccount(t *testing.T) {
 
 func TestImagePullSecrets(t *testing.T) {
 	username, password := "foo", "bar"
+	specificUser, specificPass := "very", "specific"
 	client := fakeclient.NewSimpleClientset(&corev1.ServiceAccount{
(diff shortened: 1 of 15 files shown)


pkg/authn/k8schain/k8schain_test.go
```diff
@@ -122,8 +123,9 @@ func TestImagePullSecrets(t *testing.T) {
 		Type: corev1.SecretTypeDockercfg,
 		Data: map[string][]byte{
 			corev1.DockerConfigKey: []byte(
-				fmt.Sprintf(`{"fake.registry.io": {"auth": "%s"}}`,
-					base64.StdEncoding.EncodeToString([]byte(username+":"+password))),
+				fmt.Sprintf(`{"fake.registry.io": {"auth": "%s"}, "fake.registry.io/more/specific": {"auth": "%s"}}`,
```
mattmoor:
```suggestion
				fmt.Sprintf(`{"fake.registry.io": {"auth": %q}, "fake.registry.io/more/specific": {"auth": %q}}`,
```

pkg/authn/keychain.go
```diff
@@ -24,13 +24,17 @@ import (
 	"runtime"
 
 	"github.com/google/go-containerregistry/pkg/logs"
-	"github.com/google/go-containerregistry/pkg/name"
 )
 
+type Target interface {
```
mattmoor: Target isn't a great name, document this and members?

pkg/authn/keychain.go
```diff
@@ -24,13 +24,17 @@ import (
...
+	String() string
+	RegistryStr() string
+}
+
 // Keychain is an interface for resolving an image reference to a credential.
 type Keychain interface {
 	// Resolve looks up the most appropriate credential for the specified registry.
-	Resolve(name.Registry) (Authenticator, error)
+	Resolve(Target) (Authenticator, error)
```
mattmoor: stale comment

pkg/authn/keychain.go
jonjohnsonjr: Better idea than `Target`? `Principal`? `Authority`?

jonjohnsonjr: `authn.Target` seems fine when qualified. I'm open to better suggestions.

PR #512, opened 2019-09-06
gcrane: use a task queue when copying images

### LINKED ISSUE #424: Gcrane: unrecognized HTTP status: 429 Too Many Requests

jonjohnsonjr:
This will naturally throttle us based on our concurrency.

Also adds retry with backoff in case we hit 429s, 5xxs or temporary
errors with GCR.

Fixes #424

### DIFF
--- a/pkg/gcrane/copy.go
+++ b/pkg/gcrane/copy.go
@@ -18,11 +18,15 @@ import (
 	"context"
 	"fmt"
 	"log"
+	"net/http"
+	"runtime"
 	"strings"
+	"time"
 
 	"github.com/google/go-containerregistry/pkg/authn"
 	"github.com/google/go-containerregistry/pkg/crane"
 	"github.com/google/go-containerregistry/pkg/internal/legacy"
+	"github.com/google/go-containerregistry/pkg/internal/retry"
 	"github.com/google/go-containerregistry/pkg/logs"
 	"github.com/google/go-containerregistry/pkg/name"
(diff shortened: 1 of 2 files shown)


pkg/gcrane/copy.go
```diff
@@ -33,6 +37,27 @@ import (
...
 
+// These numbers are based on GCR's posted quotas:
+// https://cloud.google.com/container-registry/quotas
+// -  30k requests per 10 minutes.
+// - 500k requests per 24 hours.
+//
+// On error, we will wait for:
+// - 6 seconds (in case of very short term 429s from GCS), then
+// - 1 minute (in case of temporary network issues), then
+// - 10 minutes (to get around GCR 10 minute quotas), then fail.
```
ImJasonH: What's the `Cap: 1 * time.Hour` for?

jonjohnsonjr: Basically to make sure I don't have an off-by-one error in my understanding of ExponentialBackoff. We don't really want to sit around for any longer than an hour, ever.

PR #513, opened 2019-09-06
Undo breaking change

jonjohnsonjr:


pkg/v1/tarball/write.go
```diff
@@ -46,27 +40,51 @@ func WriteToFile(p string, ref name.Reference, img v1.Image) error {
 }
 
 // MultiWriteToFile writes in the compressed format to a tarball, on disk.
-func MultiWriteToFile(p string, refToImage map[name.Reference]v1.Image) error {
+// This is just syntactic sugar wrapping tarball.MultiWrite with a new file.
+func MultiWriteToFile(p string, tagToImage map[name.Tag]v1.Image) error {
+	var refToImage map[name.Reference]v1.Image = make(map[name.Reference]v1.Image, len(tagToImage))
```
ImJasonH: nit: Could use `:=` instead of `var`

pkg/v1/tarball/write.go
```diff
@@ -46,27 +40,51 @@ func WriteToFile(p string, ref name.Reference, img v1.Image) error {
...
 }
 
 // MultiWrite writes the contents of each image to the provided reader, in the compressed format.
 // The contents are written in the following format:
 // One manifest.json file at the top level containing information about several images.
 // One file for each layer, named after the layer's SHA.
 // One file for the config blob, named after its SHA.
-func MultiWrite(refToImage map[name.Reference]v1.Image, w io.Writer) error {
+func MultiWrite(tagToImage map[name.Tag]v1.Image, w io.Writer) error {
+	var refToImage map[name.Reference]v1.Image = make(map[name.Reference]v1.Image, len(tagToImage))
```
ImJasonH: ditto

PR #514, opened 2019-09-06
Fix linter issues

jonjohnsonjr:
Also add linter to travis

### DIFF
--- a/pkg/internal/retry/retry.go
+++ b/pkg/internal/retry/retry.go
@@ -22,6 +22,7 @@ import (
 	"github.com/google/go-containerregistry/pkg/internal/retry/wait"
 )
 
+// Backoff is an alias for wait.Backoff.
 type Backoff = wait.Backoff
 
 // This is implemented by several errors in the net package as well as our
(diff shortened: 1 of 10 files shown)


pkg/internal/retry/retry.go
```diff
@@ -22,6 +22,7 @@ import (
 	"github.com/google/go-containerregistry/pkg/internal/retry/wait"
 )
 
+// Backoff is an alias for wait.Backoff.
```
ImJasonH: giphy (2)

If it's for backward-compatibility, users shouldn't be depending in `internal` packages anyway.

pkg/v1/remote/descriptor.go
```diff
@@ -43,6 +43,7 @@ type ErrSchema1 struct {
 	schema string
 }
 
+// NewErrSchema1 returns an ErrSchema1 with the unexpected MediaType.
 func NewErrSchema1(schema types.MediaType) error {
```
ImJasonH: (Didn't catch it in the last review) Does this need to be exported?

pkg/internal/retry/retry.go
jonjohnsonjr: I didn't know until I went to fix it, see comment.

PR #516, opened 2019-09-09
Add staticcheck to travis

jonjohnsonjr:
And fix all of its suggestions

Coincidentally, one of the issues in my comment of the "empty branch" tag path of remote.Descriptor was closed today, so drop that issue from the `Do nothing for tags; I give up.` comment 😄


pkg/crane/crane_test.go
```diff
@@ -107,7 +107,7 @@ func TestCraneRegistry(t *testing.T) {
 		t.Fatal(err)
 	}
 
-	pulled, err = crane.Pull(dst)
+	_, err = crane.Pull(dst)
```
ImJasonH:
```
if _, err := crane.Pull(...); err != nil {
```

Issue #349, opened 2019-01-15
Optimize remote tagging


jonjohnsonjr:
We don't currently expose an easy way to just add a tag to an existing image, e.g. something like:

```go
package remote

func Tag(img v1.Image, auth authn.Authenticator, transport http.Roundtripper, tags string...) error { ... }
```

This can be implemented as a single PUT of the manifest per tag (+ the token handshake).

This would be useful here:
https://github.com/GoogleContainerTools/skaffold/blob/da2090d35d95aa7c6722215a235d8e16b2186fb3/pkg/skaffold/docker/remote.go#L30

And here:
https://github.com/google/go-containerregistry/blob/686906b30117e91502533cb0569798af65a1769f/pkg/ko/publish/default.go#L103-L108

Alternatively, we can implement this logic:
https://github.com/google/containerregistry/blob/5b4a2cd1adeb4cc04cd30b89490f8748f407665e/client/v2_2/docker_session_.py#L297-L304

Perhaps also race the layer uploads in `remote.Write`, cancelling them if the image already exists?

jonjohnsonjr:
cc @nkubala

PR #517, opened 2019-09-11
Add ability to cache/push compressed layers

choo-stripe:
Hey!

I did a bunch of profiling on the library and made some optimizations. For the image I was experimenting with, push timing went from 2mins, 25 seconds to 49 seconds.

I know I've done some bad things here, but would love to work together to get some of these savings merged in some form or another.

I'll leave some comments throughout to share some thoughts so we can move forward.

### DIFF
--- a/pkg/v1/image.go
+++ b/pkg/v1/image.go
@@ -25,6 +25,11 @@ type Image interface {
 	// The order of the list is oldest/base layer first, and most-recent/top layer last.
 	Layers() ([]Layer, error)
 
+	// CompressedLayers returns the ordered collection of compressed filesystem layers that
+	// comprise this image. The order of the list is oldest/base layer first, and
+	// most-recent/top layer last. CompressedLayers are useful for optimized pushing.
+	CompressedLayers() ([]Layer, error)
+
 	// MediaType of this image's manifest.
 	MediaType() (types.MediaType, error)
 
--- a/pkg/v1/partial/uncompressed.go
+++ b/pkg/v1/partial/uncompressed.go
@@ -17,11 +17,13 @@ package partial
 import (
 	"bytes"
 	"io"
+	"io/ioutil"
 	"sync"
 
 	v1 "github.com/google/go-containerregistry/pkg/v1"
 	"github.com/google/go-containerregistry/pkg/v1/types"
 	"github.com/google/go-containerregistry/pkg/v1/v1util"
+	"golang.org/x/sync/errgroup"
 )
 
 // UncompressedLayer represents the bare minimum interface a natively
@@ -108,6 +110,9 @@ func UncompressedToImage(uic UncompressedImageCore) (v1.Image, error) {
 type uncompressedImageExtender struct {
 	UncompressedImageCore
 
+	// this is used to cache compressedLayers if the layers ever get compressed, because compression can be expensive
+	compressedLayers []v1.Layer
+
 	lock     sync.Mutex
 	manifest *v1.Manifest
 }
@@ -148,7 +153,13 @@ func (i *uncompressedImageExtender) Manifest() (*v1.Manifest, error) {
 		},
 	}
 
-	ls, err := i.Layers()
+	// TODO this shouldn't depend on layers being populated
+	var ls []v1.Layer
+	if i.compressedLayers != nil {
+		ls, err = i.CompressedLayers()
+	} else {
+		ls, err = i.Layers()
+	}
 	if err != nil {
 		return nil, err
 	}
@@ -207,6 +218,48 @@ func (i *uncompressedImageExtender) Layers() ([]v1.Layer, error) {
 	return ls, nil
 }
 
+// CompressedLayers implements v1.Image
+// This function compresses layers in parallel, so it is fairly expensive computationally.
+func (i *uncompressedImageExtender) CompressedLayers() ([]v1.Layer, error) {
+	if i.compressedLayers != nil {
+		return i.compressedLayers, nil
+	}
+
+	diffIDs, err := DiffIDs(i)
+	if err != nil {
+		return nil, err
+	}
+
+	ls := make([]v1.Layer, len(diffIDs), len(diffIDs))
+
+	var g errgroup.Group
+	for idx, d := range diffIDs {
+		idx := idx
+		d := d
+
+		g.Go(func() error {
+			l, err := i.LayerByDiffID(d)
+			if err != nil {
+				return err
+			}
+
+			ls[idx], err = ToCompressedLayer(l)
+			if err != nil {
+				return err
+			}
+
+			return nil
+		})
+	}
+	if err := g.Wait(); err != nil {
+		return nil, err
+	}
+
+	i.compressedLayers = ls
+
+	return ls, nil
+}
+
 // LayerByDiffID implements v1.Image
 func (i *uncompressedImageExtender) LayerByDiffID(diffID v1.Hash) (v1.Layer, error) {
 	ul, err := i.UncompressedImageCore.LayerByDiffID(diffID)
@@ -223,4 +276,59 @@ func (i *uncompressedImageExtender) LayerByDigest(h v1.Hash) (v1.Layer, error) {
 		return nil, err
 	}
 	return i.LayerByDiffID(diffID)
+}
+
+// compressedLayerFromUncompressedLayer implements partial.CompressedLayer
+type compressedLayerFromUncompressedLayer struct {
+	digest    v1.Hash
+	contents  []byte
+	mediaType types.MediaType
+}
+
+// ToCompressedLayer converts an uncompressed layer to a compressed layer
+func ToCompressedLayer(l v1.Layer) (v1.Layer, error) {
+	mediaType, err := l.MediaType()
+	if err != nil {
+		return nil, err
+	}
+	uncompressedReader, err := l.Uncompressed()
+	if err != nil {
+		return nil, err
+	}
+	gzipReadCloser := v1util.GzipReadCloser(uncompressedReader)
+
+	// TODO probably shouldnt cache the contents, should probably just re-gzip but aint nobody got time for that
+	buf := new(bytes.Buffer)
+	digest, _, err := v1.SHA256(io.TeeReader(gzipReadCloser, buf))
+	if err != nil {
+		return nil, err
+	}
+
+	cl := &compressedLayerFromUncompressedLayer{
+		mediaType: mediaType,
+		contents:  buf.Bytes(),
+		digest:    digest,
+	}
+
+	return CompressedToLayer(cl)
+}
+
+// Digest implements partial.CompressedLayer
+func (clft *compressedLayerFromUncompressedLayer) Digest() (v1.Hash, error) {
+	return clft.digest, nil
+}
+
+// Compressed implements partial.CompressedLayer
+func (clft *compressedLayerFromUncompressedLayer) Compressed() (io.ReadCloser, error) {
+	return ioutil.NopCloser(bytes.NewReader(clft.contents)), nil
+}
+
+// MediaType implements partial.CompressedLayer
+func (clft *compressedLayerFromUncompressedLayer) MediaType() (types.MediaType, error) {
+	return clft.mediaType, nil
+}
+
+// Size implements partial.CompressedLayer
+func (clft *compressedLayerFromUncompressedLayer) Size() (int64, error) {
+	return int64(len(clft.contents)), nil
 }
--- a/pkg/v1/remote/write.go
+++ b/pkg/v1/remote/write.go
@@ -41,17 +41,91 @@ type manifest interface {
 
 // Write pushes the provided img to the specified image reference.
 func Write(ref name.Reference, img v1.Image, options ...Option) error {
-	ls, err := img.Layers()
+	return WriteMulti([]name.Reference{ref}, img, options...)
+}
+
+// WriteMulti pushes the provided img to multiple specified image references.
+func WriteMulti(refs []name.Reference, img v1.Image, options ...Option) error {
+	// dedupe registries
+	dedupedRefs := map[string]name.Reference{}
+	for _, ref := range refs {
+		dedupedRefs[ref.Context().Registry.Name()] = ref
+	}
+
+	var g errgroup.Group
+	for _, ref := range dedupedRefs {
+		ref := ref
+		g.Go(func() error {
+			return writeLayers(ref, img, options...)
+		})
+	}
+	err := g.Wait()
 	if err != nil {
 		return err
 	}
 
+	g = errgroup.Group{}
+	for _, ref := range refs {
+		ref := ref
+		g.Go(func() error {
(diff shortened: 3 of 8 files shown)


pkg/v1/partial/uncompressed.go
```diff
@@ -148,7 +153,13 @@ func (i *uncompressedImageExtender) Manifest() (*v1.Manifest, error) {
 		},
 	}
 
-	ls, err := i.Layers()
+	// TODO this shouldn't depend on layers being populated
```
choo-stripe: We don't pass options into manifest generation, so we do some hackery here. Calling `i.Layers()` is slow, so this saved 10s of seconds.

pkg/v1/partial/uncompressed.go
```diff
@@ -207,6 +218,48 @@ func (i *uncompressedImageExtender) Layers() ([]v1.Layer, error) {
...
+
+			ls[idx], err = ToCompressedLayer(l)
+			if err != nil {
+				return err
+			}
+
+			return nil
+		})
+	}
+	if err := g.Wait(); err != nil {
```
choo-stripe: Parallelizing here saved >1 minute.

pkg/v1/partial/uncompressed.go
```diff
@@ -224,3 +277,58 @@ func (i *uncompressedImageExtender) LayerByDigest(h v1.Hash) (v1.Layer, error) {
...
+		return nil, err
+	}
+	uncompressedReader, err := l.Uncompressed()
+	if err != nil {
+		return nil, err
+	}
+	gzipReadCloser := v1util.GzipReadCloser(uncompressedReader)
+
+	// TODO probably shouldnt cache the contents, should probably just re-gzip but aint nobody got time for that
+	buf := new(bytes.Buffer)
```
choo-stripe: I was thinking about users passing in a `io.ReadWriter` to allocate their own storage or something. Any feedback welcome.

pkg/v1/remote/write.go
```diff
@@ -41,17 +41,91 @@ type manifest interface {
 
 // Write pushes the provided img to the specified image reference.
 func Write(ref name.Reference, img v1.Image, options ...Option) error {
-	ls, err := img.Layers()
+	return WriteMulti([]name.Reference{ref}, img, options...)
+}
+
+// WriteMulti pushes the provided img to multiple specified image references.
+func WriteMulti(refs []name.Reference, img v1.Image, options ...Option) error {
```
choo-stripe: Writing multiple tags seemed to be expensive (digesting, compressing, whatever). Adding the ability to just push refs cut about 20 seconds at the end.

pkg/v1/remote/write.go
```diff
@@ -41,17 +41,91 @@ type manifest interface {
...
+	// dedupe registries
```
choo-stripe: deduping is important for rate limits and for the network calls for checking if blobs exist. A lot of layers can really have an impact here - especially when pushing a bunch of tags to one repo.

pkg/v1/image.go
```diff
@@ -25,6 +25,11 @@ type Image interface {
 	// The order of the list is oldest/base layer first, and most-recent/top layer last.
 	Layers() ([]Layer, error)
 
+	// CompressedLayers returns the ordered collection of compressed filesystem layers that
+	// comprise this image. The order of the list is oldest/base layer first, and
+	// most-recent/top layer last. CompressedLayers are useful for optimized pushing.
+	CompressedLayers() ([]Layer, error)
```
jonjohnsonjr: Why is this better than just calling `Compressed` on each layer that gets returned? It's unclear to me why this is needed.

pkg/v1/partial/uncompressed.go
jonjohnsonjr: Why is `Layers` slow?

pkg/v1/partial/uncompressed.go
jonjohnsonjr: I think we might actually want to move the parallelization into here?

https://github.com/google/go-containerregistry/blob/b02d448a3705facf11018efff34f1d2830be5724/pkg/v1/mutate/mutate.go#L153-L190

pkg/v1/partial/uncompressed.go
jonjohnsonjr: I'm not 100% sure what's going on here.

pkg/v1/remote/write.go
jonjohnsonjr: I think we can partially solve some of this by just implementing an explicit tagging operation, see: https://github.com/google/go-containerregistry/issues/349

pkg/v1/remote/write.go
jonjohnsonjr: Deduping per registry is interesting... I hadn't thought of that. I think this is mostly addressed by https://github.com/google/go-containerregistry/issues/349, but I can imagine deduping blobs across multiple images to avoid extraneous existence checks (but that's not covered here).

jonjohnsonjr:
> Hey!

Hello! 👋 

> I did a bunch of profiling on the library and made some optimizations. For the image I was experimenting with, push timing went from 2mins, 25 seconds to 49 seconds.

This is awesome, thanks for taking the time to experiment with this 😄 

> I know I've done some bad things here, but would love to work together to get some of these savings merged in some form or another.
>
> I'll leave some comments throughout to share some thoughts so we can move forward.

Absolutely. I'm really interested in how you're using this package -- we might be able to optimize things further, or I could point you in a better direction. Could you put together a small, representative program that demonstrates some of the slowness you're seeing? I'd love to be able to have a reproducible test to benchmark any changes against. We have a small `registry` package for testing that implements most of the registry protocol, if that helps.

jonjohnsonjr:
There's also https://godoc.org/github.com/google/go-containerregistry/pkg/v1/cache which might be useful for caching gzipped things.

pkg/v1/image.go
choo-stripe: Through profiling today, I found out that `Layers()` was called multiple times and returned _new_ layers. So the speedup was because `CompressedLayers` cached the compressed layers, while `Layers` did not.

pkg/v1/partial/uncompressed.go
choo-stripe: I imagine it was calculating digests for the first time (since we called `CompressedLayers` above but I'm not 100% sure.

pkg/v1/partial/uncompressed.go
choo-stripe: Do uncompressed images back on to `mutate`? I haven't traced it deep enough into the stack.

pkg/v1/partial/uncompressed.go
choo-stripe: This reads the compressed layer into RAM - which comes with tradeoffs. Could run out of RAM, but also get super fast layer contents access.

pkg/v1/remote/write.go
choo-stripe: Yep!

pkg/v1/remote/write.go
choo-stripe: yep!

choo-stripe:
@jonjohnsonjr given your questions and ideas, I tried to come up with a less invasive way to speed things up (how to make the existing implementation fast instead of replacing it entirely). Given that, I did a bunch of profiling and clearly wrote out my findings for each commit. For the test harness that I wrote here, we went from ~1 minute to 20 seconds - which matches my old implementation.

I hope that by reading commit messages from oldest to newest, you'll be able to see at each stage what was slow and how the optimization helped.

I think this PR is now in a much better place.

Remaining issues that concern me:
1) We now push the config layer compressed instead of uncompressed. We need to verify that this doesn't break any assumptions (i.e. making sure the manifest doesn't use the uncompressed sha - I don't think it should)
2) We may not want to pull layers into ram before uploading. That `Pull layers into RAM before sending them` sped things up quite a bit for our specific test harness here, but I'm not convinced it would materially impact perf with a more robust/parallellizable registry. I image we want to revert d16f7f9.

choo-stripe:
I made some minor modifications to the cache and added an in-memory cache which gave ~the same perf gains because it also dedupes a lot of the compression/digesting/etc - https://github.com/choo-stripe/go-containerregistry/compare/choo/simplifying...choo-stripe:choo/inmemcache

jonjohnsonjr:
> Btw, I really appreciate you taking the time to work through this with me.

Of course! Thanks for taking the time to do this!

Going to just explain a few things that might answer most of your questions, let me know if I missed anything:

In the happy case, we know the digest ahead of time, and our upload sequence should look like this:
image

That `HEAD` is cheap if we already know the digest; we just get a 404 and subsequently do the complete upload. If we get a 200, we can skip the entire upload, so that's a huge win.

In the sad case (what you're hitting, I believe), we have uncompressed blobs, and we *don't* know the digest ahead of time. I think we're still hitting the above strategy because we're eagerly calculating the digests when we read layers from the tarball. Ideally, we wouldn't do that at all! We would just skip the first `HEAD` request and hash the contents as we upload them:
image


We actually have an implementation for this, [`stream.Layer`](https://godoc.org/github.com/google/go-containerregistry/pkg/v1/stream#Layer), which @ImJasonH implemented, but it has some problems and we're not using it anywhere by default.

*Sometimes* the best thing to do is digest the content locally before uploading, so you can do an existence check first. E.g. if your internet is slow, it's better trade CPU time just in case the registry already has the blob. In most cases, though, it's better to just upload it.

> How do you feel about the updated PR?

You're [pre-calculation of digests](https://github.com/google/go-containerregistry/pull/517/files#diff-e24ea77b0ff8bbbb37a8b713172d1bd8R71-R87) runs afoul of [this code in `remote`](https://github.com/google/go-containerregistry/blob/bb2eda47dcc0c4cb5d554415f786c8fbe45e42fb/pkg/v1/remote/write.go#L83-L97), so I'd be reluctant to merge that bit.

It might be helpful to explain what needs to happen during an image upload. Here's a diagram of data flow through the system for an image with one layer, expressed via analogous tools:
image

(this is basically the implementation of `stream.Layer`)

We need to hash the uncompressed contents, compress them, hash the compressed contents, then upload them to the registry. The hash of the uncompressed contents is called the `diffid`, which ends up in a config file. The hash of the compressed contents is usually called the `layer digest`, which ends up in the image manifest. The hash of the config file also ends up in the manifest. It's a bit of a juggling act.

How I imagine this being fixed is that we cache the uncompressed contents as we read them, and we cache the compressed contents as we read them. We should also me memoizing the digest and diffid so that we avoid re-compressing or re-hashing anything:

image

E.g., when we call `Digest()` on a layer, it might have to read the layer from disk, gzip it, and sha256sum it, just to compute the digest. We throw away all that work, though! Ideally, that call to `Digest()` should write the gzipped layer to a cache (on disk or in memory) and memoize the sha256 value for both its uncompressed and compressed contents. Subsequent calls for `Digest()` and `DiffID()` would then just return the memoized values. Subsequent calls of `Compressed()` should read the compressed contents from the cache instead of re-zipping everything.

> How would you feel about a separate PR with an in-memory cache that could be pre-populated/pre-compressed so compression and digestion become constant-time ops?

I'm somewhat surprised that you think your disk is the bottleneck here -- I would suspect it's actually gzip. If you do have a slow disk, though, being able to supply an in-memory cache would help a lot with that!

So here's what I'm thinking:
1. An in-memory cache implementation sounds great, I'd love that in a separate PR.
2. Let's figure out *why* we're re-compressing things even with a cache. I think this is because the tarball implementation is too eager. We shouldn't be calculating the digest until we need it.

For 2, I think we probably want to switch over to using `stream.Layer` for the tarball image, but it might just be that we need to make some operations lazy that are currently eager. I haven't looked at it recently, but I have a hunch that we could make it more amenable to composition with `cache.Image` than it is today.

I think some of your issues could have been solved by better documentation, which is mostly my fault 😄 if there's anything that could be clarified, doc issues and doc PRs are also super helpful.

Thanks again, @choo-stripe

jmillikin-stripe:
(quick drive-by comment; I haven't deeply reviewed the code yet)

> Sometimes the best thing to do is digest the content locally before uploading, so you can do an existence check first. E.g. if your internet is slow, it's better trade CPU time just in case the registry already has the blob. In most cases, though, it's better to just upload it.

In our case both compression and uploading are slow, and pre-existing layers are common. We'd ideally like to compress the layers and compute all relevant checksums at build time, so that the push process is a stream of `HEAD`s, then fetches of missing layers from our datastore, then pushing those (already compressed) layers to the image registry.

> I'm somewhat surprised that you think your disk is the bottleneck here -- I would suspect it's actually gzip. If you do have a slow disk, though, being able to supply an in-memory cache would help a lot with that!

It depends on the image, but in general we're nearly maxed out on both disk IO and CPU time. For gzip we use a parallel implementation (https://zlib.net/pigz/), which can saturate a 48-thread CPU.

> We need to hash the uncompressed contents, compress them, hash the compressed contents, then upload them to the registry.

In previous discussions with the maintainer of the Python containerregistry code, we brought up the idea of a more optimized image format -- one that stored compressed layers directly, and stored all of the relevant checksums. Do you think it would be practical to implement support for such a format in this new Go implementation?

jonjohnsonjr:
> In previous discussions with the maintainer of the Python containerregistry code, we brought up the idea of a more optimized image format -- one that stored compressed layers directly, and stored all of the relevant checksums. Do you think it would be practical to implement support for such a format in this new Go implementation?

Two things worth mentioning:
1. When _we_ write out tarballs, they are actually in the compressed form. We have the digests and the config, already, so we can derive the diffids from the config without any additional work. (I'm assuming we don't do anything stupid like re-hash things, but I have not confirmed this). When _docker_ writes out tarballs, it uses an old format that has uncompressed tar files and doesn't record the digests. Some more info in https://github.com/google/go-containerregistry/issues/526
2. IMO, the right answer to this is using an [OCI image layout](https://github.com/opencontainers/image-spec/blob/master/image-layout.md). This stores things (manifest, blobs, config) on disk in the same format as it's stored in the registry, so there's not any conversion that has to happen. We have [an implementation of this](https://godoc.org/github.com/google/go-containerregistry/pkg/v1/layout). It has all the same benefits as the ["faaaaast"](https://github.com/google/containerregistry/blob/da03b395ccdc4e149e34fbb540483efce962dc64/tools/fast_puller_.py#L43) format in containerregistry (I think this is what you're referring to?), with the added bonus of being an actual standard thing instead of something @mattmoor pulled out of his... head. Docker doesn't support this natively, but there's an [open issue](https://github.com/moby/moby/issues/25779) if you want to stir up trouble.

I'm not sure how y'all are consuming this code internally. If you're using `crane`, we could trivially add support for OCI layout in pull/push to crane. Let me know if that would help. If you have your own tools that rely on this library, please try it out!

mattmoor:
> with the added bonus of being an actual standard thing instead of something @mattmoor pulled out of his... head

I absolutely did not pull it out of my *head*  🙄

jonjohnsonjr:
I had a long conversation with @ImJasonH about how we could fix this "for real" and it's a bit involved. I'll write something up when I get a chance (possibly this week but probably next week).

PR #518, opened 2019-09-11
Implement remote.Tag

jonjohnsonjr:
Fixes https://github.com/google/go-containerregistry/issues/349


pkg/v1/remote/index_test.go
```diff
@@ -52,7 +52,7 @@ func mustChild(t *testing.T, idx v1.ImageIndex, h v1.Hash) v1.Image {
 	return img
 }
 
-func mustMediaType(t *testing.T, man manifest) types.MediaType {
+func mustMediaType(t *testing.T, man Taggable) types.MediaType {
```
ImJasonH: Bit weird that it's named `man` now...

pkg/v1/remote/write.go
```diff
@@ -366,7 +367,7 @@ func (w *writer) uploadOne(l v1.Layer) error {
 }
 
 // commitImage does a PUT of the image's manifest.
-func (w *writer) commitImage(man manifest) error {
+func (w *writer) commitImage(man Taggable) error {
```
ImJasonH: And here

jonjohnsonjr:
I thought about this a bit more, and I might add something like this:

```go
func WithTag(tags ...string) Option {
	return func(o *options) error {
		o.tags = append(o.tags, tags...)
		return nil
	}
}
```

This would allow you to specify multiple tags for `Write`, `WriteIndex`, and `Tag`. If we had this, then the main benefit or `remote.Tag` would be to skip the existence checks.

One drawback is that the option isn't generic. It doesn't make sense for `List`, `Delete`, `WriteLayer`, `Get`, `Image`, or `ImageIndex`. Is that okay? Should we have multiple sets of options? Should we fail if you pass a nonsensical option or just ignore it?

Issue #474, opened 2019-06-24
Add index mutation methods to pkg/v1/mutate


jonjohnsonjr:
It's currently really difficult to programmatically create an image index.

This is mostly my fault because I had tunnel vision when implementing pkg/v1/layout, but we need [these](https://github.com/google/go-containerregistry/blob/ca8b66cb1b793f7fb0e3c191ad9a37eb3dc80ea0/pkg/v1/layout/write.go#L34-L128) to work on an abritrary ImageIndex.

Ideally, we could do something like:

```go
idx := empty.Index
idx, _ = mutate.AppendImage(idx, foo)
idx, _ = mutate.AppendImage(idx, bar)
```

Which would produce an index that references both images foo and bar. I'd like the same [options](https://github.com/google/go-containerregistry/blob/ca8b66cb1b793f7fb0e3c191ad9a37eb3dc80ea0/pkg/v1/layout/options.go#L11-L42) to work, and we'll probably want `AppendIndex` as well.

I suspect we'd be able to refactor most of the `layout` package implementations to use the `mutate` package without making it inefficient.

PR #520, opened 2019-09-11
Implement index mutation

jonjohnsonjr:
Add an `AppendManifests` function that adds entries to the `manifests` clause of an index.

Fixes https://github.com/google/go-containerregistry/issues/474

### DIFF
--- a/pkg/v1/validate/index.go
+++ b/pkg/v1/validate/index.go
@@ -21,6 +21,7 @@ import (
 	"strings"
 
 	"github.com/google/go-cmp/cmp"
+	"github.com/google/go-containerregistry/pkg/logs"
 	v1 "github.com/google/go-containerregistry/pkg/v1"
 	"github.com/google/go-containerregistry/pkg/v1/types"
 )
@@ -75,7 +76,7 @@ func validateChildren(idx v1.ImageIndex) error {
 				errs = append(errs, fmt.Sprintf("failed to validate image MediaType[%d](%s): %v", i, desc.Digest, err))
 			}
 		default:
-			return fmt.Errorf("todo: validate index Blob()")
+			logs.Warn.Printf("Unexpected manifest: %s", desc.MediaType)
 		}
 	}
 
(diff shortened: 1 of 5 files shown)


jonjohnsonjr:
An alternative to having two methods is something like:

```go
type Appendable interface {
	RawManifest() ([]byte, error)
	MediaType() (types.MediaType, error)
}

type IndexAddendum struct {
	Add v1.ImageIndex
	v1.Descriptor
}

func AppendManifest(idx v1.ImageIndex, adds ...IndexAddenedum) v1.ImageIndex {
  switch v := i.(type) {
    case v1.ImageIndex:
      // add to indexMap
    case v1.Image:
      // add to imageMap
    case v1.Layer:
      // TODO: how can we do this, actually?
    default:
      // log warning for unexpected addendum
  }
  // add descriptor to manifest
}
```

We could also do type checking against something that just implements `MediaType`? It's hard to come up with something reasonable here if an index can reference arbitrary objects.

I suppose we could also append anything that implements:

```go
type Appendable interface {
	Size() (int64, error)
	MediaType() (types.MediaType, error)
        Digest() (v1.Hash, error) 
}
```

Neither `v1.Image` nor `v1.ImageIndex` implement `Size`, but it would be trivial to add. We could also add a method that returns a `v1.Descriptor`, which would allow objects to maintain annotations if they needed. This would be useful for e.g. a foreign `v1.Layer` that knows what `URLs` it should have.

pkg/v1/validate/index.go
```diff
@@ -75,7 +76,7 @@ func validateChildren(idx v1.ImageIndex) error {
 				errs = append(errs, fmt.Sprintf("failed to validate image MediaType[%d](%s): %v", i, desc.Digest, err))
 			}
 		default:
-			return fmt.Errorf("todo: validate index Blob()")
+			logs.Warn.Printf("Unexpected manifest: %s", desc.MediaType)
```
ImJasonH: Do we want to return here?

jonjohnsonjr: We need to add a `Blob` method to `v1.ImageIndex` if I want to fix that. I can send a PR for that as well, but as it is this will break our tests if I return an error. I'm pretty sure I *do* want to add `Blob` to that interface, but I'd like someone to tell me it's a good idea first :)

Issue #488, opened 2019-07-25
registry: add disk-backed storage


ammario:
The current implementation is a big memory hog for my tests.

I'm thinking we create `NewMemory()` and `NewDisk(root string)` functions, and deprecate the current `New()` function.

If this idea is OK with you @jonjohnsonjr, I can implement it.

jonjohnsonjr:
I've wanted to take advantage of the handler code (and registry) but have swappable implementations of the actual storage. There's more than just disk vs memory, e.g. I'd like to be able to use this to run a proxy registry.

We could start by defining an actual interface for [the content](https://github.com/google/go-containerregistry/blob/273af77a08b28b49cc2cff2dd8ae50a5094dac74/pkg/registry/registry.go#L66-L72) (instead of just maps) and use that interface from the handlers.

Something like:

```go
type blobStore interface {
  Get(key string) (io.ReadSeeker, error)
  Put(key string, blob io.ReadCloser) error
  Delete(key string) error
}

// Wrapper for current implementation.
type memBlobStore struct {
  contents map[string][]byte
}

func (bs *memBlobStore) Get(key string) (io.ReadSeeker, error) {
  if b, ok := bs.contents[key]; ok {
    return bytes.NewReader(b), nil
  }
  return nil, errNotFound
}

func (bs *memBlobStore) Put(key string, blob io.ReadCloser) error {
  b, err := ioutil.ReadAll(blob)
  if err != nil {
    return err
  }
  bs.contents[key] = b
  return nil
}

func (bs *memBlobStore) Delete(key string) error {
  delete(bs.contents, key)
  return nil
}
```

(I'm not 100% sold on this as a good blobs API, just a strawman.)

If that is easy enough, we can figure out an API for the manifests as well.

I'm not actually that familiar with this code, so I'd defer to @clrprod's thoughts.

clrprod:
sgtm - as long as you're adding a second implementation (i.e. postgres or another registry). I generally don't like unused abstraction.

clrprod:
Ah didn't see the original comment, yeah a disk backed version is totally doable. Maybe default to ioutil.TempDir(), but allow for an alternate path?

clrprod:
I'd probably do something like
```
struct option {
 disk string
}
type Option func(o *option)

func DIr(path string) Option {
  return func(o *option) {
     o.disk = path
}
}

func TempDir() Option {
   d, err := ioutil.TempDir
   return Dir(d)
}

New(o ...Option)
[...]

PR #521, opened 2019-09-11
registry: Allow configuring the logger

nhooyr:


jonjohnsonjr:
cc @clrprod

clrprod:
Sorry for the delay taking a look now.

pkg/registry/registry.go
```diff
@@ -54,18 +55,44 @@ func (v *v) v2(resp http.ResponseWriter, req *http.Request) *regError {
...
+	return NewWithOptions(nil)
+}
+
+// Options describes the available options
+// for creating the registry.
+type Options struct {
+	// Log is used to log requests.
+	// If nil, the global logger is used.
+	Log *log.Logger
+}
```
clrprod: Looks awesome - I think this is definitely a pretty reasonable thing to change and thank you for sending in the change.

They style in go is generally to define a option type - i.e.

type Option func(o *registry)

And then have New take variadic arguments - i.e.
New(o Option...)

and in the body do something like

opts := &registry{log:defaultLog}
for _, o := range options {
 o(opts)
}

And define the logger option like
func WithLogger(l log.Type) Option

so you can call New() to get default behaviour, New(WithLogger(log.SomethingComplicated)) to get more interesting behaviour.

Let me know if that makes sense - I'm happy to have the log be configurable, just want to make sure the option pattern is nice for the user. I'm also happy to make these changes if I'm asking for too much work from you.

nhooyr: > They style in go is generally to define a option type - i.e.

Its another pattern for sure but I wouldn't say its generally better. The stdlib for example doesn't use it at all and even though later packages did, the new go-cloud package made by the Go team explicitly decided against using them in favour of optional structs.

See https://github.com/google/go-cloud/issues/908#issuecomment-445034143

nhooyr: > I'm also happy to make these changes if I'm asking for too much work from you.

Feel free to take over.

jonjohnsonjr: I think both approaches are reasonable. There are definitely some hard to solve problems with the functional options approach, but it looks so much nicer in some cases 😄 

We should keep in mind any other extension points e.g. https://github.com/google/go-containerregistry/issues/488

clrprod: I pushed a version of this in https://github.com/google/go-containerregistry/pull/537 - let me know what you think :D

nhooyr:
Closed in favour of #537

Issue #527, opened 2019-09-13
Option to not inject defaults in objects defined in the pkg/name package


smukherj1:
For `Tag` and `Reference` created in the pkg/name package using [NewTag](https://github.com/google/go-containerregistry/blob/bb2eda47dcc0c4cb5d554415f786c8fbe45e42fb/pkg/name/tag.go#L74) and [ParseReference](https://github.com/google/go-containerregistry/blob/bb2eda47dcc0c4cb5d554415f786c8fbe45e42fb/pkg/name/ref.go#L40), for image strings that don't specify the registry name or name the repository in a certain way, some default values defined [here](https://github.com/google/go-containerregistry/blob/bb2eda47dcc0c4cb5d554415f786c8fbe45e42fb/pkg/name/registry.go#L26) and [here](https://github.com/google/go-containerregistry/blob/bb2eda47dcc0c4cb5d554415f786c8fbe45e42fb/pkg/name/repository.go#L23) get injected into the generated objects. I would like the ability to specify an option to the Tag & Reference creator functions such that they never insert any defaults. So if an image name like `foo/bar:latest` is specified, I would prefer the Registry entry in the created object to be left blank instead of defaulting it to `docker.io`. The default behavior of these functions can still be inserting these values if necessary.

smukherj1:
This is currently blocking https://github.com/bazelbuild/rules_docker/issues/580

jonjohnsonjr:
A couple options:
1. Add a `name.Option` to *not* add the defaulted values when we call `String()`
2. Expose something like `func Original() string` on `Reference` that returns the input string, so you can avoid throwing away the input.

Either one of these should also apply to references of the form `<repo>:<tag>@<digest>`. These are currently transformed into just `<repo>@<digest>` (the tag is thrown away), but we should be able to pass through the tag.

smukherj1:
> A couple options:
> 
> 1. Add a `name.Option` to _not_ add the defaulted values when we call `String()`
> 2. Expose something like `func Original() string` on `Reference` that returns the input string, so you can avoid throwing away the input.
> 
> Either one of these should also apply to references of the form `<repo>:<tag>@<digest>`. These are currently transformed into just `<repo>@<digest>` (the tag is thrown away), but we should be able to pass through the tag.

If we decide to go with 2, the v1.tarball package will need an option to use the `Original` function instead of the `String` function when generating the image manifest for the image tarball.

jonjohnsonjr:
[...]

PR #532, opened 2019-09-17
Preserve input strings for name.Reference

jonjohnsonjr:
This modifies String() to return the original input and Name() returns
the defaulted string.

Fixes https://github.com/google/go-containerregistry/issues/527

### DIFF
--- a/pkg/name/digest.go
+++ b/pkg/name/digest.go
@@ -55,6 +55,7 @@ func (d Digest) Name() string {
 	return d.Repository.Name() + digestDelim + d.DigestStr()
 }
 
+// String returns the original input string.
 func (d Digest) String() string {
 	return d.original
 }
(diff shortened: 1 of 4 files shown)


pkg/name/digest.go
```diff
@@ -55,7 +56,7 @@ func (d Digest) Name() string {
 }
 
 func (d Digest) String() string {
```
ImJasonH: Needs a doc comment

pkg/name/ref.go
```diff
@@ -45,5 +45,6 @@ func ParseReference(s string, opts ...Option) (Reference, error) {
 		return d, nil
 	}
 	// TODO: Combine above errors into something more useful?
```
ImJasonH: Is this TODO still valid?

pkg/name/tag.go
```diff
@@ -58,7 +59,7 @@ func (t Tag) Name() string {
 }
 
 func (t Tag) String() string {
```
ImJasonH: Doc comment

PR #535, opened 2019-09-18
Implement ability to generate v1 image tarballs

smukherj1:
Fixes #526 
Add functions that can generate docker image tarballs in the v1 format. This is what `docker save` generates by default in version `18.09.3` today.

PR #536, opened 2019-09-18
Implement v1 image tarball generation in new legacy package

### LINKED ISSUE #526: Add repositories files to image tarballs
Create a `repositories` file for image tarballs generated by the `v1.tarball` package. Image tarballs generated by `docker save` includes this file and our internal repo requires this because this file is part of the v1 & v2 schema.

For the image `l.gcr.io/google/bazel@sha256:97bfeed0303cae14af7e8f66aad6c13f00b2b33081c59d0f4258717b8b94efec`, the `repositories` files looks like:
```
{"l.gcr.io/google/bazel":{"latest":"626b494fdfba1950ebdf1ad5cc2e799879ec78ab8bb1bae11de6c9491fbab6cf"}}
```

This basically appears to be a map from the image name to the tag to the digest of the top most layer. This is currently blocking https://github.com/bazelbuild/rules_docker/issues/580

smukherj1:
Fixes #526
Add functions that can generate docker image tarballs in the v1 format. This is what docker save generates by default in version 18.09.3 today.

This is the same as #535 but adds the code to a new `legacy` package instead of the `v1` package.

### DIFF
--- a/pkg/legacy/config.go
+++ b/pkg/legacy/config.go
@@ -1,0 +1,40 @@
+// Copyright 2018 Google LLC All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package legacy
+
+import (
+	v1 "github.com/google/go-containerregistry/pkg/v1"
+)
+
+// LayerConfigFile is the configuration file that holds the metadata describing
+// a v1 layer. See:
+// https://github.com/moby/moby/blob/master/image/spec/v1.md
+type LayerConfigFile struct {
+	Architecture    string       `json:"architecture"`
+	Author          string       `json:"author,omitempty"`
+	Container       string       `json:"container,omitempty"`
+	Created         v1.Time      `json:"created,omitempty"`
+	Comment         string       `json:"comment,omitempty"`
+	DockerVersion   string       `json:"docker_version,omitempty"`
+	History         []v1.History `json:"history,omitempty"`
+	OS              string       `json:"os"`
+	RootFS          v1.RootFS    `json:"rootfs"`
+	Config          v1.Config    `json:"config"`
+	ContainerConfig v1.Config    `json:"container_config,omitempty"`
+	OSVersion       string       `json:"osversion,omitempty"`
+	ID              string       `json:"id,omitempty"`
+	Parent          string       `json:"parent,omitempty"`
+	Throwaway       bool         `json:"throwaway,omitempty"`
+}
--- a/pkg/legacy/config.go
+++ b/pkg/legacy/config.go
@@ -22,19 +22,9 @@ import (
 // a v1 layer. See:
 // https://github.com/moby/moby/blob/master/image/spec/v1.md
 type LayerConfigFile struct {
-	Architecture    string       `json:"architecture"`
-	Author          string       `json:"author,omitempty"`
-	Container       string       `json:"container,omitempty"`
-	Created         v1.Time      `json:"created,omitempty"`
-	Comment         string       `json:"comment,omitempty"`
-	DockerVersion   string       `json:"docker_version,omitempty"`
-	History         []v1.History `json:"history,omitempty"`
-	OS              string       `json:"os"`
-	RootFS          v1.RootFS    `json:"rootfs"`
-	Config          v1.Config    `json:"config"`
-	ContainerConfig v1.Config    `json:"container_config,omitempty"`
-	OSVersion       string       `json:"osversion,omitempty"`
-	ID              string       `json:"id,omitempty"`
-	Parent          string       `json:"parent,omitempty"`
-	Throwaway       bool         `json:"throwaway,omitempty"`
+	v1.ConfigFile
+	ID        string `json:"id,omitempty"`
+	Parent    string `json:"parent,omitempty"`
+	Throwaway bool   `json:"throwaway,omitempty"`
+	Comment   string `json:"comment,omitempty"`
 }
--- a/pkg/legacy/config.go
+++ b/pkg/legacy/config.go
@@ -1,4 +1,4 @@
-// Copyright 2018 Google LLC All Rights Reserved.
+// Copyright 2019 Google LLC All Rights Reserved.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
--- a/pkg/legacy/tarball/write.go
+++ b/pkg/legacy/tarball/write.go
@@ -1,0 +1,360 @@
+package tarball
+
+import (
+	"archive/tar"
+	"bytes"
+	"crypto/sha256"
+	"encoding/hex"
+	"encoding/json"
+	"fmt"
+	"io"
+	"io/ioutil"
+	"os"
+
+	"github.com/google/go-containerregistry/pkg/legacy"
+	"github.com/google/go-containerregistry/pkg/name"
+	v1 "github.com/google/go-containerregistry/pkg/v1"
+	"github.com/google/go-containerregistry/pkg/v1/partial"
+	"github.com/pkg/errors"
+)
+
+// singleImageTarDescriptor is the struct used to represent a single image inside a `docker save` tarball.
+type singleImageTarDescriptor struct {
+	Config   string
+	RepoTags []string
+	Layers   []string
+
+	// Tracks foreign layer info. Key is DiffID.
+	LayerSources map[v1.Hash]v1.Descriptor `json:",omitempty"`
+}
+
+// tarDescriptor is the struct used inside the `manifest.json` file of a `docker save` tarball.
+type tarDescriptor []singleImageTarDescriptor
+
+// repositoriesTarDescriptor represents the repositories file inside a `docker save` tarball.
+type repositoriesTarDescriptor map[string]map[string]string
+
+// WriteToFile writes in the compressed format to a tarball, on disk.
+// This is just syntactic sugar wrapping tarball.Write with a new file.
+func WriteToFile(p string, ref name.Reference, img v1.Image) error {
+	w, err := os.Create(p)
+	if err != nil {
+		return err
+	}
+	defer w.Close()
+
+	return Write(ref, img, w)
+}
+
+// MultiWriteToFile writes in the V1 image tarball format to a tarball, on disk.
+// This is just syntactic sugar wrapping tarball.MultiWrite with a new file.
+func MultiWriteToFile(p string, tagToImage map[name.Tag]v1.Image) error {
+	refToImage := make(map[name.Reference]v1.Image, len(tagToImage))
+	for i, d := range tagToImage {
+		refToImage[i] = d
+	}
+	return MultiRefWriteToFile(p, refToImage)
+}
+
+// MultiRefWriteToFile writes in the V1 image tarball format to a tarball, on disk.
+// This is just syntactic sugar wrapping tarball.MultiRefWrite with a new file.
+func MultiRefWriteToFile(p string, refToImage map[name.Reference]v1.Image) error {
+	w, err := os.Create(p)
+	if err != nil {
+		return err
+	}
+	defer w.Close()
+
+	return MultiRefWrite(refToImage, w)
+}
+
+// Write is a wrapper to write a single image in V1 format and tag to a tarball.
+func Write(ref name.Reference, img v1.Image, w io.Writer) error {
+	return MultiRefWrite(map[name.Reference]v1.Image{ref: img}, w)
+}
+
+// MultiWrite writes the contents of each image to the provided reader, in the V1 image tarball format.
+// The contents are written in the following format:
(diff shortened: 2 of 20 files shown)


pkg/legacy/config.go
```diff
@@ -0,0 +1,40 @@
...
+package legacy
+
+import (
+	v1 "github.com/google/go-containerregistry/pkg/v1"
+)
+
+// LayerConfigFile is the configuration file that holds the metadata describing
+// a v1 layer. See:
+// https://github.com/moby/moby/blob/master/image/spec/v1.md
+type LayerConfigFile struct {
```
jonjohnsonjr: Can you just [embed](https://golang.org/doc/effective_go.html#embedding) the v1.ConfigFile for most of this?

pkg/v1/config.go
```diff
@@ -35,6 +35,7 @@ type ConfigFile struct {
 	Config          Config    `json:"config"`
 	ContainerConfig Config    `json:"container_config,omitempty"`
 	OSVersion       string    `json:"osversion,omitempty"`
+	Throwaway       bool      `json:"throwaway,omitempty"`
```
jonjohnsonjr: Do we need this?

pkg/legacy/tarball/write.go
```diff
@@ -0,0 +1,313 @@
...
+	"github.com/google/go-containerregistry/pkg/v1/tarball"
+	"github.com/pkg/errors"
+)
+
+// repositoriesTarDescriptor represents the repositories file inside a `docker save` tarball.
+type repositoriesTarDescriptor map[string]map[string]string
+
+// WriteToFile writes in the compressed format to a tarball, on disk.
+// This is just syntactic sugar wrapping tarball.Write with a new file.
+func WriteToFile(p string, ref name.Reference, img v1.Image) error {
```
jonjohnsonjr: This might be a good time to drop some of these permutations. I'd like to see _just_ `Write` and `MultiWrite` if we can (replace the tag map with a ref map).

pkg/v1/tarball/write.go
```diff
@@ -173,10 +173,12 @@ func MultiRefWrite(refToImage map[name.Reference]v1.Image, w io.Writer) error {
...
 		return err
 	}
-	return writeTarEntry(tf, "manifest.json", bytes.NewReader(tdBytes), int64(len(tdBytes)))
+	return WriteTarEntry(tf, "manifest.json", bytes.NewReader(tdBytes), int64(len(tdBytes)))
 }
 
-func dedupRefToImage(refToImage map[name.Reference]v1.Image) map[v1.Image][]string {
+// DedupRefToImage generates an v1.Image to list of tags from the given map from
+// fully qualified image name to the image itself.
+func DedupRefToImage(refToImage map[name.Reference]v1.Image) map[v1.Image][]string {
```
jonjohnsonjr: I don't know if this is worth exposing.

pkg/v1/tarball/write.go
```diff
@@ -196,8 +198,8 @@ func dedupRefToImage(refToImage map[name.Reference]v1.Image) map[v1.Image][]stri
 	return imageToTags
 }
 
-// write a file to the provided writer with a corresponding tar header
-func writeTarEntry(tf *tar.Writer, path string, r io.Reader, size int64) error {
+// WriteTarEntry writes a file to the provided writer with a corresponding tar header
+func WriteTarEntry(tf *tar.Writer, path string, r io.Reader, size int64) error {
```
jonjohnsonjr: I wouldn't expose this either.

pkg/legacy/config.go
smukherj1: Done

pkg/v1/config.go
smukherj1: Removed

pkg/v1/tarball/write.go
smukherj1: Reverted

pkg/v1/tarball/write.go
smukherj1: Reverted

pkg/legacy/tarball/write.go
smukherj1: Done

pkg/legacy/tarball/write.go
```diff
@@ -299,15 +249,50 @@ func MultiRefWrite(refToImage map[name.Reference]v1.Image, w io.Writer) error {
...
 		return err
 	}
-	if err := tarball.WriteTarEntry(tf, "repositories", bytes.NewReader(reposBytes), int64(len(reposBytes))); err != nil {
+	if err := writeTarEntry(tf, "repositories", bytes.NewReader(reposBytes), int64(len(reposBytes))); err != nil {
 		return err
 	}
 	return nil
 }
+
+func dedupRefToImage(refToImage map[name.Reference]v1.Image) map[v1.Image][]string {
```
jonjohnsonjr: You should be able to simplify this a lot now (no need to typecheck `name.Tag`).

pkg/legacy/config.go
```diff
@@ -22,19 +22,9 @@ import (
 // a v1 layer. See:
 // https://github.com/moby/moby/blob/master/image/spec/v1.md
 type LayerConfigFile struct {
```
jonjohnsonjr: Maybe just `ConfigFile` instead of `LayerConfigFile`? Not sure on this.

smukherj1: This struct is specifically used to only generated layer metadata which is why I kept the `Layer` in the name. The legacy package still uses `v1.ConfigFile` for image configs.

pkg/legacy/tarball/write.go
smukherj1: Why would that be the case? Can't caller still pass in references to `name.Digest` to `MultiWrite` or `Write` which will need to be handled here?

pkg/v1/tarball/image.go
```diff
@@ -328,6 +348,11 @@ type compressedLayerFromTarball struct {
 	filePath string
 }
 
+// Desc implements partial.CompressedLayer
```
jonjohnsonjr: This is unused.

pkg/v1/tarball/image.go
```diff
@@ -248,6 +259,15 @@ func (i *uncompressedImage) LayerByDiffID(h v1.Hash) (partial.UncompressedLayer,
 			if bd, ok := i.imgDescriptor.LayerSources[h]; ok {
 				// Overwrite the mediaType for foreign layers.
 				mt = bd.MediaType
```
jonjohnsonjr: You can drop `mt` now and move the `types.DockerLayer` down into the `uncompressedLayerFromTarball` struct literal. Make sure you move the comment with it.

pkg/v1/tarball/image.go
```diff
@@ -220,6 +220,17 @@ type uncompressedLayerFromTarball struct {
 	filePath  string
 }
 
+// foreignUncompressedLayer implements partial.UncompressedLayer but returns
+// a custom descriptor.
+type foreignUncompressedLayer struct {
```
jonjohnsonjr: I think this wrapper is actually not necessary anymore. You could add `Descriptor` to `uncompressedLayerFromTarball` and replace [these](https://github.com/google/go-containerregistry/blob/5340dbaba4cf3d6a62218976684c9cbfc1628d32/pkg/v1/partial/uncompressed.go#L45-L46) with a `v1.Descriptor`.

pkg/v1/tarball/image.go
smukherj1: done

pkg/v1/tarball/image.go
smukherj1: Dropped mt. The struct merging has issues as discussed over IM

pkg/v1/tarball/image.go
smukherj1: The struct merging has issues as discussed over IM

pkg/v1/partial/uncompressed.go
```diff
@@ -81,6 +81,25 @@ func (ule *uncompressedLayerExtender) calcSizeHash() {
 	})
 }
 
+func (ule *uncompressedLayerExtender) Descriptor() (v1.Descriptor, error) {
```
jonjohnsonjr: A comment explaining why we have to do this would be good.

pkg/v1/partial/with.go
```diff
@@ -281,3 +281,7 @@ func DiffIDToBlob(wm WithManifestAndConfigFile, h v1.Hash) (v1.Hash, error) {
 type WithDiffID interface {
 	DiffID() (v1.Hash, error)
 }
+
+type withDescriptor interface {
```
jonjohnsonjr: a short comment here for some context would help

pkg/legacy/tarball/write.go
```diff
@@ -0,0 +1,296 @@
...
+			layerFiles[i] = fmt.Sprintf("%s/layer.tar", cur.config.ID)
+			u, err := l.Uncompressed()
+			if err != nil {
+				return err
+			}
+			defer u.Close()
+			// Reads the entire uncompressed blob into memory! Can be avoided
+			// for some layer implementations where the uncompressed blob is
+			// stored on disk and the layer can just stat the file.
+			uncompressedBlob, err := ioutil.ReadAll(u)
```
jonjohnsonjr: This is really unfortunate.

pkg/legacy/tarball/write.go
```diff
@@ -0,0 +1,296 @@
+package tarball
```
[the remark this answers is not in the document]
smukherj1: Done

pkg/legacy/config.go
```diff
@@ -0,0 +1,30 @@
+// Copyright 2018 Google LLC All Rights Reserved.
```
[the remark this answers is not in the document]
smukherj1: Done

pkg/legacy/tarball/write_test.go
```diff
@@ -0,0 +1,359 @@
+package tarball
```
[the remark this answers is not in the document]
smukherj1: done

pkg/v1/tarball/image.go
```diff
@@ -220,6 +220,17 @@ type uncompressedLayerFromTarball struct {
 	filePath  string
 }
 
+// foreignUncompressedLayer implements partial.UncompressedLayer but returns
+// a custom descriptor.
```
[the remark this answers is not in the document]
smukherj1: Done

pkg/name/tag.go
```diff
@@ -72,19 +72,24 @@ func checkTag(name string) error {
...
 
-// NewTag returns a new Tag representing the given name, according to the given strictness.
-func NewTag(name string, opts ...Option) (Tag, error) {
-	opt := makeOptions(opts...)
-	base := name
-	tag := ""
-
+// SplitTag splits the given tagged image name <registry>/<repository>:<tag>
+// into <registry>/<repository> and <tag>.
+func SplitTag(name string) (string, string) {
```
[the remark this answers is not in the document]
smukherj1: Unexposed. It is used in legacy/tarball/write.go where I duplicated the functionality.

pkg/v1/partial/uncompressed.go
```diff
@@ -81,6 +81,25 @@ func (ule *uncompressedLayerExtender) calcSizeHash() {
...
+	if withDesc, ok := ule.UncompressedLayer.(withDescriptor); ok {
+		return withDesc.Descriptor()
+	}
+	ule.calcSizeHash()
+	if ule.hashSizeError != nil {
+		return v1.Descriptor{}, ule.hashSizeError
```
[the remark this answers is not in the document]
smukherj1: done

pkg/v1/partial/uncompressed.go
smukherj1: Done

pkg/v1/partial/with.go
smukherj1: Done

pkg/legacy/tarball/write.go
smukherj1: Agreed. However, the python impl does this too so it's not any worse for us than before ¯\_(ツ)_/¯

pkg/v1/partial/uncompressed.go
```diff
@@ -81,6 +81,30 @@ func (ule *uncompressedLayerExtender) calcSizeHash() {
 	})
 }
 
+// Descriptor returns the layer manifest descriptor for this uncompressed layer.
+// The embedded UncompressedLayer is checked to see if it has a Descriptor
+// function which is returned if it exists. The underlying descriptor provides
+// foreign layer data like URLs. Otherwise, a new descriptor is
+// generated.
+func (ule *uncompressedLayerExtender) Descriptor() (*v1.Descriptor, error) {
```
jonjohnsonjr: coverage thinks this isn't tested?

pkg/v1/tarball/image.go
```diff
@@ -247,7 +259,15 @@ func (i *uncompressedImage) LayerByDiffID(h v1.Hash) (partial.UncompressedLayer,
 			mt := types.DockerLayer
 			if bd, ok := i.imgDescriptor.LayerSources[h]; ok {
 				// Overwrite the mediaType for foreign layers.
-				mt = bd.MediaType
+				return &foreignUncompressedLayer{
```
jonjohnsonjr: coverage thinks this isn't tested too

pkg/v1/partial/uncompressed.go
```diff
@@ -155,6 +179,15 @@ func (i *uncompressedImageExtender) Manifest() (*v1.Manifest, error) {
 
 	m.Layers = make([]v1.Descriptor, len(ls))
 	for i, l := range ls {
+		// Check if the layer implementation provides the descriptor directly.
```
jonjohnsonjr: or this 🤔

PR #537, opened 2019-09-18
registry: Add support for a custom logger

clrprod:


pkg/registry/registry.go
```diff
@@ -74,5 +78,18 @@ func New() http.Handler {
...
+		o(r)
+	}
+	return http.HandlerFunc(r.root)
+}
+
+// Options describes the available options
+// for creating the registry.
+type Option func(r *registry)
+
+func Logger(l *log.Logger) Option {
```
ImJasonH: `WithLogger`?

nhooyr: I'd prefer `Logger` as it's cleaner.

clrprod: Most options don't seem to add a verb at the front, and it makes them shorter. This also only pollutes the registry namespace which has New(), TLS() and now Logger()

nhooyr:
@googlebot I consent.

Issue #540, opened 2019-09-19
Preserve tag when writing to a tagged and digested image reference


glyn:
Currently the tag is discarded. For example:
```
$ crane copy glyn/cranetest:1@sha256:c1224bb80cf7ce69efe55c438bbbd0ac7d73ab0bcb8548dabc402e3b791cbacc glyn/cranetestcopy:1@sha256:c1224bb80cf7ce69efe55c438bbbd0ac7d73ab0bcb8548dabc402e3b791cbacc
$ docker pull glyn/cranetestcopy:1
Error response from daemon: manifest for glyn/cranetestcopy:1 not found: manifest unknown: manifest unknown
```

Since ggcr preserves repo digests, the digest can safely be discarded when both tag and digest are provided in the target image reference.

Without this, at least one downstream consumer is having to add [support for this feature](https://github.com/pivotal/image-relocation/issues/33) and it would be better to make the semantics consistent.

jonjohnsonjr:
So my interpretation [here](https://github.com/google/go-containerregistry/issues/351#issuecomment-459431865) was short-sighted. What is the _intent_ of `foo:tag@digest`? It seems like a comment when reading, but it's actually important when writing now. Great!

|                  | Read           | Write  |
| ------------- |:-------------:| :-----:|
| foo:tag      | tag | tag |
| foo@digest      | digest      |   digest |
| foo:tag@digest | digest      |    tag |

We've already encoded some behavior in these types:
1. `name.Digest` - I care about security.
2. `name.Tag` -  Merkle _what?_

Tags are useful for documenting intent, so this is definitely something we should fix, but I'm wondering if we have to actually expose this to users.

Neither of the proposed PRs really jump out to me as being obviously the right thing to do. E.g. adding an API for `ParseWriteReference` is weird, since the code that uses the reference will know if it's reading or writing. Adding `WriteTarget` is a bit better, but again I'm really reluctant to increase the API surface if we can solve it another way.

This change is somewhat relevant: https://github.com/google/go-containerregistry/pull/532. Since we're preserving the original input, we can re-frame the user's input in a given context:

```go
if tag, err := name.NewTag(ref.String()); err == nil {
  ref = tag
}
```

I can think of two places we'd want to do this:
1.  [remote writes](https://github.com/google/go-containerregistry/blob/53e1ac5a7d185672cfa09b847b4fa32d229899d1/pkg/v1/remote/write.go#L369-L407), and
[...]

PR #541, opened 2019-09-19

glyn:
This is a less integrated solution which essentially fixes the problem for crane and leaves it up to users of the library to opt in.

### DIFF
--- a/pkg/crane/copy.go
+++ b/pkg/crane/copy.go
@@ -32,7 +32,7 @@ func Copy(src, dst string) error {
 		return fmt.Errorf("parsing reference %q: %v", src, err)
 	}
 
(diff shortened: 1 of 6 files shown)


pkg/crane/copy.go
```diff
@@ -32,7 +32,7 @@ func Copy(src, dst string) error {
 		return fmt.Errorf("parsing reference %q: %v", src, err)
 	}
 
-	dstRef, err := name.ParseReference(dst)
+	dstRef, err := name.ParseWriteReference(dst)
```
glyn: Since we discard the digest from the destination reference, we need to validate that the digest matches that of the image and return an error if it does not.

PR #542, opened 2019-09-19

glyn:
This is a more integrated solution which fixes the problem for crane and all users of the library, without giving them the ability to opt out. Compare the less integrated solution https://github.com/google/go-containerregistry/pull/541.

### DIFF
--- a/pkg/name/digest.go
+++ b/pkg/name/digest.go
@@ -28,8 +28,9 @@ const (
 // Digest stores a digest name in a structured form.
 type Digest struct {
 	Repository
-	digest   string
-	original string
+	digest      string
+	original    string
+	writeTarget Reference // nil if original has no tag
 }
 
 // Ensure Digest implements Reference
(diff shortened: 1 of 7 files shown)


pkg/name/digest.go
```diff
@@ -79,18 +88,23 @@ func NewDigest(name string, opts ...Option) (Digest, error) {
 		return Digest{}, err
 	}
 
+	var writeTarget Reference = nil
 	tag, err := NewTag(base, opts...)
 	if err == nil {
 		base = tag.Repository.Name()
+		if tag.explicitTag() {
+			writeTarget = tag
```
glyn: Since we discard the digest from the write target, write operations need to validate that the digest matches that of the image and return an error if it does not.

afbjorklund:
Seems like I ended up re-implementing this, slightly different: #703

glyn:
> @glyn : Seems like I ended up re-implementing this, slightly different: #703

Thanks @afbjorklund! The ball is still in @jonjohnsonjr's court as to whether or not he wants the issue fixed.

afbjorklund:
For minikube, the workaround is to call `docker pull` (again) on the image afterwards.

1) Check with daemon, if name:tag@digest exists - if so, exit early (avoid re-import)
2) Download with tag and digest
3) Import with tag (digest: none)
4) Pull with tag (will set digest)

Trying to avoid the extra step with the forking of docker/podman, and use the library...

Currently step 1 is failing to match, since the digest (or tag) is removed during the import.

See https://github.com/kubernetes/minikube/pull/7242

PR #544, opened 2019-09-20
Convert build system to go modules.

clrprod:


clrprod:
The build failure is a preexisting lint failure

jonjohnsonjr:
Should [this](https://github.com/google/go-containerregistry/blob/38d665cadff75b725c843e04ae79c7354e430a69/hack/update-deps.sh#L26) do `go mod vendor` now?

clrprod:
I'd prefer to not retain vendor support, but if you need it we can do it. either way I should update that script

clrprod:
Ended up just adding go1.12 back and adding vendor dir. Should make migrations easier for people.

jonjohnsonjr:
> I'd prefer to not retain vendor support, but if you need it we can do it. either way I should update that script

I think if we don't vendor, it will break downstream (e.g. knative/serving) until they move to go modules.

clrprod:
Well it's now added back, so the discussion is academic. 1.12 (no modules) and 1.13 (modules) are both working.

PR #545, opened 2019-09-20
tarball: Fix lint failure

clrprod:
Without this newline, lint + go think the license is a package comment


clrprod:
nvm - we should just merge https://github.com/google/go-containerregistry/pull/543

PR #548, opened 2019-09-20
adds support for the /v2/_catalog API

jmakinen-ncc:


jmakinen-ncc:
CLA-wise I just use a different email for github than what I used to commit that

pkg/v1/remote/catalog.go
```diff
@@ -0,0 +1,69 @@
...
+	"github.com/google/go-containerregistry/pkg/name"
+	"github.com/google/go-containerregistry/pkg/v1/remote/transport"
+)
+
+type catalog struct {
+	Repos []string `json:"repositories"`
+}
+
+// GetCatalog calls /_catalog, returning the list of repositories on the registry
+func GetCatalog(target name.Registry, options ...Option) ([]string, error) {
```
jonjohnsonjr: Given that the results of catalog can be paginated, there might be a better return type than `[]string` -- I'm not sure what the most go idiomatic thing to do here is...

We could return a channel? Or do callbacks for each "page"?

Or maybe this is just fine...

jmakinen-ncc:
How about now?

pkg/v1/remote/catalog.go
jmakinen-ncc: what about now? (sorry to double comment)

pkg/v1/remote/catalog.go
```diff
@@ -0,0 +1,75 @@
...
+	}
+	scopes := []string{target.Scope(transport.PullScope)}
+	tr, err := transport.New(target, o.auth, o.transport, scopes)
+	if err != nil {
+		return nil, err
+	}
+
+	var query string
+
+	query = fmt.Sprintf("last=%s&n=%d", url.QueryEscape(last), n)
```
ImJasonH: `query := fmt.Sprintf(...)`

pkg/v1/remote/catalog.go
```diff
@@ -0,0 +1,75 @@
...
+	"github.com/google/go-containerregistry/pkg/v1/remote/transport"
+)
+
+type catalog struct {
+	Repos []string `json:"repositories"`
+}
+
+// GetCatalog calls /_catalog, returning the list of repositories on the registry
+func GetCatalog(target name.Registry, last string, n int, options ...Option) ([]string, error) {
+
```
ImJasonH: nit: unnecessary blank line

pkg/v1/remote/catalog.go
```diff
@@ -0,0 +1,75 @@
...
+	if err := transport.CheckError(resp, http.StatusOK); err != nil {
+		return nil, err
+	}
+
+	parsed := catalog{}
+	if err := json.NewDecoder(resp.Body).Decode(&parsed); err != nil {
+		return nil, err
+	}
+
+	//TKTK:JM iterate through results with "last"
```
ImJasonH: Is this meant to be a TODO?

pkg/v1/remote/catalog.go
```diff
@@ -0,0 +1,75 @@
...
+	return parsed.Repos, nil
+}
+
+//TKTK:JM write tests
```
ImJasonH: Yes, please do. 😄

pkg/v1/remote/catalog.go
```diff
@@ -0,0 +1,75 @@
...
+	if err != nil {
+		return nil, err
+	}
+	defer resp.Body.Close()
+
+	if err := transport.CheckError(resp, http.StatusOK); err != nil {
+		return nil, err
+	}
+
+	parsed := catalog{}
```
ImJasonH: supernit: I typically see this as `var parsed catalog` which has the same effect but is infinitesimally more idiomatic.

pkg/v1/remote/catalog.go
ImJasonH: I think it'd make sense for `GetCatalog` to do pagination itself, and return all the results it could find from however many requests were necessary to find them. That seems simpler than making callers deal with pagination, or pass a channel, or whatever.

jmakinen-ncc:
Thanks for the comments, back atcha

pkg/v1/remote/catalog.go
```diff
@@ -0,0 +1,70 @@
...
+	"github.com/google/go-containerregistry/pkg/name"
+	"github.com/google/go-containerregistry/pkg/v1/remote/transport"
+)
+
+type catalog struct {
+	Repos []string `json:"repositories"`
+}
+
+// GetCatalog calls /_catalog, returning the list of repositories on the registry
+func GetCatalog(target name.Registry, last string, n int, options ...Option) ([]string, error) {
```
ImJasonH: How do you feel about `GetCatalog` just doing pagination itself and returning all the values it can find? How many pages of results do we expect to get from this? Tens of thousands?

jmakinen-ncc: That's why I was hesitant to implement it because I'm not sure of the scale.
I was planning on writing a helper for my own purposes that I can add later, but I just needed this at a minimum.

ImJasonH: The scale might be a concern, but for (most?) users who should expect only a couple pages, it sucks that this makes them handle pagination themselves.

I'd prefer to expose the pagination-handling version for the users that don't want to deal with it, and maybe if we hear that this method is a pain for whatever scaling reason, we can export the DIY method later. If we _never_ hear that we need to expose the guts, that's great.

Does that make any sense?

jonjohnsonjr: If we're going to handle the pagination internally, we should probably make `GetCatalog` take a `context.Context` so that you can cancel it.

> it sucks that this makes them handle pagination themselves

This is basically why the `crane` package exists. IMO we should (within reason) expose all the complexity that the API itself has, then we can expose the happy path in `crane`.

jonjohnsonjr:
@jmakinen-ncc so, let's just punt on figuring out a reasonable API for paginated requests. I'd say just add your original, easy-to-use implementation under `pkg/crane` for now. We can come up with a better way to expose pagination later, and we'll actually want to do that for listing tags as well. Filed https://github.com/google/go-containerregistry/issues/549 to track that.

Sorry for the back and forth!

jmakinen-ncc:
Sorry, so does that mean that this can be merged?

pkg/v1/remote/catalog.go
clrprod: Another pattern I've seen that is nice, is to have the GetCatalog take a callback function that gets called per page. It allows users to abort early (via returning an error / cancelling the context), but it hides the pagination detail from the user.

jonjohnsonjr: @ImJasonH thinks they're not idiomatic go. There's [this presentation](https://drive.google.com/file/d/1nPdvhB0PutEJzdCq5ms6UI58dp50fcAN/view) for inspiration.

I'm not sure if we could borrow something from that, but it also doesn't like callbacks (from [here](https://about.sourcegraph.com/go/gophercon-2018-rethinking-classical-concurrency-patterns)):

> Bryan briefly mentions asynchronous callbacks as something programmers from other certain languages sometimes try to use. But he notes that most Go programmers already know not to use them in Go, and moves on to talking about Futures and Producer–Consumer Queues.

clrprod: This is a synchronous callback - not a async one. The next page wouldn't be fetched until the callback returns.

jmakinen-ncc: Ok, how do we feel about my new `crane.GetCatalog`?

pkg/v1/remote/catalog.go
```diff
@@ -0,0 +1,70 @@
...
+	"github.com/google/go-containerregistry/pkg/name"
+	"github.com/google/go-containerregistry/pkg/v1/remote/transport"
+)
+
+type catalog struct {
+	Repos []string `json:"repositories"`
+}
+
+// GetCatalogPage calls /_catalog, returning the list of repositories on the registry
+func GetCatalogPage(target name.Registry, last string, n int, options ...Option) ([]string, error) {
```
clrprod: I think the idea is to just make this a private method in the crane/catalog.go file. At some point in the future it can get moved to the remote library when we have interface concensus.

jmakinen-ncc:
Would anyone be able to help me figure out why the build is failing here?
do I need to write docs for the added command or something?
Also, I feel like keeping the most generic interface in the remote package and then having the pretty one in crane makes more sense right?
So that if anyone wants to do the dirty work of working with pages directly but still having the benefits of having auth and everything handled then they can. I would think, we'd want remote to be the most 1:1 representation of the API possible
but if someone just wants a full list without having to think, they can call the user-friendly crane

jonjohnsonjr:
> Would anyone be able to help me figure out why the build is failing here?
> do I need to write docs for the added command or something?

It's complaining that you haven't run `./hack/update-code-gen.sh` to generate the docs. We could probably make that more evident.

> Also, I feel like keeping the most generic interface in the remote package and then having the pretty one in crane makes more sense right?
> So that if anyone wants to do the dirty work of working with pages directly but still having the benefits of having auth and everything handled then they can. I would think, we'd want remote to be the most 1:1 representation of the API possible

Yeah, definitely. I'm just not sure if this is the API we want to support forever. It has the same drawback as [this](https://github.com/google/go-containerregistry/blob/ff1ac7f97758299e13f746209d8c0f199923d89a/pkg/v1/remote/write.go#L523-L528), where we can't reuse the auth handshake results across multiple pages. I'm also not sure if we want to return just the `[]string` or the whole struct. There's some other stuff we're not really exposing [here](https://github.com/docker/distribution/blob/master/docs/spec/api.md#catalog-fetch-paginated) (e.g. the `Link` header, which seems like we should respect it as a client, and there's a `next` field in the response there which seems like a bug in the spec actually).

On the other hand, the crane API is pretty straightforward, so I think it's reasonable to merge that now (so you can use it) until we can figure out the best way to represent this. Do you need the ability to list individual pages? Or is the crane API sufficient for now?

cmd/crane/cmd/catalog.go
```diff
@@ -0,0 +1,45 @@
...
+	"github.com/google/go-containerregistry/pkg/crane"
+	"github.com/spf13/cobra"
+)
+
+func init() { Root.AddCommand(NewCmdGetCatalog()) }
+
+// NewCmdGetCatalog creates a new cobra.Command for the repos subcommand.
+func NewCmdGetCatalog() *cobra.Command {
+	return &cobra.Command{
+		Use:   "repos",
```
jonjohnsonjr: I think I'd rather this just be `catalog` than `repos`

jonjohnsonjr:
@jmakinen-ncc any interest in pushing this over the finish line? I feel like it's pretty close to merge-able.

jmakinen-ncc:
Yeah, sorry been looking for the free time in my personal life to close up some loose threads.
Will do asap.


On Oct 18, 2019 2:00 PM, jonjohnsonjr <notifications@github.com> wrote:

@jmakinen-ncc<https://github.com/jmakinen-ncc> any interest in pushing this over the finish line? I feel like it's pretty close to merge-able.

-

jonjohnsonjr:
> sorry

No worries! I was just running through old issues and PRs trying to clean things up, there's no rush :smile:

jmakinen-ncc:
Do we have a winner?

cmd/crane/cmd/catalog.go
```diff
@@ -0,0 +1,45 @@
...
+	"log"
+
+	"github.com/google/go-containerregistry/pkg/crane"
+	"github.com/spf13/cobra"
+)
+
+func init() { Root.AddCommand(NewCmdGetCatalog()) }
+
+// NewCmdGetCatalog creates a new cobra.Command for the repos subcommand.
+func NewCmdGetCatalog() *cobra.Command {
```
jonjohnsonjr: Nit: I'd name this "NewCmdCatalog" to be consistent with the rest of the commands.

pkg/crane/catalog.go
```diff
@@ -0,0 +1,51 @@
...
+
+package crane
+
+import (
+	"github.com/google/go-containerregistry/pkg/authn"
+	"github.com/google/go-containerregistry/pkg/name"
+	"github.com/google/go-containerregistry/pkg/v1/remote"
+)
+
+// GetCatalog returns the repositories in a registry's catalog
```
jonjohnsonjr: supernit: add a period at the end of this sentence

pkg/crane/catalog.go
```diff
@@ -0,0 +1,51 @@
...
+func GetCatalog(src string) (res []string, err error) {
+	reg, err := name.NewRegistry(src)
+	if err != nil {
+		return nil, err
+	}
+
+	n := 100
+	last := ""
+	for {
+
```
jonjohnsonjr: supernit: drop this line

pkg/crane/catalog.go
```diff
@@ -0,0 +1,51 @@
...
+		if len(page) > 0 {
+			last = page[len(page)-1]
+			res = append(res, page...)
+		}
+
+		if len(page) < n {
+			break
+		}
+
```
jonjohnsonjr: supernit: drop this line

pkg/v1/remote/catalog.go
```diff
@@ -0,0 +1,70 @@
...
+
+	"github.com/google/go-containerregistry/pkg/name"
+	"github.com/google/go-containerregistry/pkg/v1/remote/transport"
+)
+
+type catalog struct {
+	Repos []string `json:"repositories"`
+}
+
+// GetCatalogPage calls /_catalog, returning the list of repositories on the registry
```
jonjohnsonjr: supernit: add a period at the end of this sentence

pkg/v1/remote/catalog.go
jonjohnsonjr: I think I'm actually okay with leaving this in. If we ever come up with a good API for this, we can name it `GetCatalog`, and `GetCatalogPage` would still be useful.

I am a little bit unhappy with this because it doesn't allow clients to adhere to [the spec](https://docs.docker.com/registry/spec/api/#pagination):

> Compliant client implementations should always use the Link header value when proceeding through results linearly.

However, this does enable you to skip ahead, which is immediately after:

> The client may construct URLs to skip forward in the catalog.

I think we're basically making it impossible (sometimes, depends on implementation) for a registry to implement this efficiently by handing us a cursor in the Link header, but maybe we don't care.

pkg/v1/remote/catalog_test.go
```diff
@@ -0,0 +1,83 @@
...
+		name:         "success",
+		responseBody: []byte(`{"repositories":["test/test","foo/bar"]}`),
+		wantErr:      false,
+		wantRepos:    []string{"test/test", "foo/bar"},
+	}, {
+		name:         "not json",
+		responseBody: []byte("notjson"),
+		wantErr:      true,
+	}}
+	//TODO: add test cases for pagination
```
jonjohnsonjr: Planning to do this later? :smile: 

supernit: add a space between `//` and `TODO`

jmakinen-ncc: Potentially, but I just wanted to note that they are missing if anyone runs into issues there later and wants to write some.
Do you need this?

jmakinen-ncc:
Thanks everyone!

cmd/crane/cmd/catalog.go
```diff
@@ -0,0 +1,45 @@
+// Copyright 2018 Google LLC All Rights Reserved.
```
jonjohnsonjr: nit: 2019

pkg/crane/catalog.go
```diff
@@ -0,0 +1,49 @@
+// Copyright 2018 Google LLC All Rights Reserved.
```
jonjohnsonjr: nit: 2019

pkg/v1/remote/catalog.go
```diff
@@ -0,0 +1,70 @@
+// Copyright 2018 Google LLC All Rights Reserved.
```
jonjohnsonjr: nit: 2019

pkg/v1/remote/catalog_test.go
```diff
@@ -0,0 +1,83 @@
+// Copyright 2018 Google LLC All Rights Reserved.
```
jonjohnsonjr: nit: 2019

jmakinen-ncc:
> git is my favorite chat protocol

Ask @clrprod about "Big changes, no promises"

PR #550, opened 2019-09-24

glyn:


jonjohnsonjr:
Closing this in favor of https://github.com/google/go-containerregistry/pull/542 -- we can iterate on that PR instead.

PR #551, opened 2019-09-24
transport: Fix ping code to wrap errors in a transport.Error

clrprod:
I added a high level test that the operation returns an error for a
server which only returns one status code. I'll probably add a few more
cases to help find other pathways that are swallowing the error and
contained status code.


pkg/v1/remote/error_roundtrip_test.go
```diff
@@ -0,0 +1,54 @@
+package remote_test
```
jonjohnsonjr: missing copyright header

clrprod: I think we need a presubmit for it - not all files have it.

clrprod: Also added

PR #553, opened 2019-10-01
Implement pkg/internal/compare, use it

jonjohnsonjr:
This drops a lot of testing boilerplate.

This is under pkg/internal for now because I'm not sure if I love the interface.

### DIFF
--- a/pkg/internal/compare/image.go
+++ b/pkg/internal/compare/image.go
@@ -99,7 +99,7 @@ func Images(a, b v1.Image) error {
 	for i := 0; i < len(layerss[0]); i++ {
 		if err := Layers(layerss[0][i], layerss[1][i]); err != nil {
 			// Wrap the error in newlines to delineate layer errors.
-			errs = append(errs, fmt.Sprintf("\n%v\n", err))
+			errs = append(errs, fmt.Sprintf("Layers[%d]: %v\n", i, err))
 		}
 	}
 
--- a/pkg/internal/compare/image.go
+++ b/pkg/internal/compare/image.go
@@ -24,6 +24,8 @@ import (
 	"github.com/google/go-containerregistry/pkg/v1/types"
 )
 
+// Images compares the given images to each other and returns an error if they
+// differ.
 func Images(imgs ...v1.Image) error {
 	if len(imgs) < 2 {
 		return fmt.Errorf("comparing %d images makes no sense", len(imgs))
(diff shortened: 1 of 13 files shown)


pkg/internal/compare/image.go
```diff
@@ -0,0 +1,121 @@
...
+	"reflect"
+	"strings"
+
+	v1 "github.com/google/go-containerregistry/pkg/v1"
+	"github.com/google/go-containerregistry/pkg/v1/types"
+)
+
+// Images compares the given images to each other and returns an error if they
+// differ.
+func Images(imgs ...v1.Image) error {
```
ImJasonH: This seems like it would be a lot simpler to implement if it only accepted two images, and probably a lot easier to understand resulting differences. What's the use case for wanting to compare 3+ images at once?

If you need to compare 3+ images at once, at least add tests to cover the 3+ case.

jonjohnsonjr: I think it's the same amount of code to compare 2 or N, so I think implementing the more flexible thing makes sense.

Updated the tests to cover 3+ -- we can revisit this decision if we ever want to move out of pkg/internal/

ImJasonH: Can you document that errors will be phrased in terms of comparisons of image N vs N-1?

This means that `compare.Images(a, a, b, a)` will tell you `images[2] != images[1]` and `images[3] != images[2]` but won't make it very clear that images 0, 1, 3 were all the same.

The code for this is roughly the same, but the semantics of how you phrase differences can get very complex (to write, and to read) pretty quickly.

jonjohnsonjr: Dropped down to just 2 things.

pkg/internal/compare/image.go
```diff
@@ -0,0 +1,111 @@
...
+	if len(layerss[0]) != len(layerss[1]) {
+		// If we have fewer layers than the first image, abort with an error so we don't panic.
+		return errors.New("len(a.Layers()) != len(b.Layers())")
+	}
+
+	// Compare each layer.
+	for i := 0; i < len(layerss[0]); i++ {
+		if err := Layers(layerss[0][i], layerss[1][i]); err != nil {
+			// Wrap the error in newlines to delineate layer errors.
+			errs = append(errs, fmt.Sprintf("\n%v\n", err))
```
ImJasonH: Should this error message indicate which layer index (`i`) differed?

PR #562, opened 2019-10-08
Run go mod tidy

jonjohnsonjr:
Maybe this will fix travis failures.


jonjohnsonjr:
> Done.

Nevermind! We explicitly test this stuff with go modules disabled, which causes this to fail.

Issue #563, opened 2019-10-09
Insecure option doesn't work if the image is a fat image


ktock:
On fetching an image, if the image is a child of a [fat image](https://docs.docker.com/registry/spec/manifest-v2-2/) the `Insecure` option doesn't work.

For example, assume we have:
- Normal image: `registry2:5000/ubuntu:18.04`
- [Fat image(indexed image)](https://docs.docker.com/registry/spec/manifest-v2-2/): `registry2:5000/ubuntu:fat`

When we run:
```go
src := "registry2:5000/ubuntu:18.04"
srcRef, err := name.ParseReference(src, name.Insecure)
srcImg, err := remote.Image(srcRef)
_, err = srcImg.ConfigFile()
if err != nil {
	fmt.Printf("Failed: Thin manifest: %v\n", err)
} else {
	fmt.Printf("Succeeded: Thin manifest\n")
}

src = "registry2:5000/ubuntu:fat"
srcRef, err = name.ParseReference(src, name.Insecure)
srcImg, err = remote.Image(srcRef)
_, err = srcImg.ConfigFile()
if err != nil {
	fmt.Printf("Failed: Fat manifest:%v\n", err)
} else {
	fmt.Printf("Succeeded: Fat manifest\n")
}
```

This goes:
```
$ go run foo.go
Succeeded: Thin manifest
Failed: Fat manifest:Get https://registry2:5000/v2/ubuntu/blobs/sha256:2ca708c1c9ccc509b070f226d6e4712604e0c48b55d7d8f5adc9be4a4d36029a: http: server gave HTTP response to HTTPS client
```

jonjohnsonjr:
Oh wow that's unfortunate. I'll take a look.

PR #564, opened 2019-10-09
Enable pass options to name package in case of fat image

ktock:
Fixes: #563

Currently, we can't specify options(i.e. insecure options) of name package to
child images referenced by a fat image by index.
This commit solves this issue by enabling to pass options to name packages even
for references of child images of a fat image.

Signed-off-by: Kohei Tokunaga <ktokunaga.mail@gmail.com>


jonjohnsonjr:
Thanks for the PR!

I took a different approach in https://github.com/google/go-containerregistry/pull/567 to avoid passing around name options and is a small cleanup.

ktock:
OK, thank you!

PR #567, opened 2019-10-09
Add name.Repository helpers

### LINKED ISSUE #563: Insecure option doesn't work if the image is a fat image

jonjohnsonjr:
Fixes #563

These are useful for creating a child digest/tag from a repository, which is something we need to do in remote. Before, we were throwing away the insecure option, which this approach fixes.

### DIFF
--- a/pkg/name/repository_test.go
+++ b/pkg/name/repository_test.go
@@ -120,3 +120,18 @@ func TestRepositoryBadDefaulting(t *testing.T) {
 		t.Errorf("IsBadErrName == false: %v", err)
 	}
 }
+
+func TestRepositoryChildren(t *testing.T) {
+	repo, err := NewRepository("ubuntu", Insecure)
+	if err != nil {
+		t.Fatal(err)
+	}
+	tag := repo.Tag("foo")
+	if got, want := tag.Scheme(), "http"; got != want {
+		t.Errorf("tag.Scheme(): got %s want %s", got, want)
+	}
+	digest := repo.Digest("badf00d")
+	if got, want := digest.Scheme(), "http"; got != want {
+		t.Errorf("digest.Scheme(): got %s want %s", got, want)
+	}
+}
--- a/pkg/name/repository_test.go
+++ b/pkg/name/repository_test.go
@@ -122,7 +122,7 @@ func TestRepositoryBadDefaulting(t *testing.T) {
 }
 
 func TestRepositoryChildren(t *testing.T) {
-	repo, err := NewRepository("ubuntu", Insecure)
+	repo, err := NewRepository("example.com/repo", Insecure)
 	if err != nil {
 		t.Fatal(err)
 	}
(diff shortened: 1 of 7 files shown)


jonjohnsonjr:
> + Coverage   72.67%   72.94%   +0.27%     

👌

pkg/name/repository_test.go
```diff
@@ -120,3 +120,18 @@ func TestRepositoryBadDefaulting(t *testing.T) {
...
 	}
 }
+
+func TestRepositoryChildren(t *testing.T) {
+	repo, err := NewRepository("ubuntu", Insecure)
+	if err != nil {
+		t.Fatal(err)
+	}
+	tag := repo.Tag("foo")
+	if got, want := tag.Scheme(), "http"; got != want {
```
ImJasonH: Why not compare the whole `tag.String()`?

pkg/v1/remote/index.go
```diff
@@ -162,16 +162,13 @@ func (r *remoteIndex) childByHash(h v1.Hash) (*Descriptor, error) {
 	return nil, fmt.Errorf("no child with digest %s in index %s", h, r.Ref)
 }
 
-func (r *remoteIndex) childRef(h v1.Hash) (name.Reference, error) {
-	return name.ParseReference(fmt.Sprintf("%s@%s", r.Ref.Context(), h), name.StrictValidation)
+func (r *remoteIndex) childRef(h v1.Hash) name.Reference {
+	return r.Ref.Context().Digest(h.String())
```
ImJasonH: This method might be simple enough to just inline with a comment.

pkg/name/repository_test.go
jonjohnsonjr: I mostly wanted to make sure options were preserved, but added a String() check too.

pkg/v1/remote/index.go
jonjohnsonjr: Done.

PR #568, opened 2019-10-10
Re-read $DOCKER_CONFIG when we load the config file

jonjohnsonjr:
See https://github.com/bazelbuild/rules_docker/issues/1193#issuecomment-540712704

This broke because:
https://github.com/bazelbuild/rules_docker/blob/72fe2bb93e6d59c1f8823421a1b914a3058b90e2/container/go/cmd/pusher/pusher.go#L66

and 

https://github.com/docker/cli/blob/3e07fa728a3025cf4205e3dd071da703c34cf45f/cli/config/config.go#L25

don't work together. We used to parse the config file ourselves and read from the env every time we called into the keychain, but docker does this once on startup.

### DIFF
--- a/pkg/authn/keychain.go
+++ b/pkg/authn/keychain.go
@@ -15,6 +15,8 @@
 package authn
 
 import (
+	"os"
+
 	"github.com/docker/cli/cli/config"
 	"github.com/docker/cli/cli/config/types"
 	"github.com/google/go-containerregistry/pkg/name"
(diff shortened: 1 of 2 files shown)


jonjohnsonjr:
Inspired by https://github.com/docker/cli/blob/3e07fa728a3025cf4205e3dd071da703c34cf45f/cli/config/config.go#L47-L49

PR #569, opened 2019-10-11
Fix panic in legacy tarball writer when config has no history

smukherj1:


pkg/legacy/tarball/write.go
```diff
@@ -214,6 +214,14 @@ func MultiWrite(refToImage map[name.Reference]v1.Image, w io.Writer) error {
 		if err != nil {
 			return err
 		}
+		history := cfg.History
+		if len(history) != 0 && len(layers) != len(history) {
+			return errors.Errorf("image config had layer history which did not match the number of layers, got len(history)=%d, len(layers)=%d, want len(history)=len(layers)", len(history), len(layers))
+		}
```
clrprod: NIt: If you else this off the if statement below, you can just say len(layers) != len(history).

pkg/legacy/tarball/write_test.go
```diff
@@ -320,6 +320,48 @@ func TestWriteForeignLayers(t *testing.T) {
 	}
 }
 
+func TestMultiWriteNoHistory(t *testing.T) {
```
clrprod: Since you added code to handle history and layers >0 but history != layers, is there a test which handles that? Or should a test be added for that?

pkg/legacy/tarball/write_test.go
```diff
@@ -320,6 +320,48 @@ func TestWriteForeignLayers(t *testing.T) {
...
+	cfg.History = nil
+	tag, err := name.NewTag("gcr.io/foo/bar:latest", name.StrictValidation)
+	if err != nil {
+		t.Fatalf("Error creating test tag: %v", err)
+	}
+	o, err := os.Create(fp.Name())
+	if err != nil {
+		t.Fatalf("Error creating %q to write image tarball: %v", fp.Name(), err)
+	}
+	defer o.Close()
```
clrprod: Can't you just use fp directly here? Why do you need to zero the already zero file + make a second file pointing to it?

PR #570, opened 2019-10-14
Respect mediaType of index in layout

glyn:
... rather than assuming all indexes have media type application/vnd.oci.image.index.v1+json.


glyn:
Add a hand-crafted test which should be reworked once the testdata is made reproducible.

PR #573, opened 2019-10-16
Legacy tarball writer should skip empty layer in layer history

smukherj1:


pkg/legacy/tarball/write_test.go
```diff
@@ -357,6 +357,55 @@ func TestMultiWriteNoHistory(t *testing.T) {
...
+	if err != nil {
+		t.Fatalf("Error creating random image: %v", err)
+	}
+	cfg, err := img.ConfigFile()
+	if err != nil {
+		t.Fatalf("Error getting image config: %v", err)
+	}
+	// Build a history for 2 layers that is interspersed with empty layer
+	// history.
+	cfg.History = []v1.History{
```
ImJasonH: I think it's a bit easier to scan and see what's being described if it's indented like:

```
cfg.History = []v1.History{
  {EmptyLayer: true},
  {EmptyLayer: false},
  {EmptyLayer: true},
  {EmptyLayer: false},
  {EmptyLayer: true},
}
```

Alternatively, make vars for `empty` and `nonEmpty` and use those, either is fine.

Even though `EmptyLayer: false` is the default, it's nice to be explicit, since the alternative is an "empty struct" that describes a "history with non-empty layer", and that can be confusing at a glance.

pkg/legacy/tarball/write_test.go
```diff
@@ -357,6 +357,55 @@ func TestMultiWriteNoHistory(t *testing.T) {
 	}
 }
 
+func TestMultiWriteHistoryEmptyLayers(t *testing.T) {
+	// Make a random image with 2 layers.
```
ImJasonH: Does the number of layers in this image have to correspond to the number of non-empty history items below? If so, document that or make them both refer to the same variable so it's documented in code too.

PR #574, opened 2019-10-16
Add compression option to tarball.Layer and stream.Layer

### LINKED ISSUE #572: Layer digests don't match pushed digests using `docker`
It seems like there have been a few issues regarding Digest and compression (#150, #153, #158) yet I've come here with an unexpected issue brought on by the current compression. 

It seems like the current compression used is great in regards of performance but our expectation was that it would produce a layer digest which resembles that of what `docker` would produce when pushing to a registry. That is not the case. [`docker` uses default compression](https://sourcegraph.com/github.com/moby/moby@4866f5139a0b0bc026aa4b0c916ee1071c695134/-/blob/distribution/push.go#L155) making calls to `layer.Digest()` unusable in our case and requires us to compress the layers yet again to get the appropriate and expected layer digest.

This issue appears affects both [`tarball.Layer`](https://github.com/google/go-containerregistry/blob/master/pkg/v1/tarball/layer.go#L124:6) and [`stream.Layer`](https://github.com/google/go-containerregistry/blob/master/pkg/v1/stream/layer.go#L124).

**Proposal:** Allow configuration of compression used at digest computation but default to `gzip.DefaultCompression`.

jromero:
Resolves #572

Signed-off-by: Javier Romero <jromero@pivotal.io>

### DIFF
--- a/pkg/v1/stream/layer.go
+++ b/pkg/v1/stream/layer.go
@@ -39,8 +39,9 @@ var (
 
 // Layer is a streaming implementation of v1.Layer.
 type Layer struct {
-	blob     io.ReadCloser
-	consumed bool
+	blob        io.ReadCloser
+	consumed    bool
+	compression int
 
 	mu             sync.Mutex
 	digest, diffID *v1.Hash
@@ -49,8 +50,29 @@ type Layer struct {
 
 var _ v1.Layer = (*Layer)(nil)
 
+// LayerOption applies options to layer
+type LayerOption func(*Layer)
+
+// WithCompressionLevel sets the gzip compression. See `gzip.NewWriterLevel` for possible values.
+func WithCompressionLevel(level int) LayerOption {
+	return func(l *Layer) {
+		l.compression = level
+	}
+}
+
 // NewLayer creates a Layer from an io.ReadCloser.
-func NewLayer(rc io.ReadCloser) *Layer { return &Layer{blob: rc} }
+func NewLayer(rc io.ReadCloser, opts ...LayerOption) *Layer {
+	layer := &Layer{
+		blob:        rc,
+		compression: gzip.BestSpeed,
+	}
+
+	for _, opt := range opts {
+		opt(layer)
+	}
+
+	return layer
+}
 
 // Digest implements v1.Layer.
 func (l *Layer) Digest() (v1.Hash, error) {
@@ -121,7 +143,7 @@ func newCompressedReader(l *Layer) (*compressedReader, error) {
 	// capture compressed digest, and a countWriter to capture compressed
 	// size.
 	pr, pw := io.Pipe()
-	zw, err := gzip.NewWriterLevel(io.MultiWriter(pw, zh, count), gzip.BestSpeed)
+	zw, err := gzip.NewWriterLevel(io.MultiWriter(pw, zh, count), l.compression)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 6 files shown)


pkg/v1/tarball/layer.go
```diff
@@ -68,16 +71,30 @@ func (l *layer) MediaType() (types.MediaType, error) {
...
 }
 
+// LayerOption applies options to layer
+type LayerOption func(*layer) error
+
+// WithCompressionLevel sets the gzip compression. See `gzip.NewWriterLevel` for possible possible values.
+func WithCompressionLevel(level int) LayerOption {
+	return func(l *layer) error {
+		if level < gzip.HuffmanOnly || level > gzip.BestCompression {
+			return errors.Errorf("invalid compression level: %d", level)
```
jonjohnsonjr: I've been meaning to drop this package as a dependency (it's only used in a couple places) -- can this just be `fmt.Errorf`?

pkg/v1/tarball/layer.go
```diff
@@ -68,16 +71,30 @@ func (l *layer) MediaType() (types.MediaType, error) {
 	return types.DockerLayer, nil
 }
 
+// LayerOption applies options to layer
+type LayerOption func(*layer) error
+
+// WithCompressionLevel sets the gzip compression. See `gzip.NewWriterLevel` for possible possible values.
+func WithCompressionLevel(level int) LayerOption {
+	return func(l *layer) error {
+		if level < gzip.HuffmanOnly || level > gzip.BestCompression {
```
jonjohnsonjr: This seems premature, I'd probably just let it fail later.

jonjohnsonjr: (In that the gzip package should be responsible for this validation, and not us)

jromero: I'm okay with that. 

I'm curious what you would think about the `LayerOption` function signature in that case? Since I wouldn't be using `error` return type anymore should it be removed or should it be kept as some future proofing to prevent breaking the API?

jonjohnsonjr: I'm pretty sure the interface doesn't matter, since it's impossible for anyone but us to implement it (I think), so we won't break anybody that depends on us (we can only break ourselves), so I'd remove the error. We can add it back later if we need.

jromero:
Alright, think I've got everything squared away. I've implemented the change to `stream.Layer` as well to truly resolve the issue mentioned.

I did notice (and followed) the slightly different testing convention so the tests will look different between them.

pkg/v1/tarball/layer.go
```diff
@@ -68,16 +69,26 @@ func (l *layer) MediaType() (types.MediaType, error) {
 	return types.DockerLayer, nil
 }
 
+// LayerOption applies options to layer
+type LayerOption func(*layer)
+
+// WithCompressionLevel sets the gzip compression. See `gzip.NewWriterLevel` for possible possible values.
```
jonjohnsonjr: stutter here on "possible"

pkg/v1/stream/layer.go
```diff
@@ -49,8 +50,29 @@ type Layer struct {
 
 var _ v1.Layer = (*Layer)(nil)
 
+// LayerOption applies options to layer
+type LayerOption func(*Layer)
+
+// WithCompressionLevel sets the gzip compression. See `gzip.NewWriterLevel` for possible possible values.
```
jonjohnsonjr: stutter here on "possible"

jromero:
@jonjohnsonjr I can totally see the need and use case of different compression algorithms but that goes a bit beyond our scope at the moment. I'd be happy to take a swing at it via a different PR.

jonjohnsonjr:
Thanks!

PR #578, opened 2019-10-17
Support compressed image tarballs

siddharthab:
It saves the users an additional step of uncompressing the tarball themselves, or providing a wrapper to os.Open.

See https://github.com/bazelbuild/rules_docker/pull/1223 for an alternative to this change in one use case.


jonjohnsonjr:
I'm not sure about this. Aren't the layers generally compressed within the tarball already? What's the advantage of further compressing the tarball?

From a performance perspective, it would be nice if we could avoid calling the opener twice. Also, we're already doing a bunch of seeking within the tarball, and I'm really worried that it's too easy for this to become accidentally quadratic (I've seen this before with the python library -- a gzipped tarball of gzipped tarballs took hours to process because each file access had to uncompress the outer gzip, seek through the tarball, uncompress the layer, seek through that tarball, the finally read the contents).

siddharthab:
The layers may or may not be compressed.

Your concerns are valid though. I did not go for it in the python library because of performance reasons, but in the Go library, I thought the performance hit would be acceptable if the users really want the convenience.

Happy to take this PR back if you are still unsure about it.

jonjohnsonjr:
> I thought the performance hit would be acceptable if the users really want the convenience.
>
> Happy to take this PR back if you are still unsure about it.

If it's not too hard to work around it, I'd like to avoid maintaining this. I'm planning to rewrite some of how the tarball reader works, and I'm worried that I would have to break this to do what I want... maybe we can revisit this later, though.

siddharthab:
Fair enough, this was just an idea. Thank you for your prompt responses.

PR #591, opened 2019-10-22
Add crane tag

jonjohnsonjr:
Why is this better than `crane cp`?

Less typing! Doesn't do blob existence checks, so you can guarantee it's quick! It's not immediately apparent that `crane cp` is useful for tagging images.


pkg/crane/crane_test.go
```diff
@@ -282,6 +293,9 @@ func TestBadInputs(t *testing.T) {
 		crane.Copy(invalid, invalid),
 		crane.Copy(valid, invalid),
 		crane.Copy(valid, valid), // 404
+		crane.Tag(invalid, invalid),
+		crane.Tag(valid, invalid),
+		crane.Tag(valid, valid), // 404
```
ImJasonH: It's not immediately apparent to me why this should return a 404. Is it because `valid` (while valid) doesn't exist? Maybe a better name would help?

Also, I noticed that if any of these don't return an error, the test failure message is going to be pretty unhelpful. Can you make this a `[]struct{desc string, err error}` instead, and at least log which case resulted in the nil error?

pkg/crane/tag.go
```diff
@@ -0,0 +1,50 @@
...
+	if err != nil {
+		return fmt.Errorf("fetching %q: %v", img, err)
+	}
+
+	dst := ref.Context().Tag(tag)
+
+	return remote.Tag(dst, &wrap{desc}, remote.WithAuthFromKeychain(authn.DefaultKeychain))
+}
+
+// wrap coerces a remote.Descriptor into a remote.Taggable.
```
ImJasonH: Is there a reason `remote.Descriptor` doesn't just have these methods and satisfy this interface already?

jonjohnsonjr: Name collision with the fields in v1.Descriptor that gets embedded... I have a lot of regrets.

jonjohnsonjr: How do you feel about this? https://github.com/google/go-containerregistry/compare/master...jonjohnsonjr:slim-taggable

pkg/crane/crane_test.go
jonjohnsonjr: Done

Issue #598, opened 2019-10-28
Crane broke again in GitLab CI :(


max-wittig:
I guess one should consider the master of this repository as not stable anymore.

Maybe an issue with the `go.mod`?

> go get github.com/google/go-containerregistry/cmd/crane
cd /go/src/github.com/google/go-containerregistry; git submodule update --init --recursive
fatal: No url found for submodule path 'vendor/k8s.io/code-generator' in .gitmodules
package github.com/google/go-containerregistry/cmd/crane: exit status 128

What's the alternative? Would it be possible for you to provide an docker image, similar to `kaniko:debug` with a shell? Thanks 😃

jonjohnsonjr:
> I guess one should consider the master of this repository as not stable anymore.

Sorry :(

I'm kind of shocked that our tests didn't catch this.

> Maybe an issue with the go.mod?

I'm guessing this was broken by https://github.com/google/go-containerregistry/pull/548/files#diff-87ee15ebb32e08fc4ef22a736ce9000d

I'll send a PR to revert that file, but I'd like to figure out how to catch this regression. I can't seem to reproduce your issue. If you have any ideas for how we'd catch this in this future, I'd gladly take a PR :smile: 

> Would it be possible for you to provide an docker image, similar to kaniko:debug with a shell?

Are you using the images we publish ([here](gcr.io/go-containerregistry/crane))? These don't seem broken, at least:

```
$ docker run gcr.io/go-containerregistry/crane digest ubuntu
2019/10/28 18:01:56 No matching credentials were found, falling back on anonymous
sha256:a7b8b7b33e44b123d7f997bd4d3d0a59fafc63e203d17efedf09ff3f6f516152
```

We could pretty easily add a `:debug` image to track `:latest`, but I'm not sure that it would help in this case? cc @ImJasonH wdyt?

jonjohnsonjr:
@max-wittig can you describe a bit more about how this is failing for you? I don't believe we use git submodules at all, so I'm not sure where that error message is coming from.

max-wittig:
@jonjohnsonjr Thanks for the response!

Not sure, if you're familar with GitLab CI, but we're basically just running this.

1. Spawning a new container with golang:1.12.4-stretch
2. Installing crane
3. Running a crane cp

```yml
crane-test:
  image: golang:1.12.4-stretch
  stage: crane
  script:
    - go get github.com/google/go-containerregistry/cmd/crane
[...]

PR #599, opened 2019-10-28
Tweak dep management to handle vendored tools

jonjohnsonjr:
This pulls in a ton of tool dependencies, but doesn't affect the compiled artifacts at all.

This adds a `tools.go` file, based on https://github.com/golang/go/issues/25922

I'm not sure if this is the best way to do it, but it also fixes https://github.com/google/go-containerregistry/issues/598 (I think)


hack/update-code-gen.sh
```diff
@@ -28,7 +28,7 @@ export GOPATH=$(go env GOPATH)
 export PATH="${GOPATH}/bin:${PATH}"
 
 
-go install k8s.io/code-generator/cmd/deepcopy-gen
-go install github.com/maxbrunsfeld/counterfeiter
+go get k8s.io/code-generator/cmd/deepcopy-gen
+go get github.com/maxbrunsfeld/counterfeiter/v6
```
ImJasonH: What's this for?

jonjohnsonjr: Partially just reverts https://github.com/google/go-containerregistry/pull/548/files#diff-87ee15ebb32e08fc4ef22a736ce9000d but our switch over to go modules basically broke assumptions in this file (Gopkg.toml does not vendor these packages anymore), and go modules require appending that v6 to the end of counterfeiter for reasons outside of my ability to comprehend. This + `go mod vendor` should fix the regression (I _think_).

PR #603, opened 2019-11-06
Cache google.Keychain results

### LINKED ISSUE #602: Cache authenticators in google.Keychain
We can reduce the amount of time spent waiting for gcloud if we cache what we get back in `google.Keychain`, see https://github.com/GoogleContainerTools/skaffold/pull/3007

jonjohnsonjr:
Fixes #602

### DIFF
--- a/pkg/gcrane/list.go
+++ b/pkg/gcrane/list.go
@@ -48,19 +48,14 @@ func ls(root string, recursive bool) {
 		log.Fatalln(err)
 	}
 
-	auth, err := google.Keychain.Resolve(repo.Registry)
-	if err != nil {
-		log.Fatalf("getting auth for %q: %v", root, err)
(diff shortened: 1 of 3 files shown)


jonjohnsonjr:
> What are the implications of the cache never being invalidated?

A long-running process that re-fetched credentials every so often wouldn't pick up changes to its environment. If you broke the environment (e.g. unset APPLICATION_DEFAULT_CREDENTIALS), this would keep returning the environment authenticator instead of falling back to gcloud... seems unlikely.

jonjohnsonjr:
I'm not convinced that dynamically changing the source of credentials at runtime is a very common scenario 🤔

dgageot:
I'd love to use that in Skaffold :-)

PR #604, opened 2019-11-07
Prevent asking pull scope for cross-repo mounting

### LINKED ISSUE #600: Multiple scope generation with a base image got from an openshift registry fails
Hi,

I am using [kaniko](https://github.com/GoogleContainerTools/kaniko) and ended on [an issue with openshift registries](https://github.com/GoogleContainerTools/kaniko/issues/808).

Going deeper, I spotted this was due to the fact I use a internal docker proxy

Here is my `Dockerfile`

```
FROM docker-internal-proxy/busybox:1.31.0

....
```

The `push` fails when `kaniko` does the bearer request with the scopes it deduced from the `Layer` of the image in the [Write function](https://github.com/google/go-containerregistry/blob/50b26ee28691d22901fa3f55c52bdedf6c6a0be1/pkg/v1/remote/write.go#L55)

Indeed, given the above `Dockerfile`, it tries to authenticate with the given scopes:

- `repository:test-kaniko/test-kaniko:push,pull`
- `repository:busybox:pull` (unauthorized as it should have format `<repo>/<image>`)

If I use the following `Dockerfile` (without using the internal proxy

```
FROM busybox:1.31.0

....

```

It gives the scopes

- `repository:test-kaniko/test-kaniko:push,pull` 
- `repository:library/busybox:pull` (given image `busybox` url is resolved as `index.docker.io/library/busybox`)

I removed the sending of all `pull` scope (the one for the base image) and kaniko push works well 

My question: why are we sending also `pull` scope? If necessary, could we add for the write operation a fallback **without** the `pull` scopes deduced from the layers of the image ? Or even uniformizing the scopes to match the required format?

Thanks

antechrestos:
@jonjohnsonjr As promised, I implemented the fix we discussed. I also added a test to enforce the implementation.

See #600

### DIFF
--- a/pkg/v1/remote/write.go
+++ b/pkg/v1/remote/write.go
@@ -412,9 +412,10 @@ func scopesForUploadingImage(ref name.Reference, layers []v1.Layer) []string {
 
 	for _, l := range layers {
 		if ml, ok := l.(*MountableLayer); ok {
-			// we add push scope for ref.Context() after the loop
-			if ml.Reference.Context() != ref.Context() {
-				scopeSet[ml.Reference.Context().Scope(transport.PullScope)] = struct{}{}
+			// we will add push scope for ref.Context() after the loop.
+			// for now we ask pull scope for references of the same registry
+			if ml.Reference.Context() != ref.Context() && ml.Reference.Context().Registry == ref.Context().Registry {
+				scopeSet[ml.Reference.Scope(transport.PullScope)] = struct{}{}
 			}
 		}
 	}
--- a/pkg/v1/remote/write_test.go
+++ b/pkg/v1/remote/write_test.go
@@ -907,12 +907,22 @@ func TestScopesForUploadingImage(t *testing.T) {
 		t.Fatalf("name.NewTag() = %v", err)
 	}
 
+	sameReference, err := name.NewTag("example.com/sample/sample:previous", name.WeakValidation)
+	if err != nil {
+		t.Fatalf("name.NewTag() = %v", err)
+	}
+
 	anotherRepo1, err := name.NewTag("example.com/sample/another_repo1:latest", name.WeakValidation)
 	if err != nil {
 		t.Fatalf("name.NewTag() = %v", err)
 	}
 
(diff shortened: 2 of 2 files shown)


pkg/v1/remote/write.go
```diff
@@ -412,8 +412,9 @@ func scopesForUploadingImage(ref name.Reference, layers []v1.Layer) []string {
 
 	for _, l := range layers {
 		if ml, ok := l.(*MountableLayer); ok {
-			// we add push scope for ref.Context() after the loop
-			if ml.Reference.Context() != ref.Context() {
+			// we will add push scope for ref.Context() after the loop.
+			// for now we ask pull scope for references of the same registry
+			if ml.Reference.Context().Registry == ref.Context().Registry {
```
jonjohnsonjr: I think we want to keep both the old check and the new check, otherwise you could end up with scopes like:

`repository:test-kaniko/test-kaniko:push,pull&repository:test-kaniko/test-kaniko:pull`

... which might confuse a registry.

antechrestos: @jonjohnsonjr Ok. I also added a test to validate this behaviour. 

In the mean time, I find (but I am not an expert) the two lines confusing:

```
scopeSet[ml.Reference.Context().Scope(transport.PullScope)] = struct{}{}
...
scopes = append(scopes, ref.Scope(transport.PushScope))
```

Why don't we do

```
scopeSet[ml.Reference.Context().Scope(transport.PullScope)] = struct{}{}
scopes = append(scopes, ref.Context().Scope(transport.PushScope))
...

(or without `Context()` in both instruction?

jonjohnsonjr: > Why don't we do

Good question!

> (or without `Context()` in both instruction?

I'd prefer this. I'm not sure why we call `Context` on the `MountableLayer` -- I'm guessing the author was thinking "I need read scopes for this layer's _repository_, so I'll call Context", but I believe there's no different. You could just call `ml.Reference.Scope(transport.PullScope)` instead, which I think is cleaner.

antechrestos: Ok I've re pushed this change

Issue #549, opened 2019-09-23
Paginate tags and catalog


jonjohnsonjr:
I punted on this in https://godoc.org/github.com/google/go-containerregistry/pkg/v1/remote#List and probably shouldn't have.

We should figure out a method of paginating both tags listing and catalog.

jonjohnsonjr:
Fixed by https://github.com/google/go-containerregistry/pull/607

PR #607, opened 2019-11-10
Use Link header for pagination

dsbrng25b:
* Adds support for pagination on /v2/repo/tags/list
* Apply same logic for catalog /v2/_catalog

Instead of setting the url parameters `n` and `last` manually the logic only relies on the `Link` header as described in https://docs.docker.com/registry/spec/api/#pagination.

#549

### DIFF
--- a/pkg/internal/retry/retry.go
+++ b/pkg/internal/retry/retry.go
@@ -17,6 +17,7 @@
 package retry
 
 import (
+	"context"
 	"fmt"
 
 	"github.com/google/go-containerregistry/pkg/internal/retry/wait"
@@ -35,6 +36,9 @@ type temporary interface {
 
 // IsTemporary returns true if err implements Temporary() and it returns true.
 func IsTemporary(err error) bool {
+	if err == context.Canceled || err == context.DeadlineExceeded {
+		return false
+	}
 	if te, ok := err.(temporary); ok && te.Temporary() {
 		return true
 	}
--- a/pkg/internal/retry/retry.go
+++ b/pkg/internal/retry/retry.go
@@ -36,7 +36,7 @@ type temporary interface {
 
 // IsTemporary returns true if err implements Temporary() and it returns true.
 func IsTemporary(err error) bool {
-	if err == context.Canceled || err == context.DeadlineExceeded {
+	if err == context.DeadlineExceeded {
 		return false
 	}
 	if te, ok := err.(temporary); ok && te.Temporary() {
--- a/pkg/v1/remote/list.go
+++ b/pkg/v1/remote/list.go
@@ -69,10 +69,10 @@ func ListWithContext(ctx context.Context, repo name.Repository, options ...Optio
 		}
 
 		req, err := http.NewRequest("GET", uri.String(), nil)
-		req = req.WithContext(ctx)
 		if err != nil {
 			return nil, err
 		}
+		req = req.WithContext(ctx)
 
 		resp, err := client.Do(req)
 		if err != nil {
--- a/pkg/v1/remote/list.go
+++ b/pkg/v1/remote/list.go
@@ -68,7 +68,8 @@ func ListWithContext(ctx context.Context, repo name.Repository, options ...Optio
 		default:
 		}
 
-		req, err := http.NewRequestWithContext(ctx, "GET", uri.String(), nil)
+		req, err := http.NewRequest("GET", uri.String(), nil)
+		req = req.WithContext(ctx)
 		if err != nil {
 			return nil, err
 		}
--- a/pkg/v1/remote/list.go
+++ b/pkg/v1/remote/list.go
@@ -31,7 +31,7 @@ type tags struct {
 	Tags []string `json:"tags"`
 }
 
-// List wraps ListWithContext using the backround context
+// List wraps ListWithContext using the background context.
 func List(repo name.Repository, options ...Option) ([]string, error) {
 	return ListWithContext(context.Background(), repo, options...)
 }
--- a/pkg/v1/remote/list.go
+++ b/pkg/v1/remote/list.go
@@ -50,9 +50,10 @@ func ListWithContext(ctx context.Context, repo name.Repository, options ...Optio
 	}
 
 	uri := &url.URL{
-		Scheme: repo.Registry.Scheme(),
-		Host:   repo.Registry.RegistryStr(),
-		Path:   fmt.Sprintf("/v2/%s/tags/list", repo.RepositoryStr()),
+		Scheme:   repo.Registry.Scheme(),
+		Host:     repo.Registry.RegistryStr(),
+		Path:     fmt.Sprintf("/v2/%s/tags/list", repo.RepositoryStr()),
+		RawQuery: "n=10000",
 	}
 
 	client := http.Client{Transport: tr}
--- a/pkg/v1/remote/list.go
+++ b/pkg/v1/remote/list.go
@@ -15,6 +15,7 @@
 package remote
 
 import (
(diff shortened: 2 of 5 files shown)


jonjohnsonjr:
Thanks for the PR!

I'm not sure that we want exactly this signature for `remote.Catalog`, since it ends up buffering all the results before returning to the caller, see the discussion from this PR: https://github.com/google/go-containerregistry/pull/548

I think we'll want to do something with channels or maybe even a callback function instead, but it's unclear to me what the best approach is. I'm leaning towards a signature like this:

```go
func Catalog(ctx context.Context, target name.Registry, options ...Option) (<-chan string, <-chan error)
```

Some benefits:
1. You can cancel the catalog iteration via the context.
2. You get streaming results, so you don't have to wait for everything to buffer.
3. You can ignore individual errors and keep going instead of failing all or nothing.

dsbrng25b:
I added support for passing a context.Context to the list actions.
Before it works correctly the retry mechanism in the RoundTripper has to be changed/fixed because currently it does not respect the context of the http.Request.

Further I set the page size (query parameter `n`) to 10'000. This greatly improves the performance if you list catalog which contains a lot of images. I tested this with the `registry:2` with over 6000 images. If you don't set a page size the registry responds with 100 images per page and it takes over 5 seconds to return all images. With `n` set to 10000 it takes less than 0.3s.
```
$ time ./crane catalog registry.local | wc -l    
6023    
    
real    0m5.315s    
user    0m0.103s    
sys     0m0.015s    
$ git checkout pagination                        
Switched to branch 'pagination'    
$ go build                       
$ time ./crane catalog registry.local | wc -l                       
6023    
    
real    0m0.247s    
user    0m0.070s    
sys     0m0.032s 
```

Concerning the return value I would stay with `([]string, error)`.  In my opinion you either want the whole list or nothing. Also I can't think of a use case where the return time is very important. Or what use cases do you think of where you need to have control over these things?
I would publish Catalog() and List() with ([]string, error) and if people need more control over it introduce further functions later.

dsbrng25b:
I changed the IsTemporary function to always return false if the error is a DeadlineExceeded or a Canceled error. Now the context should work correctly in List and Catalog.

jonjohnsonjr:
> In my opinion you either want the whole list or nothing. Also I can't think of a use case where the return time is very important. Or what use cases do you think of where you need to have control over these things?

Specifically for me, I can't use catalog with this interface, so it will be hard to convince me otherwise :) This is partially why we're [removing catalog](https://github.com/opencontainers/distribution-spec/pull/69) from the OCI spec, but to explain:

All registry operations are scoped to a namespace except for catalog. Catalog is specifically `<host>/v2/_catalog`. This is important for later.

There are two major ways most people segment registries into namespaces: by domain or by path. DockerHub and GCR both use path, whereas ACR and ECR (I think) segment by domain. For example, I have a registry `gcr.io/jonjohnson`, but for ACR it would be something like `jonjohnson.azurecr.io`.

For ACR, I agree with you, `GET jonjohnson.azurecr.io/v2/_catalog` would return everything in that single registry, and partial results don't make any sense.

For GCR, things get a bit tricky. The namespace here (jonjohnson) actually maps to a GCP project. Ideally, a catalog operation would be something like `GET gcr.io/v2/jonjohnson/_catalog`, but that's _not_ how catalog works. You have to `GET gcr.io/v2/_catalog`, which presents a problem. My project ID is dropped from the path, so there's no way map back to my registry.

How does GCR solve this? They list every project I'm a member of, and return results for all of them. This presents some problems:

1. As a google employee, I'm a member of a lot of projects. Trying to finish a catalog operation would probably take me hours. I actually can't control which projects I'm a part of, because others can add me to their projects without my knowledge or consent.
2. For a large majority of those projects, I don't have read access to GCR, so when I try to actually list repositories in that project, it will fail.

I can mostly work around this by using service accounts that only have access to certain projects, but folks could still add that service account to their project and break my catalog listing without my knowledge :( so... being able to ignore errors here is really important for me, and anyone else that uses GCR.

That being said... I do agree with your point that returning ([]string, error) is much easier to use. My hope here was that pkg/v1/remote would expose as much useful complexity as we could (within reason), whereas pkg/crane would be a really simple to use package that covers most use cases. If you needed more than that, you'd drop down into remote and deal with the unwieldy API of reality.

> Further I set the page size (query parameter n) to 10'000. This greatly improves the performance if you list catalog which contains a lot of images.

👍 

> I changed the IsTemporary function to always return false if the error is a DeadlineExceeded or a Canceled error. Now the context should work correctly in List and Catalog.

`context.DeadlineExceeded` makes sense but I don't think `context.Canceled` implements `Temporary`, does it?

dsbrng25b:
My main goal was to fix the List function since it did not return all tags. At the same time I changed Catalog that it uses the same logic as List. I see that there are some difficulties with _catalog on GCR or in general on container registries with a lot of images, but I don't think that these problems can be fixed by revealing more implementation details. Most likely you have to use a solution which is specific to your container registry? Or which specific details would you like to control Catalog?

> `context.DeadlineExceeded` makes sense but I don't think `context.Canceled` implements `Temporary`, does it?
I'm not sure what you mean? Neither does implement Temporary but if it is a DeadlineExceeded or a Canceled error it is never temporary. Once a context gets canceled or it times out it stays that way.

dsbrng25b:
@jonjohnsonjr How can we get ahead here? I would like to see the fixed List function merged if possible. I could remove the changes in the catalog part that  it can be discussed later or even removed in a v2 since _catalog will be removed from the oci spec either way. What do you think?

pkg/v1/remote/list.go
```diff
@@ -29,9 +31,14 @@ type tags struct {
 	Tags []string `json:"tags"`
 }
 
-// List calls /tags/list for the given repository, returning the list of tags
-// in the "tags" property.
+// List wraps ListWithContext using the backround context
```
jonjohnsonjr: nits: back*g*round, end comment with a period

jonjohnsonjr:
>I'm not sure what you mean? Neither does implement Temporary but if it is a DeadlineExceeded or a Canceled error it is never temporary.

`context.DeadlineExceeded` implements `Temporary` [here](https://github.com/golang/go/blob/0ad368675bae1e3228c9146e092cd00cfb29ac27/src/context/context.go#L167), which means it's necessary to ignore it (assuming we don't want to retry it, which I'm not so sure about), but `context.Canceled` [doesn't (currently) implement `Temporary`](https://github.com/golang/go/blob/0ad368675bae1e3228c9146e092cd00cfb29ac27/src/context/context.go#L156-L157), so we don't need to specifically check it, since it will fail [this typecheck](https://github.com/google/go-containerregistry/blob/d448c55e6b46c1d58522a2e4ef0d5a4efda9f28c/pkg/internal/retry/retry.go#L35-L37).

pkg/internal/retry/retry.go
```diff
@@ -35,6 +36,9 @@ type temporary interface {
 
 // IsTemporary returns true if err implements Temporary() and it returns true.
 func IsTemporary(err error) bool {
+	if err == context.Canceled || err == context.DeadlineExceeded {
```
jonjohnsonjr: You can drop the `context.Canceled` check, since tit will fail the typecheck below.

jonjohnsonjr:
Sorry for the delay, I missed that comment while OOO and this PR got lost in a nearly infinite sea of tabs.

dsbrng25b:
Thank you for reviewing.
I rebased the branch, fixed the typos, removed context.Canceled from the IsTemporary check and added a test for the retry transport.

PR #619, opened 2019-11-22
docs(readme): add debug image info

max-wittig:


max-wittig:
@googlebot I signed it!

jonjohnsonjr:
Thanks for the PR!

I think it might make more sense to have this info as sections under `### crane` and `### gcrane` -- that would also be a reasonable place to document the regular images (I just realized they are totally undocumented).

Perhaps a couple `#### Images` sections with copy-pastable snippets like:

```
docker run gcr.io/go-containerregistry/crane ls ubuntu
```

and

```
docker run -it --entrypoint "/busybox/sh" gcr.io/go-containerregistry/crane:debug

/ko-app/crane ls ubuntu | grep bionic$
2019/12/02 18:21:41 No matching credentials were found, falling back on anonymous
bionic
```

PR #628, opened 2019-12-10
Only use oauth flow with IdentityTokens

### LINKED ISSUE #626: When token request is forbidden, the 403 is not propagated back to the client.
Recently, the library has started to fall back to oauth to receive a token.  In the case where the original token request is forbidden due to permissions, it will then try the oauth method. If this fails, e.g. if the server doesn't support it, then the returned error does not include the forbidden info.

E.g. 1st request:
HTTP GET https://somewhere.docker.com/v2/token?scope=something&service=somewhere

Response: 403 Forbidden

Then, 2nd request:
HTTP POST https://somewhere.docker.com/v2/token

Response: 405 Method not allowed

The 405 is returned as the status code, not the 403.

jonjohnsonjr:
Fixes #626


jonjohnsonjr:
cc @kevinmdavis

PR #630, opened 2019-12-12
Separate credentialprovider into files with buildtag 🏷

vdemeester:
Each files for each provider (aws, azure, gcp, …) have a "negative"
build tag, this means by default, no behaviour change, they are all
compiled — but they can be disabled using `disable-…`.
This allows a library to depend on k8schain but not having those
automatic import, or at least choose the one to enable/disable.

Signed-off-by: Vincent Demeester <vincent@sbr.pm>

### DIFF
--- a/pkg/authn/k8schain/k8schain.go
+++ b/pkg/authn/k8schain/k8schain.go
@@ -23,14 +23,6 @@ import (
 	"k8s.io/client-go/rest"
 	"k8s.io/kubernetes/pkg/credentialprovider"
 	credentialprovidersecrets "k8s.io/kubernetes/pkg/credentialprovider/secrets"
-
-	// Credential providers
-	_ "k8s.io/kubernetes/pkg/credentialprovider/aws"
-	_ "k8s.io/kubernetes/pkg/credentialprovider/azure"
-	_ "k8s.io/kubernetes/pkg/credentialprovider/gcp"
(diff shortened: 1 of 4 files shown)


jonjohnsonjr:
This seems reasonable to me, can you update the copyright year to 2019?

pkg/authn/k8schain/k8schain_rancher.go
```diff
@@ -0,0 +1,21 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package k8schain
+
+// TODO(mattmoor): This doesn't seem to build, figure out why `dep ensure`
```
jonjohnsonjr: I think it might be worth just dropping this entirely.

vdemeester:
> This seems reasonable to me, can you update the copyright year to 2019?
> 
> Also this seems like a bug in kubernetes/kubernetes with the gcp credentialprovider. Is there an upstream bug?

Done :angel: Didn't find an upstream bug, so created one: https://github.com/kubernetes/kubernetes/issues/86245

PR #634, opened 2019-12-15
Add insecure flag to push and pull

### LINKED ISSUE #633: Force Crane to use http instead of https
I went through the code and documentation but couldn't find any option to force Crane to use http instead of https.
We've decided to use Crane inside our E2E tests and since every service is using http, when we want to pull an image from the testing server, Crane returns a connection refused error since it tries to lookup :443.
```
2019/12/14 12:54:26 Get https://pull.dockerv2.dockerv2.example.com/v2/: dial tcp 172.20.10.2:443: connect: connection refused
```

Is there any option to use http instead or we have to serve the services on https too?

erbesharat:
**What this PR does / why we need it**:
Adds the ability to push and pull from insecure registries to the crane cmd.
The flag also had to be added to a few other subcommands like export, append and etc.

**Which issue this PR fixes**:
fixes #633 

**Special notes for your reviewer**:
I wasn't sure if it's ok to change the handler map in the validate subcommand without any prior discussions, so I just added the options argument to `makeTarball` as a temporary fix so we don't have failing tests.
Let me know if it's ok to change the handler map and also what do you think is a good approach to change it.
https://github.com/google/go-containerregistry/pull/634/files#diff-67c9d71faf9a3b5faa49b6f666170f54R72

### DIFF
--- a/pkg/crane/pull.go
+++ b/pkg/crane/pull.go
@@ -30,7 +30,7 @@ import (
 const iWasADigestTag = "i-was-a-digest"
 
 // Pull returns a v1.Image of the remote image src.
-func Pull(src string, opts []name.Option) (v1.Image, error) {
+func Pull(src string, opts ...name.Option) (v1.Image, error) {
 	ref, err := name.ParseReference(src, opts...)
 	if err != nil {
 		return nil, fmt.Errorf("parsing tag %q: %v", src, err)
--- a/pkg/crane/pull.go
+++ b/pkg/crane/pull.go
@@ -30,8 +30,8 @@ import (
 const iWasADigestTag = "i-was-a-digest"
 
 // Pull returns a v1.Image of the remote image src.
-func Pull(src string) (v1.Image, error) {
-	ref, err := name.ParseReference(src)
+func Pull(src string, opts []name.Option) (v1.Image, error) {
+	ref, err := name.ParseReference(src, opts...)
 	if err != nil {
 		return nil, fmt.Errorf("parsing tag %q: %v", src, err)
 	}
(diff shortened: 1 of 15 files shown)


erbesharat:
@googlebot I signed it!

pkg/crane/pull.go
```diff
@@ -30,8 +30,8 @@ import (
 const iWasADigestTag = "i-was-a-digest"
 
 // Pull returns a v1.Image of the remote image src.
-func Pull(src string) (v1.Image, error) {
-	ref, err := name.ParseReference(src)
+func Pull(src string, opts []name.Option) (v1.Image, error) {
```
ImJasonH: This could be `opts name.Option...` and callers wouldn't have to change their signatures. Like how `ParseReference` takes zero or more options.

pkg/crane/push.go
```diff
@@ -31,8 +31,8 @@ func Load(path string) (v1.Image, error) {
 }
 
 // Push pushes the v1.Image img to a registry as dst.
-func Push(img v1.Image, dst string) error {
-	tag, err := name.NewTag(dst)
+func Push(img v1.Image, dst string, options []name.Option) error {
```
ImJasonH: Ditto

jonjohnsonjr:
You'll want to run `./hack/update-code-gen.sh` to re-generate the docs, otherwise lgtm.

> I wasn't sure if it's ok to change the handler map in the validate subcommand without any prior discussions, so I just added the options argument to makeTarball as a temporary fix so we don't have failing tests.

I don't see any problem with what you've done here, that looks like what I'd have done.

PR #639, opened 2019-12-17
Improve transport.Error

jonjohnsonjr:
Include http request in error message to provide context for what
failed.

Format multiple errors a bit better.

Add test coverage for when reading the response body fails.

### DIFF
--- a/pkg/v1/remote/transport/error.go
+++ b/pkg/v1/remote/transport/error.go
@@ -19,8 +19,26 @@ import (
 	"fmt"
 	"io/ioutil"
 	"net/http"
+	"net/url"
 	"strings"
 )
+
+// The set of query string keys that we expect to send as part of the registry
+// protocol. Anything else is potentially dangerous to leak, as it's probably
+// from a redirect. These redirects often included tokens or signed URLs.
+var paramWhitelist = map[string]struct{}{
+	// Token exchange
+	"scope":   struct{}{},
+	"service": struct{}{},
+	// Cross-repo mounting
+	"mount": struct{}{},
+	"from":  struct{}{},
+	// Layer PUT
+	"digest": struct{}{},
+	// Listing tags and catalog
+	"n":    struct{}{},
+	"last": struct{}{},
+}
 
 // Error implements error to support the following error specification:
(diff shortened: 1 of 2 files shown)


pkg/v1/remote/transport/error.go
```diff
@@ -30,13 +30,23 @@ type Error struct {
...
 }
 
 // Check that Error implements error
 var _ error = (*Error)(nil)
 
 // Error implements error
 func (e *Error) Error() string {
+	prefix := ""
+	if e.request != nil {
+		prefix = fmt.Sprintf("%s %s: ", e.request.Method, e.request.URL)
```
ImJasonH: Will this potentially include URL params related to auth?

ImJasonH: wdyt about adding `Request: ` to the beginning (`Request: GET http://blah.com`)

jonjohnsonjr: I don't love it -- seems like mostly just noise, it's pretty easy (at least for me) to recognize the $HTTP_VERB $URL pattern.

jonjohnsonjr: Possibly -- GCR includes an access token in the query parameters when redirecting to GCS (unless you're using a json key to authenticate, in which case it uses signed URLs). It's a short-lived token so the window is pretty small.

I'm also not sure if this would really be a problem. The http client issues a `GET /v2/<whatever>/blobs/<sha>` and the handles the redirect internally. The godoc for `http.Response.Request` says:

>     // This is only populated for Client requests.

So I'm guessing this is just the original request, not the redirected one.

jonjohnsonjr: Alright just tested this and it does set `Request` to the redirected URL, do you think we should drop query parameters?

jonjohnsonjr: Let's maintain a whitelist of expected query string keys from the distribution spec and REDACT anything else to avoid leaking tokens or signed URLs.

jonjohnsonjr: Done.

PR #642, opened 2019-12-18
Add crane.Option and make --insecure global

jonjohnsonjr:
I realized I want to be able to inject my own transport for crane, but
that's going to be impossible if we only accept name options, so this
wraps this options (again, sigh) in a way that lets us extend to other
packages.

This also drops -i as an option for insecure (breaking, but so briefly).

### DIFF
--- a/cmd/crane/cmd/append.go
+++ b/cmd/crane/cmd/append.go
@@ -18,7 +18,6 @@ import (
 	"log"
 
 	"github.com/google/go-containerregistry/pkg/crane"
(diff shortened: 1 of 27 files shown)


jonjohnsonjr:
This is half cleanup for https://github.com/google/go-containerregistry/pull/634

jmakinen-ncc:
+1 for wanting this approved (haven't reviewed code deeply, but I was just about to do this)

PR #643, opened 2019-12-20
Add a --json flag to gcrane ls

jonjohnsonjr:
Sometimes, I want structured output or more info than just digest +
tags. This just returns everything so that you can use e.g. jq to
process the results.

### DIFF
--- a/cmd/gcrane/cmd/list.go
+++ b/cmd/gcrane/cmd/list.go
@@ -15,6 +15,7 @@
 package cmd
 
 import (
+	"encoding/json"
 	"fmt"
 	"log"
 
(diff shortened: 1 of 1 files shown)


cmd/gcrane/cmd/list.go
```diff
@@ -28,28 +29,30 @@ func init() { Root.AddCommand(NewCmdList()) }
 // NewCmdList creates a new cobra.Command for the ls subcommand.
 func NewCmdList() *cobra.Command {
 	recursive := false
+	json := false
```
ImJasonH: wdyt about `--format=json` to sort-of match with #656 ?

jonjohnsonjr: What other formats might there be?

PR #644, opened 2019-12-20
crane append: multiple layers, empty base

jonjohnsonjr:
Allow multiple layers to be appended (via multiple -f flags, a la
kubectl apply).

Make "base" optional, which then uses empty.Image as the base.

### DIFF
--- a/cmd/crane/cmd/append.go
+++ b/cmd/crane/cmd/append.go
@@ -18,6 +18,9 @@ import (
 	"log"
 
 	"github.com/google/go-containerregistry/pkg/crane"
+	"github.com/google/go-containerregistry/pkg/logs"
+	v1 "github.com/google/go-containerregistry/pkg/v1"
+	"github.com/google/go-containerregistry/pkg/v1/empty"
 	"github.com/spf13/cobra"
 )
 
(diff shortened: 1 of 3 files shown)


cmd/crane/cmd/append.go
```diff
@@ -18,28 +18,45 @@ import (
...
+	"github.com/google/go-containerregistry/pkg/v1/empty"
 	"github.com/spf13/cobra"
 )
 
 func init() { Root.AddCommand(NewCmdAppend()) }
 
 // NewCmdAppend creates a new cobra.Command for the append subcommand.
 func NewCmdAppend() *cobra.Command {
-	var baseRef, newTag, newLayer, outFile string
+	var (
```
ImJasonH: supernit: I find the `var (` format to be wasteful of vertical space except where it's more consistent at the top-level.

```go
var baseRef, newTag, outFile string
var newLayers []string
```

(up to you)

cmd/crane/cmd/append.go
```diff
@@ -18,28 +18,45 @@ import (
...
 
 	appendCmd := &cobra.Command{
 		Use:   "append",
 		Short: "Append contents of a tarball to a remote image",
 		Args:  cobra.NoArgs,
 		Run: func(_ *cobra.Command, args []string) {
-			base, err := crane.Pull(baseRef, options...)
-			if err != nil {
-				log.Fatalf("pulling %s: %v", baseRef, err)
+			var (
```
ImJasonH: ditto

cmd/crane/cmd/append.go
jonjohnsonjr: This is the kind of pedantry I can get behind!

PR #649, opened 2020-01-07
Use vdemeester/k8s-pkg-credentialprovider temporarly 🌬

vdemeester:
This removes the dependency to `k8s.io/kubernetes` and use a "frozen
in time" version of this package.

This is temporary in the sense, that once
https://github.com/kubernetes/enhancements/pull/1406 is done, this
will not be needed anymore and we should switch back to using it
instead.

/cc @jonjohnsonjr 

Signed-off-by: Vincent Demeester <vincent@sbr.pm>


vdemeester:
The second commit updates the `vendor` folder that didn't seem up-to-date :sweat_smile: 
Not sure if we should keep it or removing it :angel:

jonjohnsonjr:
@vdemeester I think this may have broken our image publishing stuff... `go get github.com/google/ko/cmd/ko` fails with:

> Step #0: build github.com/google/ko/cmd/ko: cannot load k8s.io/api/admissionregistration/v1alpha1: module k8s.io/api@latest found (v0.17.1), but does not contain package k8s.io/api/admissionregistration/v1alpha1

We might just need to update `ko`?

vdemeester:
oh right, I guees so :thinking: I can try to take a look tomorrow :angel:

jonjohnsonjr:
https://github.com/google/ko/pull/119 might fix it

PR #653, opened 2020-01-13
Adds WithAuth functional option to pkg/crane to allow for alternate m…

LGUG2Z:
To allow for alternate methods of authenticating with GCR when consuming as a library.

Use case: In a Slack bot deployed to a GKE cluster, I am using crane as a library to validate that a given image/tag combination exists in a private GCR registry. I am handling the access control to GCR already per-node pool via node pool service accounts, so it makes sense to be able to leverage that instead of trying to mount a Docker config into a pod.

### DIFF
--- a/pkg/crane/options.go
+++ b/pkg/crane/options.go
@@ -56,7 +56,7 @@ func Insecure(o *options) {
 }
 
 // WithAuth is a functional option for overriding the default authentication
-// mechanism for GCR operations.
+// mechanism for remote operations.
 func WithAuth(a authn.Authenticator) Option {
 	return func (o *options) {
 		o.remote = append(o.remote, remote.WithAuth(a))
--- a/pkg/crane/options.go
+++ b/pkg/crane/options.go
@@ -59,6 +59,6 @@ func Insecure(o *options) {
 // mechanism for remote operations.
 func WithAuth(a authn.Authenticator) Option {
 	return func (o *options) {
-		o.remote = append(o.remote, remote.WithAuth(a))
+		o.remote = []remote.Option{remote.WithAuth(a)}
 	}
 }
--- a/pkg/crane/options.go
+++ b/pkg/crane/options.go
@@ -22,6 +22,8 @@ import (
 	"github.com/google/go-containerregistry/pkg/v1/remote"
 )
 
+var defaultRemoteOption = remote.WithAuthFromKeychain(authn.DefaultKeychain)
+
 type options struct {
 	name   []name.Option
 	remote []remote.Option
@@ -30,7 +32,7 @@ type options struct {
 func makeOptions(opts ...Option) options {
 	opt := options{
 		remote: []remote.Option{
-			remote.WithAuthFromKeychain(authn.DefaultKeychain),
+			defaultRemoteOption,
 		},
 	}
 	for _, o := range opts {
@@ -59,6 +61,10 @@ func Insecure(o *options) {
 // mechanism for remote operations.
 func WithAuth(a authn.Authenticator) Option {
 	return func (o *options) {
-		o.remote = []remote.Option{remote.WithAuth(a)}
+		if &o.remote[0] == &defaultRemoteOption {
+			o.remote[0] = remote.WithAuth(a)
+		} else {
+			o.remote = append(o.remote, remote.WithAuth(a))
+		}
 	}
 }
--- a/pkg/crane/options.go
+++ b/pkg/crane/options.go
@@ -22,8 +22,6 @@ import (
 	"github.com/google/go-containerregistry/pkg/v1/remote"
 )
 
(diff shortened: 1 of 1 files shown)


pkg/crane/options.go
```diff
@@ -54,3 +54,11 @@ func WithTransport(t http.RoundTripper) Option {
 func Insecure(o *options) {
 	o.name = append(o.name, name.Insecure)
 }
+
+// WithAuth is a functional option for overriding the default authentication
+// mechanism for GCR operations.
```
jonjohnsonjr: small nit: can you change `GCR` to just `remote`? This isn't GCR-specific.

LGUG2Z:
@googlebot I signed it!

pkg/crane/options.go
LGUG2Z: Updated from `GCR` to `remote`

jonjohnsonjr:
Thanks!

Have you tested this to see if it actually works? I'm skeptical given this implementation in `remote`:
https://github.com/google/go-containerregistry/blob/f4fb41bf00a3cb49242cd8cdaf1accc6ecd159dd/pkg/v1/remote/options.go#L36-L58

Since the default option is to always set the keychain to `authn.DefaultKeychain`, I'm pretty sure whatever `WithAuth` sets will get overridden. We may have to add a guard [here](https://github.com/google/go-containerregistry/blob/f4fb41bf00a3cb49242cd8cdaf1accc6ecd159dd/pkg/v1/remote/options.go#L49) to only invoke the `if o.keychain != nil && o.auth == authn.Anonymous` and document that `WithAuth` overrides `WithAuthFromKeychain`.

This is an unfortunate side effect of my decision to always use the default keychain in `crane`, but it felt like the right thing to do at the time :P

LGUG2Z:
You are correct that the original implementation wasn't working! I think that is due to the fact that the default keychain is always the first item in the `options.remote` slice: https://github.com/google/go-containerregistry/blob/8f601eedee18470817be5559e719a408621c66e4/pkg/crane/options.go#L30-L40

In the latest commit (which I have actually tested is working as expected in the context of the Slackbot I am developing this time), `WithAuth` creates a new slice for `options.remote` and ensures that it is the first and only authenticator parsed from that slice. This could potentially be an issue if somebody wants to use both `WithTransport` and `WithAuth` and passes `WithAuth` as the final argument as it will lose the changes made by the previous function...

Would it be safe (enough) to always remove the first element in the `options.remote` slice when calling passing `WithAuth`, given that the `options` object always prepopulates it with `remote.WithAuthFromKeychain(authn.DefaultKeychain)` as the first element?

jonjohnsonjr:
> Would it be safe (enough) to always remove the first element in the options.remote slice when calling passing WithAuth, given that the options object always prepopulates it with remote.WithAuthFromKeychain(authn.DefaultKeychain) as the first element?

I don't love it... what we could do is pull [this line](https://github.com/google/go-containerregistry/blob/f4fb41bf00a3cb49242cd8cdaf1accc6ecd159dd/pkg/crane/options.go#L33) out into a `var defaultRemoteOption` or something, then when you call `WithAuth` we can first check that the first element is what we expect (`options.remote[0] == defaultRemoteOption`) before removing it. Otherwise calling `WithAuth` twice could remove the wrong thing. WDYT?

LGUG2Z:
I thought to try this but the `remote.Option` type is not comparable (it is a func that returns an error); I tried to extract to `var defaultRemoteOption` and then do a pointer comparison but unfortunately that condition does not pass when run so the override does not get executed:

```
		if &o.remote[0] == &defaultRemoteOption {
			o.remote[0] = remote.WithAuth(a)
		}
```

Do you have any other ideas how else we could try to validate whether or not to remove or override the `defaultRemoteOption` that prepopulates the `options.remote` slice?

jonjohnsonjr:
> the remote.Option type is not comparable

Ah, interesting. TIL. For the sake of simplicity, let's just go back to your idea of replacing `o.remote[0]` with `auth`. Maybe document that passing in multiple `WithAuth`s can have unexpected behavior? If it's an issue we can fix it "for real" in the `remote` package.

pkg/crane/options.go
```diff
@@ -54,3 +54,11 @@ func WithTransport(t http.RoundTripper) Option {
 func Insecure(o *options) {
 	o.name = append(o.name, name.Insecure)
 }
+
+// WithAuth is a functional option for overriding the default authentication
+// mechanism for remote operations.
+func WithAuth(a authn.Authenticator) Option {
+	return func (o *options) {
+		o.remote = []remote.Option{remote.WithAuth(a)}
```
jonjohnsonjr, line 62:
```suggestion
		o.remote[0] = remote.WithAuth(a)
```

jonjohnsonjr (commented):
I know it's been forever, but would you like to rebase this to fix conflicts and then we could merge?

PR #654, opened 2020-01-13
Add partial.Descriptor

jonjohnsonjr:
This consolidates a bit of logic for generating a v1.Descriptor, but
also works around some of the nastiness that the partial package
introduces.

Because partial wraps layers and images with an extra level of
implementation, it shadows any methods that the underlying object
implements, so we can't expose "shortcuts". E.g. for tarball, we may
want to expose the URLs from LayerSources to callers, but the partial
wrapping gets in the way.

This is going to be a useful pattern in general unless we get rid of the
partial package entirely. Accessor methods that can introspect the
(private) partial structs will be necessary if we want to support
"optional" methods like UncompressedSize or Descriptor.


jonjohnsonjr:
cc @smukherj1

jonjohnsonjr:
We can use this pattern to add a `partial.UncompressedSize`, which rules_docker could use to solve https://github.com/bazelbuild/rules_docker/issues/1349

Given the unfortunate nature of the tar format, we need to know the size of a layer before we can start writing it, which involves scanning the entire thing. `v1.Layer` provides a `Size` (for `Compressed` bytes) but no `UncompressedSize` (for `Uncompressed` bytes). Since we probably don't want to force all layer implementations to add an `UncompressedSize` (impossible for remote, AFAIK), this has to be "optional".

There is some precedent for doing this in the standard library, e.g. [`bufio.Reader`](https://golang.org/pkg/bufio/#Reader.WriteTo):

> WriteTo implements io.WriterTo. This may make multiple calls to the Read method of the underlying Reader. If the underlying reader supports the WriteTo method, this calls the underlying WriteTo without buffering.

PR #655, opened 2020-01-14
Implement partial.UncompressedSize

jonjohnsonjr:
This allows access to an optional method that v1.Layers can implement to
cheaply get the size of the uncompressed contents of a layer. This is
useful in situations where you need to know the size of the contents
before writing them (e.g. pkg/legacy/tarball.Write).

I dropped the `ReadAll` call in the "slow path". This means that the `Uncompressed` bytes aren't buffered (yay), but if using a streaming layer, we might accidentally consume it. The only known uses of the legacy package _don't_ use streaming layers, so we can cross that bridge when we get to it.

### DIFF
--- a/pkg/v1/partial/with.go
+++ b/pkg/v1/partial/with.go
@@ -129,12 +129,6 @@ func RawConfigFile(i WithConfigFile) ([]byte, error) {
 	return json.Marshal(cfg)
 }
 
-// WithUncompressedLayer defines the subset of v1.Image used by these helper methods
-type WithUncompressedLayer interface {
-	// UncompressedLayer is like UncompressedBlob, but takes the "diff id".
-	UncompressedLayer(v1.Hash) (io.ReadCloser, error)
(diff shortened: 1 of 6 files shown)


pkg/v1/partial/with.go
```diff
@@ -348,3 +342,41 @@ func Descriptor(d Describable) (*v1.Descriptor, error) {
...
+			return wus.UncompressedSize()
+		}
+	}
+	if cle, ok := l.(*compressedLayerExtender); ok {
+		if wus, ok := cle.CompressedLayer.(withUncompressedSize); ok {
+			return wus.UncompressedSize()
+		}
+	}
+
+	// The layer doens't implement CompressedSize, we need to compute it.
```
smukherj1: nit: UncompressedSize

jonjohnsonjr: Thanks :)

PR #656, opened 2020-01-14
Add --format flag to `crane pull`

### LINKED ISSUE #621: Add different formats for crane pull
There are multiple reasonable on-disk representations of an image:
* [v1/layout](https://github.com/google/go-containerregistry/tree/master/pkg/v1/layout)
* [v1/tarball](https://github.com/google/go-containerregistry/tree/master/pkg/v1/tarball)
* [legacy/tarball](https://github.com/google/go-containerregistry/tree/master/pkg/legacy/tarball)

Crane currently just pulls v1/tarball for everything, which is compatible with `docker load`, but apparently incompatible with [dive](https://github.com/wagoodman/dive). It would be nice if we could specify the format when pulling to be broadly compatible with different tools.

What I'm thinking:

```
# v1/tarball (also default)
crane pull --format=tarball ubuntu ubuntu.tar

# legacy/tarball
crane pull --format=legacy ubuntu ubuntu.tar

# v1/layout
crane pull --format=oci ubuntu ubuntu/
```

jonjohnsonjr:
Fixes #621

This adds the --format flag with values "tarball" (default), "legacy"
(for saving pkg/legacy/tarball images, for compatibility with some
software that assumes the old format), and "oci", which saves the image
as an OCI image layout.

### DIFF
--- a/pkg/crane/crane_test.go
+++ b/pkg/crane/crane_test.go
@@ -23,6 +23,7 @@ import (
 	"net/http"
 	"net/http/httptest"
 	"net/url"
+	"os"
 	"testing"
 
 	"github.com/google/go-containerregistry/pkg/crane"
@@ -199,6 +200,8 @@ func TestCraneTarball(t *testing.T) {
 	if err != nil {
 		t.Fatal(err)
 	}
+	defer os.Remove(tmp.Name())
+
 	img, err := random.Image(1024, 5)
 	if err != nil {
 		t.Fatal(err)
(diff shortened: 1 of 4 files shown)


pkg/crane/crane_test.go
```diff
@@ -288,7 +328,7 @@ func TestCraneFilesystem(t *testing.T) {
 
 func TestBadInputs(t *testing.T) {
 	t.Parallel()
-	invalid := "@@@@@@"
+	invalid := "/dev/null/@@@@@@"
```
ImJasonH: Is `@@@@@@` no longer invalid?

jonjohnsonjr: This makes it both an invalid file to read and an invalid file to write.

Issue #651, opened 2020-01-09
Include manifest in the v1 tarball? Or latest OCI image-spec?


deitch:
Per #255 and the reference PR, support for v1 tarball was added, which steps in this direction. However, it does not include multiple blobs, or the original manifest or index, which the spec at https://github.com/opencontainers/image-spec/blob/master/image-layout.md does do. This breaks the confirmation of software supply chain a little, makes it hard to see the entire piece.

For now, I have been hacking the v1 tar file by adding the original manifest after pulling, but I would far prefer to follow the spec.

jonjohnsonjr:
There's a separate package for the OCI image layout: https://godoc.org/github.com/google/go-containerregistry/pkg/v1/layout

The tarball packages are for dealing with the `docker save` style image tarballs.

deitch:
I had used that for on disk layout, I was thinking more as a tarball. I guess I always could lay it out to disk and then tar it up, or turn it into a stream.

Still, be nice to be able to save an image to an OCI layout in a tarball. Or did I miss that in the godoc?

If it makes sense, I can try and find a little time to write a PR for it, so v1.layout can work against a Path or a tarball. 

> docker save style image tarballs

Got it. And to be compatible, you need the manifest.json, which is distinct from how the image layout does it, although I think everything else would work if the paths in the manifest.jaon were correct. 

A few questions:

* doesn’t the docker save format require the `repositories` file? We have that in the legacy tarball but not the v1 tarball. There are tools that choke without
* can we include the registry manifest? Any docker save or load tools ignore any file that doesn’t fit with its manifest, so it doesn’t hurt, but is helpful for full provenance 
* if we took the v1 layout on disk, tarred it up, and added a `manifest.json` (with the proper paths under `blob/sha256/` of course), would it not also be compatible with `docker save`?

deitch:
> if we took the v1 layout on disk, tarred it up, and added a manifest.json (with the proper paths under blob/sha256/ of course), would it not also be compatible with docker save?

I can confirm that it would. However, it would be a bit misleading, as it would have _all_ of the images under the index in it, and the `manifest.json` would point just at one. However, you could have a universal format by including just the one you wanted.

jonjohnsonjr:
> If it makes sense, I can try and find a little time to write a PR for it, so v1.layout can work against a Path or a tarball.

[...]

PR #665, opened 2020-01-26
v1.layout.WriteFile public, so can use it for other things

deitch:
As discussed in #651

### DIFF
--- a/pkg/v1/layout/write.go
+++ b/pkg/v1/layout/write.go
@@ -124,15 +124,19 @@ func (l Path) AppendDescriptor(desc v1.Descriptor) error {
 		return err
 	}
 
-	return l.writeFile("index.json", rawIndex)
+	return l.WriteFile("index.json", rawIndex, os.ModePerm)
 }
 
-func (l Path) writeFile(name string, data []byte) error {
+// WriteFile write a file with arbitrary data at an arbitrary location in a v1
+// layout. Used mostly internally to write files like "oci-layout" and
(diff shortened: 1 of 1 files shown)


pkg/v1/layout/write.go
```diff
@@ -124,10 +124,14 @@ func (l Path) AppendDescriptor(desc v1.Descriptor) error {
...
-	return l.writeFile("index.json", rawIndex)
+	return l.WriteFile("index.json", rawIndex)
 }
 
-func (l Path) writeFile(name string, data []byte) error {
+// WriteFile write a file with arbitrary data at an arbitrary location in a v1
+// layout. Used mostly internally to write files like "oci-layout" and
+// "index.json", also can be used to write other arbitrary files. Do *not* use
+// this to write blobs. Use only WriteBlob() for that.
+func (l Path) WriteFile(name string, data []byte) error {
```
jonjohnsonjr: Let's expose the `perm os.FileMode` parameter here to match `ioutil.WriteFile`, in case we don't always want `os.ModePerm`.

Otherwise LGTM.

deitch: That's a good idea. Doing it now.

deitch: Done. I had to then add the `os.ModePerm` to all of the other calls to it, but that was easy.

jonjohnsonjr:
Thanks!

deitch:
With pleasure. We can continue the rest of the conversation in #651

PR #667, opened 2020-01-30
Make gcrane work with Artifact Registry

halvards:
- Use googleKeychain to authenticate to pkg.dev registries

- Handle schema difference for the list operation. GCR returns
  timeUploadedMs as a string, while AR returns the same field
  as int64

### DIFF
--- a/pkg/v1/google/list.go
+++ b/pkg/v1/google/list.go
@@ -98,12 +98,14 @@ func (l *lister) list(repo name.Repository) (*Tags, error) {
 	return &tags, nil
 }
 
+// Uploaded uses json.Number to work around GCR returning timeUploaded
+// as a string, while AR returns the same field as int64.
 type rawManifestInfo struct {
-	Size      string   `json:"imageSizeBytes"`
-	MediaType string   `json:"mediaType"`
-	Created   string   `json:"timeCreatedMs"`
-	Uploaded  string   `json:"timeUploadedMs"`
-	Tags      []string `json:"tag"`
+	Size      string      `json:"imageSizeBytes"`
+	MediaType string      `json:"mediaType"`
+	Created   string      `json:"timeCreatedMs"`
+	Uploaded  json.Number `json:"timeUploadedMs"`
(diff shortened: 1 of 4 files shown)


pkg/v1/google/list_test.go
```diff
@@ -67,6 +67,64 @@ func TestRoundtrip(t *testing.T) {
 	}
 }
 
+// GCR returns timeUploaded as string
+func TestTimeUploadedMsAsString(t *testing.T) {
+	data := []byte(`
+		{
+			"imageSizeBytes": "100",
```
ImJasonH: Indentation is off here and below.

halvards: Oops, should be fixed now, thanks!
Please let me know if you want me to squash the commits.

pkg/v1/google/list.go
```diff
@@ -98,12 +98,14 @@ func (l *lister) list(repo name.Repository) (*Tags, error) {
...
 type rawManifestInfo struct {
-	Size      string   `json:"imageSizeBytes"`
-	MediaType string   `json:"mediaType"`
-	Created   string   `json:"timeCreatedMs"`
-	Uploaded  string   `json:"timeUploadedMs"`
-	Tags      []string `json:"tag"`
+	Size      string      `json:"imageSizeBytes"`
+	MediaType string      `json:"mediaType"`
+	Created   string      `json:"timeCreatedMs"`
+	Uploaded  json.Number `json:"timeUploadedMs"`
```
kevinmdavis: We intend for AR to return the same set of fields as GCR, so we can fix this issue on the server side rather than on the client.

halvards: That's great! Until then, given that it's a minor change that handles both GCR and AR (and you could argue that json.Number is the appropriate type for a number returned in a JSON document...), we could leave this in place so gcrane works for early AR users?

jonjohnsonjr: Seems reasonable to me!

PR #674, opened 2020-02-14
Add WithClient to daemon.Image

### LINKED ISSUE #139: daemon library shouldn't rely on environment variables.
I think the daemon logic should be using `NewClientWithOpts` and should be expressing the configuration for how to reach the Daemon via our own `Options` struct.

We can expose a variant for populating our own `Options` from the environment for CLIs, but requiring usage of the environment for passing this information seems strange for a library.

jonjohnsonjr:
Fixes #139

Rather than wrapping the docker client options (as proposed in #139),
this just allows you to pass in an entire client implementation. This
makes testing a nicer, since we can pass in a MockImageSaver, and
callers can configure the client however they'd like without us having
to keep up with the docker client dependency.

By default, we'll still use the same options (client.FromEnv) to reach
the daemon (for compatibility).

### DIFF
--- a/pkg/v1/daemon/image.go
+++ b/pkg/v1/daemon/image.go
@@ -36,6 +36,7 @@ var _ v1.Image = (*image)(nil)
 type imageOpener struct {
 	ref      name.Reference
 	buffered bool
+	client   Client
 }
 
 // ImageOption is a functional option for Image.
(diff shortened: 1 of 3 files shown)


jonjohnsonjr:
cc @briandealwis

Issue #670, opened 2020-02-04
download progress?


deitch:
When doing a pull, is it possible to get download progress? It is fairly straightforward to calculate the entire size from an image by adding up the layers sizes. If it is possible to get download updates, it would be easy to calculate percentage done.

jonjohnsonjr:
I've avoided this because I'm not sure how to implement it without pulling in a ton of dependencies. Doing cross-platform tty detection properly seemed nontrivial as well. I'm also not sure how I'd plumb it through the library and what should be responsible for managing it.

I'm open to a PR if it makes sense, but I wouldn't really know where to start.

If you wanted to implement this without needing to change go-containerregistry, you could probably implement a clever http.RoundTripper that looks for `*/manifests/*` and `*/blobs/*` in the path of http requests/responses to get a signal for what's going on by parsing the manifests and using a TeeReader when downloading blobs.

deitch:
Oh quite. I think that _outputting_ the status should be a client usage problem. But if the download had a way to _report_ it without dealing with ttys and such, it would be easier.

I was thinking something like:

```go
downloaded := make(chan int64)
err := v1tarball.WriteToFileWithUpdates(path, ref, img, downloaded)
```

And the various write funcs would send periodic "I have downloaded this much cumulatively" to the channel over time. It is up to the consumer to use it or not, and decide if it wants to send the output to a tty or something else.

This at least would provide a reliable basis for it.

deitch:
Any thoughts on this? I had some time to mull it over, and I like the approach. It basically lets the library wash its hands of how it is used, just says, "give me a channel, and I will send updates there."

jonjohnsonjr:
I'm not really sure where this should live.

We could pass something into `remote.Image` that could tell us progress when downloading layers from the registry, or we could pass something into `tarbal.Write...` that could tell us progress when writing to the tarball.

It's unclear which is more correct or efficient (especially given the matrix of source -> sink combinations).

If the progress bar is in the reader, we can more efficiently calculate the Total size of a given layer. E.g. with remote, we know we can access the image manifest to get the blob size. With a tarball, we know we can stat the files to determine the file size.

On the other hand, the writer knows if it's going to skip certain layers or not e.g. if the tarball already has some base image layers present.

It seems to me that accurately (and efficiently) calculating progress requires some context from both the source and the sink.

| Source | Blob Set | Blob Sizes | Read Progress |
|---|---|---|---|
[...]

PR #675, opened 2020-02-17
add download progress option for v1.tarball

deitch:
Initial implementation of download progress, per [the discussion here](https://github.com/google/go-containerregistry/issues/670#issuecomment-584952872) with @jonjohnsonjr .

This only does it for `v1.tarball`; it still will need to be added to `legacy.tarball` and `v1.layout`. I also have given it zero testing. The goal is to have a basis for discussion.

I chose to change the calls to `Write/MultiWrite/MultiWriteRef` rather than entirely new funcs because:

* it would create an over-proliferation of functions and be very messy
* you always can pass in `nil` to a channel arg if you don't want it (which is what I did for almost all consumers
* with go modules, versioning of consumers is easy, so breaking changes on a new commit is not a big deal.

### DIFF
--- a/pkg/crane/pull.go
+++ b/pkg/crane/pull.go
@@ -63,6 +63,7 @@ func Save(img v1.Image, src, path string) error {
 		tag = d.Repository.Tag(iWasADigestTag)
 	}
 
+	// no progress channel (for now)
 	return tarball.WriteToFile(path, tag, img)
 }
 
--- a/pkg/v1/progress.go
+++ b/pkg/v1/progress.go
@@ -1,0 +1,24 @@
+// Copyright 2020 Google LLC All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package v1
+
+// Update representation of an update of transfer progress. Some functions
+// in this module can take a channel to which updates will be sent while a
+// transfer is in progress.
+type Update struct {
+	Total    int64
+	Complete int64
+	Error    error
+}
--- a/pkg/v1/tarball/write.go
+++ b/pkg/v1/tarball/write.go
@@ -18,6 +18,7 @@ import (
 	"archive/tar"
 	"bytes"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"io"
 	"os"
@@ -31,41 +32,41 @@ import (
 
 // WriteToFile writes in the compressed format to a tarball, on disk.
 // This is just syntactic sugar wrapping tarball.Write with a new file.
-func WriteToFile(p string, ref name.Reference, img v1.Image) error {
+func WriteToFile(p string, ref name.Reference, img v1.Image, opts ...WriteOption) error {
 	w, err := os.Create(p)
 	if err != nil {
 		return err
 	}
 	defer w.Close()
 
-	return Write(ref, img, w)
+	return Write(ref, img, w, opts...)
 }
 
 // MultiWriteToFile writes in the compressed format to a tarball, on disk.
 // This is just syntactic sugar wrapping tarball.MultiWrite with a new file.
-func MultiWriteToFile(p string, tagToImage map[name.Tag]v1.Image) error {
+func MultiWriteToFile(p string, tagToImage map[name.Tag]v1.Image, opts ...WriteOption) error {
 	refToImage := make(map[name.Reference]v1.Image, len(tagToImage))
 	for i, d := range tagToImage {
 		refToImage[i] = d
 	}
-	return MultiRefWriteToFile(p, refToImage)
+	return MultiRefWriteToFile(p, refToImage, opts...)
 }
 
 // MultiRefWriteToFile writes in the compressed format to a tarball, on disk.
 // This is just syntactic sugar wrapping tarball.MultiRefWrite with a new file.
-func MultiRefWriteToFile(p string, refToImage map[name.Reference]v1.Image) error {
+func MultiRefWriteToFile(p string, refToImage map[name.Reference]v1.Image, opts ...WriteOption) error {
 	w, err := os.Create(p)
 	if err != nil {
 		return err
 	}
 	defer w.Close()
 
-	return MultiRefWrite(refToImage, w)
+	return MultiRefWrite(refToImage, w, opts...)
 }
 
 // Write is a wrapper to write a single image and tag to a tarball.
-func Write(ref name.Reference, img v1.Image, w io.Writer) error {
-	return MultiRefWrite(map[name.Reference]v1.Image{ref: img}, w)
+func Write(ref name.Reference, img v1.Image, w io.Writer, opts ...WriteOption) error {
+	return MultiRefWrite(map[name.Reference]v1.Image{ref: img}, w, opts...)
 }
 
 // MultiWrite writes the contents of each image to the provided reader, in the compressed format.
@@ -73,12 +74,12 @@ func Write(ref name.Reference, img v1.Image, w io.Writer) error {
 // One manifest.json file at the top level containing information about several images.
 // One file for each layer, named after the layer's SHA.
 // One file for the config blob, named after its SHA.
-func MultiWrite(tagToImage map[name.Tag]v1.Image, w io.Writer) error {
+func MultiWrite(tagToImage map[name.Tag]v1.Image, w io.Writer, opts ...WriteOption) error {
 	refToImage := make(map[name.Reference]v1.Image, len(tagToImage))
 	for i, d := range tagToImage {
 		refToImage[i] = d
 	}
-	return MultiRefWrite(refToImage, w)
+	return MultiRefWrite(refToImage, w, opts...)
 }
 
 // MultiRefWrite writes the contents of each image to the provided reader, in the compressed format.
@@ -86,45 +87,128 @@ func MultiWrite(tagToImage map[name.Tag]v1.Image, w io.Writer) error {
 // One manifest.json file at the top level containing information about several images.
 // One file for each layer, named after the layer's SHA.
 // One file for the config blob, named after its SHA.
-func MultiRefWrite(refToImage map[name.Reference]v1.Image, w io.Writer) error {
-	tf := tar.NewWriter(w)
-	defer tf.Close()
-
-	m, err := processImages(refToImage, tf)
-	if err != nil {
-		return err
+func MultiRefWrite(refToImage map[name.Reference]v1.Image, w io.Writer, opts ...WriteOption) error {
+	// process options
+	o := &writeOptions{
+		updates: nil,
+	}
+	for _, option := range opts {
+		if err := option(o); err != nil {
+			return err
+		}
 	}
 
+	m, err := calculateManifest(refToImage)
+	if err != nil {
+		return fmt.Errorf("error calculating manifest: %v", err)
+	}
 	mBytes, err := json.Marshal(m)
 	if err != nil {
-		return err
+		return fmt.Errorf("could not marshall manifest to bytes: %v", err)
 	}
-	return writeTarEntry(tf, "manifest.json", bytes.NewReader(mBytes), int64(len(mBytes)))
+
+	size, err := calculateTarballSize(refToImage, mBytes)
+	if err != nil {
+		return fmt.Errorf("error calculating tarball size: %v", err)
+	}
+
+	return writeImagesToTar(refToImage, mBytes, size, w, o)
 }
 
-// processImages writes the images to the provider tar.Writer. Returns the manifest
-// for the written images as Manifest.`If the tar.Writer is nil, will just return
-// the Manifest.
-func processImages(refToImage map[name.Reference]v1.Image, tf *tar.Writer) (Manifest, error) {
+// writeImagesToTar writes the images to the tarball
+func writeImagesToTar(refToImage map[name.Reference]v1.Image, m []byte, size int64, w io.Writer, o *writeOptions) (err error) {
+	if w == nil {
+		return errors.New("must pass valid writer")
+	}
 	imageToTags := dedupRefToImage(refToImage)
-	var m Manifest
+
+	tw := w
+
+	// we only calculate the sizes and use a progressWriter if we were provided
+	// an option with a progress channel
+	if o != nil && o.updates != nil {
+		tw = &progressWriter{
+			w:       w,
+			updates: o.updates,
+			size:    size,
+		}
+	}
+
+	tf := tar.NewWriter(tw)
+	defer tf.Close()
 
 	seenLayerDigests := make(map[string]struct{})
 
-	for img, tags := range imageToTags {
+	for img := range imageToTags {
 		// Write the config.
 		cfgName, err := img.ConfigName()
 		if err != nil {
-			return nil, err
+			return err
 		}
-		if tf != nil {
-			cfgBlob, err := img.RawConfigFile()
+		cfgBlob, err := img.RawConfigFile()
+		if err != nil {
+			return err
+		}
+		if err := writeTarEntry(tf, cfgName.String(), bytes.NewReader(cfgBlob), int64(len(cfgBlob))); err != nil {
+			return err
+		}
+
+		// Write the layers.
+		layers, err := img.Layers()
+		if err != nil {
+			return err
+		}
+		layerFiles := make([]string, len(layers))
+		for i, l := range layers {
+			d, err := l.Digest()
 			if err != nil {
-				return nil, err
+				return err
 			}
-			if err := writeTarEntry(tf, cfgName.String(), bytes.NewReader(cfgBlob), int64(len(cfgBlob))); err != nil {
-				return nil, err
+			// Munge the file name to appease ancient technology.
+			//
+			// tar assumes anything with a colon is a remote tape drive:
+			// https://www.gnu.org/software/tar/manual/html_section/tar_45.html
+			// Drop the algorithm prefix, e.g. "sha256:"
+			hex := d.Hex
+
+			// gunzip expects certain file extensions:
+			// https://www.gnu.org/software/gzip/manual/html_node/Overview.html
+			layerFiles[i] = fmt.Sprintf("%s.tar.gz", hex)
+
+			if _, ok := seenLayerDigests[hex]; ok {
+				continue
 			}
+			seenLayerDigests[hex] = struct{}{}
+
+			r, err := l.Compressed()
+			if err != nil {
+				return err
+			}
+			blobSize, err := l.Size()
+			if err != nil {
+				return err
+			}
+
+			if err := writeTarEntry(tf, layerFiles[i], r, blobSize); err != nil {
+				return err
+			}
+		}
+	}
+	if err := writeTarEntry(tf, "manifest.json", bytes.NewReader(m), int64(len(m))); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+// calculateManifest calculates the manifest and optionally the size of the tar file
+func calculateManifest(refToImage map[name.Reference]v1.Image) (m Manifest, err error) {
+	imageToTags := dedupRefToImage(refToImage)
+
+	for img, tags := range imageToTags {
+		cfgName, err := img.ConfigName()
+		if err != nil {
+			return nil, err
 		}
 
 		// Store foreign layer info.
@@ -152,11 +236,6 @@ func processImages(refToImage map[name.Reference]v1.Image, tf *tar.Writer) (Mani
 			// https://www.gnu.org/software/gzip/manual/html_node/Overview.html
 			layerFiles[i] = fmt.Sprintf("%s.tar.gz", hex)
 
-			if _, ok := seenLayerDigests[hex]; ok {
-				continue
-			}
-			seenLayerDigests[hex] = struct{}{}
-
 			// Add to LayerSources if it's a foreign layer.
 			desc, err := partial.BlobDescriptor(img, d)
 			if err != nil {
@@ -168,21 +247,6 @@ func processImages(refToImage map[name.Reference]v1.Image, tf *tar.Writer) (Mani
 					return nil, err
 				}
 				layerSources[diffid] = *desc
-			}
-
-			if tf != nil {
-				r, err := l.Compressed()
-				if err != nil {
-					return nil, err
-				}
-				blobSize, err := l.Size()
-				if err != nil {
-					return nil, err
-				}
-
-				if err := writeTarEntry(tf, layerFiles[i], r, blobSize); err != nil {
-					return nil, err
-				}
 			}
 		}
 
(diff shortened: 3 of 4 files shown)


deitch:
Also, I know that this just sends updates after each layer, rather than periodically, which won't be great for very large layers. I still prefer to get the basic structure right and in, and then we can iterate on and improve it with a separate go routine doing periodic updates, perhaps by reading the http client or some other method.

pkg/v1/tarball/write.go
```diff
@@ -29,83 +29,112 @@ import (
...
+	var sendUpdate = func() {
+		if updates != nil {
+			updates <- v1.Update{
+				Total:    size,
+				Complete: complete,
+			}
+		}
+	}
+
+	for img, name := range imageToTags {
```
jonjohnsonjr: I didn't think about this before, but we need to keep in mind how to avoid breaking streaming images. This code assumes it's cheap to calculate an image's `Manifest()`, but that's not always the case. We might need to rethink this design a bit so that it's possible for us to send updates cheaply. E.g. maybe the total size is something callers should calculate themselves?

pkg/v1/tarball/write.go
```diff
@@ -29,83 +29,112 @@ import (
...
+	return MultiRefWrite(refToImage, w, updates)
 }
 
 // MultiRefWrite writes the contents of each image to the provided reader, in the compressed format.
 // The contents are written in the following format:
 // One manifest.json file at the top level containing information about several images.
 // One file for each layer, named after the layer's SHA.
 // One file for the config blob, named after its SHA.
-func MultiRefWrite(refToImage map[name.Reference]v1.Image, w io.Writer) error {
+func MultiRefWrite(refToImage map[name.Reference]v1.Image, w io.Writer, updates chan<- v1.Update) error {
```
jonjohnsonjr: This is a breaking change for callers, which I'd like to avoid if possible. We've been using [functional options](https://dave.cheney.net/2014/10/17/functional-options-for-friendly-apis) for this, so I'd probably change this to something like:

```
func MultiRefWrite(refToImage map[name.Reference]v1.Image, w io.Writer, opts... Option) error {
```

and add a `WithProgress` option (or something, names are hard).

You can see how we've done this with remote: https://godoc.org/github.com/google/go-containerregistry/pkg/v1/remote#Option

pkg/v1/tarball/write.go
deitch: > assumes it's cheap to calculate an image's Manifest()

Is it not? By the time someone gets to `MultiRefWrite()`, have they not already got the manifest? Doesn't the map in the parameters to the func of `refToImage map[name.Reference]v1.Image` already contain the actual images, which contain the manifest? 

Or do we have cases where calling `img.Manifest()` actually will pull at that moment? Either way, manifest certainly is tiny cost compared with the layers themselves, and will have to be called anyways.

> maybe the total size is something callers should calculate themselves?

I would be hesitant about that. One of the beauties of this library is that it makes it so much easier to "just get images" (and manipulate them, to some degree). Making them dig in to calculate would be harder.

pkg/v1/tarball/write.go
deitch: > I'd probably change this to something like

I am ok with variadic options, makes lots of sense. Not only does it make it non-breaking, but makes it easy to expand going forward.

> and add a WithProgress option 

I am not overly enamoured of functional options (although I give credit to Dave Cheney; he really laid out the challenges and the solutions nicely in that blog post). I get what they are trying to solve, and do so quite well, but they can lead to additional complexity.

* It is harder to visualize and reason about functions as options that just straight parameters or config structs, and I really believe builder complexity is the most expensive cost
* It requires the passed object to expose a lot of members that might remain private (see Cheney's examples where he has to modify `srv.timeout` or `srv.listener`), or expose a potential slew of new methods on the object (e.g. `srv.SetTimeout()` or `srv.SetListener()`; they also might be called later, when it makes no sense (a point he raises early on the post). That is what config opts are good for.

That having been said (i.e. discussing it, maybe I will get some new insights from someone smart), if you already are using them, we can just adopt it here. I will modify, let us see.

deitch: I will say one thing; it makes the tests unchanged, which is nice.

deitch: I updated it to use variadic function opts. Take a look now and see.

pkg/v1/tarball/write.go
jonjohnsonjr: > Is it not? By the time someone gets to MultiRefWrite(), have they not already got the manifest?

Not always: https://github.com/google/go-containerregistry/tree/master/pkg/v1/stream

Basically, with stream.Layer and mutating images, the manifest might be lazily calculated, which depends on the streaming layer being consumed already. I'm not sure if this should be blocking, though.

> I would be hesitant about that. One of the beauties of this library is that it makes it so much easier to "just get images" (and manipulate them, to some degree). Making them dig in to calculate would be harder.

Seems reasonable -- another way to address my concern would be to only calculate these values if the caller passes in WithProgress so that we avoid paying the cost if we can.

pkg/v1/tarball/write.go
```diff
@@ -161,6 +200,10 @@ func MultiRefWrite(refToImage map[name.Reference]v1.Image, w io.Writer) error {
 			if err := writeTarEntry(tf, layerFiles[i], r, blobSize); err != nil {
 				return err
 			}
+
+			// at this point, we have the config file, update the info
+			complete += blobSize
```
jonjohnsonjr: Rather than just sending updates at the end of each blob write, we could add a wrapper to the `tar.Writer` that counts bytes as we write them? Then the updates would have much higher resolution.

pkg/v1/tarball/write.go
jonjohnsonjr: > I am not overly enamoured of functional options

Yeah, I have mixed feelings about them too :/ we might overuse them, but they're nice to avoid breaking changes!

pkg/v1/tarball/write.go
deitch: > another way to address my concern would be to only calculate these values if the caller passes in WithProgress

And yet another argument in favour of your functional options. :-)

pkg/v1/tarball/write.go
deitch: For sure. In the project I am working on to consume this, someone did precisely that. I will update the PR to provide that.

pkg/v1/tarball/write.go
deitch: Take a look at the latest commit. I think this addresses it.

pkg/v1/tarball/write.go
deitch: Done. Take a look and let me know what you think.

pkg/crane/pull.go
```diff
@@ -63,7 +63,8 @@ func Save(img v1.Image, src, path string) error {
 		tag = d.Repository.Tag(iWasADigestTag)
 	}
 
-	return tarball.WriteToFile(path, tag, img)
+	// no progress channel (for now)
+	return tarball.WriteToFile(path, tag, img, nil)
```
jonjohnsonjr: You can just drop the final nil here.

pkg/registry/compatibility_test.go
```diff
@@ -57,7 +57,7 @@ func TestPushAndPullContainer(t *testing.T) {
 	}
 
 	b := &bytes.Buffer{}
-	if err := tarball.Write(ref, ri, b); err != nil {
+	if err := tarball.Write(ref, ri, b, nil); err != nil {
```
jonjohnsonjr: You can just drop the final nil here.

pkg/v1/daemon/write.go
```diff
@@ -62,7 +62,7 @@ func Write(tag name.Tag, img v1.Image) (string, error) {
 
 	pr, pw := io.Pipe()
 	go func() {
-		pw.CloseWithError(tarball.Write(tag, img, pw))
+		pw.CloseWithError(tarball.Write(tag, img, pw, nil))
```
jonjohnsonjr: You can just drop the final nil here.

pkg/v1/tarball/write.go
```diff
@@ -29,80 +30,115 @@ import (
...
 		if err != nil {
 			return err
 		}
+
 		cfgBlob, err := img.RawConfigFile()
 		if err != nil {
 			return err
 		}
-		if err := writeTarEntry(tf, cfgName.String(), bytes.NewReader(cfgBlob), int64(len(cfgBlob))); err != nil {
+		cfgSize := int64(len(cfgBlob))
```
jonjohnsonjr: I'd undo this just to have a cleaner diff

pkg/v1/progress.go
```diff
@@ -0,0 +1,7 @@
+package v1
```
jonjohnsonjr: Can you add a copyright header here?

pkg/v1/tarball/write.go
```diff
@@ -213,3 +249,38 @@ func writeTarEntry(tf *tar.Writer, path string, r io.Reader, size int64) error {
 	_, err := io.Copy(tf, r)
 	return err
 }
+
+type WriteOption func(*writeOptions) error
+type writeOptions struct {
+	updates chan<- v1.Update
+}
+
+func WithProgress(updates chan<- v1.Update) WriteOption {
```
jonjohnsonjr: nit: add godoc

pkg/v1/tarball/write.go
```diff
@@ -213,3 +249,38 @@ func writeTarEntry(tf *tar.Writer, path string, r io.Reader, size int64) error {
 	_, err := io.Copy(tf, r)
 	return err
 }
+
+type WriteOption func(*writeOptions) error
```
jonjohnsonjr: nit: add godoc

pkg/v1/tarball/write.go
```diff
@@ -213,3 +249,38 @@ func writeTarEntry(tf *tar.Writer, path string, r io.Reader, size int64) error {
...
+
+func (pw progressWriter) Write(p []byte) (int, error) {
+	n, err := pw.w.Write(p)
+	if err != nil {
+		return n, err
+	}
+
+	atomic.AddInt64(&pw.complete, int64(n))
+
+	pw.updates <- v1.Update{
```
jonjohnsonjr: I'm interested to see how this affects performance. It shouldn't block this PR but we should definitely look into it before we add this to crane

pkg/v1/progress.go
```diff
@@ -0,0 +1,7 @@
...
+
+// Update representation of download update
```
jonjohnsonjr: This comment could be a little more helpful -- calling out tarball.WithProgress here would be great.

I think "download update" is a bit too specific, since this could really represent any kind of progress update, independent of the image source.

pkg/crane/pull.go
jonjohnsonjr: In fact, this nil here causes a panic in travis

deitch: Yes, good point. I will update it.

> this nil here causes a panic 

So, technically, that is a "nil panic", which means "no panic", so there isn't a panic? :-)

pkg/registry/compatibility_test.go
deitch: Done.

pkg/v1/tarball/write.go
deitch: No problem.

pkg/v1/daemon/write.go
deitch: Done

pkg/v1/progress.go
deitch: Sure. Feels strange to add a google copyright to it when I wrote it, but doesn't matter all that much.

pkg/v1/tarball/write.go
deitch: Dang! How did I miss that? I must have forgotten to lint it. Adding now.

pkg/v1/tarball/write.go
deitch: Done

pkg/v1/tarball/write.go
deitch: I would imagine it wouldn't change much of anything; downloading is far slower than sending a message on a channel. But that, of course, is theory, that should be tested (as you said).

pkg/v1/progress.go
deitch: Sure, we can make this godoc a little more user-friendly and rational.

deitch: Done

deitch:
I think I addressed everything you raised.

jonjohnsonjr:
This looks great to me, I'd just like to see a test to cover this path. If you could flesh out your [godoc example](https://github.com/google/go-containerregistry/pull/675/files#diff-a5d4705c8e8955883203e33739b0d31dR21-R32) and pull it out into a [testable example](https://blog.golang.org/examples), you could kill two birds with one stone. I think it just has to have this signature:

```
package tarball_test

func ExampleWithProgress(t *testing.T) {
}
```

And then we can link to the tarball godoc from the v1.Progress godoc.

deitch:
@jonjohnsonjr wrote [here](https://github.com/google/go-containerregistry/pull/675#issuecomment-594139457):

> I'd just like to see a test to cover this path. If you could flesh out your godoc example and pull it out into a testable example, you could kill two birds with one stone.

Like the latest? FYI, I added an extra `fmt.Printf()`, even though it makes the example test fail, because it was failing anyways. Interestingly, I think the progress writer is giving incorrect information, so good to catch it.

deitch:
OK, almost got it. I cheated a bit on the example test. Rather than setting up a goroutine and synchronizing, I just made the channel buffer size 200, which is plenty big for the test.

Last remaining issue is sizes. Here are the actual sizes in bytes, and therefore why it is failing:

* size of config + layers in manifest, which we calculate [here](https://github.com/google/go-containerregistry/pull/675/files#diff-ad81893e03d553fe2a423b9b5050c232R107-R117): **2788474**. This we calculate correctly.
* size of added manifest file that we add when writing: **232**, for total bytes inside tar of **2788706**
* size of tarfile when written: **2791936**. The difference is **3230** bytes, all of which would be tar headers and potential padding

This leaves the remaining question of how to handle this. There are two issues: tar header and padding, manifest. For the manifest, we already calculate it in `MultiWriteRef()` when writing the tarfile; I will move it up a bit and leverage it for calculating size. I suggested exposing it [here](https://github.com/google/go-containerregistry/issues/680), but that is a separate question.

For the tar headers and padding, any ideas?

deitch:
Hmm, it is a little more surgery in `MultiWriteRef()` than I thought. Not hard, but still. We would need to take [this loop](https://github.com/google/go-containerregistry/blob/6950943e71a1805ce590c2b5ee50d761c09acc02/pkg/v1/tarball/write.go#L95-L173) and do two passes:

1. Loop through each image and each layer, extracting metadata and saving it for the size and manifest
2. Loop through each again and pull the data and write to the tarfile

It isn't a big deal, but would like it if you are onboard with it before starting.

jonjohnsonjr:
> size of tarfile when written: 2791936. The difference is 3230 bytes, all of which would be tar headers and potential padding

These numbers don't seem right to me -- the test is creating a random image with 8 layers at 256 bytes each, how is this getting inflated to almost 3 MB? When I run this, I get `512/5331`.

It seems like the progressWriter isn't updating `Complete` effectively?

> This leaves the remaining question of how to handle this. There are two issues: tar header and padding, manifest.

I'm not too thrilled about the complexity of implementing this...

It seems somewhat reasonable to expose something that calculates the `manifest.json` file (i.e. https://github.com/google/go-containerregistry/issues/680), but trying to account for the padding and tar headers seems like trouble. Just accounting for the layers/config seems good enough for a progress bar, given that the manifest will be so small.

pkg/v1/tarball/write_test.go
```diff
@@ -373,3 +374,44 @@ func TestWriteSharedLayers(t *testing.T) {
 		}
 	}
 }
+
+func ExampleWriteToFileWithProgress() {
```
jonjohnsonjr: I think this function has to be named exactly `ExampleWithProgress` in order to show up under `WithProgress` in godoc.

deitch: I was wondering about that....

deitch: Yup. Now the linter stopped complying.

deitch:
> These numbers don't seem right to me -- the test is creating a random image with 8 layers at 256 bytes each, how is this getting inflated to almost 3 MB? When I run this, I get 512/5331

I tested with alpine for something larger. I will go back to your random image when this all works correctly. :-)

deitch:
> I'm not too thrilled about the complexity of implementing this...

Join the club, as I am the one doing it. :-)

> trying to account for the padding and tar headers seems like trouble

Agreed. My focus on this is iterative. I will open a PR for #680 just to see how it works, then this can be good enough.

deitch:
I opened #685 as an idea for getting the manifest. If that works, or something like it, it will be pretty easy to do 2 passes inside `MultiRefWrite()`:

1. Get the manifest, calculate its size (only if `o.updates != nil`, of course)
2. Do the usual write

deitch:
I ended up doing some pretty deep refactoring on `processImages()` - broke it out into 3 parts:

1. `calculateManifest()`
1. `calculateTarballSize()`
1. `writeImagesToTar()`

There remains a bit of duplication between them, nut not bad.

Can you take a look at it, @jonjohnsonjr @ImJasonH ? Particularly:

* the refactoring
* the test for `ExampleWithProgress()`. That one uses the image `docker.io/library/alpine:3.10`. I am fine working with any image, as long as I can predict reliably the exact size of the file in advance. I calculated it manually here.

deitch:
Hi again @jonjohnsonjr and @ImJasonH ; can we get a look at this? I really would like to finish it up.

deitch:
One thing to think about is [in the example here](https://github.com/google/go-containerregistry/pull/675/files#diff-5672be5247d8ec2431d25da290b57d35R532-R539). We use the provided channel to send updates, but create a separate channel to pass the error back. Because we are using channels, and hence separate goroutines, we cannot easily handle the `err := tarball.WriteToFile()` result. 

The question, is the example good enough, or should we also have an error property on the `update` struct itself, so we can use it to pass back errors as well?

ImJasonH:
The Travis failure is a lint check:

```
pkg/v1/tarball/write_test.go:532:30: should use make(chan error) instead (S1019)
```

re: `errchan`, I think that's going to be somewhat unavoidable in cases where you want to be able to handle download progress too. Perhaps we could change tthe semantics so that if a download progress chan is provided, then `WriteToFile` _never_ returns err, and the download progress chan is responsible for also signaling any errors. I worry that's a bit easier for users to get wrong though. And this change doesn't preclude us making that semantic change in the future, so let's stick with this for now I think.

deitch:
> re: errchan, I think that's going to be somewhat unavoidable in cases where you want to be able to handle download progress too

I think so too.

> Perhaps we could change tthe semantics so that if a download progress chan is provided, then WriteToFile never returns err, and the download progress chan 

I definitely agree, and would leave it as it is, returning the `error`. It is ok _also_ to send the error via the channel, if it supports it, but I wouldn't remove it from the return.

> let's stick with this for now I think

Sounds good. I will update to fix the linter and push.

deitch:
Pushed.

pkg/v1/tarball/write.go
```diff
@@ -31,115 +32,198 @@ import (
...
+}
+
+// calculateManifest calculates the manifest and optionally the size of the tar file
+func calculateManifest(refToImage map[name.Reference]v1.Image) (m Manifest, err error) {
+	imageToTags := dedupRefToImage(refToImage)
+
+	for img, tags := range imageToTags {
+		cfgName, err := img.ConfigName()
+		if err != nil {
+			return m, err
```
ImJasonH, line 211: Sorry I didn't catch this earlier: why is this returning `m, err` now? It's pretty unidiomatic for Go to return something with `err`. AFAICT the only place this is called in this PR doesn't consider the `m` return value anyway, and callers of exported `ComputeManifest` might be surprised by the behavior change.

pkg/v1/tarball/write_test.go
```diff
@@ -484,3 +485,70 @@ func getLayersFilenames(hashes []string) []string {
...
+			errchan <- err
+		}
+		errchan <- nil
+	}()
+	var update v1.Update
+forloop:
+	for {
+		select {
+		case update = <-c:
+			//fmt.Printf("receive update: %#v\n", update)
```
ImJasonH, line 545: Can we uncomment these?

pkg/v1/tarball/write_test.go
```diff
@@ -484,3 +485,70 @@ func getLayersFilenames(hashes []string) []string {
...
+		case err = <-errchan:
+			//fmt.Printf("receive error message: %v\n", err)
+			break forloop
```
ImJasonH, line 548: If this just prints the final complete / total, then this can be a `return` and we can avoid the break label.

pkg/v1/tarball/write_test.go
deitch: Sure, but then I have to print to `os.Stderr`, or it messed up the results.

pkg/v1/tarball/write_test.go
deitch: Nice. Much cleaner. Updating.

pkg/v1/tarball/write.go
deitch: > It's pretty unidiomatic for Go to return something with err

I must have misunderstood you. The signature is `(m Manifest, err error)`, so you always have to return two values.

> callers of exported ComputeManifest might be surprised by the behavior change.

I still don't get it. What behavioural change? The return signature is unchanged.

deitch:
I still am using `docker.io/library/alpine:3.10` at a specific hash for the example. Two points about it:

* I have no idea why it is giving an erroneous count for downloaded. I tested on Mac and on Linux, both are accurate, but we cannot let this fail on future CI runs. I need to figure it out
* are we ok with using that image in the example?

deitch:
> I have no idea why it is giving an erroneous count for downloaded

tested on linux alpine (`go test -v`) and on Debian buster (`go test -v -race`), both behave exactly as expected.

deitch:
Pushed again to kick off CI. In previous run, the matrix was:

* go1.12+modules=off: success
* go1.12+modules=on: success
* go1.13+modules=off: success
* go1.13+modules=on: fail

deitch:
Ha! It passed!

pkg/v1/tarball/write.go
ImJasonH: The return signature is unchanged, but the behavior is different now -- it used to return `nil, err`, and now it returns a non-nil `m`. This might surprise some callers. If there's a reason to return a potentially partial manifest in the case of an error, we should take care to document it better. Otherwise most callers will likely expect that the return is either-or.

ImJasonH:
> I still am using `docker.io/library/alpine:3.10` at a specific hash for the example. Two points about it:
> 
> * are we ok with using that image in the example?

I'm fine with using that image, by hash. If anything we could use a smaller image to make the test slightly faster? I don't think it practically matters though.

pkg/v1/tarball/write.go
deitch: Oh is that what you meant? I completely misunderstood. Fixing it right now.

deitch: Done. Fixed all the other places it does it in there as well.

deitch:
> I'm fine with using that image, by hash. If anything we could use a smaller image to make the test slightly faster? I don't think it practically matters though.

Then let's leave it. If it starts to bother us afterwards, we can push our own image.

deitch:
That should resolve it.

deitch:
Thanks @ImJasonH .

And then I realized (referring back to `errchan` [here](https://github.com/google/go-containerregistry/pull/675#issuecomment-620033742) that I actually _did_ include an `Error` on the `type Update`, see [here](https://github.com/google/go-containerregistry/pull/675/files#diff-a5d4705c8e8955883203e33739b0d31dR20-R23). It didn't change anything, but it is there.

Should we add a quick PR to eliminate the need for a separate `errchan` and have `MultiRefWrite()` also send its error (or `nil`) when complete to there and send an update? It certainly cannot hurt, and it simplifies things.

Thoughts?

ImJasonH:
That sounds good to me, sorry for missing that

deitch:
Coming in 5 mins...

deitch:
Done. See #716

PR #684, opened 2020-03-04
Preserve Descriptor in remote where possible

### LINKED ISSUE #683: crane rebase and mutate.Rebase fail with foreign layer in new_base image
It appears `crane rebase` fails with an image whose base layers are foreign layers. Below is hopefully a simple example using Docker Hub `hello-world:nanoserver-1809` and its `nanoserver-1809` base image (whose registry happens to be `mcr.microsoft.com`) as both new/old base.

Failure reproduction:
```
## Image-ref for `hello-world:nanoserver-1809`
app_image=index.docker.io/library/hello-world@sha256:63c287625c2b0b72900e562de73c0e381472a83b1b39217aef3856cd398eca0b

## Image-ref for `mcr.microsoft.com/windows/nanoserver:1809` (base image for above)
app_base=mcr.microsoft.com/windows/nanoserver@sha256:06281772b6a561411d4b338820d94ab1028fdeb076c85350bbc01e80c4bfa2b4

## Confirm app image layers 
crane manifest $app_image | jq ".layers[] | .digest"
## Output:
# "sha256:a35da61c356213336e646756218539950461ff2bf096badf307a23add6e70053" # <- same base layer
# "sha256:be21f08f670160cbae227e3053205b91d6bfa3de750b90c7e00bd2c511ccb63a" # COPY layer
# "sha256:10d1439be4eb8819987ec2e9c140d44d74d6b42a823d57fe1953bd99948e1bc0" # ENTRYPOINT layer

## Confirm base image layers
crane manifest $app_base | jq ".layers[] | .digest"
## Output:
# "sha256:a35da61c356213336e646756218539950461ff2bf096badf307a23add6e70053" # <- same base layer

crane rebase \
--original $app_image \
--old_base $app_base \
--new_base $app_base \
--rebased localhost:5000/rebased  # <- needs a writable registry
## Output
# 2020/03/03 16:17:35 existing blob: sha256:10d1439be4eb8819987ec2e9c140d44d74d6b42a823d57fe1953bd99948e1bc0
# 2020/03/03 16:17:35 existing blob: sha256:be21f08f670160cbae227e3053205b91d6bfa3de750b90c7e00bd2c511ccb63a
# 2020/03/03 16:17:35 existing blob: sha256:e488c333e5cb990d61a1be8d4446050befe3d7a8ffc69220eb030694d70e7beb
# 2020/03/03 16:17:35 pushing localhost:5000/rebased: PUT http://localhost:5000/v2/rebased/manifests/latest: multiple errors returned: UNKNOWN: unknown error; UNKNOWN: unknown error; map[]; MANIFEST_BLOB_UNKNOWN: blob unknown to registry; sha256:a35da61c356213336e646756218539950461ff2bf096badf307a23add6e70053 # <- failed on foreign base layer
```

Readable error:
> `2020/03/03 16:17:35 pushing localhost:5000/rebased: PUT http://localhost:5000/v2/rebased/manifests/latest: multiple errors returned: UNKNOWN: unknown error; UNKNOWN: unknown error; map[]; MANIFEST_BLOB_UNKNOWN: blob unknown to registry; sha256:a35da61c356213336e646756218539950461ff2bf096badf307a23add6e70053`

## Additional info
I don't think the separate registries are at fault since I reproduced this all on a local registry (with differing bases and many more steps and output) in this [gist](https://gist.github.com/micahyoung/c515a3c45b6a09773be08530b0ca1bed).

I chose the issue example since it is pretty copy-pastable and doesn't require a Docker Desktop WCOW daemon to reproduce, but the Docker file `hello-world:nanoserver-1809` is built from is [here](https://github.com/docker-library/hello-world/blob/9c93e37114a7fe99b5fc0d776e0b8dff99cbbb75/amd64/hello-world/nanoserver-1809/Dockerfile). 

I couldn't find any other non-Windows foreign layer images in the wild. But if needed, there is a Windows nanoserver image with foreign layers on gcr at`gcr.io/kubernetes-e2e-test-images/windows-nanoserver:v1` [origin here](https://github.com/kubernetes/kubernetes/issues/74134#issuecomment-467666811).

jonjohnsonjr:
Fixes #683 

If we resolve an index to an image, we can potentially lose a lot of
information, e.g. annotations or anything in the platform that's not in
the config, which only has os/arch.

If we return a layer from an image, we can look up the original
descriptor from that image to return as well, which might include
annotations or urls. Previously, the Mountable wrapping made this not
work.

### DIFF
--- a/pkg/v1/remote/image.go
+++ b/pkg/v1/remote/image.go
@@ -29,6 +29,11 @@ import (
 	"github.com/google/go-containerregistry/pkg/v1/v1util"
 )
 
+var acceptableImageMediaTypes = []types.MediaType{
+	types.DockerManifestSchema2,
+	types.OCIManifestSchema1,
+}
+
 // remoteImage accesses an image from a remote registry
 type remoteImage struct {
 	fetcher
@@ -37,6 +42,7 @@ type remoteImage struct {
 	configLock   sync.Mutex // Protects config
 	config       []byte
 	mediaType    types.MediaType
+	descriptor   *v1.Descriptor
 }
 
 var _ partial.CompressedImageCore = (*remoteImage)(nil)
@@ -68,15 +74,14 @@ func (r *remoteImage) RawManifest() ([]byte, error) {
 	// NOTE(jonjohnsonjr): We should never get here because the public entrypoints
 	// do type-checking via remote.Descriptor. I've left this here for tests that
 	// directly instantiate a remoteImage.
-	acceptable := []types.MediaType{
-		types.DockerManifestSchema2,
-		types.OCIManifestSchema1,
-	}
-	manifest, desc, err := r.fetchManifest(r.Ref, acceptable)
+	manifest, desc, err := r.fetchManifest(r.Ref, acceptableImageMediaTypes)
 	if err != nil {
 		return nil, err
 	}
 
+	if r.descriptor == nil {
+		r.descriptor = desc
+	}
 	r.mediaType = desc.MediaType
 	r.manifest = manifest
 	return r.manifest, nil
@@ -105,6 +110,15 @@ func (r *remoteImage) RawConfigFile() ([]byte, error) {
 		return nil, err
 	}
 	return r.config, nil
+}
+
+// Descriptor retains the original descriptor from an index manifest.
+// See partial.Descriptor.
+func (r *remoteImage) Descriptor() (*v1.Descriptor, error) {
+	// kind of a hack, but RawManifest does appropriate locking/memoization
+	// and makes sure r.descriptor is populated.
+	_, err := r.RawManifest()
+	return r.descriptor, err
 }
 
 // remoteImageLayer implements partial.CompressedLayer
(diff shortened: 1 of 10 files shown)


pkg/v1/remote/image.go
```diff
@@ -107,7 +109,19 @@ func (r *remoteImage) RawConfigFile() ([]byte, error) {
 	return r.config, nil
 }
 
+func (r *remoteImage) Descriptor() (*v1.Descriptor, error) {
+	if r.descriptor != nil {
+		return r.descriptor, nil
+	}
+
+	// NOTE(jonjohnsonjr): We should never get here because the public entrypoints
```
ImJasonH: Why not change the tests that directly deal with remoteImage?

jonjohnsonjr: I'm kinda torn on this because my comment just so happens to be true, and there's nothing in the type system to enforce this invariant :/

jonjohnsonjr: Alright I came up with something gross but slightly better

Issue #680, opened 2020-02-26
expose manifest for file before writing?


deitch:
When I download an image, once I have the manifest, I can calculate the size of the final download by looking at the layers and the config; the manifest has sizes of both of those.

However, the final `v1.tarball` format includes those _and_ the `manifest.json`, which is written to to the file [here](https://github.com/google/go-containerregistry/blob/6950943e71a1805ce590c2b5ee50d761c09acc02/pkg/v1/tarball/write.go#L166-L172). This makes it very difficult to calculate the file size in advance, which I might need for various types of validation, size calculations, etc. Sure, it is small, but it throws things off.

Would we be willing to expose this manifest, since it is easily to pre-calculate once you have the manifest in hand? 

Essentially, it would be taking the code and making it an exposed func , that is consumed by that same line of code. I already do it outside, but would not like to be so dependent on the internals.

Ideally, I would love a `v1tarball.CalculateSize(v1.Image) int64` which would do this, but maybe I am asking for too much? 

Happy to open a PR for either.

ImJasonH:
`image.RawManifest()` returns the raw JSON bytes of the manifest, you can `len` that and add it to the sizes of all the layers to get the full size. If that's all you need that seems simple enough to write in your own code, unless I'm misunderstanding the issue.

deitch:
That isn’t quite it. Look at the lines I added, that is another manifest that is added by the lib when the tar file is created. That one is not in the raw manifest. 

I stumbled across this because I do exactly what you suggested and was off by a few bytes (to give credit, one of my colleagues found it). 

Ideally, there would just be a size func that gives the whole expected size, but enough if I can get the size of that internal manifest to add to the calc you listed above.

ImJasonH:
Oh I see, the issue is the tar header overhead. Is the overhead a constant number or are you seeing it fluctuate? Could you add some buffer to account for the overhead?

I'm reluctant to add this to the library if it's fairly straightforward to calculate/approximate.

deitch:
Fluctuates. It depends on quite a bit. I tried to calculate it by copying the code and trying to replicate it (awful violation of DRY and coupling, guaranteed to break with first library change, I know), but even then missed something. 

Why the hesitancy to add something that doesn’t affect existing APIs?

jonjohnsonjr:
[...]

PR #685, opened 2020-03-05
extract v1.tarball manifest creation from writing, make open

deitch:
Addresses #680 . It doesn't change how anything works, but makes it possible to get the tarball `manifest.json` without actually writing the tarball. 

I tried to extract it as a separate func, but there was so much duplication to what is in `MultiRefWrite()` that it just made sense to pull the code out of there into a separate func that always returns the manifest. If a `*tar.Writer` is passed, then it also writes via the writer.

I also provided three separate convenience functions to parallel the three separate `Write()`/`MultiWrite()`/`MultiRefWrite()`

This is a work in progress, so looking for thoughts.

### DIFF
--- a/pkg/v1/tarball/write.go
+++ b/pkg/v1/tarball/write.go
@@ -88,6 +88,22 @@ func MultiRefWrite(refToImage map[name.Reference]v1.Image, w io.Writer) error {
 	tf := tar.NewWriter(w)
 	defer tf.Close()
 
+	m, err := processImages(refToImage, tf)
+	if err != nil {
+		return err
+	}
+
+	mBytes, err := json.Marshal(m)
+	if err != nil {
+		return err
+	}
+	return writeTarEntry(tf, "manifest.json", bytes.NewReader(mBytes), int64(len(mBytes)))
+}
+
+// processImages writes the images to the provider tar.Writer. Returns the manifest
+// for the written images as Manifest.`If the tar.Writer is nil, will just return
+// the Manifest.
+func processImages(refToImage map[name.Reference]v1.Image, tf *tar.Writer) (Manifest, error) {
 	imageToTags := dedupRefToImage(refToImage)
 	var m Manifest
 
@@ -96,14 +112,16 @@ func MultiRefWrite(refToImage map[name.Reference]v1.Image, w io.Writer) error {
 		// Write the config.
 		cfgName, err := img.ConfigName()
 		if err != nil {
-			return err
+			return nil, err
 		}
-		cfgBlob, err := img.RawConfigFile()
-		if err != nil {
-			return err
-		}
-		if err := writeTarEntry(tf, cfgName.String(), bytes.NewReader(cfgBlob), int64(len(cfgBlob))); err != nil {
-			return err
+		if tf != nil {
+			cfgBlob, err := img.RawConfigFile()
+			if err != nil {
+				return nil, err
+			}
+			if err := writeTarEntry(tf, cfgName.String(), bytes.NewReader(cfgBlob), int64(len(cfgBlob))); err != nil {
+				return nil, err
+			}
 		}
 
 		// Store foreign layer info.
@@ -112,13 +130,13 @@ func MultiRefWrite(refToImage map[name.Reference]v1.Image, w io.Writer) error {
 		// Write the layers.
 		layers, err := img.Layers()
 		if err != nil {
-			return err
+			return nil, err
 		}
 		layerFiles := make([]string, len(layers))
 		for i, l := range layers {
 			d, err := l.Digest()
 			if err != nil {
-				return err
+				return nil, err
 			}
 			// Munge the file name to appease ancient technology.
 			//
@@ -139,27 +157,29 @@ func MultiRefWrite(refToImage map[name.Reference]v1.Image, w io.Writer) error {
 			// Add to LayerSources if it's a foreign layer.
 			desc, err := partial.BlobDescriptor(img, d)
 			if err != nil {
-				return err
+				return nil, err
 			}
 			if !desc.MediaType.IsDistributable() {
 				diffid, err := partial.BlobToDiffID(img, d)
 				if err != nil {
-					return err
+					return nil, err
 				}
 				layerSources[diffid] = *desc
 			}
 
-			r, err := l.Compressed()
-			if err != nil {
-				return err
-			}
-			blobSize, err := l.Size()
-			if err != nil {
-				return err
-			}
+			if tf != nil {
+				r, err := l.Compressed()
+				if err != nil {
+					return nil, err
+				}
+				blobSize, err := l.Size()
+				if err != nil {
+					return nil, err
+				}
 
-			if err := writeTarEntry(tf, layerFiles[i], r, blobSize); err != nil {
-				return err
+				if err := writeTarEntry(tf, layerFiles[i], r, blobSize); err != nil {
+					return nil, err
+				}
 			}
 		}
 
@@ -171,12 +191,7 @@ func MultiRefWrite(refToImage map[name.Reference]v1.Image, w io.Writer) error {
 			LayerSources: layerSources,
 		})
 	}
-
-	mBytes, err := json.Marshal(m)
-	if err != nil {
-		return err
-	}
-	return writeTarEntry(tf, "manifest.json", bytes.NewReader(mBytes), int64(len(mBytes)))
+	return m, nil
 }
 
 func dedupRefToImage(refToImage map[name.Reference]v1.Image) map[v1.Image][]string {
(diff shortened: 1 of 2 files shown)


deitch:
Thoughts on this @jonjohnsonjr ?

pkg/v1/tarball/write.go
```diff
@@ -171,45 +223,26 @@ func MultiRefWrite(refToImage map[name.Reference]v1.Image, w io.Writer) error {
...
-		Size:     size,
-		Name:     path,
-	}
-	if err := tf.WriteHeader(hdr); err != nil {
-		return err
-	}
-	_, err := io.Copy(tf, r)
-	return err
+// TarManifest is a wrapper to get the Manifest for a single image and tag
+func TarManifest(ref name.Reference, img v1.Image) (Manifest, error) {
```
jonjohnsonjr: I'd like to avoid adding too many of these entrypoints to avoid having to maintain them. I'd rename this to `ComputeManifest` (to avoid the stutter with `tarball.TarManifest`) and change it to use the signature from `MultiRefTarManifest` (which I think is generic enough to cover all use cases).

pkg/v1/tarball/write.go
```diff
@@ -88,6 +88,54 @@ func MultiRefWrite(refToImage map[name.Reference]v1.Image, w io.Writer) error {
...
+		Name:     path,
+	}
+	if err := tf.WriteHeader(hdr); err != nil {
+		return err
+	}
+	_, err := io.Copy(tf, r)
+	return err
+}
+
+func processImages(refToImage map[name.Reference]v1.Image, tf *tar.Writer) (Manifest, error) {
```
jonjohnsonjr: Can you add a comment here indicating that it just computes the Manifest if `tf` is nil (and why that's useful)?

deitch: Yup, good point. I shouldn't have missed that. :-(

pkg/v1/tarball/write.go
deitch: I have no argument with this; actually, I think it makes sense to rename them, as they really are just computing the manifest.

I am going to do a bit of rejiggering on the basis of this. Expect it shortly.

deitch: Changed the func names to match but haven't yet reduced the entrypoints. I put those in to parallel the `Write`/`MultiWrite`/`MultiRefWrite` and `WriteToFile`/`MultiWriteToFile`/`MultiRefWriteToFile`. I am fine if you want just `ComputeManifest`, but it will look a bit off given that the arguments are similar.

Totally your call; let me know.

deitch:
> I tried to extract it as a separate func, but there was so much duplication to what is in MultiRefWrite() that it just made sense to pull the code out of there into a separate func that always returns the manifest. If a *tar.Writer is passed, then it also writes via the writer.

Can you think of any sane way to do it? the "if tf == nil do not write" is kind of ugly. Writing to writer and calculating a manifest are 2 distinct activities. I tried, [here](https://gist.github.com/deitch/b15a6eeba736c400f132015ad591a619) is a gist with the file (not the greatest).

I really like how [this part](https://gist.github.com/deitch/b15a6eeba736c400f132015ad591a619#file-write-go-L88-L105) works, separating getting the manifest from writing the images, but the amount of duplication between [computeManifest](https://gist.github.com/deitch/b15a6eeba736c400f132015ad591a619#file-write-go-L165-L222) and [writeImages](https://gist.github.com/deitch/b15a6eeba736c400f132015ad591a619#file-write-go-L227-L280) is, shall we say, sub-optimal.

jonjohnsonjr:
> I really like how this part works, separating getting the manifest from writing the images, but the amount of duplication between computeManifest and writeImages is, shall we say, sub-optimal.

Agree.

> Can you think of any sane way to do it?

The only thing I could think of would be to add fields/methods to tarball.Descriptor,  but we'd still end up with a lot of duplication. I think what you have in this PR is pretty easy to follow, so I wouldn't worry about it.

pkg/v1/tarball/write.go
jonjohnsonjr: Yeah I think we should just reduce this down to one. The MultiWrite/MultiRefWrite thing is a historical mistake (I don't think we should ever have had the tag version), and calling ComputeManifest should be a niche enough thing that I don't think we need any sugar on top of it -- it's not too hard for callers to just wrap an `Image` in a map.

pkg/v1/tarball/write.go
```diff
@@ -88,6 +88,57 @@ func MultiRefWrite(refToImage map[name.Reference]v1.Image, w io.Writer) error {
 	tf := tar.NewWriter(w)
 	defer tf.Close()
 
+	m, err := processImages(refToImage, tf)
```
jonjohnsonjr: This diff ended up being really drastic :/ 

This is super pedantic, but if it's possible to rearrange the ordering of some of these functions to help git minimize the diff, it will make me a lot happier in the future when I `git blame` this file 😄

pkg/v1/tarball/write.go
deitch: Sure. Consider it done.

deitch:
> so I wouldn't worry about it.

OK

pkg/v1/tarball/write.go
deitch: No problem.

deitch: That should do it.

deitch:
Going to add a simple test or two to this, as you requested (and I wanted to do), then it will be good to go.

deitch:
And one test added, and reordered the funcs to make it easier to read the diff. Back at you @jonjohnsonjr

deitch:
Oops. I missed a linter check and didn't get the email warning of the failure. Just fixed it @jonjohnsonjr

deitch:
Hi @jonjohnsonjr , hope you are holding up well in these crazy times. Can you review this so we can get it in?

deitch:
Thanks @ImJasonH ! :-)

deitch:
I think I can make some good progress on #675 now. Will head back to it.

ImJasonH:
This seems to cause failing/flaky tests:

```
  write_test.go:458: mismatched manifests.
        Actual: [{"Config":"sha256:6d3ce0d6bbbd013f648d0b82b099982b643c0195326c0748d77b8ea2540f1e21","RepoTags":["gcr.io/baz/bat:latest"],"Layers":["a6b9e78e1c85dc9fd5092f80840460b82ac477a8f6b37f9d1f8238f6414d3ef3.tar.gz","28e7473d053100c4e7b673e02f172a2e67febb4758b5881bc7153eb3c4e7a7e6.tar.gz"]},{"Config":"sha256:a980bf1fd7ae6e2a8436287ad82e51f5be9d0075efdd36f95c27ebd05525ed57","RepoTags":["gcr.io/foo/bar:latest"],"Layers":["a6b9e78e1c85dc9fd5092f80840460b82ac477a8f6b37f9d1f8238f6414d3ef3.tar.gz"]}]
        Expected: [{"Config":"sha256:a980bf1fd7ae6e2a8436287ad82e51f5be9d0075efdd36f95c27ebd05525ed57","RepoTags":["gcr.io/foo/bar:latest"],"Layers":["a6b9e78e1c85dc9fd5092f80840460b82ac477a8f6b37f9d1f8238f6414d3ef3.tar.gz"]},{"Config":"sha256:6d3ce0d6bbbd013f648d0b82b099982b643c0195326c0748d77b8ea2540f1e21","RepoTags":["gcr.io/baz/bat:latest"],"Layers":["a6b9e78e1c85dc9fd5092f80840460b82ac477a8f6b37f9d1f8238f6414d3ef3.tar.gz","28e7473d053100c4e7b673e02f172a2e67febb4758b5881bc7153eb3c4e7a7e6.tar.gz"]}]
```

See https://travis-ci.org/github/google/go-containerregistry/jobs/674377815

Please fix this ASAP to unblock other PRs. Let me know if you need any help.

deitch:
> seems to cause failing/flaky tests

Don't say that! :-)

Funny, CI passed before we could merge it in. I will take a look. Thanks for the link.

deitch:
Got it. It is an ordering problem. It is not consistently ordering them. I will figure something out and post a new PR for it. Hang tight.

deitch:
@ImJasonH here is the problem. We have two items in the manifest in an array. Because we populate it with (simplifying):

```go
        var imageToTags map[v1.Image][]string
	for img, tags := range imageToTags {
            // lots of stuff
        }
```

the order of `for := range map` is not consistent, sometimes it matches, sometimes it does not. It doesn't affect the manifest in the tar file, anyways.

There are two ways to fix this:

* have the test accept the returns manifest as long as the length is the same and all items match up
* have the func in `write.go` return a manifest consistently ordered based on image name (or anything else; it doesn't matter as long as its s consistent).

Do you have a preference @ImJasonH ? I lean towards the latter, but will defer to maintainers here.

ImJasonH:
I prefer the latter too, because it means the manifests will be stable and reproducible.

deitch:
OK, shortly.

deitch:
See #711

PR #691, opened 2020-03-11
Fix GitLab CI example

Mayeu:
The command to login should be `crane auth login`.


Mayeu:
@googlebot I signed it!

PR #692, opened 2020-03-11
Add --oci option to image export

zmackie:
Allows export images as oci layouts


jonjohnsonjr:
Do you want _export_ to do this? That flattens the filesystem.

I added something like this to `crane pull` recently: https://github.com/google/go-containerregistry/pull/656

zmackie:
ah that difference between export and pull was a bit lost on me.

PR #693, opened 2020-03-12
Canonical should keep OS/ARCH

Vlatombe:
These fields should be considered in canonical image. Upstream of a kaniko PR (tbd)


pkg/v1/mutate/mutate.go
```diff
@@ -279,6 +279,8 @@ func Time(img v1.Image, t time.Time) (v1.Image, error) {
 	cfg := cf.DeepCopy()
 
 	// Copy basic config over
+	cfg.Architecture = ocf.Architecture
+	cfg.OS = ocf.OS
```
jonjohnsonjr: Can you do OSVersion as well?

Vlatombe: done.

PR #696, opened 2020-03-18
Add ImageFromReader

knqyf263:
`tarball.Layer` has `LayerFromReader`, while `tarball.Image` doesn't have `ImageFromReader`. This PR just copies `LayerFromReader` to `ImageFromReader`.
https://godoc.org/github.com/google/go-containerregistry/pkg/v1/tarball#LayerFromReader

### DIFF
--- a/pkg/v1/tarball/image.go
+++ b/pkg/v1/tarball/image.go
@@ -68,6 +68,18 @@ func ImageFromPath(path string, tag *name.Tag) (v1.Image, error) {
 	return Image(pathOpener(path), tag)
 }
 
+// ImageFromReader returns a v1.Image given a io.Reader.
+func ImageFromReader(reader io.Reader, tag *name.Tag) (v1.Image, error) {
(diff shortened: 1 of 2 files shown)


jonjohnsonjr:
Do you have a need for this, or are you just adding it for consistency? I think the in-memory buffering that `tarball.LayerFromReader` does was a mistake -- the signature kind of implies that you can use it in a streaming manner, but it's really inefficient under the hood.

knqyf263:
Thank you for the response. I'd like to use `ImageFromReader` to read the image content from `os.Stdout` like the following.

```
$ docker save alpine:3.11 | my-tool-using-google-go-contianerregistry
```

But if there is another way, I don't need it. Of course, we can implement `opener` for io.Reader myself, but I realized `Layer` can take `io.Reader` directly in `LayerFromReader`. Then, I tried to add the same feature to `Image` as well.

PR #697, opened 2020-03-19
Handle manifest.json containing null in tarball

knqyf263:
If the content of manifest.json is null, it causes panic. I added the tar file causing panic. You can test it by `TestNullManifest`.

Also, this is PoC.
https://play.golang.org/p/RVdS6KY40-m

I faced this problem when I used Kaniko.

```
$ cat Dockerfile
FROM alpine:3.11
$ docker run -v $PWD:/workspace gcr.io/kaniko-project/executor:latest --tarPath /workspace/image.tar --dockerfile /workspace/Dockerfile --context dir:///workspace/ --no-push
```

I know this is a bug of Kaniko and it doesn't generate null manifest if we add `--destination`.

```
$ docker run -v $PWD:/workspace gcr.io/kaniko-project/executor:latest --tarPath /workspace/image.tar --dockerfile /workspace/Dockerfile --context dir:///workspace/ --no-push --destination=image
```

But I feel this library should handle this panic.


pkg/v1/tarball/image.go
```diff
@@ -162,6 +162,10 @@ func (i *image) loadTarDescriptorAndConfig() error {
 		return err
 	}
 
+	if i.manifest == nil{
```
jonjohnsonjr:
```suggestion
	if i.manifest == nil {
```

pkg/v1/tarball/image_test.go
```diff
@@ -47,6 +47,13 @@ func TestManifestAndConfig(t *testing.T) {
 	}
 }
 
+func TestNullManifest(t *testing.T) {
+	img, err := ImageFromPath("testdata/null_manifest.tar", nil)
+	if err == nil {
+		t.Fatalf("Error expected loadig image: %v", img)
```
jonjohnsonjr:
```suggestion
		t.Fatalf("Error expected loading null image: %v", img)
```

pkg/v1/tarball/image.go
knqyf263: Thanks. My linter didn't work...I've fixed it.

pkg/v1/tarball/image_test.go
knqyf263: fixed

jonjohnsonjr:
Thanks!

PR #700, opened 2020-03-25
Extend the remote tag listing to include more tag details from the manifest

ccojocar:
In addition, the cran command was extended to print these details when the `-a` option
is provided.

The `cran list -a` command returns the following output:
```
./crane ls -a gcr.io/jenkinsxio/builder-jx
TAGS                            SHA                                                                             CREATE TIME                     UPLOAD TIME
2.0.1253-593,latest             sha256:521546feceeca228fea2e42c181361f6aa7d5e0e89f7eefe6f369cfb2ad7b73b         24 Mar 20 20:30 CET             24 Mar 20 20:31 CET
0.0.0-SNAPSHOT-PR-6961-2        sha256:786f17061022643ffb03c73905b483706202b22436f81d598d268b1840af8316         24 Mar 20 20:02 CET             24 Mar 20 20:03 CET
```

cc @madchap @mestrade

### DIFF
--- a/cmd/crane/cmd/list.go
+++ b/cmd/crane/cmd/list.go
@@ -17,6 +17,11 @@ package cmd
 import (
 	"fmt"
 	"log"
+	"os"
+	"sort"
+	"strings"
+	"text/tabwriter"
+	"time"
 
 	"github.com/google/go-containerregistry/pkg/crane"
 	"github.com/spf13/cobra"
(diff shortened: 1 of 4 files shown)


jonjohnsonjr:
Hi, thanks for the PR!

The extra details you're parsing here are actually a GCR-specific thing and not something other registries would support. For that reason, it's not something we'd likely add to `crane`, but that is exactly the purpose of [`gcrane`](https://github.com/google/go-containerregistry/tree/master/cmd/gcrane) 😄 (the google-specific stuff in this repo is under [pkg/v1/google](https://github.com/google/go-containerregistry/tree/master/cmd/gcrane)).

`gcrane ls` doesn't support this kind of human-readable formatting, but it does expose everything if you use the `--json` flag. For human readable formatting, I'd probably just use [gcloud](https://cloud.google.com/sdk/gcloud/reference/container/images/list-tags).

ccojocar:
I see. It seems to be implemented by `google.List` function. Actually, I need  to use only this package in another tool, but extended the `crane` to make this functionality useful 😄.

PR #703, opened 2020-03-29
Allow Reference also for daemon write

### LINKED ISSUE #702: Allowing writing images by both tag and digest
Missed adding the feature to daemon.Write, when adding to tarball.Write (#404)

And currently the tag is **removed**, when specifying _both_ tag and digest like so:

```
busybox:latest@sha256:b26cd013274a657b86e706210ddd5cc1f82f50155791199d29b9e86e935ce135
```

The `crane` tool adds an amusing "i-was-a-digest" tag, even when one exists.

afbjorklund:
I thought this was going to be a simple one-liner, like before:

```diff
-func Write(tag name.Tag, img v1.Image) (string, error) {
+func Write(ref name.Reference, img v1.Image) (string, error) {
```

But I was wrong, since I actually wanted to keep the tag too...

Closes #702

### DIFF
--- a/pkg/name/digest.go
+++ b/pkg/name/digest.go
@@ -28,6 +28,7 @@ const (
 // Digest stores a digest name in a structured form.
 type Digest struct {
 	Repository
+	tag      string
 	digest   string
 	original string
 }
@@ -45,6 +46,11 @@ func (d Digest) Identifier() string {
 	return d.DigestStr()
 }
 
+// TagStr returns the tag component of the Digest.
+func (d Digest) TagStr() string {
+	return d.tag
+}
+
 // DigestStr returns the digest component of the Digest.
 func (d Digest) DigestStr() string {
 	return d.digest
@@ -52,7 +58,19 @@ func (d Digest) DigestStr() string {
 
 // Name returns the name from which the Digest was derived.
 func (d Digest) Name() string {
-	return d.Repository.Name() + digestDelim + d.DigestStr()
+	name := d.Repository.Name()
+	if d.tag != "" {
+		name += tagDelim + d.TagStr()
+	}
+	name += digestDelim + d.DigestStr()
+	return name
+}
+
+// Tag returns the tag component of the Digest (if any)
+func (d Digest) Tag() Tag {
+	name := d.Repository.Name() + tagDelim + d.TagStr()
+	tag, _ := NewTag(name, StrictValidation)
+	return tag
 }
 
 // String returns the original input string.
@@ -90,6 +108,7 @@ func NewDigest(name string, opts ...Option) (Digest, error) {
 	}
 	return Digest{
 		Repository: repo,
+		tag:        tag.tag,
 		digest:     digest,
 		original:   name,
 	}, nil
--- a/pkg/crane/pull.go
+++ b/pkg/crane/pull.go
@@ -60,7 +60,11 @@ func Save(img v1.Image, src, path string) error {
 		if !ok {
 			return fmt.Errorf("ref wasn't a tag or digest")
 		}
-		tag = d.Repository.Tag(iWasADigestTag)
+		if d.TagStr() != "" {
+			tag = d.Tag()
+		} else {
+			tag = d.Repository.Tag(iWasADigestTag)
+		}
 	}
 
 	// no progress channel (for now)
--- a/pkg/name/digest_test.go
+++ b/pkg/name/digest_test.go
@@ -63,8 +63,10 @@ func TestNewDigestStrictValidation(t *testing.T) {
 	}
 
 	for _, name := range goodStrictValidationTagDigestNames {
-		if _, err := NewDigest(name, StrictValidation); err != nil {
+		if digest, err := NewDigest(name, StrictValidation); err != nil {
 			t.Errorf("`%s` should be a valid Digest name, got error: %v", name, err)
+		} else if digest.Name() != name {
+			t.Errorf("`%v` .Name() should reproduce the original name. Wanted: %s Got: %s", digest, name, digest.Name())
 		}
 	}
 
@@ -95,9 +97,10 @@ func TestDigestComponents(t *testing.T) {
 	t.Parallel()
 	testRegistry := "gcr.io"
 	testRepository := "project-id/image"
+	testTag := "latest"
 	fullRepo := path.Join(testRegistry, testRepository)
 
-	digestNameStr := testRegistry + "/" + testRepository + "@" + validDigest
+	digestNameStr := testRegistry + "/" + testRepository + ":" + testTag + "@" + validDigest
 	digest, err := NewDigest(digestNameStr, StrictValidation)
 	if err != nil {
 		t.Fatalf("`%s` should be a valid Digest name, got error: %v", digestNameStr, err)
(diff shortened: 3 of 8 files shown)


medyagh:
this PR is very helpful for minikube project. any chance we could get a review on this ? this will save seconds on our minikube start.

@mattmoor @jonjohnsonjr @ImJasonH

afbjorklund:
As far as I know, this is still blocked on maintainer feedback: #540

pkg/name/digest.go
```diff
@@ -45,14 +46,31 @@ func (d Digest) Identifier() string {
...
+		name += tagDelim + d.TagStr()
+	}
+	name += digestDelim + d.DigestStr()
+	return name
+}
+
+// Tag returns the tag component of the Digest (if any)
+func (d Digest) Tag() Tag {
+	name := d.Repository.Name() + tagDelim + d.TagStr()
+	tag, _ := NewTag(name, StrictValidation)
```
mattmoor, line 72: Does this work properly when `TagStr()` returns `""`?

pkg/name/digest.go
```diff
@@ -45,14 +46,31 @@ func (d Digest) Identifier() string {
 	return d.DigestStr()
 }
 
+// TagStr returns the tag component of the Digest.
+func (d Digest) TagStr() string {
+	return d.tag
```
mattmoor, line 51: It looks like this is only tested through `Name()`, perhaps add some focused tests?

pkg/name/digest.go
afbjorklund: No,  it returns a bogus result ending in `name@`. So it is tested before use (slightly ugly)

pkg/name/digest.go
afbjorklund: Probably a good idea, even though it was just a simple accessor method.

pkg/name/digest.go
afbjorklund: Returning a `Tag{}` here would still require testing below, and returning `nil` would make it crash.
Adding an `error` only makes things worse, then it is no longer possible to use it as a one-liner.

afbjorklund: So basically what happens here, is that the new the `name` will fail the `StrictValidation`

And then the error will be ignored (`_`), and it (NewTag) will return the mentioned `Tag[}`

medyagh:
@afbjorklund  are you done updating the PR ? or are you still working on this PR ?

medyagh:
@jonjohnsonjr any chance we could have this PR merged ? it will boost the perforamnce of minikube by considerable amount of seconds.

ImJasonH:
It looks like we're waiting on Travis to figure out it needs to build and approve this. Just in case this breaks Travis (which I doubt) can you make some dummy commit message change using `git commit --amend` and force-push it to get Travis to pick it up?

afbjorklund:
Rebased the branch now, hope that helps Travis

medyagh:
once this PR is merged we can update mnikube to use the master https://github.com/kubernetes/minikube/pull/8355

pkg/name/digest.go
jonjohnsonjr: This seems kind of error prone :/

jonjohnsonjr: Also, this implementation doesn't preserve the original Digest options, see https://github.com/google/go-containerregistry/issues/563

afbjorklund:
I'm considering replacing this PR with some docker-specific hack, since the whole API is limited to docker anyway...

For the other runtimes we just throw the digest away, since they don't support having both (tag and digest) at once.

```
pkg/minikube/registry/drvs/docker/docker.go:            ImageDigest:       cc.KicBaseImage,
pkg/minikube/registry/drvs/podman/podman.go:            ImageDigest:       strings.Split(cc.KicBaseImage, "@")[0], // for podman does not support docker images references with both a tag and digest.
```

Could do the same for the cache image file save (tarball.Write), since the digest isn't actually _used_ there anyway.

Tired of carrying stale upstream patches downstream, only means that we are missing out on the other updates.

----

There are already similar hacks elsewhere, so it won't really make a big difference in minikube.

```go
// WriteImageToDaemon write img to the local docker daemon
func WriteImageToDaemon(img string) error {
        klog.Infof("Writing %s to local daemon", img)
        ref, err := name.ParseReference(img)
        if err != nil {
                return errors.Wrap(err, "parsing reference")
        }
        klog.V(3).Infof("Getting image %v", ref)
        i, err := remote.Image(ref)
        if err != nil {
                if strings.Contains(err.Error(), "GitHub Docker Registry needs login") {
                        ErrGithubNeedsLogin = errors.New(err.Error())
                        return ErrGithubNeedsLogin
                } else if strings.Contains(err.Error(), "UNAUTHORIZED") {
                        ErrNeedsLogin = errors.New(err.Error())
                        return ErrNeedsLogin
                }

                return errors.Wrap(err, "getting remote image")
        }
        klog.V(3).Infof("Writing image %v", ref)
        _, err = daemon.Write(ref, i)
        if err != nil {
                return errors.Wrap(err, "writing daemon image")
        }

        return nil
}
```

Just add the tag afterwards, so that the docker-specific check code is happy about the image.

```go
// ExistsImageInDaemon if img exist in local docker daemon
func ExistsImageInDaemon(img string) bool {
        // Check if image exists locally
        cmd := exec.Command("docker", "images", "--format", "{{.Repository}}:{{.Tag}}@{{.Digest}}")
        if output, err := cmd.Output(); err == nil {
                if strings.Contains(string(output), img) {
                        klog.Infof("Found %s in local docker daemon, skipping pull", img)
                        return true
                }
        }
        // Else, pull it
        return false
}
```

afbjorklund:
> > EDIT: Right, now I remember why we didn't do that in the first place. It's not possible without network access.
> 
> Do we need to do anything on our side? I've kind of lost context on this change.

The context is in the issue, not in the pull request. Anyway, the image digest was lost when using `go-containerregistry` to pull.

It happened when we tried to replace `docker pull` with go code, and was causing minikube to reject the image and pull again...

afbjorklund:
But it seems like the bot closed the go-containerregistry issue, and that we failed to link it to the _original_ minikube issue...

https://github.com/kubernetes/minikube/issues/7226

jonjohnsonjr:
> `refusing to create a tag with a digest reference`

Looks like this happens if you call `cli.ImageTag`, as in our `daemon.Tag` path? So we never want to include a digested reference there, but we _do_ want to include it for the `daemon.Write` path...

So we need some way to say "include both" (this is `.String()`) and "give me only the tag, if it's present", which perhaps should always be the semantics of `name.Tag`?

I'm sorry -- I guess I still don't understand why we need to change much of anything here. Doesn't `name.Digest.String()` do what you need already?

Is it okay for tarballs to include references like `ubuntu:latest@sha256:...`? I'm a little concerned about all the rules for tags within tarballs, since I am not super familiar with the expectations of docker/container/podman/etc -- does this interact with https://github.com/google/go-containerregistry/issues/890 in any way?

I'd need to look through this PR again, but last I looked I wasn't happy with this: https://github.com/google/go-containerregistry/pull/703#discussion_r434290854

afbjorklund:
I will close this PR, and do the workarounds (`docker pull`) in minikube instead: https://github.com/kubernetes/minikube/pull/12184

PR #710, opened 2020-04-12
Allow empty hash for layout.Image

knqyf263:
`tarball` allows an empty tag. I need it for `layout` as well because OCI layout contains only 1 image in our many cases. Currently, we have to parse `index.json` in advance to know the hash of an image, but I'd like to avoid it.
https://github.com/google/go-containerregistry/blob/b8d7918fda8497bccfaa470932127112d0cc0021/pkg/v1/tarball/image.go#L119-L124

I tried changing `v1.Hash` to `*v1.Hash` like `*name.Tag`, but this method is defined in the interface. I gave up it and used an empty hash, `v1.Hash{}`.
https://github.com/google/go-containerregistry/blob/be3bb4a63da1230374dc617421699e553fcf9897/pkg/v1/layout/index.go#L84

It is an open discussion. Let me know if you have any better idea.


knqyf263:
I don't know why Travis CI says this build is in progress. It looks like all tests have finished normally.
https://travis-ci.org/github/google/go-containerregistry/builds/674024388

I'd like someone from maintainers to re-run this job.

ImJasonH:
Sorry for this delay, I'm not actually sure how to kick Travis into action here. Can you maybe `git commit --amend` and make some dummy change to the commit message, then force push it to get this unblocked?

knqyf263:
@ImJasonH Thank you for the advice. It works now.

PR #711, opened 2020-04-13
Consistent ordering when processing images to create a tarball manifest

deitch:
Resolves the issue raised in [this comment](https://github.com/google/go-containerregistry/pull/685#issuecomment-612885360) by @ImJasonH . 

When creating a v1/tarball manifest of multiple images, _always_ provide the order in order of the repo tags. It doesn't matter to the tarball, but it allows for us to have consistent results. It ensures tests pass, but consistency is good overall.

The sort key is just the repo tags joined together by `","`. It is simple, but it does work. I considered using the hash of the specific manifest, but that would make it too hard to eyeball. Why use complexity unnecessarily?

### DIFF
--- a/pkg/v1/tarball/write.go
+++ b/pkg/v1/tarball/write.go
@@ -21,6 +21,8 @@ import (
 	"fmt"
 	"io"
 	"os"
+	"sort"
+	"strings"
 
 	"github.com/google/go-containerregistry/pkg/name"
 	v1 "github.com/google/go-containerregistry/pkg/v1"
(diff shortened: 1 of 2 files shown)


pkg/v1/tarball/write.go
```diff
@@ -191,6 +193,10 @@ func processImages(refToImage map[name.Reference]v1.Image, tf *tar.Writer) (Mani
 			LayerSources: layerSources,
 		})
 	}
+	// sort by name of the repotags so it is consistent. Alternatively, we could sort by hash of the
```
ImJasonH, line 196: It's a small thing, but I find `sort.Slice` simpler and easier to read -- it doesn't require satisfying an interface, it uses reflection instead. It shouldn't really affect performance while there are only going to be a handful of items.

deitch: Sure, I don't mind. Hang tight.

deitch: Done

deitch:
> Thanks!

With pleasure. :-)

PR #712, opened 2020-04-14
Drain ping body for connection reuse

ktock:
Currenty `ping` for registries doesn't drain the response body so the connection
isn't reused. But in many cases, reusing the `ping` connection is valuable for
performance because it's general to access the `ping`ed registry soon after the
`ping`ing for fetching the contents.

### DIFF
--- a/pkg/v1/remote/transport/ping.go
+++ b/pkg/v1/remote/transport/ping.go
@@ -16,6 +16,8 @@ package transport
 
 import (
 	"fmt"
+	"io"
+	"io/ioutil"
 	"net/http"
 	"strings"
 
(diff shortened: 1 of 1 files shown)


pkg/v1/remote/transport/ping.go
```diff
@@ -83,7 +85,10 @@ func ping(reg name.Registry, t http.RoundTripper) (*pingResp, error) {
 			// Potentially retry with http.
 			continue
 		}
-		defer resp.Body.Close()
+		defer func() {
```
ImJasonH, line 88: Can you add a comment describing why we do this? It might be confusing, and future readers would save time not having to trace the change back to this PR.

ktock:
@ImJasonH Thanks for the review! Added the comment.

PR #714, opened 2020-04-22
Deduplicate layers in legacy tarballs

djfinlay:
Addresses https://github.com/bazelbuild/rules_docker/issues/1420

### DIFF
--- a/pkg/legacy/tarball/write_test.go
+++ b/pkg/legacy/tarball/write_test.go
@@ -15,8 +15,12 @@
 package tarball
 
 import (
+	"archive/tar"
+	"fmt"
+	"io"
 	"io/ioutil"
 	"os"
+	"path/filepath"
 	"strings"
 	"testing"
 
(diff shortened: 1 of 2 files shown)


pkg/legacy/tarball/write_test.go
```diff
@@ -456,3 +460,147 @@ func TestUncompressedSize(t *testing.T) {
...
+	// Create two tags, one pointing to each image created.
+	tag1, err := name.NewTag("gcr.io/foo/bar:latest", name.StrictValidation)
+	if err != nil {
+		t.Fatalf("Error creating test tag1: %v", err)
+	}
+	tag2, err := name.NewTag("gcr.io/baz/bat:latest", name.StrictValidation)
+	if err != nil {
+		t.Fatalf("Error creating test tag2: %v", err)
+	}
+	refToImage := make(map[name.Reference]v1.Image)
```
ImJasonH, line 503: This might be easier to read as:

```
refToImage := map[name.Reference]v1.Image{
  tag1: baseImage,
  tag2: extendedImage,
}
```

pkg/legacy/tarball/write_test.go
```diff
@@ -456,3 +460,147 @@ func TestUncompressedSize(t *testing.T) {
 		t.Errorf("expected UncompressedSize to be called, but it wasn't")
 	}
 }
+
+func TestWriteSharedLayers(t *testing.T) {
```
ImJasonH, line 464: Can you add a comment describing the scenario this tests?

Something like: "TestWriteSharedLayers tests that writing a tarball of multiple images that share some layers only writes those shared layers once." (feel free to reword)

I'm not sure I'd understand this test at a glance outside the context of this PR and a comment would help a lot.

PR #715, opened 2020-04-27
Avoid long delays on http-only registries

nicolaferraro:
This fixes a number of reported slow push time on Minikube.

The reason is that it did a first ping on https, even if the registry is declared as insecure. But connection to port 443 is kept idle by kubernetes until a timeout occurs (which is a typical firewall behavior as well).

If the registry is insecure, the go-containerregistry lib should not try https as first attempt.

Eg.
- https://github.com/GoogleContainerTools/kaniko/issues/970
- https://github.com/GoogleContainerTools/kaniko/issues/1013
- https://github.com/apache/camel-k/issues/1416

### DIFF
--- a/pkg/v1/remote/transport/ping.go
+++ b/pkg/v1/remote/transport/ping.go
@@ -68,13 +68,16 @@ func parseChallenge(suffix string) map[string]string {
 func ping(reg name.Registry, t http.RoundTripper) (*pingResp, error) {
 	client := http.Client{Transport: t}
 
-	// This first attempts to use "https" for every request, falling back to http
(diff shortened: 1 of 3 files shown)


darkdragon-001:
What happened to this?

jonjohnsonjr:
This PR somehow slipped by me. It's not an unreasonable change, but it just flips the problem around for folks running a secure registry on localhost: https://github.com/google/go-containerregistry/issues/279

An alternative could be using an `errgroup` to ping _both_ `http` and `https` concurrently, following whichever returns successfully first and cancelling the other?

PR #716, opened 2020-04-29
send error messages along with progress

deitch:
Finishes off #675 by sending any errors on the update channel in addition to (i.e. without changing) the usual return code.

Error messages along channel can be one of:

* `nil` - no error, more to come (i.e. not finished)
* not-`EOF` - error, finished
* `EOF` - no error, finished


deitch:
I definitely like how this simplifies thinking about getting errors @ImJasonH . Glad we decided to do it.

PR #719, opened 2020-05-01
create CalculateSize()

deitch:
We already pull out the size of the file to use when reporting updates; might as well expose it and simplify our routines.

### DIFF
--- a/pkg/v1/tarball/write.go
+++ b/pkg/v1/tarball/write.go
@@ -98,18 +98,9 @@ func MultiRefWrite(refToImage map[name.Reference]v1.Image, w io.Writer, opts ...
 		}
 	}
 
-	m, err := calculateManifest(refToImage)
+	size, _, mBytes, err := getSizeAndManifest(refToImage)
 	if err != nil {
-		return sendUpdateReturn(o, fmt.Errorf("error calculating manifest: %v", err))
-	}
-	mBytes, err := json.Marshal(m)
-	if err != nil {
-		return sendUpdateReturn(o, fmt.Errorf("could not marshall manifest to bytes: %v", err))
-	}
-
-	size, err := calculateTarballSize(refToImage, mBytes)
-	if err != nil {
-		return sendUpdateReturn(o, fmt.Errorf("error calculating tarball size: %v", err))
+		return sendUpdateReturn(o, err)
 	}
 
 	return writeImagesToTar(refToImage, mBytes, size, w, o)
@@ -293,6 +284,29 @@ func calculateManifest(refToImage map[name.Reference]v1.Image) (m Manifest, err 
 	return m, nil
 }
 
+// CalculateSize calculates the expected complete size of the output tar file
+func CalculateSize(refToImage map[name.Reference]v1.Image) (size int64, err error) {
+	size, _, _, err = getSizeAndManifest(refToImage)
+	return size, err
+}
+
+func getSizeAndManifest(refToImage map[name.Reference]v1.Image) (size int64, m Manifest, mBytes []byte, err error) {
+	m, err = calculateManifest(refToImage)
+	if err != nil {
+		return 0, nil, nil, fmt.Errorf("unable to calculate manifest: %v", err)
+	}
+	mBytes, err = json.Marshal(m)
+	if err != nil {
+		return 0, nil, nil, fmt.Errorf("could not marshall manifest to bytes: %v", err)
+	}
+
+	size, err = calculateTarballSize(refToImage, mBytes)
+	if err != nil {
+		return 0, nil, nil, fmt.Errorf("error calculating tarball size: %v", err)
+	}
+	return size, m, mBytes, nil
+}
+
 // calculateTarballSize calculates the size of the tar file
 func calculateTarballSize(refToImage map[name.Reference]v1.Image, mBytes []byte) (size int64, err error) {
 	imageToTags := dedupRefToImage(refToImage)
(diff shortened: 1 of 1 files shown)


deitch:
Hey @ImJasonH and @jonjohnsonjr ; this is the final follow-up from those other ones, notably #675 and its predecessors.

I might start doing some of these progress and other things for the v1layout format as well, once this is in.

pkg/v1/tarball/write.go
```diff
@@ -293,6 +284,29 @@ func calculateManifest(refToImage map[name.Reference]v1.Image) (m Manifest, err
 	return m, nil
 }
 
+// CalculateSize calculates the expected size of the output tar file
+func CalculateSize(refToImage map[name.Reference]v1.Image) (size int64, err error) {
```
ImJasonH, line 288: Having both this and [`CalculateTarFileSize`](https://godoc.org/github.com/google/go-containerregistry/pkg/v1/tarball#CalculateTarFileSize) seems confusing. Can we rename one, or make it unexported?

deitch: Sure, I can rename `CalculateTarFileSize()`. 5 mins.

deitch: Done. Have a look.

deitch: Ouch, this is still open. Can we get a review, merge it in, close it out?

pkg/v1/tarball/write.go
```diff
@@ -416,10 +430,10 @@ func (pw *progressWriter) Close() error {
 	return io.EOF
 }
 
-// CalculateTarFileSize calculate the size a file will take up in a tar archive,
+// CalculateSingleFileInTarSize calculate the size a file will take up in a tar archive,
 // given the input data. Provided by rounding up to nearest whole block (512)
 // and adding header 512
-func CalculateTarFileSize(in int64) (out int64) {
+func CalculateSingleFileInTarSize(in int64) (out int64) {
```
jonjohnsonjr, line 436: Does this need to be exported?

pkg/v1/tarball/write.go
jonjohnsonjr: Oof, it's been a strange couple months so I've let some things slip, sorry about that.

pkg/v1/tarball/write.go
deitch: Need to? Not really, but it is such a nice convenience function that it might be useful.

You kind of wish this were built into core golang `archive/tar` package.

deitch:
> Oof, it's been a strange couple months so I've let some things slip, sorry about that.

No worries.

deitch:
While we are at it, separately, would you take a "download progress updates" on `Layer()` and similar like we did on the entire tarball?

deitch:
This has been open super long. Can we get it accepted and merged or denied for a reason and closed @ImJasonH or @jonjohnsonjr ? Thanks

pkg/v1/tarball/write.go
jonjohnsonjr: I agree this is useful, but I'd like to avoid exposing anything unless it's really necessary. They might accept a commit to expose some of `archive/tar`? https://github.com/golang/go/blob/8f2db14cd35bbd674cb2988a508306de6655e425/src/archive/tar/format.go#L141-L152

jonjohnsonjr (commented):
Going through and cleaning up old issues -- I didn't realize we never merged this. If you'd still like to get it merged, it LGTM other than exposing `CalculateSingleFileInTarSize`, which feels out of place.

pkg/v1/tarball/write.go
deitch: Sure, I will privatize it and rebase and update. Check soon.

deitch: I got it. Take another look. It has been a while, I might have made an error on the rebase, so please check carefully. Another pair of eyes is always helpful.

pkg/v1/tarball/write.go
```diff
@@ -302,13 +316,13 @@ func calculateTarballSize(refToImage map[name.Reference]v1.Image, mBytes []byte)
...
-		size += CalculateTarFileSize(manifest.Config.Size)
+		size += calculateSingleFileInTarSize(manifest.Config.Size)
 		for _, l := range manifest.Layers {
-			size += CalculateTarFileSize(l.Size)
+			size += calculateSingleFileInTarSize(l.Size)
 		}
 	}
 	// add the manifest
-	size += CalculateTarFileSize(int64(len(mBytes)))
+	size += calculateSingleFileInTarFileSize(int64(len(mBytes)))
```
jonjohnsonjr, line 325:
```suggestion
	size += calculateSingleFileInTarSize(int64(len(mBytes)))
```

deitch: Need to update the func def as well, right? I can just update the commit and re-push.

jonjohnsonjr: You have `calculateSingleFileInTarSize` in three places and `calculateSingleFileInTarFileSize` in one place. Whichever you prefer, just needs to be consistent. Also feel free to reject my suggestion and just update/rebase whatever 👍

jonjohnsonjr:
Looks like travis doesn't have very many windows machines available :(

jonjohnsonjr (approved):

jonjohnsonjr:
Travis failed but I didn't notice, retried and it works.

Issue #717, opened 2020-04-30
pkg/v1/cache is unable to handle incomplete files


chrischdi:
If using the cache and interrupting the download the following will happen:
* The in-flight layer stays as-is in the local cache directory
* On the next run, when [fsCache.Get] is getting called with the broken hash, `io.Copy` in [pkg/v1/hash.go](https://github.com/google/go-containerregistry/blob/master/pkg/v1/hash.go#L115) returns a `UnexpectedEOF` error (see below for full stacktrace)

This could get fixed by Deleting the broken file in [fsCache.Get] when `UnexpectedEOF` was returned and returning ErrNotFound, which will lead to downloading the file again.

I've pushed this change [on my fork here](https://github.com/chrischdi/go-containerregistry/commit/a264bb14a97d84343d53d887612fd4958a27ab53)

If this is valid I would be happy to open a PR for this :-)

Sidenote: I found this by using `container_pull` using the of [rules_docker](https://github.com/bazelbuild/rules_docker)

[fsCache.Get]: https://github.com/google/go-containerregistry/blob/master/pkg/v1/cache/fs.go#L103

Stacktrace to the UnexpectedEOF error
```
github.com/google/go-containerregistry/pkg/v1.SHA256 (/home/schloc/go/src/github.com/google/go-containerregistry/pkg/v1/hash.go:116)
github.com/google/go-containerregistry/pkg/v1/tarball.computeDigest (/home/schloc/go/src/github.com/google/go-containerregistry/pkg/v1/tarball/layer.go:144)
github.com/google/go-containerregistry/pkg/v1/tarball.LayerFromOpener (/home/schloc/go/src/github.com/google/go-containerregistry/pkg/v1/tarball/layer.go:113)
github.com/google/go-containerregistry/pkg/v1/tarball.LayerFromFile (/home/schloc/go/src/github.com/google/go-containerregistry/pkg/v1/tarball/layer.go:87)
github.com/google/go-containerregistry/pkg/v1/cache.(*fscache).Get (/home/schloc/go/src/github.com/google/go-containerregistry/pkg/v1/cache/fs.go:104)
github.com/google/go-containerregistry/pkg/v1/cache.(*image).Layers (/home/schloc/go/src/github.com/google/go-containerregistry/pkg/v1/cache/cache.go:64)
[...]

PR #720, opened 2020-05-01
pkg/v1/cache/fs.go delete cached layer on UnexpectedEOF

chrischdi:
This fixes an issue when the cache has to handle with interrupted downloads of previous executions.

It solves #717


chrischdi:
@googlebot I signed it!

chrischdi:
An alternative solution which came to my mind is to return maybe a new error here (e.g. `ErrCorrupt`) and let the application using the Get function handle the error.

ImJasonH:
> An alternative solution which came to my mind is to return maybe a new error here (e.g. `ErrCorrupt`) and let the application using the Get function handle the error.

That might be useful in the future, but I think for now at least cleaning up makes sense.

Thanks for this change, lgtm!

PR #721, opened 2020-05-07
Use Google keychain on gcrane commands

FGtatsuro:
This PR includes following changes.

- `gcrane gc` command uses Google keychain like `gcrane ls`.
- `gcrane delete` command uses Google keychain like `gcrane copy`.

My motivation is using above two commands on GCP services(on my case, Cloud Run).


FGtatsuro:
@googlebot I signed it!

jonjohnsonjr:
Apologies for not getting around to this sooner. I think a solution I'd prefer to see is adding a keychain option to `pkg/crane` (a la https://github.com/google/go-containerregistry/pull/653) and then just use that from gcrane.

PR #725, opened 2020-05-27
set up goreleaser with github action

yihanzhen:
Tested in [my fork](https://github.com/yihanzhen/go-containerregistry/releases).

To make it work we'll also need to add a github secret `RELEASE_TOKEN` (which requires admin permission to the repo).


yihanzhen:
@jonjohnsonjr @ImJasonH PTAL

yihanzhen:
cc/ @lindsayismith

lindsayismith:
What is the name of the release binary that it creates?  How does it relate to the tags we push?

yihanzhen:
@lindsayismith The name of the binary is `go-containerregistry_{version}_{os}.tar.gz`. The tag gets filled into the `{version}` part. See [this](https://github.com/yihanzhen/go-containerregistry/releases/tag/v0.4.0) as an example.

.goreleaser.yml
```diff
@@ -0,0 +1,41 @@
+# This is an example goreleaser.yaml file with some sane defaults.
+# Make sure to check the documentation at http://goreleaser.com
+# before:
+#   hooks:
+#     # You may remove this if you don't use go modules.
+#     - go mod download
+#     # you may remove this if you don't need go generate
+#     - go generate ./...
+builds:
+- id: registry
```
jonjohnsonjr, line 10: Let's drop the registry binary from this -- it's primarily for testing and I don't want to implicitly endorse its use by including it in releases.

yihanzhen: Done

PR #735, opened 2020-07-06
Add support for squashing images.

tomprince:


ImJasonH:
First of all sorry for the long delay in responding.

The change itself lgtm and I think there are some cool use cases it unlocks. So yay!

The only nit I think is with the name `squash` -- in other contexts where this has been done it's been called `flatten`, which I think is slightly more descriptive.

If you're open to renaming (and you're still interested after all this time) I think we can merge this mostly as-is.

tomprince:
I'd be happy to see this land, but I'm no longer with the company where I wanted this, so probably won't have the time to make changes.

Issue #737, opened 2020-07-07
Misleading error when creating a transport fails due to bad username/password


ewrenn8:
When [constructing a transport][0], if auth fails, we get this error: 
```GET https://auth.docker.io/token?scope=repository%3A<USER>%2F<REPO>%3Apull&service=registry.docker.io: unsupported status code 401```

I would expect 401 to be a supported error code, leading to a more meaningful and descriptive error: 
```GET https://auth.docker.io/token?scope=repository%3A<USER>%2F<REPO>%3Apull&service=registry.docker.io:  UNAUTHORIZED: authentication required ...```

It seems like this is because the error returned from the [calling the auth endpoint][1] does not unmarshal in to a [structured error][2] in a useful way.

This is the error being returned from the auth endpoint when attempting to authorize with base64 encoded `foo:bar`:
```bash
$ curl https://auth.docker.io/token\?scope\=repository%3A<USER>%2F<REPO>%3Apull\&service\=registry.docker.io -H "Authorization: Basic Zm9vOmJhcgo="
{"details":"incorrect username or password"}
```

This may be more of an api side issue as the auth endpoint doesn't match the docker [error format][3], in which case, feel free to close.

Thanks!

[0]: https://github.com/google/go-containerregistry/blob/72edbad9993e722466d28ff308cf2cdba4b675c5/pkg/v1/remote/transport/transport.go#L84
[1]: https://github.com/google/go-containerregistry/blob/72edbad9993e722466d28ff308cf2cdba4b675c5/pkg/v1/remote/transport/bearer.go#L229-L257
[2]: https://github.com/google/go-containerregistry/blob/72edbad9993e722466d28ff308cf2cdba4b675c5/pkg/v1/remote/transport/error.go#L45-L53
[3]: https://docs.docker.com/registry/spec/api/#errors

jonjohnsonjr:
> This may be more of an api side issue as the auth endpoint doesn't match the docker error format, in which case, feel free to close.

Our intention is to return the raw body if we can't parse it into a structured error:

https://github.com/google/go-containerregistry/blob/72edbad9993e722466d28ff308cf2cdba4b675c5/pkg/v1/remote/transport/error.go#L169-L171

https://github.com/google/go-containerregistry/blob/72edbad9993e722466d28ff308cf2cdba4b675c5/pkg/v1/remote/transport/error.go#L69-L73

[...]

PR #739, opened 2020-07-07
transport: preserve unexpected json errors

jonjohnsonjr:
Fixes https://github.com/google/go-containerregistry/issues/737

### DIFF
--- a/pkg/v1/remote/transport/error_test.go
+++ b/pkg/v1/remote/transport/error_test.go
@@ -91,6 +91,11 @@ func TestCheckErrorNotError(t *testing.T) {
 		msg:  "unsupported status code 400",
 	}, {
 		code: http.StatusUnauthorized,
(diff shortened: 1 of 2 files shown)


pkg/v1/remote/transport/error_test.go
```diff
@@ -89,6 +89,11 @@ func TestCheckErrorNotError(t *testing.T) {
 		code: http.StatusBadRequest,
 		body: "",
 		msg:  "unsupported status code 400",
+	}, {
```
ImJasonH, line 92: What about a test case for invalid JSON, to exercise the point you make in the comment.

jonjohnsonjr: That's the next case, with a body of `Not JSON`.

ImJasonH: _How convenient_.

Issue #741, opened 2020-07-16
crane copy does not work with non-distributable layers


dsbrng25b:
Since docker distribution version 2.7 validation is enabled by default. As far as I understand this makes it impossible to copy images with non-distributable layers, since these layers are not uploaded (https://github.com/google/go-containerregistry/blob/72597da50ce903895087828b5fc4b918636366f4/pkg/v1/remote/write.go#L76).

Steps to reproduce:
```
$ docker run -it --rm -p 443:5000 --name registry -v "$(pwd)"/tls.key:/certs/tls.key -v "$(pwd)"/tls.crt:/certs/tls.crt -e REGISTRY_HTTP_TLS_CERTIFICA=/certs/tls.crt -e REGISTRY_HTTP_TLS_KEY=/certs/tls.key registry:2.7

$ crane copy mongo@sha256:1b29fbe615ce2f0a91e8973a1aa6fca59b4aaa21bc5d6c8311e6a55cc6ff6b18 my.reg/mongo:latest
...
failed to copy index: PUT https://my.reg/v2/mongo/manifests/sha256:2487bc12c2f0933de86100d72b213e37d0ff2ef5c4419288a03e50778505a3be: multiple errors returned: UNKNOWN: unknown error; UNKNOWN: unknown error; map[]; MANIFEST_BLOB_UNKNOWN: blob unknown to registry; sha256:3889bb8d808bbae6fa5a33e07093e65c31371bcf9e4c38c21be6b9af52ad1548; UNKNOWN: unknown error; UNKNOWN: unknown error; map[]; MANIFEST_BLOB_UNKNOWN: blob unknown to registry; sha256:bf261bf8d7bc7faa62a55474c337e0c2bf6e7d24df25e7737fa97ace46a2cd9a
```
I think this is mainly a problem for windows images which I don't need either way. So for me a solution for this problem would be if one could only copy images from a certain platform. For example [skopeo](https://github.com/containers/skopeo/) supports this. By default skopeo only copies the image which matches the current platform. Another benefit of specifying the platform is that copy runs faster since images which are not needed are not copied.

jonjohnsonjr:
This is upsetting. I can't believe how poorly the non-distributable layers stuff was implemented :/

Specifying platform seems like a good thing to add. I'd also be fine adding a `--allow-nondistributable-artifacts` flag to allow pushing the blobs anyway... maybe both.

dsbrng25b:
[...]

PR #742, opened 2020-07-16
Add platform option to crane copy and digest

dsbrng25b:
Add platform option to crane copy and digest. If the remote reference is an index/list it only copies the image which matches the specified platform.

This would be a possible solution for #741

### DIFF
--- a/pkg/crane/digest.go
+++ b/pkg/crane/digest.go
@@ -16,6 +16,22 @@ package crane
 
 // Digest returns the sha256 hash of the remote image at ref.
 func Digest(ref string, opt ...Option) (string, error) {
+	o := makeOptions(opt...)
+	if o.platform != nil {
+		desc, err := getManifest(ref, opt...)
+		if err != nil {
+			return "", err
+		}
+		img, err := desc.Image()
+		if err != nil {
+			return "", err
+		}
+		digest, err := img.Digest()
+		if err != nil {
+			return "", err
+		}
+		return digest.String(), nil
+	}
 	desc, err := head(ref, opt...)
 	if err != nil {
 		return "", err
--- a/pkg/crane/manifest.go
+++ b/pkg/crane/manifest.go
@@ -20,5 +20,13 @@ func Manifest(ref string, opt ...Option) ([]byte, error) {
 	if err != nil {
 		return nil, err
 	}
+	o := makeOptions(opt...)
+	if o.platform != nil {
+		img, err := desc.Image()
+		if err != nil {
+			return nil, err
+		}
+		return img.RawManifest()
+	}
 	return desc.Manifest, nil
 }
--- a/pkg/crane/copy.go
+++ b/pkg/crane/copy.go
@@ -47,8 +47,15 @@ func Copy(src, dst string, opt ...Option) error {
 	switch desc.MediaType {
 	case types.OCIImageIndex, types.DockerManifestList:
 		// Handle indexes separately.
-		if err := copyIndex(desc, dstRef, o); err != nil {
-			return fmt.Errorf("failed to copy index: %v", err)
+		if o.platform != nil {
+			// If platform is explicitly set, don't copy the whole index, just the appropriate image.
+			if err := copyImage(desc, dstRef, o); err != nil {
+				return fmt.Errorf("failed to copy image: %v", err)
+			}
+		} else {
+			if err := copyIndex(desc, dstRef, o); err != nil {
+				return fmt.Errorf("failed to copy index: %v", err)
+			}
 		}
 	case types.DockerManifestSchema1, types.DockerManifestSchema1Signed:
 		// Handle schema 1 images separately.
--- a/pkg/crane/options.go
+++ b/pkg/crane/options.go
@@ -19,12 +19,14 @@ import (
 
 	"github.com/google/go-containerregistry/pkg/authn"
 	"github.com/google/go-containerregistry/pkg/name"
+	"github.com/google/go-containerregistry/pkg/v1"
 	"github.com/google/go-containerregistry/pkg/v1/remote"
 )
 
 type options struct {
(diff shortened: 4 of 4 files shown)


pkg/crane/get.go
```diff
@@ -41,5 +42,46 @@ func getManifest(r string, opt ...Option) (*remote.Descriptor, error) {
...
 	}
-	return remote.Get(ref, o.remote...)
+	return getRemoteDescriptor(ref, o)
+}
+
+// getRemoteDescriptor returns the remote.Descriptor of a reference. If the
+// referece points to a index and the platform option is set, instead of the
+// index descriptor, the descriptor of the image which matches the platform is
+// returned.
+func getRemoteDescriptor(ref name.Reference, o options) (*remote.Descriptor, error) {
```
jonjohnsonjr, line 52: Rather than doing this logic here, it might be easier to just do something in `pkg/crane` like:

```go
if opt.platform != nil {
  img, err := desc.Image()
  ...
}
```

in copy and manifest --  I think that would be a smaller change.

dsbrng25b: I don't get what you mean. The logic (for loop below) is required to find the matching platform if the descriptor is a index/list.

jonjohnsonjr: I'd like to avoid duplicating this logic too much -- this happens in `remote`: https://github.com/google/go-containerregistry/blob/72597da50ce903895087828b5fc4b918636366f4/pkg/v1/remote/index.go#L125-L159

Your loop here assumes there's only ever one level of nesting within an image index, which is usually true, but it's possible to have a [nested index](https://github.com/google/go-containerregistry/tree/72597da50ce903895087828b5fc4b918636366f4/pkg/v1/remote#anatomy-of-an-index), which this wouldn't handle.

If you make `crane.WithPlatform` additionally pass the option down through to `remote.WithPlatform` (similar to [this](https://github.com/google/go-containerregistry/blob/72597da50ce903895087828b5fc4b918636366f4/pkg/crane/options.go#L49)), we can reuse that logic from `remote`.

For copy, you could just change [this](https://github.com/google/go-containerregistry/blob/72597da50ce903895087828b5fc4b918636366f4/pkg/crane/copy.go#L48-L52) to:

```go
if o.platform != nil {
                // If platform is explicitly set, don't copy the whole index, just the appropriate image.
  		if err := copyImage(desc, dstRef, o); err != nil {
			return fmt.Errorf("failed to copy image: %v", err)
		}
} else {
                 if err := copyIndex(desc, dstRef, o); err != nil {
			return fmt.Errorf("failed to copy index: %v", err)
		}
}
```

Since `copyImage` call [`desc.Image`](https://github.com/google/go-containerregistry/blob/72597da50ce903895087828b5fc4b918636366f4/pkg/crane/copy.go#L69), it will do the child matching logic for you.

Fixing `Manifest` and `Digest` would require similar changes.

pkg/crane/digest.go
```diff
@@ -16,9 +16,14 @@ package crane
...
 	if err != nil {
 		return "", err
 	}
-	return desc.Digest.String(), nil
+
+	digest, err := img.Digest()
+	if err != nil {
+		return "", err
+	}
+	return "sha256:" + digest.Hex, nil
```
dsbrng25b, line 28: Is there a better way to do this?

jonjohnsonjr: `digest.String()` should work

pkg/crane/get.go
```diff
@@ -35,11 +40,20 @@ func getImage(r string, opt ...Option) (v1.Image, name.Reference, error) {
...
 		return nil, fmt.Errorf("parsing reference %q: %v", r, err)
 	}
-	return remote.Get(ref, o.remote...)
+	desc, err := remote.Get(ref, o.remote...)
+	if err != nil {
+		return nil, err
+	}
+
+	if o.platform == nil {
+		return desc.ImageIndex()
```
jonjohnsonjr, line 55: This will fail for anything that's not an index (schema1 or image) -- I'd probably just inline the logic from `getManifest` into `Manifest` and `Digest` -- the awkwardness of trying to fit this into the type system makes it tough.

This would be a lot easier if I hadn't (stupidly) embedded `v1.Descriptor` into `remote.Descriptor`, because we have a name collision between the `Digest` in the struct in the `Digest()` method 😞

dsbrng25b:
I moved the logic to the Digest and Manifest function. Does not look super nice, but is probably easier to understand than the logic I used before with the interface.

pkg/crane/manifest.go
```diff
@@ -14,11 +14,36 @@
...
+		}
+		return index.RawManifest()
+	}
+
+	img, err := desc.Image()
 	if err != nil {
 		return nil, err
 	}
-	return desc.Manifest, nil
+	return img.RawManifest()
```
jonjohnsonjr, line 48: This will still fail for schema 1 images, which we aren't able to satisfy the `v1.Image` interface for. If we keep the behavior the same here, but only try to resolve to an image when platform is specified, I think we can keep everything working, i.e.:

```go
if o.platform != nil {
  img, err := desc.Image()
  if err != nil {
    return nil, err
  }
  return img.RawManifest()
}

return desc.Manifest, nil
```

Sorry for all the back and forth on this, I really appreciate your contribution!

dsbrng25b:
Tried to change the logic according to your points in the last review, squashed the commits and rebased them on master.
Since in the meantime crane.Digest was changed and now only uses HEAD I had to change the logic that it still gets the manifest if a platform is specified. Or is there another solution.

jonjohnsonjr:
> I had to change the logic that it still gets the manifest if a platform is specified. Or is there another solution?

I think what you've done is the best we can do.

jonjohnsonjr (approved):
LGTM, thanks!

PR #744, opened 2020-07-17
Add remote.WithContext

### LINKED ISSUE #743: Support Context with remote.Image and remote.Get
I am not seeing a way to pass in a `context` when using `remote.Get()`. Under the hood, it looks like it's just using the standard `net/http` package so I don't think it'd be too difficult of an addition if this proposal makes sense.

.._Context_.. :drum: being in the event of a tcp timeout (e.g. unresposive registry), its just using the default timeout which is quite lengthy for my case. I'm defining a parent context and would like to cancel the request sooner.

jonjohnsonjr:
Fixes #743 

This allows passing a context.Context into any remote operations.


jpreese:
Wow that was fast, I'm super excited! Upgrading to this as soon as it's released 🤣

jonjohnsonjr:
The trick is to file an issue that's not very hard to fix when I'm procrastinating doing other stuff 😉

PR #745, opened 2020-07-20
[Fix] force a specific order in manifest for hermetic builds

grahamjenson:
The order of the containers in the manifest is in-determinant but does not matter. This causes builds via tools like [`rules_docker`](https://github.com/bazelbuild/rules_docker) to create non-deterministic files.

### DIFF
--- a/pkg/legacy/tarball/write.go
+++ b/pkg/legacy/tarball/write.go
@@ -22,6 +22,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
+	"sort"
 	"strings"
 
 	"github.com/google/go-containerregistry/pkg/legacy"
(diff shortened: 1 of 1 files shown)


pkg/legacy/tarball/write.go
```diff
@@ -297,10 +298,15 @@ func MultiWrite(refToImage map[name.Reference]v1.Image, w io.Writer) error {
 		addTags(repos, tags, prev.config.ID)
 	}
 
+	sort.Slice(m, func(i, j int) bool {
```
jonjohnsonjr, line 301: Do we not care about the order in which these images are written to the tarball? Just the manifest.json file that describes them?

grahamjenson: Not sure. I have tried in my use-case and this seems to work. I will have a look and validate that is the case.

grahamjenson: You were correct, I sorted all images by their `ConfigName()` so now all tarball contents are also sorted.

PR #748, opened 2020-07-28
crane: add a mirror command

josselin-c:
Mirror command can be used to mirror a whole repository.


jonjohnsonjr:
Rather than a separate command, what if this was just a flag on `crane cp`?

There's an `--all-tags` / `-a` flag for `docker pull` that we could pun on: https://docs.docker.com/engine/reference/commandline/pull/#options

E.g.:
`crane cp -a ubuntu example.com/ubuntu`

Issue #751, opened 2020-08-03
crane rebase and mutate.Rebase set empty-valued platform fields


micahyoung:
After rebasing, the `rebased` image has `platform` fields with empty values (`"os":""`, `"architecture":""`, `"os.version":""`, etc.) though the input images contained these fields with valid values.

Docker and k8s Linux containers allow pulling and running these platform-less images by adding their own platform but Windows containers fail to pull or run (which is how I noticed this issue).

It appears this is mostly implemented already in [`mutate.Time`](https://github.com/google/go-containerregistry/blob/d47c18fa2769e0d7bd121ad703abe364e978e7a2/pkg/v1/mutate/mutate.go#L282-L285) though some optional fields are missed (`os.features`, `variant`, `features` from [`platform` struct](https://github.com/google/go-containerregistry/blob/38ad4ec12b1d707307f901c6675d99e738f4d7e3/pkg/v1/platform.go#L22-L24)).

I feel the expected behavior would be to copy all platform fields with values ([spec'd under `platform`](https://github.com/opencontainers/image-spec/blob/243ea084a44451d27322fed02b682d99e2af3ba9/image-index.md#image-index-property-descriptions)) from the `new_base` image to the `rebased` image. 

I'll add a PR assuming there's no objection to the proposed behavior.

Repro:
```
# Local registry
docker run -p 5000:5000 -d --rm registry:2

# Tags to be used
original_image="localhost:5000/my-original-image"
rebased_image="localhost:5000/my-rebased-image"
old_base_image="alpine:3.11"
new_base_image="alpine:3.12"

# Build and push the original image
docker build . --tag ${original_image} -f <(cat <<EOF
FROM ${old_base_image}
RUN echo bar > /foo
EOF
)
docker push ${original_image}
docker rmi -f ${original_image}

# Rebase onto new base
crane rebase \
  --original ${original_image} \
  --old_base ${old_base_image} \
  --new_base ${new_base_image} \
  --rebased ${rebased_image}

# See missing fields on rebased image
crane config ${rebased_image} | jq ".os,.architecture"
## Output
""
""

# See existing platform fields on original and old base images
crane config ${original_image} | jq ".os,.architecture"
crane config ${old_base_image} | jq ".os,.architecture"
crane config ${new_base_image} | jq ".os,.architecture"
## Output
"linux"
"amd64"
"linux"
"amd64"
"linux"
"amd64"
```

ImJasonH:
[...]

Issue #730, opened 2020-06-09
crane rebase : problems with environment variables


eddie71:
I have two java images, one with 11.0.5 and one with 11.0.6. they are both based on the same template. For instance

FROM centos
ENV JAVA_HOME /usr/lib/jvm/java-11-openjdk-11.0.5.10-0.el8_0.x86_64

RUN yum update -y  && \
    yum install -y csh && \
    yum install -y  rng-tools && \
    yum install -y hostname && \
    yum install -y wget && \
    yum install -y passwd && \
    yum install -y nc && \
    yum install -y expect && \
    yum install -y unzip && \
    echo "password" | passwd --stdin root

USER root
RUN yum install -y java-11-openjdk-devel-11.0.5.10-0.el8_0.x86_64
RUN rm -rf /tmp/*

ENV PATH $JAVA_HOME/bin:$PATH
CMD ["/bin/bash"]

Java 11.0.6 version
-----------------
FROM centos
ENV JAVA_HOME /usr/lib/jvm/java-11-openjdk-11.0.6.10-0.el8_1.x86_64/

RUN yum update -y  && \
    yum install -y csh && \
    yum install -y  rng-tools && \
    yum install -y hostname && \
    yum install -y wget && \
    yum install -y passwd && \
    yum install -y nc && \
    yum install -y expect && \
    yum install -y unzip && \
    echo "password" | passwd --stdin root

USER root

RUN yum install -y java-11-openjdk-devel-11.0.6.10-0.el8_1.x86_64
RUN rm -rf /tmp/*

ENV PATH $JAVA_HOME/bin:$PATH
CMD ["/bin/bash"]

i created a simple app and extended 11.0.5
FROM java:11.0.5
ADD target/app-1.0-SNAPSHOT.jar /
CMD $JAVA_HOME/bin/java -cp /app-1.0-SNAPSHOT.jar com.test.Application

it works with 11.0.5 and if i build the image directly on 11.0.6, it works ok. It simply prints the java version and the value or JAVA_HOME

If i take the 11.0.5 and rebase it, i get the error
/usr/lib/jvm/java-11-openjdk-11.0.5.10-0.el8_0.x86_64/bin/java: No such file or directory
(its picking up the 11.0.5 env varaible still)

[...]

PR #752, opened 2020-08-04
Fix for rebase omitting os/arch config properties

micahyoung:
Fixes [#751] for missing `os`, `architecture` and `os.version` properties after `mutate.Rebase`

* `os`, `architecture` properties are required by config spec [[1]]

* `os.version` is unspec'd but required for Windows images to run on Windows containers [[2]]

This implementation takes these three properties from newBase, which most accurately defines the newly rebased image's os/architecture. If the properties were taken from original, it would prevent reuse of de-facto ABI-compatible top layers with differing old and new base images (i.e. data-only top layers rebased from amd64 to arm, Windows os.version upgrades). 

This does not address additional, optional config properties that are currently being omitted during rebase such as `author` and `created`, which I feel could be addressed in a separate issue.

[1]: https://github.com/opencontainers/image-spec/blob/master/config.md#properties
[2]: https://docs.microsoft.com/en-us/virtualization/windowscontainers/deploy-containers/version-compatibility#matching-container-host-version-with-container-image-versions

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -63,18 +63,16 @@ require (
 	github.com/timakin/bodyclose v0.0.0-20200424151742-cb6215831a94 // indirect
 	github.com/vdemeester/k8s-pkg-credentialprovider v1.17.4
 	github.com/xanzy/go-gitlab v0.32.0 // indirect
-	golang.org/x/crypto v0.0.0-20200510223506-06a226fb4e37 // indirect
 	golang.org/x/mod v0.3.0 // indirect
-	golang.org/x/net v0.0.0-20200520182314-0ba52f642ac2 // indirect
 	golang.org/x/oauth2 v0.0.0-20200107190931-bf48bf16ab8d
-	golang.org/x/sync v0.0.0-20200317015054-43a5402ce75a
+	golang.org/x/sync v0.0.0-20200625203802-6e8e738ad208
 	golang.org/x/sys v0.0.0-20200523222454-059865788121 // indirect
 	golang.org/x/time v0.0.0-20200416051211-89c76fbcd5d1 // indirect
-	golang.org/x/tools v0.0.0-20200527183253-8e7acdbce89d // indirect
+	golang.org/x/tools v0.0.0-20200804011535-6c149bb5ef0d // indirect
 	google.golang.org/api v0.25.0 // indirect
 	google.golang.org/genproto v0.0.0-20200527145253-8367513e4ece // indirect
 	gopkg.in/ini.v1 v1.56.0 // indirect
-	honnef.co/go/tools v0.0.1-2020.1.4 // indirect
+	honnef.co/go/tools v0.0.1-2020.1.5 // indirect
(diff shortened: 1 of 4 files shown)


jonjohnsonjr:
> If the properties were taken from original, it would prevent reuse of de-facto ABI-compatible top layers with differing old and new base images (i.e. data-only top layers rebased from amd64 to arm, Windows os.version upgrades).

Does this describe your current use case or is it just hypothetical?

I see now that my original comments might not have made much sense. We want, largely, the config from the original image to remain intact, and this change takes the os/arch from the new base such that you can re-use top layers across different architectures.

I'm wondering if this isn't universally true? For example, what if I want to use gcr.io/distroless/static as a base image for multiple architectures of a go binary? There's nothing platform-specific about the gcr.io/distroless/static base layers, and the platform-specific bits would be in the top layers (just the binary). This seems like the opposite scenario from what you've described.

I think we're looking at two separate issues:

1. Rebase drops all fields from `ConfigFile` other than `Config`, which is unexpected.
2. How do we handle conflicts between the original image and the new base image? This ties into https://github.com/google/go-containerregistry/issues/730 with environment variables, a bit.

Given that we currently don't set any platform info _at all_, I'm fine merging this and letting the new base image "win" for platform stuff, since you'd demonstrated a need for it, but we probably want some way to handle these conflicts e.g. to support the scenario I described.

@ImJasonH does what I'm saying here make sense? Do you have any problem with just merging this? I don't think there's any backward-compat issues.

micahyoung:
Hey @jonjohnsonjr I'd be happy to use original for the source of these fields if it makes more sense before we handle resolving configfile property conflicts. It will unblock me and I can handle cherry-picking the important values from new base after mutate.Rebase and letting everything else come from original.

I'm without power at home after a big storm in the eastern US so I wouldn't be able to adjust the PR until later today.

jonjohnsonjr:
> I'd be happy to use original for the source of these fields if it makes more sense to start, then later we can implement handling for resolving configfile property conflicts.

I thought about this a little bit more and I think the scenario I described is rather unlikely. Most base images contain platform-specific stuff, and distroless is a strange exception, so I'm content to merge this PR. I think it's a reasonable default, and we can provide other ways to achieve what I described.

> It will let me remove the workaround in my own codebase and in its place, I can handle cherry-picking the conflicting values from new base after mutate.Rebase and letting everything else come from original.

Let's merge this to unblock you downstream, but some follow-up that might be worthwhile:
* Indicate that platform is inherited from the new base image in the [rebase docs](https://github.com/google/go-containerregistry/blob/master/cmd/crane/rebase.md#caveats) (and maybe point to these docs from the `crane rebase` help text?)
* I've opened a separate issue for the rest of `ConfigFile` if we want to keep discussing it: https://github.com/google/go-containerregistry/issues/753

> I'm without power at home after a big storm in the eastern US so I wouldn't be able to adjust the PR until later today.

Sorry to hear that :( hope you're doing alright.

Thanks for the PR!

PR #759, opened 2020-08-24
Test go 1.14, not 1.12

jonjohnsonjr:
I noticed that go 1.12 with GO111MODULE=on takes over 20 minutes to run. Let's drop it in favor of 1.14, given that even 1.15 has been released already.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,8 +6,8 @@ language:
   - go
 
 go:
-  - "1.12"
   - "1.13"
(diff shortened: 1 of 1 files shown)


jonjohnsonjr:
Much better:

Screenshot 2020-08-24 at 10 53 07 AM

I suspect dropping 1.13 for 1.15 will cut this down to 5 min, which will be great.

PR #760, opened 2020-09-01
Fix incorrect tag size limit

alexey-igrychev:
A tag name must be valid ASCII and may contain lowercase and uppercase letters, digits, underscores, periods and dashes. A tag name may not start with a period or a dash and may contain a maximum of **128 characters**.


alexey-igrychev:
@googlebot I signed it!

PR #764, opened 2020-09-02
Add --service-account-key flag to gcrane ls

micnncim:
Added `--service-account-key` flag to `gcrane ls`. The default auth still uses the key chain but this PR allows us to optionally set a service account key.

```console
$ go run cmd/gcrane/main.go ls -h
List the contents of a repo

Usage:
  gcrane ls REPO [flags]

Flags:
  -h, --help                         help for ls
      --json                         Format the response from the registry as JSON, one line per repo
  -r, --recursive                    Whether to recurse through repos
      --service-account-key string   File path to service account key JSON

Global Flags:
  -v, --verbose   Enable debug logs

```

### DIFF
--- a/cmd/gcrane/cmd/list.go
+++ b/cmd/gcrane/cmd/list.go
@@ -17,6 +17,7 @@ package cmd
 import (
 	"encoding/json"
 	"fmt"
+	"io/ioutil"
 	"log"
 
 	"github.com/google/go-containerregistry/pkg/name"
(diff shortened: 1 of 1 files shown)


jonjohnsonjr:
I'm not strictly opposed to adding this, but does setting [`GOOGLE_APPLICATION_CREDENTIALS`](https://cloud.google.com/docs/authentication/getting-started#setting_the_environment_variable) to your service account key work instead?

It might be sufficient to just update docs if this works for you.

micnncim:
Thank you! I didn't know `GOOGLE_APPLICATION_CREDENTIALS`, whick looks good me me 👍 I'll close this PR.

PR #767, opened 2020-09-10
Update GGCR to K8s 1.18 libs

mattmoor:


mattmoor:
Beyond simply bumping the dependencies, the meat of this change is accommodating the breaking change in K8s 1.18 to expect `ctx` in all client-go calls.

ImJasonH (approved):

PR #770, opened 2020-09-16
add remote.Head

### LINKED ISSUE #769: Ability to send a HEAD request to get a manifest's digest
With Docker soon [enforcing pretty strict rate limits](https://www.docker.com/blog/scaling-docker-to-serve-millions-more-developers-network-egress/) we're scrambling a bit to prevent Concourse's image polling from grinding to a halt on November 1.

The `registry-image` resource `check` call currently uses `remote.Image` to fetch an image's digest. However this results in a `GET` request, which counts towards the rate limit.

Given that we only need the `sha256:...` digest string, which is available as the `Docker-Content-Digest` response header, would it be possible to add a `go-containerregistry` API which just sends a `HEAD` request and returns the hash? Something like `remote.ImageDigest(name.Reference, ...Option) (v1.Hash, error)`?

Happy to put together a PR if you've got any guidance.

vito:
fixes #769 

Lend me your nitpicks! There's a bit of copypasta here but I didn't want to over-aggressively DRY things up.

### DIFF
--- a/pkg/v1/remote/descriptor.go
+++ b/pkg/v1/remote/descriptor.go
@@ -22,6 +22,7 @@ import (
 	"io/ioutil"
 	"net/http"
 	"net/url"
+	"strconv"
 	"strings"
 
 	"github.com/google/go-containerregistry/pkg/logs"
@@ -76,6 +77,8 @@ func (d *Descriptor) RawManifest() ([]byte, error) {
 // Get returns a remote.Descriptor for the given reference. The response from
 // the registry is left un-interpreted, for the most part. This is useful for
 // querying what kind of artifact a reference represents.
+//
+// See Head if you don't need the response body.
 func Get(ref name.Reference, options ...Option) (*Descriptor, error) {
 	acceptable := []types.MediaType{
 		// Just to look at them.
@@ -85,6 +88,30 @@ func Get(ref name.Reference, options ...Option) (*Descriptor, error) {
 	acceptable = append(acceptable, acceptableImageMediaTypes...)
 	acceptable = append(acceptable, acceptableIndexMediaTypes...)
 	return get(ref, acceptable, options...)
+}
+
+// Head returns a v1.Descriptor for the given reference by issuing a HEAD
+// request.
+func Head(ref name.Reference, options ...Option) (*v1.Descriptor, error) {
+	acceptable := []types.MediaType{
+		// Just to look at them.
+		types.DockerManifestSchema1,
+		types.DockerManifestSchema1Signed,
+	}
+	acceptable = append(acceptable, acceptableImageMediaTypes...)
+	acceptable = append(acceptable, acceptableIndexMediaTypes...)
+
+	o, err := makeOptions(ref.Context(), options...)
+	if err != nil {
+		return nil, err
+	}
+
+	f, err := makeFetcher(ref, o)
+	if err != nil {
+		return nil, err
+	}
+
+	return f.headManifest(ref, acceptable)
 }
 
 // Handle options and fetch the manifest with the acceptable MediaTypes in the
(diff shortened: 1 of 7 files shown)


pkg/v1/remote/descriptor.go
```diff
@@ -87,6 +88,30 @@ func Get(ref name.Reference, options ...Option) (*Descriptor, error) {
 	return get(ref, acceptable, options...)
 }
 
+// Head returns a v1.Descriptor for the given reference. This is useful for
+// getting the current digest for a reference without impacting the rate limit.
+func Head(ref name.Reference, options ...Option) (*v1.Descriptor, error) {
```
jonjohnsonjr, line 93: I wouldn't mention the rate limit, since it's a registry-specific thing, just:

```
// Head returns a v1.Descriptor for the given reference by issuing a HEAD request.
```

Mentioning the HEAD request should be enough of a clue that this works around the docker hub rate limit.

I would update the comment for `Get` to mention that the response includes the manifest body,  and something like: "See Head if you don't need the response body."

pkg/v1/remote/descriptor.go
```diff
@@ -277,6 +302,55 @@ func (f *fetcher) fetchManifest(ref name.Reference, acceptable []types.MediaType
 	return manifest, &desc, nil
 }
 
+func (f *fetcher) fetchDescriptor(ref name.Reference, acceptable []types.MediaType) (*v1.Descriptor, error) {
```
jonjohnsonjr, line 305: Maybe add a short comment that this is like fetchManifest but using a HEAD request instead of GET.

jonjohnsonjr (commented):
Looks like the example test failed -- it talks to dockerhub, which we should fix.

You can just [skip](https://golang.org/pkg/testing/#hdr-Skipping) it to fix travis for now.

pkg/v1/remote/descriptor.go
jonjohnsonjr: Or maybe just rename to `headManifest`.

vito:
One thing I noticed when trying to use this: for multi-platform images, the digest you'll get back from `remote.Head(ref)` will be different from the digest you'll get from `remote.Image(ref).Digest()`, which is what we used before.

The former is the digest of the "manifest list", while the latter is the digest of the manifest for your particular platform, *contained* in said list.

I'm guessing there's no real way around this, since the platforms are discovered through the manifest, and the whole point of this is to not fetch the manifest. It's probably fine for our use case - it just means that the digest will change when any of the platforms change.

jonjohnsonjr:
> I'm guessing there's no real way around this, since the platforms are discovered through the manifest, and the whole point of this is to not fetch the manifest.

I don't believe there is a general solution to this. As you said, the platform resolution fundamentally requires parsing the manifest.

It's possible to request the amd64/linux image from the registry by not including manifest lists in the accept headers, but that only solves a subset of the problem and (IMO) makes things messier... I'm also pretty sure it only works if you request schema 1 images, which are deprecated, so probably not even a partial solution here.

> If the manifest being requested uses the new format, and the appropriate media type is not present in an Accept header, the registry will assume that the client cannot handle the manifest as-is, and rewrite it on the fly into the old format. If the object that would otherwise be returned is a manifest list, the registry will look up the appropriate manifest for the amd64 platform and linux OS, rewrite that manifest into the old format if necessary, and return the result to the client. If no suitable manifest is found in the manifest list, the registry will return a 404 error.

https://docs.docker.com/registry/spec/manifest-v2-2/#backward-compatibility

> It's probably fine for our use case - it just means that the digest will change when any of the platforms change.

In my experience, child manifests usually change together, so I imagine the false positive rate would be low. If you wanted, you could check the mediaType and resolve the list to whichever image you care about for manifest lists, but you need to know the desired platform ahead of time. I'm not familiar enough with your system to know if that's possible.

If possible, I'd just use the digest of the manifest list everywhere. Runtimes should know how to pull manifest lists, so the end result tends to be the same. I ran into issues with this on cri-o last year but they've since been resolved: https://github.com/knative/serving/issues/3997

cc @julz you may want to pick up this PR for knative digest resolution once it merges

jonjohnsonjr:
```
err: exit status 1: stderr: go: inconsistent vendoring in /home/travis/gopath/src/github.com/google/go-containerregistry:
	golang.org/x/tools@v0.0.0-20200916195026-c9a70fc28ce3: is explicitly required in go.mod, but not marked as explicit in vendor/modules.txt
	golang.org/x/tools@v0.0.0-20200911153331-7ad463ce66dd: is marked as explicit in vendor/modules.txt, but not explicitly required in go.mod
run 'go mod vendor' to sync, or use -mod=mod or -mod=readonly to ignore the vendor directory
```

`./hack/presubmit.sh` is your friend here.

vito:
> `./hack/presubmit.sh` is your friend here.

Woops! I ran it and saw the same failures but didn't expect my changes to affect any dependencies, so I assumed something was wrong on my machine. Pushed a commit to resolve this.

Side note: I haven't skipped the failing test from before but I'll push a commit for that too if it flakes out again.

Thanks for your help!

jonjohnsonjr (approved):
Thank you for doing this!

PR #776, opened 2020-09-27
Add WithContext to google.ListerOption

micnncim:
Added `WithContext` function, which implements `google.ListerOption`. This enables us to pass `context.Context` so that we could cancel sending HTTP requests. This PR won't change the default behavior since [`(*http.Client).Get`](https://golang.org/src/net/http/client.go?s=16191:16251#L460) wraps [`http.NewRequest`](https://golang.org/src/net/http/request.go?s=27051:27120#L802) and [`(*http.Client).Do`](https://golang.org/src/net/http/client.go?s=20422:20474#L575) and `http.NewRequest` wraps [`http.NewRequestWithContext`](https://golang.org/src/net/http/request.go?s=28365:28466#L828) with `context.Background`.

### DIFF
--- a/pkg/v1/google/list.go
+++ b/pkg/v1/google/list.go
@@ -15,6 +15,7 @@
 package google
 
 import (
+	"context"
 	"encoding/json"
 	"fmt"
 	"net/http"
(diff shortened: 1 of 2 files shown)


jonjohnsonjr (approved):
Looks great, thanks!

Sorry for the delay, I've been away on vacation.

PR #788, opened 2020-10-20
Add --platform CLI option for crane

dsbrng25b:
Adds a `--platform` option for the crane CLI tool.

I added a if-statement in `pkg/crane/options.go` that we don't get a nil pointer dereference if WithPlatform is called with nil. We could also catch this in cmd/crane/cmd/root.go` but I think it would in general make sense to avoid the nil pointer dereference.

I changed the error output in `pkg/v1/remote/index.go` to the form `os/arch`. This is not necessary but i think we should use the same form in the error output and the flag. I think this is quite arbitrary but I prefer os/arch over arch/os.

### DIFF
--- a/cmd/crane/cmd/root.go
+++ b/cmd/crane/cmd/root.go
@@ -13,23 +13,28 @@
 package cmd
 
 import (
+	"fmt"
 	"net/http"
 	"os"
+	"strings"
 
 	"github.com/docker/cli/cli/config"
 	"github.com/google/go-containerregistry/pkg/crane"
 	"github.com/google/go-containerregistry/pkg/logs"
+	v1 "github.com/google/go-containerregistry/pkg/v1"
 	"github.com/spf13/cobra"
 )
 
 func init() {
 	Root.PersistentFlags().BoolVarP(&verbose, "verbose", "v", false, "Enable debug logs")
 	Root.PersistentFlags().BoolVar(&insecure, "insecure", false, "Allow image references to be fetched without TLS")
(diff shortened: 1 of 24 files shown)


cmd/crane/cmd/root.go
```diff
@@ -80,3 +87,31 @@ func (ht *headerTransport) RoundTrip(in *http.Request) (*http.Response, error) {
 	}
 	return ht.inner.RoundTrip(in)
 }
+
+type platformValue struct {
+	platform *v1.Platform
+}
+
+func (pv *platformValue) Set(platform string) error {
```
jonjohnsonjr, line 95: We'll probably also want to allow a third value for the variant. We do this in [ko](https://github.com/google/ko/blob/9b910355628d9f9bb7a386fd42608ff4dc34c4a4/pkg/commands/config.go#L75-L87) and I'd like to be consistent.

[containerd](https://github.com/containerd/containerd/blob/ba8377590d326a5f73e4abe2d282c46c385329ae/platforms/platforms.go#L69) also allows this, and in the fullness of time I'd like to be compatible with what they allow, but for now let's just do the os/arch/variant form.

jonjohnsonjr:
> This is not necessary but i think we should use the same form in the error output and the flag. I think this is quite arbitrary but I prefer os/arch over arch/os.

I think that's fine, we use that same order in ko, so that works for me: https://github.com/google/ko#multi-platform-images

dsbrng25b:
Added support for variant.

jonjohnsonjr (approved):
Thanks so much!

jonjohnsonjr:
Ah the travis failures -- you'll need to run `./hack/update-codegen.sh` to regenerate the docs.

PR #791, opened 2020-10-21
Return basic auth from google.Keychain

jonjohnsonjr:
Before, we were taking advantage of the fact that GCR will allow you to
just send an access token as bearer auth, but apparently this breaks
`gcrane cp` if it's your first push to a registry, so this change
returns basic auth to force us to always do a token exchange.


jonjohnsonjr:
FYI @hosungs

hosungs (approved):
LGTM

ImJasonH (approved):

PR #793, opened 2020-10-22
Remove hidden unit test dependency on GCR creds

ImJasonH:
By having pkg/crane/copy_test.go set up its fake registry at `gcr.io`, this seems to trigger behavior that ends up looking for  default GCP credentials to authorize pushes/pulls. This works fine when tests run in GCB, and presumably on most of our local workstations, where GCP creds happen to be available, but [fails on GitHub Actions](https://github.com/google/go-containerregistry/pull/792#issuecomment-714163492), where creds are not available.

Instead of calling our fake registry "gcr.io", this change renames it to "xcr.io".

This change includes the Unit Tests action being added by #792, to check that this change fixes the unit test failure.

### DIFF
--- a/.github/workflows/knative-go-test.yaml
+++ b/.github/workflows/knative-go-test.yaml
@@ -1,0 +1,65 @@
+# Copyright 2020 The Knative Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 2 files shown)


ImJasonH:
cc @mattmoor

mattmoor (approved):

PR #797, opened 2020-10-23
Add partial.ImageByPlatform helper

ImJasonH:
This is used by ko now in a [couple](https://github.com/google/ko/pull/227) [places](https://github.com/google/ko/blob/d9cc0ca9ea6754cb57e2e63a5a1c88f0a43bd81c/pkg/publish/kind.go#L58), and it seems like a useful
helper to provide in general.

Reuse this helper in `pkg/v1/remote/`


pkg/v1/partial/index.go
```diff
@@ -0,0 +1,54 @@
...
+
+// ImageByArch returns the v1.Image belonging to the v1.ImageIndex with the
+// matching OS and Architecture.
+func ImageByArch(w WithRawIndexManifest, os, arch string) (v1.Image, error) {
+	im, err := IndexManifest(w)
+	if err != nil {
+		return nil, err
+	}
+	for _, manifest := range im.Manifests {
+		if manifest.Platform != nil &&
```
jonjohnsonjr, line 47: I'd like to consolidate this logic if we can. There's already this: https://github.com/google/go-containerregistry/blob/919148f07b7c222ede2b9cf228e946b66a955808/pkg/v1/remote/index.go#L193-L221

I'm also not 100% sure that the existing logic there is sound (haven't read the spec carefully), but at the very least we are missing variant, which our platform flags allow you to specify.

pkg/v1/partial/index.go
```diff
@@ -0,0 +1,54 @@
...
+	b, err := w.RawManifest()
+	if err != nil {
+		return nil, err
+	}
+	return v1.ParseIndexManifest(bytes.NewReader(b))
+}
+
+// ImageByArch returns the v1.Image belonging to the v1.ImageIndex with the
+// matching OS and Architecture.
+func ImageByArch(w WithRawIndexManifest, os, arch string) (v1.Image, error) {
```
jonjohnsonjr, line 41: I'd almost rather have this return a `WithRawManifest` or something given that, in theory, an index could have a child index with a platform field, but I understand that it would make the calling code more cumbersome.

Since child image is by far the common case, I'm fine merging this, but I'm not 100% happy about it :)

pkg/v1/partial/index.go
```diff
@@ -0,0 +1,54 @@
...
+	im, err := IndexManifest(w)
+	if err != nil {
+		return nil, err
+	}
+	for _, manifest := range im.Manifests {
+		if manifest.Platform != nil &&
+			manifest.Platform.OS == os &&
+			manifest.Platform.Architecture == arch {
+			return w.Image(manifest.Digest)
```
jonjohnsonjr, line 50: Should we check the descriptor's media type or just let the index return an error if this is wrong?

pkg/v1/partial/index.go
ImJasonH: How about `ImageByPlatform` that takes a [`v1.Platform`](https://godoc.org/github.com/google/go-containerregistry/pkg/v1#Platform) which must match?

pkg/v1/partial/index.go
ImJasonH: I tthink we can just let `w.Image` return an error in that case. It'd take a sub-manifest list reporting a platform to trigger it, which is either nonsense or invalid (or both!)

pkg/v1/partial/index.go
```diff
@@ -0,0 +1,84 @@
...
+	}
+	for _, manifest := range im.Manifests {
+		if manifest.Platform != nil && platformsEqual(*manifest.Platform, p) {
+			return w.Image(manifest.Digest)
+		}
+	}
+	return nil, errors.New("failed to find image for platform in index")
+}
+
+func platformsEqual(a, b v1.Platform) bool {
```
jonjohnsonjr, line 54: This is too naive -- see the equivalent logic remote.

We basically need to answer the question: is there an image that satisfies the requirements of the given platform. E.g. if there is an image with the superset of features, that's fine.

ImJasonH: Oh yeah, absolutely right. Now just wholesale stealing the logic from pkg/v1/remote/ 😄

pkg/v1/remote/index.go
```diff
@@ -122,42 +117,6 @@ func (r *remoteIndex) ImageIndex(h v1.Hash) (v1.ImageIndex, error) {
...
-//
-// But first we'd need to migrate to:
-//   github.com/opencontainers/image-spec/specs-go/v1
-func (r *remoteIndex) childByPlatform(platform v1.Platform) (*Descriptor, error) {
-	index, err := r.IndexManifest()
-	if err != nil {
-		return nil, err
-	}
-	for _, childDesc := range index.Manifests {
-		// If platform is missing from child descriptor, assume it's amd64/linux.
```
jonjohnsonjr, line 148: Do you think it's fine to drop this?

pkg/v1/remote/index.go
```diff
@@ -122,42 +117,6 @@ func (r *remoteIndex) ImageIndex(h v1.Hash) (v1.ImageIndex, error) {
...
-		p := defaultPlatform
-		if childDesc.Platform != nil {
-			p = *childDesc.Platform
-		}
-
-		if matchesPlatform(p, platform) {
-			return r.childDescriptor(childDesc, platform)
-		}
-	}
-	return nil, fmt.Errorf("no child with platform %s/%s in index %s", platform.OS, platform.Architecture, r.Ref)
```
jonjohnsonjr, line 158: Let's have partial return a similar error.

jonjohnsonjr: And maybe fix up some things to include the ref in the equivalent error from remote? I'm not sure how useful it is to include.

pkg/v1/remote/index.go
ImJasonH: Re-adding.

jonjohnsonjr (approved):

pkg/v1/partial/index.go
```diff
@@ -0,0 +1,107 @@
...
+)
+
+var defaultPlatform = v1.Platform{
+	Architecture: "amd64",
+	OS:           "linux",
+}
+
+// WithRawIndexManifest defines the subset of v1.ImageIndex used by ImageByArch.
+type WithRawIndexManifest interface {
+	RawManifest() ([]byte, error)
```
jonjohnsonjr, line 31: I'm actually concerned that this is too general because a `v1.Image` that also implements `Image(Hash)` will satisfy this, but will not work at all because `v1.Image`'s descriptors are under `Layers`. Should we just have `ImageByPlatform` take a `v1.ImageIndex` instead?

jonjohnsonjr (changes requested):

ImJasonH:
Closing in favor of https://github.com/google/go-containerregistry/pull/823

PR #798, opened 2020-10-24
Add remote.MultiWrite

ImJasonH:
This method takes multiple images or manifest lists, and attempts to
efficiently upload all of them in parallel, deduplicating shared layer
and config blobs and avoiding unnecessary auth token exchanges.

Blobs are uploaded in parallel first, then manifests are tagged in order
with as much parallelism as we can support.

This method is currently limited in that:
- all images must be getting uploaded to the same repository (to
  simplify auth and transports), and
- no image may contain streaming layers (to simplify digests), but both
  of these can be removed in future changes.
- blobs are uploaded and manifests are tagged in parallel with a
  constant max parallelism, which should be a user-controlled option.

### DIFF
--- a/pkg/registry/manifest.go
+++ b/pkg/registry/manifest.go
@@ -23,6 +23,9 @@ import (
 	"net/http"
 	"strings"
 	"sync"
+
+	v1 "github.com/google/go-containerregistry/pkg/v1"
+	"github.com/google/go-containerregistry/pkg/v1/types"
 )
 
 type manifest struct {
@@ -123,6 +126,33 @@ func (m *manifests) handle(resp http.ResponseWriter, req *http.Request) *regErro
 			blob:        b.Bytes(),
 			contentType: req.Header.Get("Content-Type"),
 		}
+
+		// If the manifest is a manifest list, check that the manifest
+		// list's constituent manifests are already uploaded.
+		// This isn't strictly required by the registry API, but some
+		// registries require this.
+		if mf.contentType == string(types.OCIImageIndex) ||
+			mf.contentType == string(types.DockerManifestList) {
+
+			im, err := v1.ParseIndexManifest(b)
+			if err != nil {
+				return &regError{
+					Status:  http.StatusNotFound,
+					Code:    "MANIFEST_UNKNOWN",
+					Message: err.Error(),
+				}
+			}
+			for _, desc := range im.Manifests {
+				if _, found := m.manifests[repo][desc.Digest.String()]; !found {
+					return &regError{
+						Status:  http.StatusNotFound,
+						Code:    "MANIFEST_UNKNOWN",
+						Message: fmt.Sprintf("Sub-manifest %q not found", desc.Digest),
+					}
+				}
+			}
+		}
+
 		// Allow future references by target (tag) and immutable digest.
 		// See https://docs.docker.com/engine/reference/commandline/pull/#pull-an-image-by-digest-immutable-identifier.
 		m.manifests[repo][target] = mf
--- a/pkg/v1/remote/multi_write.go
+++ b/pkg/v1/remote/multi_write.go
@@ -1,0 +1,245 @@
+// Copyright 2020 Google LLC All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package remote
+
+import (
+	"fmt"
+	"net/http"
+
+	"github.com/google/go-containerregistry/pkg/name"
+	v1 "github.com/google/go-containerregistry/pkg/v1"
+	"github.com/google/go-containerregistry/pkg/v1/partial"
+	"github.com/google/go-containerregistry/pkg/v1/remote/transport"
+	"github.com/google/go-containerregistry/pkg/v1/types"
+	"golang.org/x/sync/errgroup"
+)
+
+// Parallelism of blob and manifest uploads
+// TODO(jasonhall): Make this an Option.
+const jobs = 4
+
+// MultiWrite writes the given Images or ImageIndexes to the given refs, as
+// efficiently as possible, by deduping shared layer blobs and uploading layers
+// in parallel, then uploading all manifests in parallel.
+//
+// Current limitations:
+// - All refs must share the same repository.
+// - Images cannot consist of stream.Layers.
+func MultiWrite(m map[name.Reference]Taggable, options ...Option) error {
+	// Determine the repository being pushed to; if asked to push to
+	// multiple repositories, give up.
+	var repo, zero name.Repository
+	for ref := range m {
+		if repo == zero {
+			repo = ref.Context()
+		} else if ref.Context() != repo {
+			return fmt.Errorf("MultiWrite can only push to the same repository (saw %q and %q)", repo, ref.Context())
+		}
+	}
+
+	// Collect unique blobs (layers and config blobs).
+	blobs := map[v1.Hash]v1.Layer{}
+	newManifests := []map[name.Reference]Taggable{}
+	// Separate originally requested images and indexes, so we can push images first.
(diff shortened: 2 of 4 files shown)


pkg/v1/remote/multi_write.go
```diff
@@ -0,0 +1,172 @@
...
+		case types.OCIManifestSchema1, types.DockerManifestSchema2:
+			img, err := idx.Image(desc.Digest)
+			if err != nil {
+				return err
+			}
+			if err := addImageBlobs(img, blobs); err != nil {
+				return err
+			}
+		default:
+			return fmt.Errorf("unknown media type: %v", desc.MediaType)
```
jonjohnsonjr, line 141: We need something like `v1.ImageIndex.Blob` to access non-manifest artifacts that are referenced by an index. Making the interface wider would be a breaking change, which is annoying, but I doubt it would affect much code outside of this repo...

pkg/v1/remote/multi_write.go
```diff
@@ -0,0 +1,172 @@
...
+		mt, err := b.MediaType()
+		if err != nil {
+			return err
+		}
+		if !mt.IsDistributable() {
+			// TODO(jonjohnsonjr): Add "allow-nondistributable-artifacts" option.
+			continue
+		}
+
+		g.Go(func() error { return w.uploadOne(b) })
```
jonjohnsonjr, line 101: This will end up being too wide for large repositories. We'll want to limit the concurrency to something reasonable by default with an option to override (and backport that to the other Write functions).

pkg/v1/remote/multi_write.go
```diff
@@ -0,0 +1,172 @@
...
+	if err := g.Wait(); err != nil {
+		return err
+	}
+
+	// With all of the constituent elements uploaded, upload the manifests
+	// to commit the images and indexes.
+	for ref, i := range m {
+		ref := ref
+		i := i
+		g.Go(func() error { return w.commitImage(i, ref) })
```
jonjohnsonjr, line 112: Same comment as above, this is probably too wide.

pkg/v1/remote/multi_write_test.go
```diff
@@ -0,0 +1,77 @@
...
+	rl, err := random.Layer(1024, types.OCIUncompressedLayer)
+	if err != nil {
+		t.Fatalf("random.Layer: %v", err)
+	}
+	img2, err := mutate.AppendLayers(img1, rl)
+	if err != nil {
+		t.Fatalf("mutate.AppendLayers: %v", err)
+	}
+
+	// Create an index containing both images above.
```
jonjohnsonjr, line 34: You could add a random index here to get more coverage on the recursive index path.

pkg/v1/remote/multi_write.go
```diff
@@ -0,0 +1,172 @@
...
+	if err != nil {
+		return err
+	}
+	// Determine if any of the layers are Mountable, because if so we need
+	// to request Pull scope too.
+	ls := []v1.Layer{}
+	for _, l := range blobs {
+		ls = append(ls, l)
+	}
+	scopes := scopesForUploadingImage(repo, ls)
```
jonjohnsonjr, line 75: I'm a little afraid that this would end up making the URL too large if we have a ton of mountable layers from a large number of repositories, but that seems unlikely -- reasonable to ignore unless we hit it.

pkg/v1/remote/multi_write.go
```diff
@@ -0,0 +1,172 @@
...
+	var g errgroup.Group
+	for _, b := range blobs {
+		b := b
+
+		// Handle foreign layers.
+		mt, err := b.MediaType()
+		if err != nil {
+			return err
+		}
+		if !mt.IsDistributable() {
```
jonjohnsonjr, line 96: Weird edge case we probably shouldn't worry about:

What if we have two layers with the same digest but different media types? Since we dedupe by just the digest, this could lead to broken images if the first layer we encounter is foreign and the second we encounter (and skip because of dedupe) is not foreign. The second layer's manifest PUT would fail because the layer wasn't uploaded and isn't foreign.

Actually maybe we should worry about it. Just skip adding them to the map in the first place if they're not distributable instead of checking it here?

pkg/v1/remote/multi_write.go
ImJasonH: Done. Going with a const parallelism for now with a TODO.

pkg/v1/remote/multi_write.go
ImJasonH: Done.

pkg/v1/remote/multi_write_test.go
ImJasonH: Done, though this seems to break some tests. Looking into it...

pkg/v1/remote/multi_write.go
ImJasonH: Good call, but I agree we should wait to see if/how it bites us before handling it.

pkg/v1/remote/multi_write.go
ImJasonH: Good idea, done.

pkg/v1/remote/multi_write.go
```diff
@@ -0,0 +1,207 @@
...
+		repo:    repo,
+		client:  &http.Client{Transport: tr},
+		context: o.context,
+	}
+
+	// Upload individual blobs and collect any errors.
+	blobChan := make(chan v1.Layer, 2*jobs)
+	var g errgroup.Group
+	for i := 0; i < jobs; i++ {
+		// Start N workers consuminng blobs to upload.
```
jonjohnsonjr, line 94:
```suggestion
		// Start N workers consuming blobs to upload.
```

pkg/v1/remote/multi_write.go
```diff
@@ -0,0 +1,207 @@
...
+
+	// With all of the constituent elements uploaded, upload the manifests
+	// to commit the images and indexes, and collect any errors.
+	type task struct {
+		i   Taggable
+		ref name.Reference
+	}
+	taskChan := make(chan task, 2*jobs)
+	for i := 0; i < jobs; i++ {
+		// Start N workers consuminng tasks to upload manifests.
```
jonjohnsonjr, line 122:
```suggestion
		// Start N workers consuming tasks to upload manifests.
```

ImJasonH:
Talked with Jon yesterday, some of this is more complex than we'd expected, since some registries (e.g., GCR) require that all of a manifest list's referenced manifests already be uploaded, so we can't just blindly copy manifests up at once, we have to do it in an order.

Since a manifest list can reference other manifest lists, etc., we get a tree of them that has to be uploaded, leafs before roots. Each level can be uploaded in parallel, but has to complete before the next level up starts.

I'm adding this check to our test registry in `pkg/registry` so we'll be able to test it's done right.

pkg/registry/manifest.go
```diff
@@ -23,6 +23,9 @@ import (
 	"net/http"
 	"strings"
 	"sync"
+
+	v1 "github.com/google/go-containerregistry/pkg/v1"
```
jonjohnsonjr, line 27: I'd prefer it if we inlined some of this stuff to avoid taking a dependency outside of the stdlib for this package, but I might just fix it myself before we cut a new release.

pkg/registry/manifest.go
```diff
@@ -123,6 +126,33 @@ func (m *manifests) handle(resp http.ResponseWriter, req *http.Request) *regErro
 			blob:        b.Bytes(),
 			contentType: req.Header.Get("Content-Type"),
 		}
+
+		// If the manifest is a manifest list, check that the manifest
+		// list's constituent manifests are already uploaded.
+		// This isn't strictly required by the registry API, but some
+		// registries require this.
+		if mf.contentType == string(types.OCIImageIndex) ||
```
jonjohnsonjr, line 134: > 86.20% <7.69%> (-13.80%)

This used to have 100% coverage 😄

pkg/v1/remote/multi_write.go
```diff
@@ -0,0 +1,245 @@
...
+const jobs = 4
+
+// MultiWrite writes the given Images or ImageIndexes to the given refs, as
+// efficiently as possible, by deduping shared layer blobs and uploading layers
+// in parallel, then uploading all manifests in parallel.
+//
+// Current limitations:
+// - All refs must share the same repository.
+// - Images cannot consist of stream.Layers.
+func MultiWrite(m map[name.Reference]Taggable, options ...Option) error {
```
jonjohnsonjr, line 40: This API is good. I have some issues with the implementation and limitations, which we've discussed offline. Given that the interface is correct, I'm fine with merging this and improving it as we have time.

jonjohnsonjr (approved):

PR #803, opened 2020-11-02
Add remote.WithJobs and use it in remote.MultiWrite

ImJasonH:
- default parallel is 4
- an error is returned if WithParallel is called with a value <=0 or >100

We can probably reuse this to fetch and push blobs elsewhere in remote.

MultiWrite'ing 123 distroless images with `WithParallel(100)` took 4s compared to 11s for the default.

### DIFF
--- a/pkg/v1/remote/options.go
+++ b/pkg/v1/remote/options.go
@@ -16,6 +16,7 @@ package remote
 
 import (
 	"context"
+	"errors"
 	"net/http"
 
 	"github.com/google/go-containerregistry/pkg/authn"
@@ -33,7 +34,15 @@ type options struct {
 	transport http.RoundTripper
 	platform  v1.Platform
 	context   context.Context
+	jobs      int
 }
+
+var defaultPlatform = v1.Platform{
+	Architecture: "amd64",
+	OS:           "linux",
+}
+
+const defaultJobs = 4
 
 func makeOptions(target authn.Resource, opts ...Option) (*options, error) {
 	o := &options{
@@ -41,6 +50,7 @@ func makeOptions(target authn.Resource, opts ...Option) (*options, error) {
 		transport: http.DefaultTransport,
 		platform:  defaultPlatform,
 		context:   context.Background(),
+		jobs:      defaultJobs,
 	}
 
 	for _, option := range opts {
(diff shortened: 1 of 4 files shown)


pkg/v1/remote/options.go
```diff
@@ -131,3 +134,23 @@ func WithContext(ctx context.Context) Option {
 		return nil
 	}
 }
+
+// WithParallel is a functional option for setting the parallelism of remote
+// operations performed by a given function. Note that not all remote
+// operations support parallelism.
+//
+// The default parallel is 4.
+func WithParallel(parallel int) Option {
```
jonjohnsonjr, line 143: I called this WithJobs for gcrane, not sure if we want to keep this consistent: https://github.com/google/go-containerregistry/blob/ab3252bfe3ddb5e612349ff3fde459864cc6e8fb/cmd/gcrane/cmd/copy.go#L43

pkg/v1/remote/options.go
```diff
@@ -131,3 +134,23 @@ func WithContext(ctx context.Context) Option {
...
+	return func(o *options) error {
+		if parallel <= 0 {
+			return errors.New("parallel must be greater than zero")
+		}
+		if parallel > 100 {
+			return errors.New("parallel must be <= 100")
```
jonjohnsonjr, line 149: but why?

ImJasonH: Removed.

pkg/v1/remote/options.go
ImJasonH: WithJobs it is!

pkg/v1/remote/options.go
```diff
@@ -33,14 +34,18 @@ type options struct {
 	transport http.RoundTripper
 	platform  v1.Platform
 	context   context.Context
+	jobs      int
 }
 
+const defaultJobs = 4
```
jonjohnsonjr, line 40: While you're in here, can you move `defaultPlatform` to this file as well?

jonjohnsonjr (commented):
Can you add a TODO to `Write` and `WriteIndex` to respect `WithJobs`?

jonjohnsonjr (approved):

PR #804, opened 2020-11-03
Fix error when calculating tarball size of manifest

tydra-wang:
Manifest is marshalled repeatedly.


jonjohnsonjr:
cc @deitch can you take a look?

deitch (approved):
It makes sense to me. I cannot understand how it worked before, unless I simply don't understand it now.

jonjohnsonjr (approved):
Thanks!

tydra-wang:
> It makes sense to me. I cannot understand how it worked before, unless I simply don't understand it now.

The `size` is only used to calculate the progress.
And the tarball padding makes error happens occasionally.

Issue #810, opened 2020-11-04
gcrane debug container does not have shell


kgaurav2:
gcrane latest debug image does not have shell. https://console.cloud.google.com/gcr/images/go-containerregistry/GLOBAL/gcrane@sha256:bbc477dc1d0e3426f2c034f07d2c8896e1e0e6ac460e5d07d7db9881f43e8703
```
docker run --rm -it --entrypoint sh gcr.io/go-containerregistry/gcrane:debug
docker: Error response from daemon: OCI runtime create failed: container_linux.go:349: starting container process caused "exec: \"sh\": executable file not found in $PATH": unknown
```

jonjohnsonjr:
Looks like a typo in https://github.com/google/go-containerregistry/pull/802 🤦

jonjohnsonjr:
Looks like it worked, sorry about that:

```
$ crane export gcr.io/go-containerregistry/gcrane:debug - | tar -tvf - | grep busybox | head
tar: Removing leading `/' from member names
drwxr-xr-x 0/0               0 1969-12-31 16:00 ./busybox/
lrwxrwxrwx 0/0               0 1969-12-31 16:00 ./busybox/stat -> busybox
lrwxrwxrwx 0/0               0 1969-12-31 16:00 ./busybox/[ -> busybox
lrwxrwxrwx 0/0               0 1969-12-31 16:00 ./busybox/mount -> busybox
lrwxrwxrwx 0/0               0 1969-12-31 16:00 ./busybox/more -> busybox
lrwxrwxrwx 0/0               0 1969-12-31 16:00 ./busybox/xz -> busybox
lrwxrwxrwx 0/0               0 1969-12-31 16:00 ./busybox/softlimit -> busybox
lrwxrwxrwx 0/0               0 1969-12-31 16:00 ./busybox/cat -> busybox
lrwxrwxrwx 0/0               0 1969-12-31 16:00 ./busybox/wget -> busybox
lrwxrwxrwx 0/0               0 1969-12-31 16:00 ./busybox/xzcat -> busybox
```

neclimdul:
thanks for the quick fix. Got really unlucky tagging images in that short window.

kgaurav2:
thanks guys for such a quick response.

jonjohnsonjr:
No problem -- we really need tests for this 🤷

PR #811, opened 2020-11-04
Fix debug images

jonjohnsonjr:
Fixes https://github.com/google/go-containerregistry/issues/810


ImJasonH (approved):
Could add a test in cloudbuild.yaml that debug has a shell too to prevent future typos🤔

PR #815, opened 2020-11-05
expose v1/layout.WriteImage and WriteIndex

### LINKED ISSUE #813: Provide ability to update an image layout, ensuring all of its content is available
When using `v1/layout`, if I run `Path.AppendIndex()` / `Path.AppendImage()`, there is no way to say:

1. Update any missing blobs
1. Do not update the `index.json`, because there already is an entry

The `AppendIndex`/`AppendImage` already call `writeIndex`/`writeImage`, which sanely check each blob down the tree to see if it is there and, if not, then download it.

But the rest of `AppendIndex`/`AppendImage` creates the descriptor for use in `index.json` and then writes it. Most of the time this makes sense, but if the exact same descriptor already is there, it doesn't.

Practically, this means I can add a new image, or download all the missing blobs for an image and add it a second time, but I have no way to just download the missing blobs.

I see two ways to do this:

* make `writeIndex`/`writeImage` public, so someone can simply choose to call `WriteIndex` if they do not want to update `index.json`, or `AppendIndex` if they do
* make `AppendDescriptor` idempotent. If it is the _exact_ same descriptor, do not write it twice.

I am good with either of these, and happy to get a PR in if it will get reviewed and pulled.

Discussed offline with @jonjohnsonjr

deitch:
Fixes #813

### DIFF
--- a/pkg/v1/layout/write.go
+++ b/pkg/v1/layout/write.go
@@ -34,7 +34,7 @@ var layoutFile = `{
 // AppendImage writes a v1.Image to the Path and updates
 // the index.json to reference it.
 func (l Path) AppendImage(img v1.Image, options ...Option) error {
-	if err := l.writeImage(img); err != nil {
+	if err := l.WriteImage(img); err != nil {
 		return err
 	}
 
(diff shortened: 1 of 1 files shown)


jonjohnsonjr:
@mattmoor @ImJasonH this CodeQL things sucks, I often do want to log credentials for debugging reasons.

mattmoor:
Logging credentials is a huge security problem, logs are often far more accessible than the credentials potentially logged to them.  This has resulted in a large number of security incidents.

jonjohnsonjr (approved):

jonjohnsonjr:
Thanks!

PR #820, opened 2020-11-08
feat(gcrane): add gc pattern and upload date filters

b4nst:
This PR add `--before` add `--pattern` to `gcrane gc`.

`--before <timestamp>` will avoid images uploaded after `<timestamp>` being collected.
`--pattern <regex>` if set, will also collect images that have not been tagged with at least one tag matching this pattern.

### DIFF
--- a/cmd/gcrane/cmd/gc.go
+++ b/cmd/gcrane/cmd/gc.go
@@ -16,7 +16,6 @@ package cmd
 
 import (
 	"fmt"
-	"log"
 	"regexp"
 	"time"
 
@@ -36,11 +35,8 @@ func NewCmdGc() *cobra.Command {
 		Use:   "gc",
 		Short: "List images that are not tagged",
 		Args:  cobra.ExactArgs(1),
-		Run: func(_ *cobra.Command, args []string) {
-			err := gc(args[0], recursive, time.Now().Add(-before), pattern)
-			if err != nil {
-				log.Fatalln(err)
-			}
+		RunE: func(_ *cobra.Command, args []string) error {
+			return gc(args[0], recursive, time.Now().Add(-before), pattern)
 		},
 	}
 
--- a/cmd/gcrane/cmd/gc.go
+++ b/cmd/gcrane/cmd/gc.go
@@ -29,7 +29,7 @@ import (
 // NewCmdGc creates a new cobra.Command for the gc subcommand.
 func NewCmdGc() *cobra.Command {
 	recursive := false
-	before := int64(-1)
+	before := time.Duration(0)
 	pattern := ""
 
 	cmd := &cobra.Command{
@@ -37,12 +37,12 @@ func NewCmdGc() *cobra.Command {
 		Short: "List images that are not tagged",
 		Args:  cobra.ExactArgs(1),
 		Run: func(_ *cobra.Command, args []string) {
-			gc(args[0], recursive, time.Unix(before, 0), pattern)
+			gc(args[0], recursive, time.Now().Add(-before), pattern)
 		},
 	}
 
 	cmd.Flags().BoolVarP(&recursive, "recursive", "r", false, "Whether to recurse through repos")
-	cmd.Flags().Int64VarP(&before, "before", "b", time.Now().Unix(), "Oldest upload time (unix timestamp)")
+	cmd.Flags().DurationVarP(&before, "before", "b", time.Duration(0), "Age of the oldest uploaded image to print (ns, us, ms, s, m, h)")
 	cmd.Flags().StringVarP(&pattern, "pattern", "p", "", "Will also collect images with no tags matching this pattern")
 
 	return cmd
--- a/cmd/gcrane/cmd/gc.go
+++ b/cmd/gcrane/cmd/gc.go
@@ -108,6 +108,9 @@ func collector(filters []func(manifest google.ManifestInfo) bool) func(repo name
 				}
 			}
 			if collect {
+				for _, tag := range manifest.Tags {
+					fmt.Printf("%s:%s\n", repo, tag)
+				}
 				fmt.Printf("%s@%s\n", repo, digest)
 			}
 		}
--- a/cmd/gcrane/cmd/gc.go
+++ b/cmd/gcrane/cmd/gc.go
@@ -37,7 +37,10 @@ func NewCmdGc() *cobra.Command {
 		Short: "List images that are not tagged",
 		Args:  cobra.ExactArgs(1),
 		Run: func(_ *cobra.Command, args []string) {
-			gc(args[0], recursive, time.Now().Add(-before), pattern)
+			err := gc(args[0], recursive, time.Now().Add(-before), pattern)
+			if err != nil {
+				log.Fatalln(err)
+			}
 		},
 	}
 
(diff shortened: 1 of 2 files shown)


cmd/gcrane/cmd/gc.go
```diff
@@ -29,28 +31,57 @@ func init() { Root.AddCommand(NewCmdGc()) }
...
 		Short: "List images that are not tagged",
 		Args:  cobra.ExactArgs(1),
 		Run: func(_ *cobra.Command, args []string) {
-			gc(args[0], recursive)
+			gc(args[0], recursive, time.Unix(before, 0), pattern)
 		},
 	}
 
 	cmd.Flags().BoolVarP(&recursive, "recursive", "r", false, "Whether to recurse through repos")
+	cmd.Flags().Int64VarP(&before, "before", "b", time.Now().Unix(), "Oldest upload time (unix timestamp)")
```
ImJasonH, line 47: If we support this I'd love to do it in such a way that users don't have to handle Unix timestamps. Maybe an relative age instead of a date? --before=30d perhaps?

ImJasonH (commented):
Can this change also include some docs to make it easier to understand? Filter especially might be confusing if people expect the filter to operate the opposite way.

b4nst:
Wow oops, that's my bad, I didn't include the README in my commit. Gonna fix that.

cmd/gcrane/cmd/gc.go
b4nst: Agree. Sounds like a duration flag. The problem with duration flag tho is that it does only accept ["ns", "us" (or "µs"), "ms", "s", "m", "h"](https://golang.org/pkg/time/#ParseDuration) (and this seems to be a [touchy subject](https://github.com/golang/go/issues/11473) in the golang community). Accepting only day would be a bit limiting too. Do we go for a string flag with a custom parser or do we keep the `Duration` limitations?

ImJasonH: How would you feel about using cobra's [`Duration` flag](https://godoc.org/github.com/spf13/pflag#Duration) for now, which would only support up to hours, and when/if someone complains we can evaluate custom parsing based on need. 

Worst case, `--before=168h` is still better I think than whatever shell expression would provide the equivalent Unix timestamp, even though `--before=7d` would be even more idea.

jonjohnsonjr:
Part of me wants to reject this and just document the correct incantations of `gcrane ls --json ...` and `jq` to do something equivalent, but I know that's kind of a pain, especially since you could do the same thing for the untagged part.

I'd like to be extremely specific in the docs that this filters based on the `uploadedTimeMs`, _not_ `creationTimeMs`, which is something else entirely.

cmd/gcrane/cmd/gc.go
b4nst: Agree. Changes incoming.

b4nst:
@jonjohnsonjr Yeah I supposed that `gc` command was here precisely to reduce some overwhelming script tricks. In that sense, adding this 2 options ease the use of this command for my use case. That being said, if I'm the only one to see an improved UX here, that probably means that I'm somewhere wrong in my usage. 

I've updated the doc to be extremely specific on which date is used.

cmd/gcrane/cmd/gc.go
```diff
@@ -29,28 +31,57 @@ func init() { Root.AddCommand(NewCmdGc()) }
...
 		Short: "List images that are not tagged",
 		Args:  cobra.ExactArgs(1),
 		Run: func(_ *cobra.Command, args []string) {
-			gc(args[0], recursive)
+			gc(args[0], recursive, time.Now().Add(-before), pattern)
 		},
 	}
 
 	cmd.Flags().BoolVarP(&recursive, "recursive", "r", false, "Whether to recurse through repos")
+	cmd.Flags().DurationVarP(&before, "before", "b", time.Duration(0), "Age of the oldest uploaded image to print (ns, us, ms, s, m, h)")
```
jonjohnsonjr, line 47: @ImJasonH what do you think about [borrowing the flag name](https://github.com/sethvargo/gcr-cleaner#payload--parameters) `grace` from @sethvargo instead of `before`?

jonjohnsonjr:
> In that sense, adding this 2 options ease the use of this command for my use case. That being said, if I'm the only one to see an improved UX here, that probably means that I'm somewhere wrong in my usage.

You're definitely not alone. I even started writing up a doc for stuff that we could add to `gcrane gc`, but never got around to implementing it. It was pretty similar to this PR 😄

cmd/gcrane/cmd/gc.go
```diff
@@ -29,28 +31,57 @@ func init() { Root.AddCommand(NewCmdGc()) }
...
+	cmd.Flags().StringVarP(&pattern, "pattern", "p", "", "Will also collect images with no tags matching this pattern")
```
jonjohnsonjr, line 48: I don't find this behavior 100% intuitive. Given the default behavior of only untagged things being collected, this does make some sense, but I have to pause to think about it. I'd like to think about what this flag behavior should be if we made the untagged nature of `gc` a flag (`--untagged` or something, defaulting to true for backwards compat).

One thing that would be useful here is to print both the digest and tag references for this case. Usually for `gcrane gc`, I'm piping the output into `crane delete`. If we don't delete the tags that match this regex, we won't actually be able to delete the image, since it's still tagged. (This isn't required by the API spec but is a quirk of GCR.)

For the collector, let's also:

```go
for _, tag := range manifest.Tags {
  fmt.Printf("%s:%s\n", repo, tag)
}
```

cmd/gcrane/cmd/gc.go
```diff
@@ -29,28 +31,57 @@ func init() { Root.AddCommand(NewCmdGc()) }
...
 	repo, err := name.NewRepository(root)
 	if err != nil {
 		log.Fatalln(err)
 	}
 
 	auth := google.WithAuthFromKeychain(authn.DefaultKeychain)
 
+	re, err := regexp.Compile(pattern)
+	if err != nil {
+		log.Fatalln(err)
```
jonjohnsonjr, line 63: While you're in here, do you mind pulling out these fatal logs and returning an `error` instead (just fatal in the cobra `Run` method)?

That will move us a bit closer to exposing this function.

b4nst: Done. I've taken the opportunity to use [RunE](https://github.com/spf13/cobra#returning-and-handling-errors)

cmd/gcrane/cmd/gc.go
b4nst: Agree

cmd/gcrane/cmd/gc.go
b4nst: Nice catch for the tag issue. Regarding the behaviour of `--pattern` if we add an `--untagged` you're right, it's unclear. My guess is that it should collect only images tagged with only tags name that don't match the pattern. Reversing it (collecting images that match the pattern) makes it a bit more intuitive, but less convenient. Most of the time you know the pattern that you want to keep, not the one that you want to garbage collect.

cmd/gcrane/cmd/gc.go
jonjohnsonjr: Nice. TIL.

cmd/gcrane/cmd/gc.go
jonjohnsonjr: Can you accomplish what you're trying to do by just deleting the tags you want to ignore? Or do you always want to keep these ignored tags if there is another tag pointing to that image?

@ImJasonH mentioned that the regex thing is not too hard to do with `gcrane ls` and `grep` or `grep -v`.

I would suggest we go ahead and merge the `--grace` part and open an issue to keep this discussion going.

b4nst: OK works for me. I suppose I can do a first pass with `gcrane ls | grep` and then use `grcrane gc` to delete untagged images. Let me refactor everything to get rid of the pattern option.

jonjohnsonjr: Sounds good, thanks! If for some reason that doesn't work well, we can come back and have a longer discussion on naming and behavior and whatnot, but I didn't want to block the `--grace` flag on that conversation because it seems obviously good and useful 👍

PR #824, opened 2020-11-11
Add ability to override default registry and tag in pkg/name

### LINKED ISSUE #805: Add default registry option in pkg/name
Currently, if a hostname is omitted in name parsing, registry is [assumed to be `index.docker.io`](https://github.com/google/go-containerregistry/blob/5040388aee43ce720902e720f929de5a56c32a0c/pkg/name/registry.go#L25). It would be nice if this behavior was configurable via an [`Option`](https://github.com/google/go-containerregistry/blob/5040388aee43ce720902e720f929de5a56c32a0c/pkg/name/options.go#L31). This should be able to be introduced without a breaking API change by adding a private field to the `Registry` struct (will also necessitate adding to parent `Reference` / `Tag` / `Digest` / `Repository` as well I believe).

hasheddan:
Adds default registry and default tag options to `pkg/name` such that a user may configure their own defaults. Also move defaulting logic to tag / registry construction rather than methods called on the objects. See individual commits for how this maintains compatibility with existing behavior.

/cc @jonjohnsonjr @ImJasonH 

Fixes #805

### DIFF
--- a/pkg/name/ref_test.go
+++ b/pkg/name/ref_test.go
@@ -18,6 +18,35 @@ import (
 	"testing"
 )
 
+var (
+	testDefaultRegistry = "registry.upbound.io"
+	testDefaultTag      = "stable"
+	inputDefaultNames   = []string{
+		"crossplane/provider-gcp",
+		"crossplane/provider-gcp:v0.14.0",
+		"ubuntu",
+		"gcr.io/crossplane/provider-gcp:latest",
+	}
+	outputDefaultNames = []string{
+		"registry.upbound.io/crossplane/provider-gcp:stable",
+		"registry.upbound.io/crossplane/provider-gcp:v0.14.0",
+		"registry.upbound.io/ubuntu:stable",
+		"gcr.io/crossplane/provider-gcp:latest",
+	}
+)
+
+func TestParseReferenceDefaulting(t *testing.T) {
+	for i, name := range inputDefaultNames {
+		ref, err := ParseReference(name, WithDefaultRegistry(testDefaultRegistry), WithDefaultTag(testDefaultTag))
+		if err != nil {
+			t.Errorf("ParseReference(%q); %v", name, err)
+		}
+		if ref.Name() != outputDefaultNames[i] {
+			t.Errorf("ParseReference(%q); got %v, want %v", name, ref.String(), outputDefaultNames[i])
+		}
+	}
+}
+
 func TestParseReference(t *testing.T) {
 	for _, name := range goodWeakValidationDigestNames {
 		ref, err := ParseReference(name, WeakValidation)
--- a/pkg/name/registry_test.go
+++ b/pkg/name/registry_test.go
@@ -108,6 +108,24 @@ func TestDefaultRegistryNames(t *testing.T) {
 	}
 }
 
+func TestOverrideDefaultRegistryNames(t *testing.T) {
+	testRegistries := []string{"docker.io", ""}
+	expectedRegistries := []string{"index.docker.io", "gcr.io"}
+	overrideDefault := "gcr.io"
+
+	for i, testRegistry := range testRegistries {
+		registry, err := NewRegistry(testRegistry, WeakValidation, WithDefaultRegistry(overrideDefault))
+		if err != nil {
+			t.Fatalf("`%s` should be a valid Registry name, got error: %v", testRegistry, err)
+		}
+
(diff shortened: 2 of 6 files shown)


jonjohnsonjr:
> and tag

This is interesting. Do you have a use case for this in mind, or are you doing this for consistency sake?

hasheddan:
@jonjohnsonjr some of both. For instance, in @crossplane we don't use the `latest` tag. We just have semvers and release channel tags, so we would likely default to the stable channel tag if no tag is provided.

hasheddan:
I had also briefly toyed with the `WithDefaultNamespace` option, but it doesn't really make as much sense and would require larger changes.

pkg/name/registry_test.go
```diff
@@ -108,6 +108,24 @@ func TestDefaultRegistryNames(t *testing.T) {
 	}
 }
 
+func TestOverrideDefaultRegistryNames(t *testing.T) {
+	testRegistries := []string{"docker.io", ""}
+	expectedRegistries := []string{"index.docker.io", "gcr.io"}
+	overrideDefault := "gcr.io"
```
jonjohnsonjr, line 114: What happens if this is `gcr.io/myproject`?

hasheddan: It would use `gcr.io/myproject`. We do not check the default, so a user has the ability to provide something nonsensical there. We were not previously validating the default either, but there was also only one option that could ever be the default. I could go either way on this, but I lean towards letting the user do what they want if they are opting to override. Typically I imagine the override will be hardcoded, rather than changing on accepted input, so the package consumer is responsible for making sure they use a reasonable value. What do you think?

jonjohnsonjr: > I could go either way on this, but I lean towards letting the user do what they want if they are opting to override.

Yeah there is precedent for this (repo.Tag(), repo.Digest()), so I think it's fine.

I'd like to see tests for this that actually calls `ParseReference` instead of `NewRegistry` (basically nobody does that). A list of triples with (input, defaultRegistry, output) that includes this case would be great.

hasheddan: @jonjohnsonjr SGTM. Would you be open to these tests being refactored to table-driven in a follow-up PR?

hasheddan: @jonjohnsonjr I have updated with the `ParsseReference` tests here :+1:

jonjohnsonjr: > Would you be open to these tests being refactored to table-driven in a follow-up PR?

Yeah that's fine, honestly this whole package needs that treatment and some docs love.

pkg/name/ref_test.go
```diff
@@ -18,6 +18,35 @@ import (
...
+	inputDefaultNames   = []string{
+		"crossplane/provider-gcp",
+		"crossplane/provider-gcp:v0.14.0",
+		"ubuntu",
+		"gcr.io/crossplane/provider-gcp:latest",
+	}
+	outputDefaultNames = []string{
+		"registry.upbound.io/crossplane/provider-gcp:stable",
+		"registry.upbound.io/crossplane/provider-gcp:v0.14.0",
+		"registry.upbound.io/ubuntu:stable",
```
jonjohnsonjr, line 33: One thing I hadn't considered is how this interacts with dockerhub's `library/` stuff, see here: https://github.com/google/go-containerregistry/blob/ef0e499de80f001ac57881689e9a40a2497247a6/pkg/name/repository.go#L42

If I try to use this for something like mirroring to mirror.example.com, I'd need some way to make `ubuntu` produce `mirror.example.com/library/ubuntu`.

I don't know that this matters for now, perhaps we can add a `WithDefaultNamespace` option if I actually try to do this.

hasheddan: Yeah the `WithDefaultNamespace` would behave a bit different, but I think it could certainly be added. Happy to follow-up with that / also look at mirroring more holistically.

jonjohnsonjr: Let's punt on this and look at it later if we need to.

pkg/name/registry_test.go
hasheddan: @jonjohnsonjr sounds good! I'll get to work on that next 👍 Any more updates needed here?

jonjohnsonjr (approved):
LGTM

PR #831, opened 2020-11-17
fix vet complaint

deitch:
`go vet` regularly complains about these two unkeyed fields. It doesn't really affect anything, but might as well have good hygiene.


deitch:
Also, it was bothering me 😀

jonjohnsonjr (approved):
Fixing stuff just because it's bothering you is the bedrock of this project and arguably all of open source. Thanks :)

Issue #832, opened 2020-11-17
regression: crane Client.Timeout exceeded between v0.1.4 and `go get` version


hstenzel:
I had CI that was using the `go get` version of crane. It unexpectedly began failing on the current crane versions. The failure mode is the timeout message included below.

I've not seen this failure mode previously, but now jobs that would always succeed are always failing; these jobs make extensive use of crane to promote docker images in the pipeline.

The failure message is:

```
2020/11/17 15:41:13 fetching config: reading image "registry.gitlab.com/foo/bar:baz": Get "https://registry.gitlab.com/v2/": net/http: request canceled while waiting for connection (Client.Timeout exceeded while awaiting headers)
```

I was able to work around the issue by using the v0.1.4 version instead.

jonjohnsonjr:
Sorry, this was me being lazy: https://github.com/google/go-containerregistry/pull/808#pullrequestreview-527623768

Let me bump this to something much longer but non-infinite. I'd like to do this correctly but there's not currently a way to plumb a context through `transport.New` (big oversight).

jonjohnsonjr:
> I had CI that was using the go get version of crane. It unexpectedly began failing on the current crane versions. The failure mode is the timeout message included below.

If we are breaking you too much (again, sorry), I'd suggest pinning to a tag or commit, but I also understand there are trade-offs here, and it might be worth breaking every so often to get fixes without any maintenance.

I also really appreciate your reporting this.

hstenzel:
@jonjohnsonjr Thanks for such a quick turnaround & your work on a great tool!

I'm happy to verify the fix; but it's not clear to me the mechanism by which the `go get` released code is updated.

You are right about the tradeoffs for the version that drives CI. In our case it was simply that our use of crane predates the 0.1.0 release where GitHub actions based binary releases were introduced.

jonjohnsonjr:
> I'm happy to verify the fix; but it's not clear to me the mechanism by which the go get released code is updated.

I thought it grabs the most recent semver tag but that can't be right because the regression wasn't in a release.

PR #834, opened 2020-11-17
Plumb contexts through transport.New

jonjohnsonjr:
Fixes https://github.com/google/go-containerregistry/issues/832

Add transport.NewWithContext.

Use the provided context for the initial ping request and token
exchange.

Subsequent token exchanges (to refresh) will use the context of the
incoming request in RoundTrip.

Drop the hardcoded ping timeout.

Use transport.NewWithContext everywhere that's easily plumbable.

### DIFF
--- a/pkg/internal/legacy/copy.go
+++ b/pkg/internal/legacy/copy.go
@@ -58,6 +58,8 @@ func CopySchema1(desc *remote.Descriptor, srcRef, dstRef name.Reference, srcAuth
 func putManifest(desc *remote.Descriptor, dstRef name.Reference, dstAuth authn.Authenticator) error {
 	reg := dstRef.Context().Registry
 	scopes := []string{dstRef.Scope(transport.PushScope)}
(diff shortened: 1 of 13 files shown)


jonjohnsonjr:
cc @CodingJzy this will revert https://github.com/google/go-containerregistry/pull/808 and use the context from `remote.WithContext` for most cases, just FYI.

CodingJzy:
> cc @CodingJzy，这将还原[＃808](https://github.com/google/go-containerregistry/pull/808)并在`remote.WithContext`大多数情况下使用FYI中的上下文。

thank you，Your way of passing in the context parameter is correct. I just discovered the problem, and then temporarily added a 2s timeout, which did not fundamentally solve the problem.

ImJasonH (approved):

jonjohnsonjr:
Thanks for making us aware of the problem @CodingJzy 😃

PR #835, opened 2020-11-19
interface for standard processor converters

deitch:
Follow-on from discussion with @jonjohnsonjr [here](https://github.com/google/go-containerregistry/pull/828#issuecomment-728265001). This creates a "thing" in `pkg/v1` that can be resolved to either a `v1.Image` _or_ a `v1.ImageIndex`. 

`pkg/v1/remote/Descriptor` already implements this, so this creates a standard for applying it across other areas.

The name `ResolvableDescriptor` might not be the best. 😁

### DIFF
--- a/pkg/v1/layout/index.go
+++ b/pkg/v1/layout/index.go
@@ -32,6 +32,8 @@ type layoutIndex struct {
 	mediaType types.MediaType
 	path      Path
 	rawIndex  []byte
+
+	v1.ResolvableDescriptor
 }
 
 // ImageIndexFromPath is a convenience function which constructs a Path and returns its v1.ImageIndex.
--- a/pkg/v1/manifest.go
+++ b/pkg/v1/manifest.go
@@ -48,6 +48,17 @@ type Descriptor struct {
 	Platform    *Platform         `json:"platform,omitempty"`
 }
 
+// ResolvableDescriptor an item that represents a descriptor that can be resolved
+// into either an Image or an ImageIndex
+type ResolvableDescriptor interface {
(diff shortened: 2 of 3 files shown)


deitch:
Note the this works well with the root of `remote`, because our starting point is `remote.Get()`, which returns a `remote.Descriptor`, which implements `ResolvableDescriptor` and _not_ `ImageIndex`. If you want to use `ImageIndex` features (like `Image(hash)`), you first have to convert it to `ImageIndex` via `desc.ImageIndex()` and _then_ you can use `ImageIndex(hash)`.

With `layout`, the root point _is_ an `ImageIndex`, so it cannot implement _both_ `v1.ImageIndex` and `v1.ResolvableDescriptor`. But it wouldn't want to. We already know it is an `ImageIndex`.

The value in layout is a `FindDescriptor` that might return one of these.

deitch:
Continuing the conversation from #828 and above.

What should we call the interface here? I called it `ResolvableDescriptor` at first cut because it is a `Descriptor` that is resolvable to an `Image` or an `ImageIndex`. I think that `remote.Get()` works fine for that (it returns a `remote.Descriptor` which implements `v1.ResolvableDescriptor`, while we could implement a `layout.FindDescriptor()` which would return a `layout.Descriptor` which implements `v1.ResolvableDescriptor`.

We would need to:

1. Get the naming right
1. Fix the naming in this PR
1. merge this PR in
1. Add a new PR for the `layout` function 

So... naming?

deitch:
How is this in direction @jonjohnsonjr ?

PR #839, opened 2020-11-23
functions to check and validate if can retrieve all parts of an image or index

### LINKED ISSUE #838: how do I check if a layout contains all of the blobs?
Is it possible to somehow try to "resolve", or walk, my local v1 layout from an index that is there and see if all parts are there?

Use case: let's say I have an image as follows. I am using part of `docker.io/library/alpine:3.11`

```
9a839e63dad54c3a6d1834e29692c8492d93f90c59c978c1ed79109ea4fb9a54 - index
|
|- 39eda93d15866957feaee28f8fc5adb545276a64147445c64992ef69804dbf01 - manifest linux/amd64
    |- f70734b6a266dcb5f44c383274821207885b549b75c8e119404917a61335981a - config
    |-cbdbe7a5bc2a134ca8ec91be58565ec07d037386d1f1d8385412d224deafca08 - layer0
|- ad295e950e71627e9d0d14cdc533f4031d42edae31ab57a841c5b9588eacc280 - manifest linux/arm64
    |- c20d2a9ab6869161e3ea6d8cb52d00be9adac2cc733d3fbc3955b9268bfd7fc5 - config
    |- 29e5d40040c18c692ed73df24511071725b74956ca1a61fe6056a651d86a13bd - layer0
```

For the above, I have all of the parts from root index through manifests, configs, and layers for linux/amd64 and linux/arm64.

If my local layout directory has some of those, but not linux/arm64 parts, then it might look like this:

```
9a839e63dad54c3a6d1834e29692c8492d93f90c59c978c1ed79109ea4fb9a54 - index
|
|- 39eda93d15866957feaee28f8fc5adb545276a64147445c64992ef69804dbf01 - manifest linux/amd64
    |- f70734b6a266dcb5f44c383274821207885b549b75c8e119404917a61335981a - config
    |-cbdbe7a5bc2a134ca8ec91be58565ec07d037386d1f1d8385412d224deafca08 - layer0
```

How do I check if all of the parts are there without going to docker.io? I can do the following:

```go
		desc, err := remote.Get("docker.io/library/alpine:3.11")
		ii, err := desc.ImageIndex()
		err = p.AppendIndex(ii)
```

and it will go to `docker.io`, get the index, and then download any missing parts.

However, it should be possible to somehow try to "resolve" my local index and see if all parts are there.

As usual, happy to open a PR on it.

deitch:
An attempt to resolve #838 

Allows you to pass it a `v1.Image()` or a `v1.ImageIndex()`, and will then walk through all of the children, to see if you can get a reader to the layers, or actually read and process the manifests/indexes.

Includes some tests.

Not sure about the naming or package location.

### DIFF
--- a/pkg/v1/partial/validate.go
+++ b/pkg/v1/partial/validate.go
@@ -1,0 +1,87 @@
+// Copyright 2018 Google LLC All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package partial
+
+import (
+	v1 "github.com/google/go-containerregistry/pkg/v1"
+	"github.com/google/go-containerregistry/pkg/v1/types"
+)
+
+// ValidateImage takes an image and checks that it can read all of the parts.
+// - for config, tries to read the config; if it can do so, it is valid
+// - for layers, tries to read the layer; if it can do so, it is valid
+//
+// It doesn't care if the image is from remote, or local v1/layout, or anywhere.
+// It just cares that it can reach it.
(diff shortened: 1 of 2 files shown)


deitch:
This seems unnecessary in the face of [these](https://pkg.go.dev/github.com/google/go-containerregistry@v0.1.4/pkg/v1/validate), so closing.

@jonjohnsonjr if you want to use some of the testing code to check `validate.Index` and `validate.Image`, go ahead. I will leave the branch around here for a bit.

PR #841, opened 2020-11-24
match.Platforms and match.Digests

deitch:
Option to:

* `match.Platforms([]v1.Platform)` for any one of them.
* `match.Digest(v1.Hash)`
* `match.Digests([]v1.Hash)` for any one of them.

Also tests, of course.

### DIFF
--- a/pkg/v1/match/match.go
+++ b/pkg/v1/match/match.go
@@ -44,19 +44,24 @@ func Annotation(key, value string) Matcher {
 	}
 }
 
-// Platform returns a match.Matcher that matches on the provided platform.
+// Platforms returns a match.Matcher that matches on any one of the provided platforms.
 // Ignores any descriptors that do not have a platform.
-func Platform(platform v1.Platform) Matcher {
+func Platforms(platforms ...v1.Platform) Matcher {
 	return func(desc v1.Descriptor) bool {
 		if desc.Platform == nil {
 			return false
 		}
-		return desc.Platform.Equals(platform)
+		for _, platform := range platforms {
+			if desc.Platform.Equals(platform) {
+				return true
+			}
+		}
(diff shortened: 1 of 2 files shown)


deitch:
Updated to include `match.Digest` and `match.Digests`

deitch:
Review, please @jonjohnsonjr ?

jonjohnsonjr:
I'm kinda -1 on having both `match.Foo` and `match.Foos` for everything...

Option 1:

```go
func Foo(foo Foo) Matcher
func Foos(foos []Foo) Matcher
```

Option 2:

```go
// Clients have to use match.Foos([]Foo{foo})
func Foos(foos []Foo) Matcher
```

Option 3:

```go
// Works with match.Foos(foo) and match.Foos(foo, bar)
func Foos(foos ...Foo) Matcher
```

Option 4:

```go
// Works with match.Foo(foo) and match.Foo(foo, bar)
func Foo(foos ...Foo) Matcher
```

WDYT? Order of preference for me is probably 4, 3, 1, 2.

We already have `match.MediaTypes` for 2, but this has been merged for such a short window that I'd be fine making a breaking change for that to make it consistent.

deitch:
Oh, yeah, I like 4. Variadic is so nice and easy. The question is, should it be named:

```go
func Foo(foo Foo...) Matcher
```

or

```go
func Foos(foo Foo...) Matcher
```

I think the second, since it can match multiple?

jonjohnsonjr:
They're about equivalent to me, so whichever you prefer.

deitch:
OK, updated @jonjohnsonjr ; take a look. I did `MediaTypes` too, as you suggested.

jonjohnsonjr (approved):

Issue #740, opened 2020-07-14
Is there a way to reuse token/transport to prevent repeated auth requests?


vito:
Hi there, thanks for developing this package - it's pretty great!

I'm using this package in the Concourse [`registry-image` resource](https://github.com/concourse/registry-image-resource) for listing tags and detecting changes to digests that tags point to - specifically, this PR: https://github.com/concourse/registry-image-resource/pull/214

I noticed that when I point it to a repository that has many tags it can take a very long time to find everything. Judging from the debug logs, it looks like one thing contributing to this is that every `remote.Image` call is going through its own auth flow, probably from this call:

https://github.com/google/go-containerregistry/blob/72597da50ce903895087828b5fc4b918636366f4/pkg/v1/remote/descriptor.go#L199-L203

Is there any way to re-use the token and/or transport for multiple requests in order to speed this up, by skipping the ping + token request? I'm worried that sending so many requests could lead to being rate limited more quickly, though I'm not sure if rate limits apply to the auth flow or the ping call (maybe not).

Thanks!

jonjohnsonjr:
> Hi there, thanks for developing this package - it's pretty great!

Thank you for the kind words.

> listing tags and detecting changes to digests that tags point to

I know there's no way to avoid doing this because there's not a standard API for notifications, but this can be pretty painful for registries :P Joseph Schorr wrote [a proposal](https://docs.google.com/document/d/1V4vaF9mSTBFbIbUUDH0mk-4KizCSFlv7oqA_4ZkD9a8/edit) for a new registry API to support this, which obviously you wouldn't be able to use anytime soon, but I wish there was a better way to do this. It would be wonderful if some software project existed to multiplex over [the various pubsub implementations of major registries](https://github.com/opencontainers/distribution-spec/issues/22#issuecomment-417143466) and expose a common API, but that might be an eternal pipedream 😄 

> Is there any way to re-use the token and/or transport for multiple requests in order to speed this up, by skipping the ping + token request?

This is definitely an annoying problem, and I don't know if there's an easy way to fix it. There's not currently a great way to work around this, but I did have one idea:

[...]

PR #844, opened 2020-11-26
Implement remote.MultiDelete and crane.MultiDelete

### LINKED ISSUE #843: `crane delete` should be able to delete multiple refs at once
`crane delete` only accepts one image ref to delete. This leads [`gcrane gc`'s example](https://github.com/google/go-containerregistry/blob/master/cmd/gcrane/README.md#gc) to pass each gc-able ref to `crane delete` with `xargs -n1`, and each `crane delete` invocation comes with the overhead of individual CLI invocations, auth handshakes, etc.

Instead, `crane delete` should interpret all positional args as refs to parse and delete, perhaps using a `remote.MultiDelete` which uses a single auth handshake. This should produce a noticeable speedup when GCing a large repo, and let users remove the `-n1` and pass as many args to `crane delete` as the platform allows.

ImJasonH:
This adds support for `crane delete foo:one foo:two foo@sha256:...` and updates `gcrane gc` docs to use the new `crane delete` form without `xargs -n1`.

This is currently limited to deleting refs for the same repo, like `MultiWrite` is currently limited. This should support the `gcrane gc` use case at least.

Fixes #843

### DIFF
--- a/pkg/crane/delete.go
+++ b/pkg/crane/delete.go
@@ -23,11 +23,21 @@ import (
 
 // Delete deletes the remote reference at src.
 func Delete(src string, opt ...Option) error {
+	return MultiDelete([]string{src}, opt...)
+}
+
+// MultiDelete deletes the remote references in srcs.
+func MultiDelete(srcs []string, opt ...Option) error {
 	o := makeOptions(opt...)
-	ref, err := name.ParseReference(src, o.name...)
-	if err != nil {
-		return fmt.Errorf("parsing reference %q: %v", src, err)
+
+	var refs []name.Reference
+	for _, src := range srcs {
+		ref, err := name.ParseReference(src, o.name...)
+		if err != nil {
+			return fmt.Errorf("parsing reference %q: %v", src, err)
+		}
+		refs = append(refs, ref)
 	}
 
-	return remote.Delete(ref, o.remote...)
+	return remote.MultiDelete(refs, o.remote...)
 }
--- a/cmd/crane/cmd/delete.go
+++ b/cmd/crane/cmd/delete.go
@@ -25,12 +25,11 @@ import (
 func NewCmdDelete(options *[]crane.Option) *cobra.Command {
 	return &cobra.Command{
 		Use:   "delete IMAGE",
-		Short: "Delete an image reference from its registry",
-		Args:  cobra.ExactArgs(1),
+		Short: "Deletes image references from their registries",
+		Args:  cobra.MinimumNArgs(1),
 		Run: func(_ *cobra.Command, args []string) {
-			ref := args[0]
-			if err := crane.Delete(ref, *options...); err != nil {
-				log.Fatalf("deleting %s: %v", ref, err)
+			if err := crane.MultiDelete(args, *options...); err != nil {
+				log.Fatal("deleting: ", err)
 			}
 		},
 	}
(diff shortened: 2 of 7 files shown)


ImJasonH:
In addition to supporting deleting across repos, we might also want to let the user specify how many parallel jobs to use to delete (currently defaults to 4)

ImJasonH:
I tried to untangle transports to make them reusable, and got pretty lost in there. (re: https://github.com/google/go-containerregistry/issues/843#issuecomment-734395500 and  https://github.com/google/go-containerregistry/issues/740)

I think `MultiDelete` is a good compromise until we can figure out how to reuse transports better, and doesn't box us in once we do -- `MultiDelete` can just become `Delete` in a loop, instead of `Delete` being `MultiDelete` with a one-element map.

I think the benefit of being able to `crane gc | xargs crane rm` efficiently is worth proceeding with `MultiDelete`.

pkg/crane/delete.go
```diff
@@ -1,4 +1,4 @@
-// Copyright 2018 Google LLC All Rights Reserved.
+// Copyright 2020 Google LLC All Rights Reserved.
```
jonjohnsonjr, line 1: Don't update the copyright year (here and elsewhere). It expresses a weaker claim (asserting copyright since 2020) than leaving it alone (asserting copyright since 2018).

ImJasonH: Done.

jonjohnsonjr:
Rather than commenting directly on the code (as it can get lost) going to put my suggestion here, since it requires some discussion.

Deletion is, unfortunately, rather complicated. Some registries do reference counting, while others don't. Some registries allow (or even _require_) deletion of tags, while others don't. Some registries (looking at you, Docker Hub) don't even allow deletion through the API _at all_.

So, I have a couple thoughts.

Often, we're going to encounter these:

```
tag -> image -> layers
tag -> index -> images -> layers
image -> layers
index -> images -> layers
```

While tags can only point to one thing, indexes can share images, and images can share layers. Individual images within an index can also be tagged, further complicating things.

Sometimes we might encounter these:

```
tag -> index -> index -> index -> ... -> image -> layers
tag -> index -> layers | image | index
```

Now... it's possible that callers of `MultiWrite` can know something about this topology (for example, `gcrane` at least will have the mediaType available), but without actually doing some _read_ requests against the registry, all we can do is guess using heuristics.

The current implementation will likely fail, unless the network and randomness align _just so_ to delete everything in the correct order. This seems unlikely. So...

1. Should we do internal retries as "rounds" of deletions until progress stops? This is more in line with "do what I mean", but might be assuming too much on behalf of the user.
2. Should we first build a topology of the registry before attempting to delete each ref? This seems not great if it's _impossible_ to delete certain things, because we'll have wasted a bunch of effort to construct a topology that can't be deleted!

In theory, clients could just write:
```
while ! gcrane ls $REPO | xargs crane delete
```

ImJasonH:
🤯🤦‍♂️ You're right.

I'm slightly biased toward (2), even though it might end up being wasted effort if it's impossible. (1) is also possibly impossible, and will perform some real deletes while it learns that fact.

jonjohnsonjr:
Added 3 because that was what I was proposing with my bash at the end. I can imagine our errors being good enough that clients could do "the right thing" in some situations...

jonjohnsonjr:
> (1) is also possibly impossible, and will perform some real deletes while it learns that fact.

We can't stop the world for the registry, so this is possible no matter what if someone is writing concurrently.

PR #847, opened 2020-11-29
Thunk-ify layout options

jonjohnsonjr:
If we want to add anything other than descriptor mutators as options, we
need a place to do that, so this introduces an options struct that holds
options and thunk-ifies the existing descriptor mutators.


deitch (approved):
I like this (but you knew that). Looking forward to expanding this once it is in.

PR #848, opened 2020-11-29
add to layout replacedescriptor and removedescriptor

### LINKED ISSUE #842: layout.RemoveDescriptor
Any objection to having a `layout.RemoveDescriptor()` function? We have lots of `Append*()`, but no way to remove something I don't want. 

Here is just one use case:

1. On Sunday I download an image `docker.io/foo/bar:latest` and save it via `AppendIndex()`, so it ends up with an entry in `index.json` using the official name annotation
1. On Monday the mutable tag latest is updated, so its root manifest is different'
1. On Tuesday I pull `docker.io/foo/bar:latest` and save it using the only tool I have `AppendIndex()`

I now have two entries in `index.json`, each pointing to a different root manifest.

I agree wholeheartedly that this is all because of mutable tags, which may very well be a blight upon technology, but they are there and used.

Either way, we probably should have a way to remove a descriptor from `index.json`.

I actually wouldn't mind a way to `RemoveImage()` which removes all of its children blobs, except for those in use by others. But that is a much heavier lift, and probably should be done in two steps:

1. Remove the reference in `index.json`
1. Do a "garbage collection" which removes anything that isn't part of a tree rooted in `index.json`

But the latter could have negative side-effects, so I would defer on those for now.

deitch:
Fixes #842 

adds the following to `v1/layout`:

* `RemoveDescriptors(matcher match.Matcher)` - removes from `index.json` any descriptors that match 
* `ReplaceDescriptor(desc v1.Descriptor, matcher match.Matcher)` - equivalent of `RemoveDescriptors(matcher)` followed by `AppendDescriptor(desc)`
* `ReplaceImage(image v1.Image, matcher match.Matcher)` - equivalent of `RemoveDescriptors(matcher)` followed by `AppendImage(image)`
* `ReplaceIndex(index v1.ImageIndex, matcher match.Matcher)` - equivalent of `RemoveDescriptors(matcher)` followed by `AppendIndex(index)`

None of these does garbage collection in the blobs dir, which is a completely separate topic, and should be handled under separate cover.

### DIFF
--- a/pkg/v1/layout/write.go
+++ b/pkg/v1/layout/write.go
@@ -23,6 +23,9 @@ import (
 	"path/filepath"
 
 	v1 "github.com/google/go-containerregistry/pkg/v1"
+	"github.com/google/go-containerregistry/pkg/v1/match"
+	"github.com/google/go-containerregistry/pkg/v1/mutate"
+	"github.com/google/go-containerregistry/pkg/v1/partial"
 	"github.com/google/go-containerregistry/pkg/v1/types"
 	"golang.org/x/sync/errgroup"
 )
@@ -118,6 +121,85 @@ func (l Path) AppendDescriptor(desc v1.Descriptor) error {
 	}
 
 	index.Manifests = append(index.Manifests, desc)
+
+	rawIndex, err := json.MarshalIndent(index, "", "   ")
+	if err != nil {
+		return err
+	}
+
+	return l.WriteFile("index.json", rawIndex, os.ModePerm)
+}
+
+// ReplaceImage writes a v1.Image to the Path and updates
+// the index.json to reference it, replacing any existing one that matches matcher, if found.
+func (l Path) ReplaceImage(img v1.Image, matcher match.Matcher, options ...Option) error {
+	if err := l.WriteImage(img); err != nil {
+		return err
+	}
+
+	return l.replaceDescriptor(img, matcher, options...)
+}
+
+// ReplaceIndex writes a v1.ImageIndex to the Path and updates
+// the index.json to reference it, replacing any existing one that matches matcher, if found.
+func (l Path) ReplaceIndex(ii v1.ImageIndex, matcher match.Matcher, options ...Option) error {
+	if err := l.WriteIndex(ii); err != nil {
+		return err
+	}
+
+	return l.replaceDescriptor(ii, matcher, options...)
+}
+
+// replaceDescriptor adds a descriptor to the index.json of the Path, replacing
+// any one matching matcher, if found.
+func (l Path) replaceDescriptor(append mutate.Appendable, matcher match.Matcher, options ...Option) error {
+	ii, err := l.ImageIndex()
+	if err != nil {
+		return err
+	}
+
+	desc, err := partial.Descriptor(append)
+	if err != nil {
+		return err
+	}
+
+	for _, opt := range options {
+		if err := opt(desc); err != nil {
+			return err
+		}
+	}
+
+	add := mutate.IndexAddendum{
+		Add:        append,
+		Descriptor: *desc,
+	}
+	ii = mutate.AppendManifests(mutate.RemoveManifests(ii, matcher), add)
+
+	index, err := ii.IndexManifest()
+	if err != nil {
+		return err
+	}
+
+	rawIndex, err := json.MarshalIndent(index, "", "   ")
+	if err != nil {
+		return err
+	}
+
+	return l.WriteFile("index.json", rawIndex, os.ModePerm)
+}
+
+// RemoveDescriptors removes any descriptors that match the match.Matcher from the index.json of the Path.
+func (l Path) RemoveDescriptors(matcher match.Matcher) error {
+	ii, err := l.ImageIndex()
+	if err != nil {
+		return err
+	}
+	ii = mutate.RemoveManifests(ii, matcher)
+
+	index, err := ii.IndexManifest()
+	if err != nil {
+		return err
+	}
 
 	rawIndex, err := json.MarshalIndent(index, "", "   ")
 	if err != nil {
--- a/pkg/v1/mutate/index.go
+++ b/pkg/v1/mutate/index.go
@@ -20,6 +20,7 @@ import (
 
 	"github.com/google/go-containerregistry/pkg/logs"
 	v1 "github.com/google/go-containerregistry/pkg/v1"
+	"github.com/google/go-containerregistry/pkg/v1/match"
 	"github.com/google/go-containerregistry/pkg/v1/partial"
 	"github.com/google/go-containerregistry/pkg/v1/types"
 )
@@ -56,6 +57,8 @@ func computeDescriptor(ia IndexAddendum) (*v1.Descriptor, error) {
 type index struct {
 	base v1.ImageIndex
 	adds []IndexAddendum
+	// remove is removed before adds
+	remove match.Matcher
 
 	computed  bool
 	manifest  *v1.IndexManifest
@@ -90,6 +93,17 @@ func (i *index) compute() error {
 	}
 	manifest := m.DeepCopy()
 	manifests := manifest.Manifests
+
+	if i.remove != nil {
+		var cleanedManifests []v1.Descriptor
+		for _, m := range manifests {
+			if !i.remove(m) {
+				cleanedManifests = append(cleanedManifests, m)
+			}
+		}
+		manifests = cleanedManifests
+	}
+
 	for _, add := range i.adds {
 		desc, err := computeDescriptor(add)
 		if err != nil {
@@ -105,6 +119,7 @@ func (i *index) compute() error {
 			logs.Warn.Printf("Unexpected index addendum: %T", add.Add)
 		}
 	}
+
 	manifest.Manifests = manifests
 
 	// With OCI media types, this should not be set, see discussion:
--- a/pkg/v1/mutate/mutate.go
+++ b/pkg/v1/mutate/mutate.go
@@ -27,6 +27,7 @@ import (
 	v1 "github.com/google/go-containerregistry/pkg/v1"
 	"github.com/google/go-containerregistry/pkg/v1/empty"
 	"github.com/google/go-containerregistry/pkg/v1/internal/gzip"
+	"github.com/google/go-containerregistry/pkg/v1/match"
 	"github.com/google/go-containerregistry/pkg/v1/tarball"
 	"github.com/google/go-containerregistry/pkg/v1/types"
 )
@@ -89,6 +90,14 @@ func AppendManifests(base v1.ImageIndex, adds ...IndexAddendum) v1.ImageIndex {
 	return &index{
 		base: base,
 		adds: adds,
+	}
+}
+
+// RemoveManifests removes any descriptors that match the match.Matcher.
+func RemoveManifests(base v1.ImageIndex, matcher match.Matcher) v1.ImageIndex {
+	return &index{
+		base:   base,
+		remove: matcher,
 	}
 }
 
--- a/pkg/v1/layout/write_test.go
+++ b/pkg/v1/layout/write_test.go
@@ -12,6 +12,7 @@ import (
 	"github.com/google/go-cmp/cmp"
 	v1 "github.com/google/go-containerregistry/pkg/v1"
 	"github.com/google/go-containerregistry/pkg/v1/empty"
+	"github.com/google/go-containerregistry/pkg/v1/match"
 	"github.com/google/go-containerregistry/pkg/v1/random"
 	"github.com/google/go-containerregistry/pkg/v1/types"
 	"github.com/google/go-containerregistry/pkg/v1/validate"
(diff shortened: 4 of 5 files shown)


jonjohnsonjr:
I like `Remove` and `Replace` as primitives, but these should probably be part of the `mutate` package. Ideally, everything that does mutations is in the `mutate` package... but layout has a couple of exceptions already. The `Append` stuff was an oversight on my part that I wish I could walk back. The `WriteFile` etc. enable some things that would be otherwise impossible, so that was a concession of purity. For things like this PR (everything is generic except the source and the sink), the calling should look something like:

`layout.Write(path, mutate.Remove(layout.FromPath(path), matcher))` and
`layout.Write(path, mutate.AppendManifests(mutate.Remove(layout.FromPath(path), matcher), addendum))`. 

In the same way we have `mutate.AppendManifests`, we should have `mutate.RemoveManifest[s]`. You can just shove a Matcher in [here](https://github.com/google/go-containerregistry/blob/cf9d7b33e9e735ce7ad8c1446f76deb51fbfe8e0/pkg/v1/mutate/index.go#L58) and use that in `compute()`.

If we get cute with type-casting, we could just have `mutate.Remove` that acts on both `v1.Image` and `v1.ImageIndex`, but we could also do `mutate.RemoveManifests` and `mutate.RemoveLayers`.

For your replace implementation, I would really expect:

`Layout([A, B, C]).ReplaceImage(D, Matcher(B)) == Layout([A, D, C])` and also

`Layout([A, A]).ReplaceImage(B, Matcher(A)) == Layout([B, B])`,

but these are currently just equivalent to `Append(Remove(Matcher(...), X)`.

If just Remove then Append is your desired behavior, how would you feel about adding Remove to mutate, then doing the remove && append in your client code? I'd be okay with a "replace" for mutate if it did in-place and not just remove + append.

deitch:
> The Append stuff was an oversight on my part that I wish I could walk back. 

Is it really? Given that `Append*()` does:

1. Write the Image/Index contents to the layout dir
1. Append the descriptor to the root index
1. Write that index to the layout dir as an updated `index.json`

how would you place that in any other package?

> If we get cute with type-casting, we could just have mutate.Remove that acts on both v1.Image and v1.ImageIndex, but we could also do mutate.RemoveManifests and mutate.RemoveLayers.

Please not, or at least not yet. The need for distinct Image and ImageIndex in various places is an issue throughout, as you raised. I would rather be consistent here with what we do elsewhere, and if we find a way to unify it, then do it everywhere at once.

> For things like this PR (everything is generic except the source and the sink), the calling should look something like:
> 
> layout.Write(path, mutate.Remove(layout.FromPath(path), matcher)) and
> layout.Write(path, mutate.AppendManifests(mutate.Remove(layout.FromPath(path), matcher), addendum)).

I really don't understand (which leaves me concerned others might not as well). I fully get that I can open a `layout` and get a `Path` and I can `AppendImage` or `AppendIndex`, etc.

The above is complex to understand, especially if I already have a `Path`, that I want to modify. 

> If just Remove then Append is your desired behavior, how would you feel about adding Remove to mutate, then doing the remove && append in your client code? I'd be okay with a "replace" for mutate if it did in-place and not just remove + append.

I have no objection to adding `Remove` and `Replace` to mutate, but then what goes in here that makes it easy to understand.

What do I, as a newbie user of this library, call to remove or replace a manifest from a `Path`?

With this PR - or a variant - I can do:

```go
p, err := layout.FromPath(dir) 
// OR
p, err := layout.Write(dir, empty.Index)
// and then
p.ReplaceDescriptor(desc, someMatcher)
```

but now you are saying `Replace*()` goes in mutate, so I end up with:

```go
p, err := layout.FromPath(dir)
// OR
p, err := layout.Write(dir, empty.Index)
// and then
err = p.WriteImage(img)
index, err := p.ImageIndex()
// generate the descriptor from img, redoing everything that is at https://github.com/google/go-containerregistry/blob/3c3abfb82ca884fb7cc6db0efe77264f1c817f1f/pkg/v1/layout/write.go#L42-L69
desc := generateDescriptorFromImage(img)
index, err = mutate.ReplaceDescriptor(desc, someMatcher)
err = p.WriteIndex(index) // this is a function that does not exist now
```

That is a _lot_ messier and a lot more work than the original

If you say, you want to keep the conveniences in `layout`, just move some of the core logic into `mutate`, I can see that, but that will get messy with the `Option` as well.

deitch:
I think I figured out what you meant @jonjohnsonjr . Take a look at it now. All of the mutation is done in `mutate`, while the convenience functions wrap them in `layout`.

deitch:
OK, we have #841 in, which makes using this easier. Any other comments on this?

pkg/v1/layout/write.go
```diff
@@ -127,6 +129,133 @@ func (l Path) AppendDescriptor(desc v1.Descriptor) error {
...
 }
 
+// ReplaceImage writes a v1.Image to the Path and updates
+// the index.json to reference it, replacing any existing one that matches matcher, if found.
+func (l Path) ReplaceImage(img v1.Image, matcher match.Matcher, options ...Option) error {
+	if err := l.WriteImage(img); err != nil {
+		return err
+	}
+
+	mt, err := img.MediaType()
```
jonjohnsonjr, line 139: Use `partial.Descriptor` here, for some bad reasons and some good reasons.

pkg/v1/layout/write.go
```diff
@@ -127,6 +129,133 @@ func (l Path) AppendDescriptor(desc v1.Descriptor) error {
...
+}
+
+// ReplaceIndex writes a v1.ImageIndex to the Path and updates
+// the index.json to reference it, replacing any existing one that matches matcher, if found.
+func (l Path) ReplaceIndex(ii v1.ImageIndex, matcher match.Matcher, options ...Option) error {
+	if err := l.WriteIndex(ii); err != nil {
+		return err
+	}
+
+	mt, err := ii.MediaType()
```
jonjohnsonjr, line 176: Use `partial.Descriptor` here.

pkg/v1/layout/write.go
deitch: I don't understand

pkg/v1/layout/write.go
deitch: Ditto above.

pkg/v1/mutate/mutate.go
```diff
@@ -92,6 +93,49 @@ func AppendManifests(base v1.ImageIndex, adds ...IndexAddendum) v1.ImageIndex {
...
+	return &index{
+		base:     base,
+		replaces: manifests,
+	}, nil
+}
+
+// ReplaceManifests removes any descriptors that match the match.Matcher and replaces with new ones.
+// It is the equivalent of calling RemoveManifests().AppendManifests(), but a bit more efficient and
+// convenient.
+func ReplaceManifests(base v1.ImageIndex, matcher match.Matcher, adds ...IndexAddendum) (v1.ImageIndex, error) {
```
jonjohnsonjr, line 119: I'm not a huge fan of these semantics -- I'd prefer we drop this for now and just have `layout` call RemoveManifests + AppendManifests. If you're worried about efficiency, we could do some typechecking to avoid some of this work, but I don't imagine it's _that_ expensive to recompute the manifest.

jonjohnsonjr:
> I really don't understand (which leaves me concerned others might not as well).

I agree this is a problem, which maybe we can solve with docs and examples, but the "model" for how this library works will be hard to change without a big, breaking overhaul.

> think I figured out what you meant

I think we're roughly on the same page, but if there's anything from [this comment](https://github.com/google/go-containerregistry/pull/848#issuecomment-735474692) that I still need to address, let me know.

pkg/v1/mutate/mutate.go
deitch: I am not sure I understand. Are you suggesting that we just get rid of `ReplaceManifests()` and leave it to the equivalent (which is referenced in the comment)?

deitch:
> I think we're roughly on the same page

I didn't understand the three comments above. Can you clarify?

pkg/v1/layout/write.go
jonjohnsonjr: Sure, sorry I should have provided more details. It's two things:

1. This is ~9 lines of code to produce a `v1.Descriptor` for an artifact, and we've consolidated that into a `partial.Descriptor` method, just for brevity.
2. There are situations where you have some artifact, say a `v1.Image`, which was originally referenced by a `v1.ImageIndex` descriptor that contained platform information. If you just reconstruct the `v1.Descriptor` as you're doing here, we lose that information. Thus, artifacts can optionally implement `Descriptor() (v1.Descriptor, error)` to return the original `v1.Descriptor` by which they were referenced. This is especially interesting for foreign layers. See [here](https://github.com/google/go-containerregistry/tree/master/pkg/v1/partial#partialdescriptor) and https://github.com/google/go-containerregistry/pull/654.

It unfortunately needs to live in `partial` because the way we wrap things in partial means we shadow the wrapped artifact, but fortunately it makes sense to live there anyway.

pkg/v1/mutate/mutate.go
jonjohnsonjr: > Are you suggesting that we just get rid of ReplaceManifests() and leave it to the equivalent (which is referenced in the comment)?

Yeah exactly. From my original comment:

> For your replace implementation, I would really expect:
> 
> `Layout([A, B, C]).ReplaceImage(D, Matcher(B)) == Layout([A, D, C])` and also
> 
> `Layout([A, A]).ReplaceImage(B, Matcher(A)) == Layout([B, B])`,
> 
> but these are currently just equivalent to `Append(Remove(Matcher(...), X)`.
> 
> If just Remove then Append is your desired behavior, how would you feel about adding Remove to mutate, then doing the remove && append in your client code? I'd be okay with a "replace" for mutate if it did in-place and not just remove + append.

I don't think we agree on what `ReplaceManifests` should do, and since what you want to do here is achievable without it, I'd rather punt on `ReplaceManifests` and just make the `layout.ReplaceDescriptor` implementation slightly more verbose.

pkg/v1/layout/write.go
jonjohnsonjr: Oh, I see that I bungled the multiline comment thing in GitHub. This comment should apply to lines 139-158.

pkg/v1/mutate/mutate.go
deitch: So basically we are just getting rid of `ReplaceManifests()`. OK.

pkg/v1/layout/write.go
```diff
@@ -127,6 +129,133 @@ func (l Path) AppendDescriptor(desc v1.Descriptor) error {
...
+	manifest, err := ii.RawManifest()
+	if err != nil {
+		return err
+	}
+
+	desc := v1.Descriptor{
+		MediaType: mt,
+		Size:      int64(len(manifest)),
+		Digest:    d,
+	}
```
jonjohnsonjr, line 195: Resolved the other one, but same as my above comment, these lines can be replaced with:

```suggestion
	desc, err := partial.Descriptor(ii)
	if err != nil {
		return err
	} 
```

pkg/v1/mutate/mutate.go
deitch: OK, I did that and removed the one usage of it.

deitch:
OK, now I think I got what you meant. See the latest.

jonjohnsonjr (commented):
Nice, I think this is a lot cleaner -- with tests this would lgtm.

deitch:
OK, lots of tests. Back at you @jonjohnsonjr

pkg/v1/layout/write.go
```diff
@@ -127,6 +130,102 @@ func (l Path) AppendDescriptor(desc v1.Descriptor) error {
...
+			return err
+		}
+	}
+
+	return l.ReplaceDescriptor(*desc, matcher)
+}
+
+// ReplaceDescriptor adds a descriptor to the index.json of the Path, replacing
+// any one matching matcher, if found.
+func (l Path) ReplaceDescriptor(desc v1.Descriptor, matcher match.Matcher) error {
```
jonjohnsonjr, line 177: (Sorry I missed this before.)

Do you need this exposed? I don't think we need `mutate.AppendableDescriptor` unless this is exposed, since we can just pass through the index or image...

If you do need it, that's fine, I just don't want to expose anything that doesn't have an immediate need.

deitch: You didn't miss it before. I had to add it. When you pass something to mutate.AppendManifests, it has to be a mutate.IndexAddendum, which has to have an Appendable, which _returns_ a Descriptor but not _actually_ a Descriptor. So I needed something to wrap a Descriptor if you actually have one. 

I did try just using IndexAddendum and passing the Descriptor with the Add blank. It gives a SIGSEGV. 

If you have a better approach, I'm all for it.

deitch: I totally misunderstood your comment before, since I responded from my phone. Now that I read it, I understand what you are saying. I think you are right. I will try to rework it.

deitch: So in retrospect, we _can_ remove `ReplaceDescriptor` and just have `ReplaceImage` and `ReplaceIndex`.

The only reason why I can think to keep it is because for append we have:

* `layout.AppendImage`
* `layout.AppendIndex`
* `layout.AppendDescriptor`

So it would make sense to keep it parallel with

* `layout.ReplaceImage`
* `layout.ReplaceIndex`
* `layout.ReplaceDescriptor`

Your call @jonjohnsonjr . If you want to keep it parallel, we can keep it as is. If you prefer not to have the `ReplaceDescriptor`, we can kill it.

My leaning is towards parallel, but I can go either way, so your call.

Then let's close this out.

jonjohnsonjr: I am sympathetic towards the parallelism, but I'd lean towards killing it just so that we can get rid of `mutate.AppendableDescriptor`, and as you can probably tell, I bias heavily against exposing anything that we don't need to :P

deitch: Done. Take a look.

deitch:
And rebased

pkg/v1/layout/write.go
```diff
@@ -127,6 +130,103 @@ func (l Path) AppendDescriptor(desc v1.Descriptor) error {
...
+			return err
+		}
+	}
+
+	return l.replaceDescriptor(*desc, ii, matcher)
+}
+
+// replaceDescriptor adds a descriptor to the index.json of the Path, replacing
+// any one matching matcher, if found.
+func (l Path) replaceDescriptor(desc v1.Descriptor, append mutate.Appendable, matcher match.Matcher) error {
```
jonjohnsonjr, line 177:
```suggestion
func (l Path) replaceDescriptor(append mutate.Appendable, matcher match.Matcher, options ...Option) error {
```

pkg/v1/layout/write.go
```diff
@@ -127,6 +130,103 @@ func (l Path) AppendDescriptor(desc v1.Descriptor) error {
...
+	ii, err := l.ImageIndex()
+	if err != nil {
+		return err
+	}
+
```
jonjohnsonjr, line 182:
```suggestion
	desc, err := partial.Descriptor(append)
	if err != nil {
		return err
	}

	for _, opt := range options {
		if err := opt(desc); err != nil {
			return err
		}
	}
```

pkg/v1/layout/write.go
```diff
@@ -127,6 +130,103 @@ func (l Path) AppendDescriptor(desc v1.Descriptor) error {
...
+	for _, opt := range options {
+		if err := opt(desc); err != nil {
+			return err
+		}
+	}
+
+	return l.replaceDescriptor(*desc, ii, matcher)
```
jonjohnsonjr, line 172:
```suggestion
	return l.replaceDescriptor(ii, matcher, options...)
```

pkg/v1/layout/write.go
```diff
@@ -127,6 +130,103 @@ func (l Path) AppendDescriptor(desc v1.Descriptor) error {
...
+		return err
+	}
+
+	for _, opt := range options {
+		if err := opt(desc); err != nil {
+			return err
+		}
+	}
+
+	return l.replaceDescriptor(*desc, img, matcher)
```
jonjohnsonjr, line 151:
```suggestion
	return l.replaceDescriptor(img, matcher, options...)
```

jonjohnsonjr (commented):
Small suggestion to dedupe some code now that ReplaceDescriptor is removed, otherwise lgtm, feel free to push back if I'm missing something with my suggestion.

deitch:
Oh, yeah, I like that. Nice and clean.

deitch:
Updated with your suggestions @jonjohnsonjr ; back at you.

pkg/v1/mutate/mutate.go
```diff
@@ -92,6 +93,26 @@ func AppendManifests(base v1.ImageIndex, adds ...IndexAddendum) v1.ImageIndex {
 	}
 }
 
+// RemoveManifests removes any descriptors that match the match.Matcher.
+func RemoveManifests(base v1.ImageIndex, matcher match.Matcher) (v1.ImageIndex, error) {
```
jonjohnsonjr, line 97: Okay, fine, one last thing before I merge this. Sorry... thank you for your patience.

We _could_ defer this to be a little bit lazier (happens in `compute` instead of here) so that we don't return an error here, which would make using it a little easier (one less error check, matching `AppendManifests`).

This would allow for:

```go
	ii = mutate.AppendManifests(mutate.RemoveManifests(ii, matcher), add)
```

Not a _huge_ deal, but something I often find frustrating when using this library is that I am constantly checking for errors for things that don't _really_ need to return an error, and I wish I could go back and fix that.

deitch: I understand what you are trying to do. If this only returns a `v1.ImageIndex` and not `(v1.ImageIndex, error)`, then you have the option to chain them.

I just don't see how. 

The source for this is:

```go
// RemoveManifests removes any descriptors that match the match.Matcher.
func RemoveManifests(base v1.ImageIndex, matcher match.Matcher) (v1.ImageIndex, error) {
	ii, err := base.IndexManifest()
	if err != nil {
		return base, err
	}

	// create a list without any existing ones
	manifests := []v1.Descriptor{}
	for _, manifest := range ii.Manifests {
		if !matcher(manifest) {
			manifests = append(manifests, manifest)
		}
	}
	return &index{
		base:     base,
		replaces: manifests,
	}, nil
}
```

The only potential error is from getting `base.IndexManifest()`, but it _could_ return an error. How would we handle it?

jonjohnsonjr: > but it could return an error. How would we handle it?

Drop `index.replaces`, for an `index.removeMatcher` and set it to the `match.Matcher` we pass in, then instead of doing:

```go
        manifests := manifest.Manifests
	if len(i.replaces) > 0 {
		manifests = i.replaces
	}
```

in `index.compute`, do:

```go
	manifests := []v1.Descriptor
        for _, desc := range manifest.Manifests {
            if !i.removeMatcher(desc) {
                manifests = append(manifests, desc)
            }
        }
```

We're basically just deferring that `IndexManifest()` call until `compute()` is called.

`RemoveManifests` just becomes:

```go
// RemoveManifests removes any descriptors that match the match.Matcher.
func RemoveManifests(base v1.ImageIndex, matcher match.Matcher) v1.ImageIndex {
	return &index{
		base:     base,
		removeMatcher: matcher,
	}
}
```

deitch: I think I got what you had in mind. I added it as another commit, so I can roll back if that is not it. Take a look and comment here. If it is, I will squash the two commits.

pkg/v1/mutate/index.go
```diff
@@ -90,6 +95,20 @@ func (i *index) compute() error {
 	}
 	manifest := m.DeepCopy()
 	manifests := manifest.Manifests
+	if len(i.replaces) > 0 {
```
jonjohnsonjr, line 98: This is just unused now, right? We can drop `index.replaces`.

pkg/v1/mutate/index.go
```diff
@@ -56,6 +57,10 @@ func computeDescriptor(ia IndexAddendum) (*v1.Descriptor, error) {
 type index struct {
 	base v1.ImageIndex
 	adds []IndexAddendum
+	// replaces are applied before adds
+	replaces []v1.Descriptor
+	// removes are removed after replaces and before adds
+	removes []match.Matcher
```
jonjohnsonjr, line 63: I don't think this needs to be a list (yet?). In theory, we may be able to optimize some things in the future to do introspection and directly modify these fields instead of returning an entirely new `mutate.index` each time, but right now this can only be a single matcher, I think.

pkg/v1/mutate/index.go
deitch: We can, but having figured it out how to work it, maybe we should just leave it?

pkg/v1/mutate/index.go
deitch: It probably can, but it parallels adds, for a clean API, and I made it similarly variadic. Let's just leave it.

jonjohnsonjr: Eh... I'm not sure.

I do like that this mirrors AddManifests, but I think making this variadic is going to be a source of confusion. There's more than one way to compose matchers, and this interface forces the operator to only ever be "or".

You can achieve the same thing now with multiple calls to `RemoveManifests`, which is definitely less efficient, but you can also just implement a `match.Matcher` that matches multiple predicates.

Having `RemoveManifests` accept multiple matchers encodes some matcher composition within mutate, and I'd rather that live in `match` if we need it.

Of course, you know that I will advocate for adding `match.Or` and `match.And` to solve this, but we're not there yet :P

deitch: OK done. And rebased.

pkg/v1/mutate/index.go
deitch: OK done.

jonjohnsonjr (approved):
Thanks :)

Issue #866, opened 2020-12-10
Make it easy to support estargz


mattmoor:
See the proposal here: https://github.com/opencontainers/image-spec/issues/815

Given the following factors:
1. estargz is designed for backwards compatibility with existing clients,
1. images must be built with estargz to even have the option of lazy loading (they can still be eager!),
1. the volume of build tools leveraging GGCR to produce images (ko, buildpacks, kaniko, bazel(?)).

I propose that we make estargz either the new default we push for images, or trivially easy to consume.

jonjohnsonjr:
> trivially easy to consume

I'd be okay with something like `export GGCR_EXPERIMENT_ESTARGZ=1`?

mattmoor:
Fortunately the optimize code path in the stargz snapshotter sort of lays this out for us already (since they use ggcr).  I think what we effectively want is a way to make this easy: https://github.com/containerd/stargz-snapshotter/blob/a85d05d70ed851a0f4cb0744ecd0153fe3acafcf/cmd/ctr-remote/commands/optimize.go#L400-L412

mattmoor:
I think the simplest starting point would be to write a function that can convert images and image indices to estargz and harness that in `crane`.  Perhaps an `estargz` package, with functions for converting layers, images and image indices.

I suspect that we may want to create an extended `v1.Layer` interface that can pass annotations, e.g. 

```go
type AnnotatedLayer interface {
      Layer
      Annotations() map[string]string
}
```

Then here we can check whether layers implement the type and augment the addendum passed to Append: https://github.com/google/go-containerregistry/blob/94913c02dcaa92b09b913585c1d26ba2eba1d946/pkg/v1/mutate/mutate.go#L50

I suspect that just doing the conversion will force us to confront a number of the issues, and also sets things up so that downstream tooling that wants to consciously produce estargz images need not start by building those images from scratch.

---

With the ability to convert images, we could then look at `ko` as a next step.  It does its own gzipping, and given a good library for estargz I think we could simply take the resulting TOC digest and annotate it on the addendum here: https://github.com/google/ko/blob/06ebf033f04236836da625cb0477a002f6a06639/pkg/build/gobuild.go#L506

With the ability to convert the `remote.Image` base, and estargz-ify the kodata and ko-app layers, we should have `ko` in the bag, which gives us a very large corpus of real-world things to test it on.

mattmoor:
[...]

PR #870, opened 2020-12-16
Add v1util.EstargzReadCloser utility.

mattmoor:
This adds the first utility function we will need to start producing estargz layers.

Related: https://github.com/google/go-containerregistry/issues/866


mattmoor:
Now with 100% coverage 😛

ImJasonH:
Commented on Slack, but for visibility:

`v1util` is a smell, can we move this to its own package, `pkg/internal/estargz` while we add experimental support to e.g., `remote`?

pkg/v1/v1util/zip.go
```diff
@@ -80,6 +84,27 @@ func GunzipReadCloser(r io.ReadCloser) (io.ReadCloser, error) {
 	}, nil
 }
 
+// EstargzReadCloser reads uncompressed input data from the io.ReadCloser and
+// returns an io.ReadCloser from which compressed data may be read.
+// Refer to estargz for the options:
+// https://pkg.go.dev/github.com/containerd/stargz-snapshotter@v0.2.0/estargz#Option
+func EstargzReadCloser(r io.ReadCloser, opts ...estargz.Option) (io.ReadCloser, v1.Hash, error) {
```
jonjohnsonjr, line 91: Can we put this under `pkg/internal/estargz` while we're experimenting and expose it later? I'm not sure that I want to commit to this interface quite yet, and v1util is depended on by basically every package in this module, so it can't get pruned if it's in here.

Also `estargz.NewRead[Close?]er` reads nicely 😛

pkg/v1/v1util/zip.go
```diff
@@ -80,6 +84,27 @@ func GunzipReadCloser(r io.ReadCloser) (io.ReadCloser, error) {
...
+	defer r.Close()
+
+	bs, err := ioutil.ReadAll(r)
```
jonjohnsonjr, line 94: Assuming this is just for prototyping and we won't always buffer the whole thing for a final implementation? We should be teeing `r` into a hasher, right?

jonjohnsonjr (commented):
I am _really_ tempted to fork their package and remove `github.com/hashicorp/errwrap` and `github.com/hashicorp/go-multierror`, but I understand that's asking a lot 😄

jonjohnsonjr:
> v1util is a smell, can we move this to its own package, pkg/internal/estargz while we add experimental support to e.g., remote?

Are you me??

mattmoor:
> I am really tempted to fork their package and remove github.com/hashicorp/errwrap and github.com/hashicorp/go-multierror, but I understand that's asking a lot

Why not patch upstream?

pkg/v1/v1util/zip.go
mattmoor: This should be done, and I blew up the rest of v1util in: https://github.com/google/go-containerregistry/pull/872 for consistency.

pkg/v1/v1util/zip.go
mattmoor: Yeah, I don't love this implementation either, but I think the estargz stuff wants to be able to seek around to build the index and reorder things for the final estargz file.  The `v1.Hash` is not the overall hash, but the hash of the table of contents.  I owe a better function comment than this has, so let me take a crack at that now.

jonjohnsonjr:
> Why not patch upstream?

Very finite amounts of time.

mattmoor:
> Very finite amounts of time

So you want to maintain a fork of something...  right... 🤣

pkg/v1/internal/estargz/estargz.go
```diff
@@ -0,0 +1,50 @@
...
+//  * An io.ReadCloser from which compressed data may be read, and
+//  * A v1.Hash with the hash of the estargz table of contents, or
+//  * An error if the estargz processing encountered a problem.
+//
+// Refer to estargz for the options:
+// https://pkg.go.dev/github.com/containerd/stargz-snapshotter@v0.2.0/estargz#Option
+func ReadCloser(r io.ReadCloser, opts ...estargz.Option) (io.ReadCloser, v1.Hash, error) {
+	defer r.Close()
+
+	bs, err := ioutil.ReadAll(r)
```
jonjohnsonjr, line 37: I imagine that sometimes r here is an `os.File` that already implements `io.SectionReader`, can we do:

```go
if sr, ok := r.(io.SectionReader); ok {
  rc, err = estargz.Build(...)
}
```

And avoid buffering sometimes?

mattmoor: I thought about this, but `io.SectionReader` is not an interface, so this does not work.

I also thought about checking `io.ReaderAt`, but we still need to know the length.

If `os.File` had a length method (closest is `Stat`), we could define our own interface that wrapped `io.ReaderAt` and included the length method, but that doesn't work terribly well either 😞

mattmoor: https://github.com/google/go-containerregistry/issues/876

jonjohnsonjr (approved):

PR #871, opened 2020-12-16
Support a mode to emit estargz automagically.

mattmoor:
Fixes: https://github.com/google/go-containerregistry/issues/866


pkg/v1/internal/estargz/constants.go
```diff
@@ -0,0 +1,19 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package estargz
+
+// TOCDigestKey holds the key of the layer annotation used to associate
+// table-of-contents verification information to the layer.
+const TOCDigestKey = "containerd.io/snapshot/stargz/toc.digest"
```
jonjohnsonjr, line 19: Pull this from https://github.com/containerd/stargz-snapshotter/blob/af03a142bbeefb0e07e915e4e8fdedc082e5039b/estargz/types.go#L71 instead?

pkg/v1/internal/estargz/estargz.go
```diff
@@ -31,7 +34,7 @@ import (
 //
 // Refer to estargz for the options:
 // https://pkg.go.dev/github.com/containerd/stargz-snapshotter@v0.2.0/estargz#Option
-func ReadCloser(r io.ReadCloser, opts ...estargz.Option) (io.ReadCloser, v1.Hash, error) {
+func ReadCloser(r io.ReadCloser, opts ...estargz.Option) (*estargz.Blob, v1.Hash, error) {
```
jonjohnsonjr, line 37: Hmm... so all we're doing here is actually implementing `func(io.ReadCloser) (io.SectionReader, error)`...

Do we really need this? If we want to keep it, can we update the return signature to match `estargz.Build`? I'd honestly rather they change their interface and take advantage of any implemented interfaces, and then we just drop this entirely.

mattmoor: This is a projection to our idioms, e.g. taking `io.ReadCloser` and returning `v1.Hash` (they cannot adopt the latter without a dependency cycle, but we could switch to using `go-digest` everywhere).

As #876 evolves, perhaps the necessity of this will diminish, but I think it's a useful place for us to mitigate things in the interim, and since it's internal there's effectively zero risk that we can't remove it.  🤷

pkg/v1/internal/estargz/constants.go
mattmoor: done

jonjohnsonjr (approved):

PR #872, opened 2020-12-16
Towards eliminating v1util

mattmoor:
This blows apart v1util into smaller (mostly internal) packages, keeping only aliases of what we have today under `pkg/v1/v1util` to avoid breaking downstreams, but adding a deprecation notice, which should trigger `golangci-lint` to start flagging users of the deprecated methods.


pkg/v1/mutate/mutate.go
```diff
@@ -335,7 +335,7 @@ func layerTime(layer v1.Layer, t time.Time) (v1.Layer, error) {
 	b := w.Bytes()
 	// gzip the contents, then create the layer
 	opener := func() (io.ReadCloser, error) {
-		return v1util.GzipReadCloser(ioutil.NopCloser(bytes.NewReader(b))), nil
+		return gzip.GzipReadCloser(ioutil.NopCloser(bytes.NewReader(b))), nil
```
jonjohnsonjr, line 338: Drop the stutter?

jonjohnsonjr (commented):
I think the convention is `Deprecated: ` per https://github.com/golang/go/wiki/Deprecated

ImJasonH:
`pkg/v1/v1util` -> `pkg/v1/internal` has roughly the same effect, and is simply mechanical. Splitting it into separate internal packages and moving some to still-exported symbols in `v1` seems like more work than necessary.

pkg/v1/mutate/mutate.go
mattmoor: What do you want to call the Gunzip variant :)

jonjohnsonjr: You got me there.

mattmoor:
> pkg/v1/v1util -> pkg/v1/internal has roughly the same effect, and is simply mechanical

Are you suggesting we have a literal `internal` package?  I kinda thought part of moving away from `v1util` was to get better package naming.

> and moving some to still-exported symbols in v1 seems like more work than necessary

Not parsing, but exposing the things that were previously public under v1util with `Deprecated:` was intentional to not break folks immediately.

jonjohnsonjr:
> Not parsing, but exposing the things that were previously public under v1util with Deprecated: was intentional to not break folks immediately.

+1, we owe them one release of deprecated before we drop it, IMO.

mattmoor:
I can `TODO(#1234)` removing, if we want to track this better.

ImJasonH:
> Are you suggesting we have a literal `internal` package? I kinda thought part of moving away from `v1util` was to get better package naming.

I'm suggesting we stage the change first by making it `internal` (to shake off external deps), then we have the flexibility to move methods as unexported where they're used if we want.

Package naming and organization matters a lot less to me inside `internal` where it's clear it's only for our own uses.

> Not parsing, but exposing the things that were previously public under v1util with `Deprecated:` was intentional to not break folks immediately.

Yeah, that's fine, I was talking about `v1.VerifyReadCloser`, which I think we should keep internal unless we know somewhere important someone's depending on it.

mattmoor:
> I'm suggesting we stage the change first by making it internal (to shake off external deps)

Just making it internal is breaking, and I don't think we should just break folks currently using it (until a few days ago this included `ko`, are we sure Bazel isn't using this?).

The intent of staging things this way was to make it clear that the external exposure is Deprecated (and will break), while simultaneously enabling us to restructure things in a way that's consistent with what y'all are asking for with `pkg/v1/internal/estargz`.

> Package naming and organization matters a lot less to me inside internal where it's clear it's only for our own uses

I completely agree, which is why I didn't sweat the "stutter" stuff right away.

> I was talking about v1.VerifyReadCloser

I'm happy to make this private, I don't *really* care, but digest verification just felt like something folks would want, but it is not a hill I'd die on 😉

pkg/v1/mutate/mutate.go
mattmoor: I was piggybacking on the sentiment @ImJasonH had below that renaming things becomes *reeeeaaaaal* easy under `internal/`, so this is trivial to fixup later if we find a name we like. 🤷

mattmoor:
Alright, we now have `verify.ReadCloser` in an internal package (aliased as deprecated `v1util.VerifyReadCloser`)

ImJasonH:
To be clear I'm absolutely on board with aliasing + deprecating for now. It was the atomizing into internal packages I was commenting about. Sorry that wasn't clear.

mattmoor:
It feels like we are quibbling about internal structure, which we all agreed that we can now change whenever we want.  I went with this because it seemed like it was consistent with what was being suggested for estargz, and it seemed like we all agreed about this on slack:

image

The "consistent structure" I was talking about above was "the atomizing" 😅 

Perhaps we're talking past each other, so it may be simplest to hop on a zoom (🤔  now I'm wondering if this is a ploy to see the 🐶 , very :sus:)

jonjohnsonjr (approved):

mattmoor:
@ImJasonH I think this covers the key points that we discussed.  I am happy to keep making changes, which should be easy now that this stuff is under `internal/`.  Let me know if you want to chat, and I can `/zoom` you.

ImJasonH:
Lgtm, thanks for putting up with my quibbling 😆

PR #877, opened 2020-12-17
Update the philosophy

mattmoor:
I took a stab at this.  I wanted this to better reflect the immutability + function mutation + composability of the interface design.


jonjohnsonjr (approved):
Thanks. You have such a way with words.

Issue #878, opened 2020-12-18
implement `crane estargz`


mattmoor:
The intent of this is to be like `crane cp`, but to convert the image to the estargz format.

For bonus points, this could take a list of files to prioritize 🤔 

cc @ImJasonH @jonjohnsonjr for thoughts.

ImJasonH:
How's it different than `GGCR_EXPERIMENT_ESTARGZ=1 crane cp foo bar`? I thought the goal was to make eStargz optimizing transparent?

mattmoor:
I wouldn't ever expect `crane cp` to change the digest, and I would expect it to be efficient about not fetching layers when it can mount them, so I think a fundamentally different command would be useful.

ImJasonH:
`crane cp foo bar --optimize`? Just throwing out ideas.

I don't know that people will necessarily care about the specific eStargz implementation, and there even may be other versions/options in the future (like stargz and eStargz)

mattmoor:
Maybe `crane {convert,optimize} --type=estargz foo bar`?

multiple modes makes sense for sure.

ImJasonH:
> Maybe `crane {convert,optimize} --type=estargz foo bar`?

Yeah I like that.

Is there anything else {e}stargz related we might be able to add? Validating the TOCs match their digests, and actual layer contents?  Efficiently fetching a single file with `crane estargz {something} {path}` ?

mattmoor:
🤩  An optimized `crane export` w/ file list would be 🔥 

Probably worth a second issue?

mattmoor:
I'm going to go with `crane optimize` for now.  Hopefully I'll have something to show by HackyHour 🤞

jonjohnsonjr:
So this ties into https://github.com/google/go-containerregistry/issues/732 and some other issues a little bit. I want to take some inspiration from skopeo here, but matt's point about digest preservation is salient: https://github.com/containers/skopeo/issues/1102

Some form of `crane copy` that supports changing formats would be dope. We can alias `crane pull` and `crane push` to just do this. I'm okay with `crane convert` for a name. I would suggest `crane mv`, but I really want that to basically `crane tag && crane delete`, so don't use that...

The way skopeo handles `--type` flag is to just have schemes for different src/dst types, e.g. `containers-storage:docker-reference`, `docker://`, `docker-archive:path[:docker-reference]`, `docker-daemon:docker-reference`, `oci:path:tag`.

[...]

PR #879, opened 2020-12-19
Start to flesh out crane optimize.

mattmoor:
This is a hidden command, which roundtrips a remote image to a target image through `tarball.LayerFromOpener(layer.Uncompressed)`.

Right now this does nothing to force estargz (still need `GGCR_EXPERIMENT_ESTARGZ=1`) or prioritize files (need `estargz.WithPrioritizedFiles(foo)`), but want to start the convo.

Fixes: https://github.com/google/go-containerregistry/issues/878

### DIFF
--- a/cmd/crane/cmd/optimize.go
+++ b/cmd/crane/cmd/optimize.go
@@ -23,7 +23,9 @@ import (
 
 // NewCmdOptimize creates a new cobra.Command for the optimize subcommand.
 func NewCmdOptimize(options *[]crane.Option) *cobra.Command {
-	return &cobra.Command{
+	var files []string
+
+	cmd := &cobra.Command{
 		Use:     "optimize SRC DST",
 		Hidden:  true,
 		Aliases: []string{"opt"},
@@ -31,9 +33,14 @@ func NewCmdOptimize(options *[]crane.Option) *cobra.Command {
 		Args:    cobra.ExactArgs(2),
 		Run: func(_ *cobra.Command, args []string) {
 			src, dst := args[0], args[1]
-			if err := crane.Optimize(src, dst, *options...); err != nil {
+			if err := crane.Optimize(src, dst, files, *options...); err != nil {
 				log.Fatal(err)
 			}
 		},
 	}
+
+	cmd.Flags().StringSliceVar(&files, "prioritize", nil,
+		"The list of files to prioritize in the optimized image.")
+
+	return cmd
 }
--- a/cmd/crane/cmd/optimize.go
+++ b/cmd/crane/cmd/optimize.go
@@ -1,0 +1,39 @@
+// Copyright 2018 Google LLC All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package cmd
+
+import (
+	"log"
+
+	"github.com/google/go-containerregistry/pkg/crane"
+	"github.com/spf13/cobra"
+)
+
+// NewCmdOptimize creates a new cobra.Command for the optimize subcommand.
+func NewCmdOptimize(options *[]crane.Option) *cobra.Command {
+	return &cobra.Command{
+		Use:     "optimize SRC DST",
+		Hidden:  true,
+		Aliases: []string{"opt"},
+		Short:   "Optimize a remote container image from src to dst",
+		Args:    cobra.ExactArgs(2),
+		Run: func(_ *cobra.Command, args []string) {
+			src, dst := args[0], args[1]
+			if err := crane.Optimize(src, dst, *options...); err != nil {
+				log.Fatal(err)
+			}
+		},
+	}
+}
--- a/cmd/crane/cmd/optimize.go
+++ b/cmd/crane/cmd/optimize.go
@@ -1,4 +1,4 @@
-// Copyright 2018 Google LLC All Rights Reserved.
+// Copyright 2020 Google LLC All Rights Reserved.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
--- a/pkg/crane/optimize.go
+++ b/pkg/crane/optimize.go
@@ -1,0 +1,176 @@
+// Copyright 2018 Google LLC All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package crane
+
+import (
+	"errors"
+	"fmt"
+
+	"github.com/google/go-containerregistry/pkg/logs"
+	"github.com/google/go-containerregistry/pkg/name"
+	v1 "github.com/google/go-containerregistry/pkg/v1"
+	"github.com/google/go-containerregistry/pkg/v1/empty"
+	"github.com/google/go-containerregistry/pkg/v1/mutate"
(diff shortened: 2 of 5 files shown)


mattmoor:
Alright, I also have:
```
crane optimize gcr.io/distroless/static:nonroot ghcr.io/mattmoor/distroless/static:nonroot --prioritize=etc/passwd
```

... working to prioritize `/etc/passwd`, though I'm not sure there's a good way to tell whether it found the file in question, and I get different digests when I pass `/etc/passwd` vs. `etc/passwd` vs. nothing 🤔 

cc @ktock @AkihiroSuda FYI

cmd/crane/cmd/optimize.go
```diff
@@ -0,0 +1,46 @@
+// Copyright 2018 Google LLC All Rights Reserved.
```
jonjohnsonjr, line 1:
```suggestion
// Copyright 2020 Google LLC All Rights Reserved.
```

cmd/crane/cmd/optimize.go
```diff
@@ -0,0 +1,46 @@
...
+	"github.com/spf13/cobra"
+)
+
+// NewCmdOptimize creates a new cobra.Command for the optimize subcommand.
+func NewCmdOptimize(options *[]crane.Option) *cobra.Command {
+	var files []string
+
+	cmd := &cobra.Command{
+		Use:     "optimize SRC DST",
+		Hidden:  true,
```
jonjohnsonjr, line 30: Because this is hidden, I don't _really_ care that much, but I would rather this be called something like `[e]stargz[ify]` (and documented as such), because "optimize" doesn't mean much.

pkg/crane/optimize.go
```diff
@@ -0,0 +1,178 @@
+// Copyright 2018 Google LLC All Rights Reserved.
```
jonjohnsonjr, line 1:
```suggestion
// Copyright 2020 Google LLC All Rights Reserved.
```

cmd/crane/cmd/optimize.go
```diff
@@ -0,0 +1,46 @@
...
+		Aliases: []string{"opt"},
+		Short:   "Optimize a remote container image from src to dst",
+		Args:    cobra.ExactArgs(2),
```
jonjohnsonjr, line 33: It would be nice if this allowed one arg and would just retag the optimized version, with an optional `-t` flag for the dst if you want to leave src alone.

pkg/crane/optimize.go
```diff
@@ -0,0 +1,178 @@
...
+			return nil, err
+		}
+		adds = append(adds, mutate.IndexAddendum{
+			Add: oimg,
+			Descriptor: v1.Descriptor{
+				URLs:        desc.URLs,
+				MediaType:   desc.MediaType,
+				Annotations: desc.Annotations,
+				Platform:    desc.Platform,
+			},
```
jonjohnsonjr, line 168: We could probably do something in `mutate` to make this a little nicer.

pkg/crane/optimize.go
```diff
@@ -0,0 +1,178 @@
...
+
+	oimg, err := optimizeImage(img, prioritize)
+	if err != nil {
+		return err
+	}
+
+	return remote.Write(dstRef, oimg, o.remote...)
+}
+
+func optimizeImage(img v1.Image, prioritize []string) (v1.Image, error) {
```
jonjohnsonjr, line 92: What you're doing follows the patterns from `crane.Copy`, which makes sense, but it would be nice to expose this separately as well, maybe `OptimizeImage`, so it can be used with non-remote images.

It might also make sense for some of this boilerplate to live in `mutate`.

pkg/crane/optimize.go
```diff
@@ -0,0 +1,178 @@
...
+	v1 "github.com/google/go-containerregistry/pkg/v1"
+	"github.com/google/go-containerregistry/pkg/v1/empty"
+	"github.com/google/go-containerregistry/pkg/v1/mutate"
+	"github.com/google/go-containerregistry/pkg/v1/remote"
+	"github.com/google/go-containerregistry/pkg/v1/tarball"
+	"github.com/google/go-containerregistry/pkg/v1/types"
+)
+
+// Optimize optimizes a remote image or index from src to dst.
+func Optimize(src, dst string, prioritize []string, opt ...Option) error {
```
jonjohnsonjr, line 33: My "as long as it's hidden" concession for the command also kind of applies here. I don't think we should expose this in pkg/crane until we're happy with exposing it in the CLI as well.

pkg/crane/optimize.go
```diff
@@ -0,0 +1,178 @@
...
+
+	oidx, err := optimizeIndex(idx, prioritize)
+	if err != nil {
+		return err
+	}
+
+	return remote.WriteIndex(dstRef, oidx, o.remote...)
+}
+
+func optimizeIndex(idx v1.ImageIndex, prioritize []string) (v1.ImageIndex, error) {
```
jonjohnsonjr, line 143: We really need a:

```go
func Map(f func(v1.Image) (v1.Image, error), i v1.ImageOrIndex) (v1.ImageOrIndex, error)
```

pkg/crane/optimize.go
```diff
@@ -0,0 +1,178 @@
...
+	}
+
+	layers, err := img.Layers()
+	if err != nil {
+		return nil, err
+	}
+
+	olayers := make([]mutate.Addendum, 0, len(layers))
+	for i, layer := range layers {
+		olayer, err := tarball.LayerFromOpener(layer.Uncompressed,
```
jonjohnsonjr, line 113: Never even thought about the fact that this works. Nice.

pkg/crane/optimize.go
mattmoor: I believe I have one downstream somewhere 😇

cmd/crane/cmd/optimize.go
mattmoor: done

cmd/crane/cmd/optimize.go
mattmoor: I'll let you and @ImJasonH fight about this.  That was my original proposal and I care almost not at all.

cmd/crane/cmd/optimize.go
mattmoor: IDK how I feel about in-place optimization, but most of the places I want this for are things that aren't yet producing these images.

pkg/crane/optimize.go
mattmoor: https://github.com/mattmoor/mink/blob/4afdeb3398c4d56401d4638ebd5cb2cb85ad12ad/pkg/bundles/map.go#L118

pkg/crane/optimize.go
mattmoor: Given how things are structured, that isn't particularly feasible (without deviating significantly from the broader style).  What did you have in mind?

pkg/crane/optimize.go
```diff
@@ -0,0 +1,178 @@
...
+			tarball.WithEstargzOptions(estargz.WithPrioritizedFiles(prioritize)))
+		if err != nil {
+			return nil, err
+		}
+
+		olayers = append(olayers, mutate.Addendum{
+			Layer:     olayer,
+			History:   cfg.History[i],
```
mattmoor: I may drop this as I'm not sure it is actually correct.

pkg/crane/optimize.go
mattmoor: done

pkg/crane/optimize.go
jonjohnsonjr: Oh yeah you cced me on this earlier. My only issue is the pushing bit, we are missing an abstraction here for an ImageOrIndex thing (strawman here: https://github.com/google/go-containerregistry/pull/835)

pkg/crane/optimize.go
jonjohnsonjr: You could keep this under cmd/crane/cmd, which is what I've done for `gcrane` stuff I don't want to expose.

jonjohnsonjr:
Moving this down from inline comment so it doesn't get lost...

> IDK how I feel about in-place optimization, but most of the places I want this for are things that aren't yet producing these images.

Given that it's ostensibly backward-compatible, I don't see a huge issue, but I understand the reservations.

If we _do_ in-place, one thing that would be nice for one-liners is to print the resulting image to stdout:

```
$ crane optimize foo@sha256:abc
foo@sha256:def
```

So you could do stuff like:

```
$ crane export $(crane optimize $(ko publish ./cmd/ko)) | tar -tvf -
```

That might not make sense for in-place by tag, but:
```
$ crane optimize foo:latest
foo:latest@sha256:def
```

Would be kinda nice? I like the pattern in general for mutation stuff, especially if you don't want to write me a compiler so this stuff only works in-registry. I'd be okay with needing to set a flag to overwrite tags, as well.

mattmoor:
BTW, I think being able to collect missed files (https://github.com/containerd/stargz-snapshotter/pull/216) will be very useful for surfacing user errors with the prioritized file stuff (thanks @ktock!).

jonjohnsonjr (approved):

PR #880, opened 2020-12-19
Use gcrane.Keychain for not only crane-inherit commands but also gcrane-specific commands.

FGtatsuro:
https://github.com/google/go-containerregistry/pull/845 introduces `gcrane.Keychain`.

This PR uses `gcrane.Keychain` for not only crane-inherit commands(ex. pull) but also gcrane-specific commands(gc/ls).


jonjohnsonjr (approved):
Nice catch, thanks!

PR #882, opened 2020-12-21
Include target resource in credential warning

jonjohnsonjr:
This should be slightly less confusing the first time you see it, we can look at removing this entirely by skipping ping and only falling back on 401, but that might take a while to implement.

### DIFF
--- a/pkg/v1/google/options.go
+++ b/pkg/v1/google/options.go
@@ -49,7 +49,7 @@ func WithAuthFromKeychain(keys authn.Keychain) ListerOption {
 			return err
 		}
 		if auth == authn.Anonymous {
-			logs.Warn.Println("No matching credentials were found, falling back on anonymous")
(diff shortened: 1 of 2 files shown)


pkg/v1/google/options.go
```diff
@@ -49,7 +49,7 @@ func WithAuthFromKeychain(keys authn.Keychain) ListerOption {
 			return err
 		}
 		if auth == authn.Anonymous {
-			logs.Warn.Println("No matching credentials were found, falling back on anonymous")
+			logs.Warn.Printf("No matching credentials were found for %q, falling back on anonymous", l.repo.Registry)
```
mattmoor, line 52: Does this need a newline?

jonjohnsonjr: Nope, from https://golang.org/pkg/log/:

> Every log message is output on a separate line: if the message being printed does not end in a newline, the logger will add one.

mattmoor (approved):

PR #883, opened 2020-12-22
Fail without pushing when `crane optimize` misses files

mattmoor:


mattmoor:
Sample failing output (nothing pushed):
```
# crane optimize gcr.io/distroless/static:nonroot ghcr.io/mattmoor/distroless/static:nonroot --prioritize /a/b --prioritize /c/d
2020/12/22 08:40:52 Optimizing from gcr.io/distroless/static:nonroot to ghcr.io/mattmoor/distroless/static:nonroot
2020/12/22 08:40:52 No matching credentials were found for "gcr.io/distroless/static", falling back on anonymous
2020/12/22 08:40:56 failed to optimize index: the following prioritized files were missing from all images: [/a/b /c/d]
```

Sample successful output:
```
# crane optimize gcr.io/distroless/static:nonroot ghcr.io/mattmoor/distroless/static:nonroot --prioritize /etc/passwd
2020/12/22 08:41:34 Optimizing from gcr.io/distroless/static:nonroot to ghcr.io/mattmoor/distroless/static:nonroot
2020/12/22 08:41:34 No matching credentials were found for "gcr.io/distroless/static", falling back on anonymous
2020/12/22 08:41:38 existing manifest: sha256:96fd34242ad0be6d50809b9f4a2b2fdfd79e8fb4fbef36c8905e53816bd7a69d
2020/12/22 08:41:38 existing manifest: sha256:4ca465ad088e114d4c207f268cb0cfd59ff1feae16d1d8ca4e7c4857863d533e
2020/12/22 08:41:38 existing manifest: sha256:93fd2b8ecf0d102094930acef84d064253e2c9cc307354f4dbb89e1e7d35915c
2020/12/22 08:41:38 existing manifest: sha256:83564ecf2de20e3ab36aad46ab320574a54a9d149a952b0f9534a516a2707815
2020/12/22 08:41:39 existing manifest: sha256:0bd13356efab3bf5162bd2e6f074d65205c99f101155711eb68d0495773316fc
2020/12/22 08:41:39 ghcr.io/mattmoor/distroless/static:nonroot: digest: sha256:e09730b49128208050b41a6e42991ea64716af1b4ae17bb697971c31c767965d size: 1165
```

mattmoor:
@ktock thanks for the new option 🤩

mattmoor:
I am trying to pull in https://github.com/containerd/stargz-snapshotter/pull/225 to fix the Windows leg 🤞

mattmoor:
Alright, looks like this needs: https://github.com/containerd/stargz-snapshotter/pull/225 to land.

mattmoor:
@ImJasonH @jonjohnsonjr this is RFAL (upstream merged, and this is rebased)

pkg/crane/optimize.go
```diff
@@ -76,89 +77,110 @@ func Optimize(src, dst string, prioritize []string, opt ...Option) error {
...
-			tarball.WithEstargzOptions(estargz.WithPrioritizedFiles(prioritize)))
+			tarball.WithEstargzOptions(
+				estargz.WithPrioritizedFiles(prioritize.List()),
+				estargz.WithAllowPrioritizeNotFound(&missingFromLayer),
+			))
 		if err != nil {
-			return nil, err
+			return nil, nil, err
 		}
+		missingFromImage = missingFromImage.Intersection(sets.NewString(missingFromLayer...))
```
ImJasonH, line 130: If the presence of missing files is considered an error, why not just return an error if this set is non-empty? This makes the method a bit easier to use, since you don't have to remember to check for returned missing files.

mattmoor: Because they might be missing from one image, but not the index.

We only error before pushing because it is only at that point that we have no more places we might find the file.

pkg/crane/optimize.go
```diff
@@ -27,11 +27,12 @@ import (
 	"github.com/google/go-containerregistry/pkg/v1/remote"
 	"github.com/google/go-containerregistry/pkg/v1/tarball"
 	"github.com/google/go-containerregistry/pkg/v1/types"
+	"k8s.io/apimachinery/pkg/util/sets"
```
jonjohnsonjr, line 30: Not super happy about depending on this.

pkg/crane/optimize.go
jonjohnsonjr: Interesting. For an index, this only fails if a prioritized file is not found in _any_ of the images. That makes some sense to me, but isn't super intuitive...

Maybe worth logging a warning or requiring a flag that says "yeah I know, these files might not be in every image".

pkg/crane/optimize.go
mattmoor: It's fine :P

pkg/crane/optimize.go
mattmoor: My original change only warned for missing files and I found the output easily got lost in the push output, so I opted instead for making it fatal (and not pushing the image(s)).

I do think it would be interesting to add a verbose mode that displays:
1. Which files are found in which layers
2. Which images are missing which files

... might be useful, but this feels more diagnostic than something I'd want always-on...

TBH I don't know how common different files are for different architectures.  I know for muti-arch kaniko (b/c of how Bazel names outputs) I ended up making the key binary `/kaniko/executor_{arch}` because it was the only way to get multiple binaries out of Bazel for the same Go 🙃 

I suspect that if this takes off, then we're going to want to have significantly more diagnostic tooling around this (e.g. inspecting the TOC of a layer, which files are prioritized?)

jonjohnsonjr: > I suspect that if this takes off, then we're going to want to have significantly more diagnostic tooling around this (e.g. inspecting the TOC of a layer, which files are prioritized?)

Maybe an entire `estargz` subcommand? 😉

mattmoor: It's hidden, so we can always do that later 🤷

jonjohnsonjr (approved):
I'm still not happy about the set thing but it lgtm. When we isolate k8schain (move it or give it its own go.mod), I'll remove this.

PR #885, opened 2020-12-23
Peek at gzip header bytes

jonjohnsonjr:
First commit doesn't take advantage of the new interface yet, but this allows
us to avoid consuming a ReadCloser just to get the first two bytes.

Second commit uses this for tarball.Layer.

Doing this in tarball.Image will require some more work.

### DIFF
--- a/pkg/v1/internal/gzip/zip.go
+++ b/pkg/v1/internal/gzip/zip.go
@@ -83,18 +83,31 @@ func UnzipReadCloser(r io.ReadCloser) (io.ReadCloser, error) {
 	}, nil
 }
 
-// Is detects whether the input stream is gzip compressed. A small number
-// of bytes are buffered to Peek at the gzip header so that the returned
-// io.Reader can be used as a replacement for the consumed input io.Reader.
-func Is(r io.Reader) (bool, *bufio.Reader, error) {
-	buf := bufio.NewReader(r)
-	header, err := buf.Peek(2)
+// PeekReader is an io.Reader that also implements Peek a la bufio.Reader.
+type PeekReader interface {
+	io.Reader
+	Peek(n int) ([]byte, error)
+}
+
+// Is detects whether the input stream is gzip compressed.
+//
+// If r implements Peek, we will use that directly, otherwise a small number
+// of bytes are buffered to Peek at the gzip header, and the returned
+// PeekReader can be used as a replacement for the consumed input io.Reader.
+func Is(r io.Reader) (bool, PeekReader, error) {
+	var pr PeekReader
+	if p, ok := r.(PeekReader); ok {
+		pr = p
+	} else {
+		pr = bufio.NewReader(r)
+	}
+	header, err := pr.Peek(2)
 	if err != nil {
 		// https://github.com/google/go-containerregistry/issues/367
 		if err == io.EOF {
-			return false, buf, nil
+			return false, pr, nil
 		}
-		return false, buf, err
+		return false, pr, err
 	}
-	return bytes.Equal(header, gzipMagicHeader), buf, nil
+	return bytes.Equal(header, gzipMagicHeader), pr, nil
 }
--- a/pkg/v1/internal/gzip/zip.go
+++ b/pkg/v1/internal/gzip/zip.go
@@ -15,6 +15,7 @@
 package gzip
 
 import (
+	"bufio"
 	"bytes"
 	"compress/gzip"
 	"io"
@@ -82,15 +83,18 @@ func UnzipReadCloser(r io.ReadCloser) (io.ReadCloser, error) {
 	}, nil
 }
 
-// Is detects whether the input stream is compressed.
-func Is(r io.Reader) (bool, error) {
-	magicHeader := make([]byte, 2)
-	n, err := r.Read(magicHeader)
-	if n == 0 && err == io.EOF {
-		return false, nil
-	}
+// Is detects whether the input stream is gzip compressed. A small number
+// of bytes are buffered to Peek at the gzip header so that the returned
+// io.Reader can be used as a replacement for the consumed input io.Reader.
+func Is(r io.Reader) (bool, *bufio.Reader, error) {
+	buf := bufio.NewReader(r)
+	header, err := buf.Peek(2)
 	if err != nil {
-		return false, err
+		// https://github.com/google/go-containerregistry/issues/367
+		if err == io.EOF {
+			return false, buf, nil
+		}
+		return false, buf, err
 	}
-	return bytes.Equal(magicHeader, gzipMagicHeader), nil
+	return bytes.Equal(header, gzipMagicHeader), buf, nil
 }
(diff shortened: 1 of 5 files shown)


pkg/v1/internal/gzip/zip.go
```diff
@@ -82,15 +83,17 @@ func UnzipReadCloser(r io.ReadCloser) (io.ReadCloser, error) {
...
+// of bytes are buffered to Peek at the gzip header so that the returned
+// io.Reader can be used as a replacement for the consumed input io.Reader.
+func Is(r io.Reader) (bool, io.Reader, error) {
+	buf := bufio.NewReader(r)
+	header, err := buf.Peek(2)
 	if err != nil {
-		return false, err
+		if err == io.EOF {
+			return false, buf, nil
+		}
```
mattmoor, line 95: Why swallow the EOF?

pkg/v1/tarball/layer_test.go
```diff
@@ -200,9 +200,12 @@ func TestLayerFromOpenerReader(t *testing.T) {
 		tarLayer.Compressed()
 	}
 
-	// We expect three calls: gzip sniff, diffid computation, cached compression
-	if cachedCount != 3 {
-		t.Errorf("cached count = %d, wanted %d", cachedCount, 3)
+	// ~~We expect three calls: gzip sniff, diffid computation, cached compression~~
```
mattmoor, line 203: ~~?

mattmoor: stale comment

pkg/v1/tarball/layer.go
```diff
@@ -281,3 +286,21 @@ func computeDiffID(opener Opener) (v1.Hash, error) {
...
+	mu := sync.Mutex{}
+	consumed := false
+	return func() (io.ReadCloser, error) {
+		mu.Lock()
+		defer mu.Unlock()
+
+		if !consumed {
+			consumed = true
+			return rc, nil
+		}
```
mattmoor, line 303: I'd be there's a way to do this more efficiently for reopens with `sync.Once` wrt synchronization.

jonjohnsonjr: Something like this?

```go
func reopener(rc io.ReadCloser, opener Opener) Opener {
        var once sync.Once
        return func() (io.ReadCloser, error) {
                open := opener
                once.Do(func() {
                        open = func() (io.ReadCloser, error) {
                                return rc, nil
                        }
                })
                return open()
        }
}
```

I don't think that's actually safe and benchmarking reveals no substantial speedup either way. Do you have something else in mind?

mattmoor: > I don't think that's actually safe

Why not?

jonjohnsonjr: I guess the more likely path is `if consumed` so maybe it makes sense to move the `if !consumed` path into a separate function... I guess we could copy over the `sync.Once` implementation?

jonjohnsonjr: Here's `sync.Once` inlined, without the "slow path" outlined:

```go
func reopener(rc io.ReadCloser, opener Opener) Opener {
        var (
                mu       sync.Mutex
                consumed uint32
        )
        return func() (io.ReadCloser, error) {
                if atomic.LoadUint32(&consumed) == 0 {
                        mu.Lock()
                        defer mu.Unlock()

                        if consumed == 0 {
                                defer atomic.StoreUint32(&consumed, 1)
                                return rc, nil
                        }
                }

                return opener()
        }
}
```

jonjohnsonjr: > Why not?

Eh... I was thinking because the `return open()` happens outside of the sync.Once mutex, we'd have a race between goroutines, but thinking about it, we're mutating the `open` on the stack of whichever goroutine hit that first, so it doesn't matter at all? 🤷

pkg/v1/internal/gzip/zip.go
jonjohnsonjr: tl;dr the answer to "is this gzipped?" is "no" if we encounter an `io.EOF` by reading the first two bytes.

pkg/v1/tarball/layer.go
jonjohnsonjr: Pushed with this impl: https://github.com/google/go-containerregistry/pull/885#discussion_r548260635

mattmoor: I would use: https://godoc.org/sync/atomic#CompareAndSwapInt32

jonjohnsonjr: Done

pkg/v1/internal/gzip/zip.go
```diff
@@ -82,15 +83,18 @@ func UnzipReadCloser(r io.ReadCloser) (io.ReadCloser, error) {
...
-func Is(r io.Reader) (bool, error) {
-	magicHeader := make([]byte, 2)
-	n, err := r.Read(magicHeader)
-	if n == 0 && err == io.EOF {
-		return false, nil
-	}
+// Is detects whether the input stream is gzip compressed. A small number
+// of bytes are buffered to Peek at the gzip header so that the returned
+// io.Reader can be used as a replacement for the consumed input io.Reader.
+func Is(r io.Reader) (bool, *bufio.Reader, error) {
```
jonjohnsonjr, line 89: Thinking about this a little more, it is kind of cumbersome to use, I'm  okay with merging this as-is for a quick win, but what do you think about some kind of interface like this that hides all of it:

```go
type Reopener struct {
  consumed int32
  gzipped bool
  rc io.ReadCloser
  opener Opener
}

func (ro *Reopener) Open() (io.ReadCloser, error) {
  if atomic.CompareAndSwapInt32(&ro.consumed, 0, 1) {
    return ro.rc, nil
  }
  return rc.opener()
}

func (ro  *Reopener) IsGzipped() bool {
  return ro.gzipped
}

func Reopen(opener Opener) (*Reopener, error) {
  rc, err := opener()
  if err != nil {
    return nil, false err
  }

  br, ok, err := Is(rc)
  if err != nil {
    return false, nil, err
  }

  brc := &and.ReadCloser{
    Reader:    r,
    CloseFunc: rc.Close,
  }

  return &Reopener{
    gzipped: ok,
    rc: rc,
    opener: opener,
  }, nil
}
```

And in `tarball` we can just:

```go
reopener, err := ggzip.Reopen(rc)
if err != nil {
  return nil, err
}

compressed := reopener.IsGzipped()
opener = reopener.Open
```

Names are hard, so I hate them, and perhaps the "reopener" stuff doesn't make sense within the ggzip package because "Opener" is a `tarball` abstraction...

Maybe `ggzip` should just expose something that compares a `[]byte` to `gzipMagicHeader` or even just exposes that array as `ggzip.MagicHeader`? I'm not sure, but having `gzip.Is` take an `io.Reader` certainly feels goofier the more I look at this PR.

jonjohnsonjr:
Related to https://github.com/google/go-containerregistry/pull/885

PR #887, opened 2020-12-26
fail computing manifest if no refs are provided

jchorl:
This PR just adds some validation to ensure we're not producing invalid manifests. `null` might be a valid digest but I can't imagine any user intentionally producing an image with a `null` manifest (Docker can't import that or do anything useful with it).

I found a weird quirk in [kaniko](https://github.com/GoogleContainerTools/kaniko) where they were passing no refs into `WriteToFile`. Unfortunately, this produced a `nil` manifest, which json-encoded to `null` (valid json!), so I ended up with a tarball with only a `manifest.json` with the contents `null`.

Of course, ECR and Docker were happy to accept this tarball and just... silently do nothing with it.

I opened a [PR](https://github.com/moby/moby/pull/41842) to fix docker loads with a `null` manifest. I'll also open a PR to kaniko to reject no destinations.

### DIFF
--- a/pkg/v1/tarball/write.go
+++ b/pkg/v1/tarball/write.go
@@ -222,6 +222,10 @@ func writeImagesToTar(refToImage map[name.Reference]v1.Image, m []byte, size int
 func calculateManifest(refToImage map[name.Reference]v1.Image) (m Manifest, err error) {
 	imageToTags := dedupRefToImage(refToImage)
 
+	if len(imageToTags) == 0 {
+		return nil, errors.New("at least one reference must be provided to calculate a manifest")
+	}
+
 	for img, tags := range imageToTags {
 		cfgName, err := img.ConfigName()
 		if err != nil {
--- a/pkg/v1/tarball/write.go
+++ b/pkg/v1/tarball/write.go
@@ -223,7 +223,7 @@ func calculateManifest(refToImage map[name.Reference]v1.Image) (m Manifest, err 
 	imageToTags := dedupRefToImage(refToImage)
 
 	if len(imageToTags) == 0 {
-		return nil, errors.New("at least one reference must be provided to calculate a manifest")
+		return nil, errors.New("set of images is empty")
 	}
 
 	for img, tags := range imageToTags {
--- a/pkg/v1/tarball/write_test.go
+++ b/pkg/v1/tarball/write_test.go
@@ -463,12 +463,12 @@ func TestComputeManifest(t *testing.T) {
 
 func TestComputeManifest_FailsOnNoRefs(t *testing.T) {
 	_, err := tarball.ComputeManifest(nil)
-	if err == nil || !strings.Contains(err.Error(), "at least one reference must be provided to calculate a manifest") {
+	if err == nil || !strings.Contains(err.Error(), "set of images is empty") {
 		t.Error("expected calculateManifest to fail with nil input")
 	}
 
 	_, err = tarball.ComputeManifest(map[name.Reference]v1.Image{})
-	if err == nil || !strings.Contains(err.Error(), "at least one reference must be provided to calculate a manifest") {
+	if err == nil || !strings.Contains(err.Error(), "set of images is empty") {
 		t.Error("expected calculateManifest to fail with empty input")
 	}
 }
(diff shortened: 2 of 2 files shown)


pkg/v1/tarball/write.go
```diff
@@ -222,6 +222,10 @@ func writeImagesToTar(refToImage map[name.Reference]v1.Image, m []byte, size int
 func calculateManifest(refToImage map[name.Reference]v1.Image) (m Manifest, err error) {
 	imageToTags := dedupRefToImage(refToImage)
 
+	if len(imageToTags) == 0 {
+		return nil, errors.New("at least one reference must be provided to calculate a manifest")
```
jonjohnsonjr, line 226: I realize this is mostly because we are inconsistent in how we're wrapping errors, and we should do better here, but in the common path we're going to see this returned as:

```
unable to calculate manifest: at least one reference must be provided to calculate a manifest
```

That seems a bit redundant, so I might change this error to `set of images is empty` or something along those lines.

jonjohnsonjr (commented):
Thanks for the PR!

> I found a weird quirk in kaniko where they were passing no refs into WriteToFile. Unfortunately, this produced a nil manifest, which json-encoded to null (valid json!), so I ended up with a tarball with only a manifest.json with the contents null.

Interesting! Is this intentional from their side? E.g. they're unconditionally writing stuff to a tarball, even if sometimes that's `nil`? I would suspect that's a bug on their side, but failing earlier on our side is definitely reasonable.

Instead of writing out `null`, should we be writing `[]` instead? Basically, I'm not sure where the bug is... I understand this isn't expected behavior and that we need to change _something_, but does returning an error here exclude producing a valid (if useless) image-less tarball? Is there any case for that being something we'd ever want to allow?

Your PR to docker makes it fail loudly when passed a `null` manifest. Your PR here makes us fail loudly w hen passed a `nil` map. So I'm guessing you just want this to fail more loudly and either PR would do the trick?

I'm guessing it's something [in here](https://github.com/GoogleContainerTools/kaniko/blob/9f76932171bf1ea4e61216a93642653c63ba499f/pkg/executor/push.go#L197-L224) where `len(destRefs) == 0`?

jonjohnsonjr:
Also interesting that this PR essentially fixes the other half: https://github.com/google/go-containerregistry/pull/697

jchorl:
> Is this intentional from their side? E.g. they're unconditionally writing stuff to a tarball, even if sometimes that's nil? I would suspect that's a bug on their side, but failing earlier on our side is definitely reasonable.

I can't imagine this is desired. I opened https://github.com/GoogleContainerTools/kaniko/pull/1534

> Basically, I'm not sure where the bug is... I understand this isn't expected behavior and that we need to change something, but does returning an error here exclude producing a valid (if useless) image-less tarball?

I wouldn't say there's a bug. But I think we can improve UX. I can't imagine anyone would want to pack a useless `null` manifest tarball. 

> So I'm guessing you just want this to fail more loudly and either PR would do the trick?

Yeah - basically any of the PRs to docker/kaniko/go-containerregistry would have saved me a big yakshave. I submitted a PR to each layer because I could see the same issue biting other people who use different tooling (might use go-containerregistry without kaniko, might use docker with some other build tool, etc). Once I was in the weeds and knew what was going on, I figured we should prevent other folks from needing to get in those weeds.

Anyway, my vote is to just error in this case. I think this will prevent a lot more frustration than it will cause. Even `[]` is silently failing. If you're asking go-containerregistry for a tarball, and don't give it the information it needs to build a tarball, you'd probably want to know. If you think this could cause problems, would prefer `[]`, or just think there's a better approach (including not merging at all), let me know and I'll make the changes.

jonjohnsonjr:
> Yeah - basically any of the PRs to docker/kaniko/go-containerregistry would have saved me a big yakshave.

Sorry about that :/

> Once I was in the weeds and knew what was going on, I figured we should prevent other folks from needing to get in those weeds.

Indeed... this often feels like my purpose in life :)

> or just think there's a better approach (including not merging at all), let me know and I'll make the changes.

I think erroring here is fine and less surprising than what we're currently doing. I'm happy to merge if you address my comment re: the error string 👍

jchorl:
> I'm happy to merge if you address my comment re: the error string

Sorry! Missed this part. Should be done now :)

Thanks so much for working through this with me :)

jonjohnsonjr (approved):
Thanks!

PR #891, opened 2020-12-29
Improve `append` command - ease the process to push manifests to remote registries from a set of images

vboulineau:
Early PR to to get feedback on the added feature: the `merge` command eases the following workflow:
CI -> one image / ARCH and/or OS -> build registry -> single manifest in remote registry

Command generates merged manifest from images or manifests. Currently platform information are always deduced and cannot be taken from user input.

Creating the PR to get feedbacks. If there's interest to get this feature in, I can polish/rework the PR as needed.

### DIFF
--- a/pkg/crane/merge.go
+++ b/pkg/crane/merge.go
@@ -1,0 +1,173 @@
+// Copyright 2018 Google LLC All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package crane
+
+import (
+	"fmt"
+
+	"github.com/google/go-containerregistry/pkg/logs"
+	"github.com/google/go-containerregistry/pkg/name"
+	v1 "github.com/google/go-containerregistry/pkg/v1"
+	"github.com/google/go-containerregistry/pkg/v1/empty"
+	"github.com/google/go-containerregistry/pkg/v1/mutate"
+	"github.com/google/go-containerregistry/pkg/v1/remote"
+	"github.com/google/go-containerregistry/pkg/v1/types"
+)
+
+// MergeSource allows to specify a remote reference and any overrides that should be included in generated index
+// Override is only applied if Source is an image (not an index), otherwise we cannot know which child image should get the override.
+type MergeSource struct {
+	Source   string
+	Override *v1.Descriptor
+}
+
+// MakeMergeSources is a small helper to generate MergeSource when no modifications are required
+func MakeMergeSources(sources []string) []MergeSource {
+	ms := make([]MergeSource, 0, len(sources))
+	for _, s := range sources {
+		ms = append(ms, MergeSource{Source: s})
+	}
+	return ms
+}
+
+// Merge creates an index from list of paths (tarballs) and
+func Merge(sources []MergeSource, opt ...Option) (*v1.ImageIndex, error) {
+	o := makeOptions(opt...)
+
+	idx, err := appendIndex(mutate.IndexMediaType(empty.Index, types.DockerManifestList), sources, o)
+	if err != nil {
+		return nil, fmt.Errorf("unable to create merged index, err: %v", err)
+	}
+
+	return &idx, nil
+}
+
+func appendIndex(base v1.ImageIndex, sources []MergeSource, o options) (v1.ImageIndex, error) {
+	for _, src := range sources {
+		desc, err := getRemoteDescriptor(src.Source, o)
+		if err != nil {
+			return nil, err
+		}
+		adds, err := indexAddendumFromRemote(desc, src.Override)
+		if err != nil {
+			return nil, err
+		}
+
+		base = mutate.AppendManifests(base, adds...)
+	}
+
+	return base, nil
+}
+
+func indexAddendumFromRemote(desc *remote.Descriptor, override *v1.Descriptor) ([]mutate.IndexAddendum, error) {
+	adds := make([]mutate.IndexAddendum, 0)
+
+	switch desc.MediaType {
+	case types.OCIImageIndex, types.DockerManifestList:
+		idx, err := desc.ImageIndex()
+		if err != nil {
+			return nil, err
+		}
+
+		im, err := idx.IndexManifest()
+		if err != nil {
+			return nil, err
+		}
+
+		for _, imDesc := range im.Manifests {
+			// Currently only Image and ImageIndex are supported by mutate/index
+			if imDesc.MediaType.IsImage() {
+				i, err := idx.Image(imDesc.Digest)
+				if err != nil {
+					return nil, err
+				}
+
+				adds = append(adds, mutate.IndexAddendum{
+					Add: i,
+				})
+			} else if imDesc.MediaType.IsIndex() {
+				ii, err := idx.ImageIndex(imDesc.Digest)
+				if err != nil {
+					return nil, err
+				}
+
+				adds = append(adds, mutate.IndexAddendum{
+					Add: ii,
+				})
+			} else {
+				logs.Warn.Printf("Unsupported media type in Manifest: %v, won't be appended", imDesc.MediaType)
+			}
+		}
+	case types.DockerManifestSchema1, types.DockerManifestSchema1Signed:
+		return nil, fmt.Errorf("merging v1 manifest is not supported")
+	default:
+		img, err := desc.Image()
+		if err != nil {
+			return nil, err
+		}
+		cfg, err := img.ConfigFile()
+		if err != nil {
+			return nil, err
+		}
+
+		desc := v1.Descriptor{
+			URLs:        desc.URLs,
+			MediaType:   desc.MediaType,
+			Annotations: desc.Annotations,
+			Platform: &v1.Platform{
+				Architecture: cfg.Architecture,
+				OS:           cfg.OS,
+				OSVersion:    cfg.OSVersion,
+			},
+		}
+		if override != nil {
+			desc = mergeImageDescriptors(desc, *override)
+		}
+
+		adds = append(adds, mutate.IndexAddendum{
+			Add:        img,
+			Descriptor: desc,
+		})
+	}
+
+	return adds, nil
+}
+
+func getRemoteDescriptor(src string, o options) (*remote.Descriptor, error) {
+	ref, err := name.ParseReference(src, o.name...)
+	if err != nil {
+		return nil, fmt.Errorf("parsing reference for %q: %v", src, err)
+	}
+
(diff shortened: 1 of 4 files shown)


vboulineau:
@googlebot I signed it!

jonjohnsonjr:
> Creating the PR to get feedbacks. If there's interest to get this feature in, I can polish/rework the PR as needed.

Thank you for doing this, it's a good starting point for discussion.

I'm +1 on the need for something like this but not sure about this exact behavior. I have some low-level feedback but let's get to that once we've settled on higher level stuff...

My initial reaction is "we could make `append` support most of this", but maybe it would get confusing if we start producing an index or an image depending on which flags you set (though I think it's doable). Currently, `crane append` understands a list of `-f` tarball layers for appending to an image. If we added e.g. `-i` to represent an image, we could append everything to an index instead of an image and push that.

Do you have a use case for flattening an index [here](https://github.com/google/go-containerregistry/pull/891/files#diff-c1dfdc8c190affb2ce37c129fa6cdb97f526c7f2c61e4c5b739806d72af2da32R73-R99)? If not for that bit, we could just append each SRC descriptor to an index (with the platform stuff, if it's an image) and call it a day.

[This](https://github.com/google/go-containerregistry/pull/891/files#diff-c1dfdc8c190affb2ce37c129fa6cdb97f526c7f2c61e4c5b739806d72af2da32R85) also assumes that all children of an index will always be an image, which isn't necessarily true. Ideally you'd recurse on a child index (we might want to expose something like [this](https://github.com/google/go-containerregistry/blob/6544cb07415a0b92dcc4079fdb13c4540f651e11/pkg/v1/remote/index.go#L175) to make that easier) if you did want to flatten everything down to one index..

If you want to punt on the flattening bits, we could simplify this a lot, but if that's something you think is useful, I'm open to understanding why. It might also make sense to pull that out into a separate thing, depending on why it's needed.

> Currently platform information are always deduced and cannot be taken from user input.

This is reasonable for now, but we should figure out what we would want the interface to look like for overriding this. Going back to the possibility of adding this to `crane append`... I do have concerns about the flags getting messy if we want to support stuff like overriding platforms, but perhaps we could do something generic enough that it's useful for `crane append`ed images as well, e.g. a flag for overriding descriptor values, something like:

```
crane append ... -i foo -i bar \
--override '.manifests[0].platform = {"os":"linux","architecture":"amd64"}' \
--override '.manifests[1].platform = {"os":"linux","architecture":"arm64"}' \
--override '.manifests[1].annotations["org.opencontainers.image.url"] = "example.com/foo"'
```

```
crane append ... -f foo.tar \
--override '.layers[0].URLs = {"os":"linux","architecture":"amd64"}'
```

Now... that would involve implementing a subset of `jq` as a little interpreter thing, which would be fun, but probably way outside the scope of this PR 😄 cc @ImJasonH if we implemented this, it would be very useful for mutating image config as well? Also would be a nice generic tool for mutating indexes without having to merge my crazy interactive editor experiment.

We could also just borrow ideas from https://github.com/estesp/manifest-tool, e.g. `--platforms linux/amd64,linux/arm/v5,linux/arm/v7` matches what we're doing elsewhere pretty well, but I'm not sure what value would represent "derive this from config" vs "don't set any platform", y'know? There's also `os.version` that we'd need to represent somehow... 🤔 

Speaking of which, is there a reason you'd like to add this to crane instead of using manifest-tool? (I'm open to it, just curious.)

jonjohnsonjr:
> though we will need to prevent mixing image and tarballs

I'm not so sure about that. Technically, an index can reference anything, not just images, so it's totally fine to produce an index that references layers (or just arbitrary stuff, really). I guess it's not just "technically", because the buildkit folks do this, so it's definitely a valid use case.

My only concern is that I'm not sure if there's a way to easily control the ordering of different `[]string` flags using cobra, so we'd have to parse that ourselves or just have the caveat that images come before layers in the resulting index's list of descriptors.

> About the platform information, I was leaning toward the manifest-tool syntax, however it lacks support for os.version, though I am not sure this part requires override.

I'd be fine with us being "clever" with a `--platform` flag. Something like:

```
crane append \
  -i foo --platform "linux/amd64" \
  -i bar --platform "linux/arm64/v7" \
  -i baz --platform '{"os":"windows","architecture":"amd64","os.version":"10.0.14393.1066"}
```

We can check to see if `--platform` starts with `"{"`, and just parse it as JSON in that case, otherwise we try to parse the short form. This way you can supply whatever you want for the platform field. WDYT?

vboulineau:
Hello,

I've updated the PR with the implementation in `Append` function.
Some points I'd like to share that could be helpful in the review:
- The `crane.Append` is taking `string` and `[]string` as parameter, similar to `Copy` but unlike some other functions where `v1` types are used in input. I've a mixed feeling about this. On one hand it makes it easier to use Crane as a library, but on the other hand, it prevents testing a bit (as we cannot inject arbitrary structure like `random.Image`). On top of that as `base` can be an `Index` or an `Image`, it pushes significant work to `cmd` (or to users) if `Append` does not do this job.
- Due to the above, I haven't added tests for `Append` and I commented out the filesystem test. When we decide on the point above, I can update the tests.
- The index flattening is required to make proper usage of the feature, nested indexes are not supported by Docker or DockerHub. Index is valid and can be pushed, but we cannot pull the arch referenced in nested index, so it kind of defeats the purpose of the feature.
- I return `Image/Index` objects as after calling `Append` (as a library), one might want to sign these images using Notary, which requires digest and size.

About the inline `platform`, I think it's easier to have everything in `-i` option, i.e. `-i foo,linux/amd64` or `-i bar,{"os":"windows","architecture":"amd64","os.version":"10.0.14393.1066"}`
I think it's easier to add and makes sure we don't depend on ordering, WDYT?

jonjohnsonjr:
I want to respond just to make sure you don't feel left in the dark, but I haven't had much time to think deeply about this PR, and I apologize for that.

> The crane.Append is taking string and []string as parameter, similar to Copy but unlike some other functions where v1 types are used in input. I've a mixed feeling about this. On one hand it makes it easier to use Crane as a library, but on the other hand, it prevents testing a bit (as we cannot inject arbitrary structure like random.Image). On top of that as base can be an Index or an Image, it pushes significant work to cmd (or to users) if Append does not do this job.

This is a good point that I hadn't considered. Most of the stuff in `pkg/crane` was added as needed to simplify `cmd/crane` and lacks a cohesive design. Some things are composable, where other things aren't (and assume we are only interacting with remote registries). This is frustrating...

> Due to the above, I haven't added tests for Append and I commented out the filesystem test. When we decide on the point above, I can update the tests.

Makes sense.

> The index flattening is required to make proper usage of the feature, nested indexes are not supported by Docker or DockerHub. Index is valid and can be pushed, but we cannot pull the arch referenced in nested index, so it kind of defeats the purpose of the feature.

I wonder if it would be worth the effort to start pushing on this? The fact that there is no standardized mechanism for resolving an index to a manifest came up recently in the context of adding zstd compression to layers. I think it would move the whole ecosystem forward if we could agree on how clients should behave (including what to do when confronted with a nested index) and then update clients to match. See https://github.com/opencontainers/image-spec/issues/803#issuecomment-742670465 for ~some context but there's a lot in that thread that is relevant.

I understand that you're probably not interested in blocking on this PR for years until we fix this, so let me think about an alternative for now.

> I return Image/Index objects as after calling Append (as a library), one might want to sign these images using Notary, which requires digest and size.

This feels a bit strange. I think I'd rather see a `v1.Descriptor` returned here, which has all the metadata you need, and the client should have sufficient context to know where it was pushed if it needs that.

If we're going to have to change the signature of `Append` anyway, it might make sense to split this off into a separate function (please don't hate me for this whiplash 😄 I suspect there is still some more back and forth to come). We could have `cmd/append` know how to distinguish between `crane.Append` and `crane.Merge` (or whatever) based on the flags, or just split that off into a separate command too (again, sorry, I know this was your original proposal).

I am not entirely sure how to do this cleanly, so take most of this feedback as brainstorming -- hopefully we can arrive at something that feels elegant...

> About the inline platform, I think it's easier to have everything in -i option, i.e. -i foo,linux/amd64 or -i bar,{"os":"windows","architecture":"amd64","os.version":"10.0.14393.1066"}
I think it's easier to add and makes sure we don't depend on ordering, WDYT?

I do see what you mean, but I'm not a huge fan of the syntax of trying to smash this all into a flag value. I can see why `manifest-tool` takes a config file. Using `#` as a separator for platform is _probably_ fine, but I'm really reluctant to establish any new conventions like this because we'll be on the hook for maintaining that forever. This is one thing I _do_ like about the `--override` flag I proposed earlier. It's order-independent and powerful enough to express ~anything you you might want to do; e.g. if you wanted to add annotations to one of these images, we'd need some other separator and then some string separator, etc etc.

One idea I've had (and haven't quite fleshed out, so please try to poke holes in it) would be to do this via multiple commands instead of just one. We could use an image layout to build up the index you care about (via `layout.Append<whatever>`) and have flags for overriding platform, annotations, etc. Then when you've done everything you want, just push that layout as an index to your target registry.

Of course, this loses some of the benefit of doing it in one shot (e.g. eliding layer downloads/uploads becomes trickier), but we could be able to split up some of these actions into separate commands to make them easier to use.

I am imagining something like:

```
# Append each image referenced by gcr.io/distroless/static:latest
crane layout append -C output/ --index gcr.io/distroless/static:latest --flatten

# Also, append some windows images (this happens to be an index of length one)
crane layout append -C output/ --index mcr.microsoft.com/windows/nanoserver:20H2 --flatten

# If you want to manually manipulate this index, do something silly like:
jq '.manifests[0].annotations["foo"] = "bar"' < output/index.json > output/index.json

# Now we can push it wherever we want
crane push --format=layout output/ registry.example.com/whatever
```

For eliding layers, we could do something stupid like add something equivalent to [`RepoTags`](https://github.com/google/go-containerregistry/blob/859dff29bf3017e5bd7f0bdd29d716913c1131b0/pkg/v1/tarball/image.go#L108) so that it's possible to skip downloading blobs (`--sparse` or something) to the layout, but still know where to fetch them when we need them.

If you're trying to actually get something merged soon, I'm sorry. Doing this kind of thing is sorely needed in a _lot_ of contexts, and I think if we can nail the user experience, `crane` becomes a really, _really_ nice tool for doing some stuff that is otherwise ~impossible right now.

vboulineau:
I'm not in any rush to get the PR merged (thanks to Go modules replace), so PR is only here to propose things and help.

I think the `crane layout` command line is indeed a great idea, and should probably be the generic way to expose it through command line.

It'd also probably be a quite convenient API to play with when vendoring, though _maybe_ there's a space for a bit higher level APIs for simple cases. So I've reworked the PR towards this goal. I moved it back to standalone, remove the command line part (though we could easily bring back `crane merge` - but was not sure if it's wise to add something that will be replaced later shortly after)

Implementation could later be re-written with the Layout API (probably some other high-level APIs could be too), so in the end this PR might not be that valuable, I think it highly depends on the direction for higher-level APIs and the timeframe for layout command.

jonjohnsonjr:
Not sure if you're still interested in this, but I've had several people who want this now, and I think my dreams of making `crane append` more general might never happen. The name `crane amend` came to me and I kind of love it, if we just wanted to make an index version of append.

vboulineau:
Hey @jonjohnsonjr,

Yes very much. We're still using my `crane` fork to support this. I think `amend` could make sense but I am not sure limiting it to index provides a good user experience.

As noted in #1137, a practical use case is to aggregate images (or index) built on multiple OS (it's our use case as well). In this scenario I believe you should not have to know, as a end-user, if the source behind some tag is an image or an index.

Issue #895, opened 2020-12-31
Calculate digest from tarball


aelij:
Our build system produces tarballs and K8s deployment configurations, later to be deployed to multiple environments. We would like to get the image's digest at build time so we could reference it in the configuration, rather than relying on a tag, which is not stable.

Could Crane calculate the digest from a tarball? From what I gather, `crane digest` only works with a remote image.

afbjorklund:
But the digest _is_ from the remote image, so it cannot be calculated from the local image (without being saved with it)

We added some code to save the digest (unlike docker), but that doesn't work if the image has been created locally...

You will find that "RepoDigests" is empty.

It's a shortcoming of the image model...

aelij:
Sorry I must be missing something :) Doesn't the container registry just calculate a hash algorithm on the bytes it receives? After all the digest is identical if I push to multiple CRs. Why can't it be calculated the same way on the client side? Assuming one knows the hash algorithm used by the specific CR one is targeting.

afbjorklund:
I meant that you need to push the image to the registry, in order for docker to calculate the digest.

Also totally misremembered what we had patched, we still don't save the digest in the local tarball.

afbjorklund:
Was referring to https://github.com/moby/moby/issues/32016

aelij:
Can't we grab the code that calculates the digest from the container registry implementation and add it to crane? And compress the layers if necessary.

From that post, looks like Bazel does this, so it should be possible :)

afbjorklund:
The way I have understood it, is that the `Id` is based on the image itself and the `Digest` is based on random files on the server.

Similar to when we download regular files, you have your git commit and git archive and you have your .tar.gz and checksum.

And even though they reference the same files, it's not possible to "guess" the checksum without knowing the server etc*.

\* As described in "pristine-tar", a small delta remains

afbjorklund:
I'm starting to think that the Digest is useless for identifying images, and better to use unique tags instead...

We have some users that want to compare "foo:latest" with "foo:latest", and for those we will use Id (locally)

afbjorklund:
It's a common/constant source of confusion, maybe because they use the same algorithm (sha256) or something.

Like in #627, why it takes longer to look up a value locally than having to download the manifest from the registry.

jonjohnsonjr:
> We would like to get the image's digest at build time so we could reference it in the configuration, rather than relying on a tag, which is not stable.
[...]

PR #896, opened 2021-01-02
Add --tarball to crane digest

jonjohnsonjr:
Fixes https://github.com/google/go-containerregistry/issues/895

### DIFF
--- a/cmd/crane/cmd/digest.go
+++ b/cmd/crane/cmd/digest.go
@@ -24,16 +24,51 @@ import (
 
 // NewCmdDigest creates a new cobra.Command for the digest subcommand.
 func NewCmdDigest(options *[]crane.Option) *cobra.Command {
-	return &cobra.Command{
+	var tarball string
+	cmd := &cobra.Command{
 		Use:   "digest IMAGE",
 		Short: "Get the digest of an image",
-		Args:  cobra.ExactArgs(1),
-		Run: func(_ *cobra.Command, args []string) {
-			digest, err := crane.Digest(args[0], *options...)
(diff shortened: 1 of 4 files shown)


aelij:
Can you remove the required tag argument? Our tarballs have `"RepoTags": null` in their manifest.

aelij:
@jonjohnsonjr Do you want me to finish this? I can create a sub-command as you suggested.

aelij:
Ping :)

jonjohnsonjr:
Ended up just finessing this with some manual flag logic. If we want to do something fancier in the future, we can revisit this.

I should probably provide a big warning about how this doesn't do what most people would expect, but we can do that if it becomes an issue.

cmd/crane/cmd/digest.go
```diff
@@ -24,16 +24,42 @@ import (
...
+
+				img, err := crane.LoadTag(tarball, tag, *options...)
+				if err != nil {
+					log.Fatalf("loading image from %q: %v", tarball, err)
+				}
+				digest, err := img.Digest()
+				if err != nil {
+					log.Fatalf("computing digest: %v", err)
+				}
+				fmt.Println(digest.String())
```
ImJasonH, line 47: nit: return here, and unindent the `else` block. Maybe make both branches helper methods?

ImJasonH (approved):

cmd/crane/cmd/digest.go
jonjohnsonjr: You drive a hard bargain.

ImJasonH (approved):

aelij:
Thanks all

PR #901, opened 2021-01-07
Retry registry access on some server errors.

kramarz:
Sometimes the error is on proxy level, but it still will be beneficial to retry it.


jonjohnsonjr (approved):
Thanks!

PR #902, opened 2021-01-07
upgrade all dependencies

sampointer:
Upgrade all dependencies.

In particular ensure that `k8s.io/api` and `k8s.io/client-go` are upgraded to current versions. K8s doesn't adhere to the "don't break master" dictum. This library is pulled in as a transitive dependency by a large number of projects (in my case: https://github.com/gruntwork-io/terratest) leading an inability in turn to update the dependencies of those projects.


sampointer:
@googlebot I signed it!

jonjohnsonjr:
This is currently a huge thorn in my side. I think we might be able to contain some of the damage by isolating [k8schain](https://github.com/google/go-containerregistry/tree/master/pkg/authn/k8schain) into its own go module, but I am a bit worried about having multiple go modules in one repo, as I'm not sure how well the tooling supports this.

It doesn't seem like terratest uses k8schain anywhere, so this should help a lot: https://github.com/gruntwork-io/terratest/search?q=go-containerregistry

jonjohnsonjr:
I'll try to get this repo in a working state tomorrow (and possibly cut a release), assuming I can get reviewers.

PR #903, opened 2021-01-07
Drop k8s.io/apimachinery/pkg/util/sets

jonjohnsonjr:
Progress towards https://github.com/google/go-containerregistry/pull/902#issuecomment-756365992

### DIFF
--- a/pkg/crane/optimize.go
+++ b/pkg/crane/optimize.go
@@ -27,12 +27,12 @@ import (
 	"github.com/google/go-containerregistry/pkg/v1/remote"
 	"github.com/google/go-containerregistry/pkg/v1/tarball"
 	"github.com/google/go-containerregistry/pkg/v1/types"
-	"k8s.io/apimachinery/pkg/util/sets"
 )
 
 // Optimize optimizes a remote image or index from src to dst.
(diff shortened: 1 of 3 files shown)


pkg/crane/optimize.go
```diff
@@ -27,12 +28,12 @@ import (
...
 	"github.com/google/go-containerregistry/pkg/v1/tarball"
 	"github.com/google/go-containerregistry/pkg/v1/types"
-	"k8s.io/apimachinery/pkg/util/sets"
 )
 
 // Optimize optimizes a remote image or index from src to dst.
 // THIS API IS EXPERIMENTAL AND SUBJECT TO CHANGE WITHOUT WARNING.
-func Optimize(src, dst string, prioritize sets.String, opt ...Option) error {
+func Optimize(src, dst string, prioritize []string, opt ...Option) error {
+	pset := sets.NewString(prioritize...)
```
mattmoor, line 36: if you are removing it from the public interface, then why not make it a private type alias of `map[string]struct{}` and expose just what we need in `set[_test].go`?

jonjohnsonjr: Done, roughly.

mattmoor (approved):

PR #904, opened 2021-01-07
Put k8schain in its own go module

jonjohnsonjr:
This should make go-containerregistry much easier to use.

https://github.com/google/go-containerregistry/pull/902
https://github.com/google/ko/issues/287
https://github.com/google/ko/issues/258#issuecomment-756243939

WANT_LGTM=all


ImJasonH (approved):

jonjohnsonjr:
Per https://pkg.go.dev/github.com/google/go-containerregistry/pkg/authn/k8schain?tab=importedby y'all might be affected.

cc @andreyvelich 
cc @scothis
cc @matthewmcnew 
cc @hasheddan 
cc @davidewatson (also, https://github.com/davidewatson/keychain looks interesting, would love to see a README 😉)
cc @leogr

jonjohnsonjr:
Split this up into two commits:

1. go.mod changes, should be easy to review
2. go mod tidy && go mod vendor

I didn't vendor the k8schain deps because I think it might break some expectations of various tools (only one top-level vendor/ directory). That's why there's the huge negative diff. Hopefully that's fine.

mattmoor:
Did you add any presubmit safeguard that the nested module files are correct?

jonjohnsonjr:
cc @tejal29 as well

jonjohnsonjr:
> Did you add any presubmit safeguard that the nested module files are correct?

Done: https://github.com/google/go-containerregistry/pull/904/commits/52dc010d199eacf52ea60176701af16cc180bc33

hasheddan (approved):
Thanks @jonjohnsonjr! I'll test this out with our @crossplane usage :+1:

andreyvelich:
Thank you for the notice @jonjohnsonjr!
cc @gaocegege @johnugeorge

pkg/authn/k8schain/go.mod
```diff
@@ -0,0 +1,23 @@
+module github.com/google/go-containerregistry/pkg/authn/k8schain
+
+go 1.15
```
mattmoor, line 3: the parent is 1.14, these should be consistent.

jonjohnsonjr: I think it would be fine for them to diverge in the future, but I've dropped this to 1.15.

Issue #892, opened 2020-12-30
match.Platform but ignore variant


deitch:
In order to match on a platform, we have [match.Platforms](https://pkg.go.dev/github.com/google/go-containerregistry/pkg/v1/match#Platforms). Most of the time this works well.

However, the `variant` part of the platform, described in [the spec](https://github.com/opencontainers/image-spec/blob/79b036d80240ae530a8de15e1d21c7ab9292c693/image-index.md#image-index-property-descriptions) is listed as "OPTIONAL".

The current `match.Platforms` has no way of indicating, "match my OS/Architecture but ignore the variant". For that matter, I think it may be a broader question of a way to tell the platform-matcher to ignore some fields. E.g. I want all entries that have architecture=amd64, I don't care about the OS, etc.

I am happy to open a PR on this, if we can agree on the right interface.

## Options

Some options:

### match.Any

Some form of `match.Any` option, such that you could provide it to any given field. So you would do:

```go
match.Platforms(v1.Platform{Architecture: "amd64", OS: match.Any, Variant: match.Any})
```

I am not sure how to make that work; some fields in `v1.Platform` are `string`, others are `[]string`. Maybe it just works for the `string` fields?

### Wrapper

A different variant might be a wrapper, where we could do:

```go
match.IgnoreBlank(match.Platforms(platforms ...v1.Platform))
```

The above tells it to reject fields that do not match, but ignore blank. Again, not sure how to make this work. Don't we need to tell `match.Platforms` to ignore blanks?

### Alternate 

Another option is to have an alternative matcher:

```go
match.PlatformsWithIgnoreBlank(platforms ...v1.Platform)
```

I don't like the nomenclature, but the idea isn't bad.

I looked at a parameter, but the platforms arg already is variadic.

Thoughts?

jonjohnsonjr:
> However, the variant part of the platform, described in the spec is listed as "OPTIONAL".

So this is where things get a little fuzzy. The spec also says:

> This OPTIONAL property describes the minimum runtime requirements of the image.

Simiarly, for `os.features`:

> This OPTIONAL property specifies an array of strings, each specifying a mandatory OS feature.

So you want to ignore this stuff, but these are part of the minimum runtime requirements or mandatory OS features. Depending on what you're asking, you absolutely should or should not ignore this field...

When a _runtime_ encounters a set of platform-specific images, it needs to ensure that it can actually run that thing, so it definitely shouldn't ignore the variant or os.features.

[...]

PR #907, opened 2021-01-08
Fix crane digest for v1 and platform option

### LINKED ISSUE #906: crane digest does not work with specific platform on v1 schema
crane digest does not work with the platform option on v1 schema images. It returns the following error:
```
unsupported MediaType: "application/vnd.docker.distribution.manifest.v1+prettyjws", see https://github.com/google/go-containerregistry/issues/377
```

Code to reproduce:
```
crane.Digest("nginx:1.10.0", crane.WithPlatform(&v1.Platform{OS: "linux", Architecture: "amd64"}))
```

Context:
I use crane to sync images from external registries to a local registry. For that I call `crane.Digest(src)` and `crane.Digest(dst)` and if the values differ I do a `crane.Copy` and after the copy trigger some pipelines. I used to use `Copy` and `Digest` without the platform option which worked for v1 schema but this way it did not work for indexes which contain images with non-distributable layers.

dvob:
This allows to use crane.Digest on images with a v1 schema.

Fixes #906

### DIFF
--- a/pkg/crane/digest.go
+++ b/pkg/crane/digest.go
@@ -25,6 +25,8 @@ func Digest(ref string, opt ...Option) (string, error) {
 		if !desc.MediaType.IsIndex() {
 			return desc.Digest.String(), nil
 		}
+
+		// TODO: does not work for indexes which contain schema v1 manifests
 		img, err := desc.Image()
 		if err != nil {
 			return "", err
--- a/pkg/crane/digest.go
+++ b/pkg/crane/digest.go
@@ -19,6 +19,9 @@ func Digest(ref string, opt ...Option) (string, error) {
 	o := makeOptions(opt...)
 	if o.platform != nil {
 		desc, err := getManifest(ref, opt...)
+		if !desc.MediaType.IsIndex() {
+			return desc.Digest.String(), nil
+		}
 		if err != nil {
 			return "", err
 		}
--- a/pkg/crane/digest.go
+++ b/pkg/crane/digest.go
@@ -19,11 +19,11 @@ func Digest(ref string, opt ...Option) (string, error) {
 	o := makeOptions(opt...)
 	if o.platform != nil {
 		desc, err := getManifest(ref, opt...)
-		if !desc.MediaType.IsIndex() {
-			return desc.Digest.String(), nil
-		}
 		if err != nil {
 			return "", err
+		}
+		if !desc.MediaType.IsIndex() {
+			return desc.Digest.String(), nil
 		}
 		img, err := desc.Image()
 		if err != nil {


dvob:
@googlebot I signed it!

pkg/crane/digest.go
```diff
@@ -19,6 +19,9 @@ func Digest(ref string, opt ...Option) (string, error) {
 	o := makeOptions(opt...)
 	if o.platform != nil {
 		desc, err := getManifest(ref, opt...)
+		if !desc.MediaType.IsIndex() {
```
ImJasonH, line 22: It looks like desc will be `nil` if there are errors fetching the manifest (e.g., HTTP 404, 403), which will result in a nil panic at this line. Should this check happen after the err check?

ImJasonH (dismissed):

pkg/crane/digest.go
dvob: Yes this should be after the err check. I moved it.

dvob:
How often do you release new versions (tags)?

ImJasonH:
> How often do you release new versions (tags)?

We tend to release as-needed whenever there are significant changes to make available. 

cc @jonjohnsonjr to speak on whether/how we want `crane` to support v1 manifests. This change by itself seems innocuous to me at least 🤷

jonjohnsonjr:
> How often do you release new versions (tags)?

Yeah, roughly as needed. Usually when I need to re-import this package into google or some large downstream dependency requests a fix be included in a release. We could have more discipline about this (cc @hasheddan).

> This change by itself seems innocuous to me at least 🤷

This doesn't cover every base, but it's definitely an improvement, so I'm fine with it.

@dvob consider situations such as this: https://github.com/docker-library/golang/issues/269#issuecomment-729984068

```
$ crane manifest nginx:1.19.0-alpine | jq .
2021/01/08 09:18:20 No matching credentials were found for "index.docker.io/library/nginx", falling back on anonymous
{
  "manifests": [
    {
      "digest": "sha256:ee5a9b68e8d4a4b8b48318ff08ad5489bd1ce52b357bf48c511968a302bc347b",
      "mediaType": "application/vnd.docker.distribution.manifest.v2+json",
      "platform": {
        "architecture": "amd64",
        "os": "linux"
      },
      "size": 1360
    },
    {
      "digest": "sha256:88fd834bc089b5dff1f27ab0315c3b0bbdfddfca3b2c1aeb9798944d179e441b",
      "mediaType": "application/vnd.docker.distribution.manifest.v2+json",
      "platform": {
        "architecture": "arm",
        "os": "linux",
        "variant": "v6"
      },
      "size": 1360
    },
    {
      "digest": "sha256:bf8ebb8068d7ff10711c1670398eeaa83283a7987cdc9b9b014560dbba71773c",
      "mediaType": "application/vnd.docker.distribution.manifest.v2+json",
      "platform": {
        "architecture": "arm",
        "os": "linux",
        "variant": "v7"
      },
      "size": 1360
    },
    {
      "digest": "sha256:eff196a3849ad6541fd3afe676113896be214753740e567575bb562986bd2cd4",
      "mediaType": "application/vnd.docker.distribution.manifest.v1+json",
      "platform": {
        "architecture": "arm64",
        "os": "linux",
        "variant": "v8"
      },
      "size": 13297
    },
    {
      "digest": "sha256:2fa12030ffb0224e0b2e17bc4b1f1479e191e2fd65869fc60d09a8efa5b6d879",
      "mediaType": "application/vnd.docker.distribution.manifest.v2+json",
      "platform": {
        "architecture": "386",
        "os": "linux"
      },
      "size": 1360
    },
    {
      "digest": "sha256:9012a1962d3201492ea815d0b395af623cf78e4b2187f9069d5cee0c0d1345c1",
      "mediaType": "application/vnd.docker.distribution.manifest.v2+json",
      "platform": {
        "architecture": "ppc64le",
        "os": "linux"
      },
      "size": 1360
    },
    {
      "digest": "sha256:d7b9368659ff1be2bbb4267c9d45e7aa201cb282ac9bc86e09bf390087df68b6",
      "mediaType": "application/vnd.docker.distribution.manifest.v2+json",
      "platform": {
        "architecture": "s390x",
        "os": "linux"
      },
      "size": 1360
    }
  ],
  "mediaType": "application/vnd.docker.distribution.manifest.list.v2+json",
  "schemaVersion": 2
}
```

The linux/arm64/v8 image is a schema 1 image, so `desc.Image()` will fail, even though we could just pluck the digest from that descriptor. I think this is such a rare case that it's probably not worth fixing (and is currently broken anyway), but possibly add a TODO?

dvob:
Or do you see an easy way to fix it for these cases as well?

jonjohnsonjr:
> Or do you see an easy way to fix it for these cases as well?

We'd need to move this logic somewhere else:

https://github.com/google/go-containerregistry/blob/f97c411e4aa61a02a2fefd761529d14106687a83/pkg/v1/remote/index.go#L193-L237

(probably as a `match.Matcher`)

And use that inside `crane.Digest` to match the right platform. Like here:

https://github.com/google/go-containerregistry/blob/f97c411e4aa61a02a2fefd761529d14106687a83/pkg/v1/remote/index.go#L143-L157

The TODO is fine for now, I think. There is some discussion around platform matching in https://github.com/google/go-containerregistry/issues/892 that we can tack this work onto.

jonjohnsonjr (approved):

Issue #908, opened 2021-01-08
Provide detailed instructions on contributing


naveensrinivasan:
The contributing.md needs to have more detailed instructions on how to contribute and run tests.

The document just mentions to run 
```
./hack/presubmit.sh
```
And running that produces failed tests.

```
--- FAIL: TestGcloudErrors (0.00s)
    --- FAIL: TestGcloudErrors/error (0.00s)
        auth_test.go:113: wanted error, got nil
    --- FAIL: TestGcloudErrors/badoutput (0.00s)
        auth_test.go:113: wanted error, got nil
    --- FAIL: TestGcloudErrors/badexpiry (0.00s)
        auth_test.go:113: wanted error, got nil
--- FAIL: TestGcloudSuccess (0.00s)
    auth_test.go:139: wanted token "mytoken", got ""
--- FAIL: TestKeychainGCRandAR (0.00s)
    --- FAIL: TestKeychainGCRandAR/cases[0] (0.00s)
        auth_test.go:205: expected not anonymous auth for {gcr.io true}, got: &{}
        auth_test.go:216: expected not anonymous auth for {gcr.io true}, got: &{}
    --- FAIL: TestKeychainGCRandAR/cases[1] (0.00s)
        auth_test.go:205: expected not anonymous auth for {us.gcr.io true}, got: &{}
        auth_test.go:216: expected not anonymous auth for {us.gcr.io true}, got: &{}
    --- FAIL: TestKeychainGCRandAR/cases[2] (0.00s)
        auth_test.go:205: expected not anonymous auth for {eu.gcr.io true}, got: &{}
        auth_test.go:216: expected not anonymous auth for {eu.gcr.io true}, got: &{}
    --- FAIL: TestKeychainGCRandAR/cases[3] (0.00s)
        auth_test.go:205: expected not anonymous auth for {asia.gcr.io true}, got: &{}
        auth_test.go:216: expected not anonymous auth for {asia.gcr.io true}, got: &{}
    --- FAIL: TestKeychainGCRandAR/cases[4] (0.00s)
        auth_test.go:205: expected not anonymous auth for {staging-k8s.gcr.io true}, got: &{}
        auth_test.go:216: expected not anonymous auth for {staging-k8s.gcr.io true}, got: &{}
    --- FAIL: TestKeychainGCRandAR/cases[5] (0.00s)
        auth_test.go:205: expected not anonymous auth for {global.gcr.io true}, got: &{}
[...]

PR #909, opened 2021-01-08
Temporarily disable gcloud tests for darwin

jonjohnsonjr:
Workaround for https://github.com/google/go-containerregistry/issues/908

### DIFF
--- a/pkg/v1/google/auth_test.go
+++ b/pkg/v1/google/auth_test.go
@@ -1,4 +1,5 @@
 // +build !arm64
+// +build !darwin
 // Copyright 2018 Google LLC All Rights Reserved.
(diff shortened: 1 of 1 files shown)


naveensrinivasan:
I was able to test it and it fixed the issue. Thanks

naveensrinivasan:
Probably It would be good to include tests that target multiple OS. I can see if I can do that PR :)

dlorenc (approved):

jonjohnsonjr:
> Probably It would be good to include tests that target multiple OS. I can see if I can do that PR :)

I can't reproduce it myself, so I would really appreciate it. Thanks!

PR #910, opened 2021-01-08
Create codeql-analysis.yml

naveensrinivasan:
Ran Codeql on the fork and found a couple of issues.  Enabling to make sure those are being addressed.


jonjohnsonjr:
I think we already have this in https://github.com/google/go-containerregistry/blob/94913c02dcaa92b09b913585c1d26ba2eba1d946/.github/workflows/knative-security.yaml#L47

naveensrinivasan:
OK. Thanks

PR #911, opened 2021-01-08
feature - Include tests run for macos

naveensrinivasan:
Enabled test to run on macos.

### DIFF
--- a/.github/workflows/knative-go-test.yaml
+++ b/.github/workflows/knative-go-test.yaml
@@ -32,7 +32,7 @@ jobs:
     strategy:
       matrix:
         go-version: [1.15.x]
-        platform: [ubuntu-latest]
+        platform: [ubuntu-latest, macos-latest]
 
     runs-on: ${{ matrix.platform }}
 


jonjohnsonjr:
These get pulled in automatically from https://github.com/knative/.github/blob/master/workflow-templates/knative-go-test.yaml via https://github.com/mattmoor/knobots-actions/blob/588f82af0751284d46999c8cad7b55152305c732/repos.yaml#L151-L156 e.g. https://github.com/google/go-containerregistry/pull/865

I'm not sure how I feel about it, or how we can diverge... cc @mattmoor @ImJasonH

jonjohnsonjr:
The macos-latest tests failed:
```
--- FAIL: TestTimeoutContext (0.11s)
    retry_test.go:129: deadline was not recognized by transport
```

I wonder if this isn't a long enough hang (if the machine is too slow due to other tests or compilation burning all the CPU, perhaps?): https://github.com/google/go-containerregistry/blob/96d4226fd9db72907efe274af4801fcf1cea2054/pkg/v1/remote/transport/retry_test.go#L104

naveensrinivasan:
TBH having all these knative actions are confusing without any documentation as to why are these knative stuff here.

jonjohnsonjr:
I agree :)

naveensrinivasan:
> The macos-latest tests failed:
> 
> ```
> --- FAIL: TestTimeoutContext (0.11s)
>     retry_test.go:129: deadline was not recognized by transport
> ```
> 
> I wonder if this isn't a long enough hang (if the machine is too slow due to other tests or compilation burning all the CPU, perhaps?):
> 
> https://github.com/google/go-containerregistry/blob/96d4226fd9db72907efe274af4801fcf1cea2054/pkg/v1/remote/transport/retry_test.go#L104

Why would test's sleep? I think it should be refactored.

mattmoor:
> I'm not sure how I feel about it, or how we can diverge

As written this PR will be overwritten when the sync actions resume next week.  To disable all the knobots, whack [this](https://github.com/knative-sandbox/knobots/blob/045482812bcba785bdab809b4359b7c9be74a261/repos.yaml#L182-L186).  To disable just the action sync add this repo [here](https://github.com/knative-sandbox/knobots/blob/master/actions-exclude.yaml).

Diverging from Knative is fine, but it should be a hard fork and you'd need to keep your own actions up to date.  Alternatively you can tweak things upstream to make the sync job more dynamic, but today your choices are fairly binary.

jonjohnsonjr:
> Why would test's sleep? I think it should be refactored.

Makes sense -- I think the author just wanted to test that the transport respected cancellation. I'm sure there is a way to test that without sleeping.

jonjohnsonjr:
Does travis have macos support? https://github.com/google/go-containerregistry/blob/master/.travis.yml#L21

naveensrinivasan:
> Does travis have macos support? https://github.com/google/go-containerregistry/blob/master/.travis.yml#L21

I think fixing the root cause of the issue would be is to fix the tests and not looking a workaround using Travis.

imjasonh:
Now that we're not using Knative actions anymore this should be relatively straightforward if you're still interested.

naveensrinivasan:
> Now that we're not using Knative actions anymore this should be relatively straightforward if you're still interested.

Sounds good. I will rebase this against the latest main. Thanks!

PR #913, opened 2021-01-15
doc: add context.Context to examples

bonifaido:
The signature has changed since the original README was created.


jonjohnsonjr (approved):
Thank you!

PR #915, opened 2021-01-18
implement delete manifest tests

cpanato:
I was learning this repo to try to use in one personal thing and saw that some tests were pending to add.

This PR tries to implement the tests for `crane.Delete`, I'm not 100% if it is correct but it is a tryout 😄 

if sounds reasonable I will try to implement the other missing tests

/cc @jonjohnsonjr

### DIFF
--- a/pkg/crane/crane_test.go
+++ b/pkg/crane/crane_test.go
@@ -34,7 +34,6 @@ import (
 	"github.com/google/go-containerregistry/pkg/v1/remote"
 )
 
-// TODO(jonjohnsonjr): Test crane.Delete behavior.
 // TODO(jonjohnsonjr): Test crane.ListTags behavior.
 // TODO(jonjohnsonjr): Test crane.Catalog behavior.
 // TODO(jonjohnsonjr): Test crane.Copy failures.
@@ -146,6 +145,30 @@ func TestCraneRegistry(t *testing.T) {
 	}
 
 	if err := compare.Layers(pulledLayer, layer); err != nil {
+		t.Fatal(err)
+	}
+
+	// Delete the non existing image
+	if err := crane.Delete("honk-image"); err == nil {
+		t.Fatal("wanted err, got nil")
+	}
+
+	// Delete the image
+	if err := crane.Delete(src); err != nil {
+		t.Fatal(err)
+	}
+
+	// check if the image was really deleted
+	if _, err := crane.Pull(src); err == nil {
+		t.Fatal("wanted err, got nil")
+	}
+
+	// check if the copied image still exist
+	dstPulled, err := crane.Pull(dst)
+	if err != nil {
+		t.Fatal(err)
+	}
+	if err := compare.Images(dstPulled, copied); err != nil {
 		t.Fatal(err)
 	}
 }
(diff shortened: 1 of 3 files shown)


pkg/crane/crane_test.go
```diff
@@ -148,6 +147,33 @@ func TestCraneRegistry(t *testing.T) {
 	if err := compare.Layers(pulledLayer, layer); err != nil {
 		t.Fatal(err)
 	}
+
+	// Delete the non existing image
+	err = crane.Delete("honk-image")
```
ImJasonH, line 152: style nit: If a method only returns an error, you can scope that error to an if block that checks that error:

```go
if err := crane.Delete("honk-image"); err == nil {
  t.Fatal(...)
}
```

This has the benefit of not polluting outer scope, and can help readability since the reader can know the call and check won't have any effect on the rest of the code. It's also consistent with other code in the file, like `compare.Layers` above on line 147.

ImJasonH (commented):
lgtm in general, just a small style/consistency nit.

pkg/crane/crane_test.go
cpanato: thanks for the review

pkg/crane/crane_test.go
```diff
@@ -148,6 +147,31 @@ func TestCraneRegistry(t *testing.T) {
...
+		t.Fatal("wanted err, got nil")
+	}
+
+	// Delete the image
+	if err := crane.Delete(src); err != nil {
+		t.Fatal(err)
+	}
+
+	// check if the image was really deleted
+	_, err = crane.Pull(src)
```
ImJasonH, line 162: This one can actually also be:

```go
if _, err := crane.Pull(src); err == nil {
  ...
}
```

cpanato: yes, true

ImJasonH (approved):

pkg/registry/manifest.go
```diff
@@ -161,6 +161,32 @@ func (m *manifests) handle(resp http.ResponseWriter, req *http.Request) *regErro
...
+		if !ok {
+			return &regError{
+				Status:  http.StatusNotFound,
+				Code:    "MANIFEST_UNKNOWN",
+				Message: "Unknown manifest",
+			}
+		}
+
+		delete(m.manifests[repo], target)
+		resp.WriteHeader(http.StatusOK)
```
jonjohnsonjr, line 186: This should be a 202, [per the spec](https://github.com/opencontainers/distribution-spec/blob/master/spec.md#deleting-tags), not 200.

Can you also add a test here? https://github.com/google/go-containerregistry/blob/master/pkg/registry/registry_test.go

jonjohnsonjr (changes requested):
Couple nits, otherwise lgtm.

jonjohnsonjr:
> if sounds reasonable I will try to implement the other missing tests

That would be wonderful, if you have the time! Thanks :)

pkg/registry/manifest.go
cpanato: thanks for your review @jonjohnsonjr 
PTAL

jonjohnsonjr (approved):
Thanks!

PR #918, opened 2021-01-19
implement listTags tests

cpanato:
as a continuation of in intention of the work here: https://github.com/google/go-containerregistry/pull/915

This PR implements the `ListTags` tests

All feedback are welcome

/assign @jonjohnsonjr

### DIFF
--- a/pkg/registry/manifest.go
+++ b/pkg/registry/manifest.go
@@ -18,15 +18,23 @@ import (
 	"bytes"
 	"crypto/sha256"
 	"encoding/hex"
+	"encoding/json"
 	"fmt"
 	"io"
 	"net/http"
+	"sort"
+	"strconv"
 	"strings"
 	"sync"
 
 	v1 "github.com/google/go-containerregistry/pkg/v1"
 	"github.com/google/go-containerregistry/pkg/v1/types"
 )
+
+type listTags struct {
+	Name string   `json:"name"`
+	Tags []string `json:"tags"`
+}
 
 type manifest struct {
 	contentType string
@@ -48,6 +56,15 @@ func isManifest(req *http.Request) bool {
 	return elems[len(elems)-2] == "manifests"
 }
 
+func isTags(req *http.Request) bool {
+	elems := strings.Split(req.URL.Path, "/")
+	elems = elems[1:]
+	if len(elems) < 4 {
+		return false
+	}
+	return elems[len(elems)-2] == "tags"
+}
+
 // https://github.com/opencontainers/distribution-spec/blob/master/spec.md#pulling-an-image-manifest
 // https://github.com/opencontainers/distribution-spec/blob/master/spec.md#pushing-an-image
 func (m *manifests) handle(resp http.ResponseWriter, req *http.Request) *regError {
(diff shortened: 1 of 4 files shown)


pkg/registry/manifest.go
```diff
@@ -193,3 +211,59 @@ func (m *manifests) handle(resp http.ResponseWriter, req *http.Request) *regErro
...
+				Code:    "NAME_UNKNOWN",
+				Message: "Unknown name",
+			}
+		}
+
+		var tags []string
+		countTags := 0
+		for tag := range c {
+			if countTags >= n {
+				break
```
jonjohnsonjr, line 243: I'd be satisfied with just a TODO here that we need to include the `Link` header for the next page, but you could also just implement it, if you'd like!

https://github.com/opencontainers/distribution-spec/blob/b505e9cc53ec499edbd9c1be32298388921bb705/detail.md#tags-paginated

pkg/registry/manifest.go
```diff
@@ -48,6 +56,15 @@ func isManifest(req *http.Request) bool {
 	return elems[len(elems)-2] == "manifests"
 }
 
+func isTags(req *http.Request) bool {
+	elems := strings.Split(req.URL.Path, "/")
+	elems = elems[1:]
+	if len(elems) < 4 {
+		return false
+	}
+	return elems[len(elems)-2] == "tags"
```
jonjohnsonjr, line 65: This is _probably_ fine, but checking that it ends with `/tags/list` is probably safer.

pkg/registry/manifest.go
cpanato: @jonjohnsonjr thanks for the feedback, i will implement that

cpanato: @jonjohnsonjr currently the crane does not support pagination, this is something we would like to have as input flags in the command?

cpanato: I tried to run the list tags against docker API and I think I'm doing something wrong because when I set the `n` to be something like 5 or 100 it returns the same as not setting the `n`

```
TOKEN=$(curl -s "https://auth.docker.io/token?scope=repository%3Alibrary%2Fgolang%3Apull&service=registry.docker.io" | jq -r .token)
curl -v -s -H "Authorization: Bearer ${TOKEN}" "https://index.docker.io/v2/library/golang/tags/list?n=5"
```

am I doing something very wrong?

cpanato: tried in gcr as well same results :/

jonjohnsonjr: > am I doing something very wrong?

Nope! It's just that neither docker hub nor GCR implement this (for different reasons).

Looking for a random public image on the internet, it seems that quay.io does implement this:

```
$ curl -v https://quay.io/v2/kubernetes_incubator/nfs-provisioner/tags/list?n=5
```

cpanato: @jonjohnsonjr I added a TODO, because I'm afraid I cannot implement this due to lack of time and would not have this PR hanging around for a long time.

When I have more time I will comeback if this was not implemented by another person.

PTAL

jonjohnsonjr: Sounds good, thanks!

jonjohnsonjr (approved):

Issue #920, opened 2021-01-22
Pushing large images to Dockerhub 504


dwillist:
I'm consistently getting 504 errors when pushing large layers to Dockerhub using this library. It seems like this has to do with the interplay between the following:
- `gzip` writers do compression on demand.
- http requests send whatever data is available after a read.
- `io.Pipe`created [here](https://github.com/google/go-containerregistry/blob/859dff29bf3017e5bd7f0bdd29d716913c1131b0/pkg/v1/internal/gzip/zip.go#L50) to read the compressed layer and upload it [here](https://github.com/google/go-containerregistry/blob/859dff29bf3017e5bd7f0bdd29d716913c1131b0/pkg/v1/remote/write.go#L365) matches single reads to single writes.

As a result we send a bunch of small chunked pieces of a `PATCH` request which Dockerhub does not like. 

Attached is a program I can used to consistently reproduce. by running `go run . <dockerhub-image-tag> <highly-compressible-file>`. If I bump up the `BufferSize` enough I can get these uploads to consistently pass. 

[ggcr-bug.tar.gz](https://github.com/google/go-containerregistry/files/5856307/ggcr-bug.tar.gz)

I am using the `enwiki8` or `enwiki9` as my `highly compressible file` both are quite large and can be downloaded from: http://mattmahoney.net/dc/textdata

jonjohnsonjr:
> Attached is a program I can used to consistently reproduce.

Thank you for this. It saves me a lot of time. I'm taking a few days vacation, but let me see if I can reproduce this next week. In the meantime, enabling verbose logging _might_ be helpful: https://github.com/google/go-containerregistry/blob/859dff29bf3017e5bd7f0bdd29d716913c1131b0/cmd/crane/cmd/root.go#L49

Though, I recently started omitting binary bodies, so that might not help :)

jonjohnsonjr:
This is fascinating. I wasn't able to reproduce at first (internet is too fast, fortunately), but adding a `time.Sleep(100 * time.Millisecond)` in the `Read` was sufficient to trigger this.

Docker Hub will reply with a 504 after 30 seconds, every time. I'll keep poking at it, but I'm guessing there's some threshold where they close the connection after 30 seconds if they haven't received enough data (to prevent something like a [Slowloris attack](https://en.wikipedia.org/wiki/Slowloris_(computer_security))) -- we can fix this by doing some buffering client side (which will probably speed up everything, a bit), but I'd like to understand more about this before just fixing it.

[...]

PR #923, opened 2021-01-28
Buffer the output of gzip.Writer to avoid stalling

jonjohnsonjr:
Use a bufio.Writer to buffer gzipped output while we are reading from
the other end of an io.Pipe to allow gzip to keep compressing its input.

A 64K buffer was chosen for its humor value. The default size of
bufio.Writer was too small when testing against a local registry.
Increasing beyond 64K didn't seem to have any noticeable effect. It
might make sense to make this smaller, but I don't see a reason to worry
about it ATM.

Fixes https://github.com/google/go-containerregistry/issues/920

In local testing, this improved write speed from ~8MB/s to ~30MB/s. Against an actual registry (Docker Hub or GCR), this seems to improve things by ~30%.

### DIFF
--- a/pkg/v1/internal/gzip/zip.go
+++ b/pkg/v1/internal/gzip/zip.go
@@ -15,6 +15,7 @@
 package gzip
 
 import (
+	"bufio"
 	"bytes"
 	"compress/gzip"
 	"io"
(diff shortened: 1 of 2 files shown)


jonjohnsonjr:
Thinking about it a little more, 64K is probably pretty good, actually. The `Read()`s seem to be happening in ~32K chunks, so having that plus some additional buffer should remove gzip as the bottleneck for a while.

ImJasonH (approved):
💯

jonjohnsonjr:
We might want to look at using a `sync.Pool` if we find that this causes the garbage collector to churn too much: https://github.com/containerd/containerd/blob/4d7d63f390e73f62676cd4994233805a6f821ab7/archive/compression/compression.go#L58

I was poking around containerd and docker to see if they do anything like this, in the hopes of landing a fix against them too, and it seems they do even better :P

PR #925, opened 2021-01-28
Update terms

jonjohnsonjr:
master -> main (where meaningful)

whitelist -> allowlist

exotic -> strange


ImJasonH (approved):
💯 

Thanks for doing this!

PR #928, opened 2021-02-02
Update docker version to v20.10.0-beta

### LINKED ISSUE #916: Docker version upgrade?
Is there any plan to upgrade the version of docker used by this project? It's currently at `github.com/docker/docker v1.4.2-0.20190924003213-a8608b5b67c7`, which is a bit old.

For context, our project imports GGCR and tries to stay on the same version of docker to avoid dependency pain (https://github.com/buildpacks/imgutil/issues/70) however this is now causing us new pain (https://github.com/buildpacks/lifecycle/issues/504).

natalieparellano:
Fixes #916 

Signed-off-by: Natalie Arellano <narellano@vmware.com>


vendor/github.com/containerd/containerd/platforms/cpuinfo.go
```diff
@@ -0,0 +1,117 @@
...
+
+	"github.com/containerd/containerd/errdefs"
+	"github.com/containerd/containerd/log"
+	"github.com/pkg/errors"
+)
+
+// Present the ARM instruction set architecture, eg: v7, v8
+var cpuVariant string
+
+func init() {
```
ImJasonH, line 33: @jonjohnsonjr This was the init-time lookup we were trying to avoid pulling in.

ImJasonH: The fix in containerd is included in v1.5.0-beta.0: https://github.com/containerd/containerd/blob/v1.5.0-beta.0/platforms/cpuinfo.go

If we get moby/moby to update to that release of containerd/containerd we can pull in that version of moby/moby: https://github.com/moby/moby/blob/8d3179546e79065adefa67cc697c09d0ab137d30/vendor.conf#L133

jonjohnsonjr: So something else is pulling in containerd than what we expected (this was supposed to be before the problematic commit that your fix helps with, IIUC).

natalieparellano: Apologies for the loooong delay in following up on this PR. I think we were mistaken that `v20.10.0-beta1` doesn't have the problematic commit, because it appears in the list of tags for https://github.com/moby/moby/commit/7a9cb29fb980c0ab3928272cdc24c7089b2fcf64. I did manage to verify that `v19.03.15` will "work" in the sense that it doesn't contain `vendor/github.com/containerd/containerd/platforms/cpuinfo.go` and also plays nicely with our repositories. I don't really understand the meaning of these versions (`v19.03.15` seems to be equivalent to `v17.12.0-ce-rc1`? At least go.mod updates to that version...) so I am not sure if updating to this version makes sense.

imjasonh:
Superceded by #1036

PR #929, opened 2021-02-02
Don't consider go.mod as generated, because it's not

ImJasonH:
[I recommended this in a code review in September 2019](https://github.com/google/go-containerregistry/pull/462#discussion_r328430853). What a fool I was then.


jonjohnsonjr (approved):

jonjohnsonjr:
> I recommended this in a code review in September 2019. What a fool I was then.

Never to be trusted again!

PR #930, opened 2021-02-04
Support non-distributable layers in Write and MultiWrite

StevenLocke:
Implements #741

### DIFF
--- a/pkg/v1/remote/options.go
+++ b/pkg/v1/remote/options.go
@@ -180,9 +180,7 @@ func WithUserAgent(ua string) Option {
 // https://github.com/opencontainers/image-spec/blob/master/layer.md#non-distributable-layers
 //
 // The default behaviour is to skip these layers
-func WithNondistributable() Option {
-	return func(opt *options) error {
-		opt.allowNondistributableArtifacts = true
-		return nil
-	}
+func WithNondistributable(o *options) error {
+	o.allowNondistributableArtifacts = true
+	return nil
 }
--- a/pkg/v1/remote/options.go
+++ b/pkg/v1/remote/options.go
@@ -175,11 +175,12 @@ func WithUserAgent(ua string) Option {
 	}
 }
 
-// WithIncludeNondistributable is a functional option for allowing
-// non-distributable (foreign) layers
+// WithNondistributable includes non-distributable (foreign) layers
+// when writing images, see:
+// https://github.com/opencontainers/image-spec/blob/master/layer.md#non-distributable-layers
 //
-// The default behaviour is not to include them.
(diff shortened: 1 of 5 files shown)


pkg/v1/remote/options.go
```diff
@@ -173,3 +174,14 @@ func WithUserAgent(ua string) Option {
 		return nil
 	}
 }
+
+// WithIncludeNondistributable is a functional option for allowing
+// non-distributable (foreign) layers
+//
+// The default behaviour is not to include them.
+func WithIncludeNondistributable() Option {
```
jonjohnsonjr, line 182: This is a mouthful, but I don't have a better suggestion. Maybe `WithNondistributable` is good enough?

This could also take a `bool` if we want `WithNondistributable(false)` to work as an explicit "don't push these" option, to override any other options.

If that doesn't seem plausible, we can drop the thunk if this will only ever be `true`.

cc @ImJasonH bikeshed naming here?

jonjohnsonjr (commented):
Neat, thanks! Once CLA is sorted out this looks pretty good.

I'll want to expose this option in pkg/crane and as a flag for `--crane`, but I can do that later myself.

pkg/v1/remote/options.go
ImJasonH: +1 to `WithNonDistributable` and a clear docstring. If we find a better alternative we can deprecate this and call it something else later.

Also +1 to making it only true-able, and not make people think about what `WithNonDistributable(false)` means. 😄

StevenLocke:
@googlebot I signed it!

pkg/v1/remote/options.go
jonjohnsonjr: > Also +1 to making it only true-able, and not make people think about what WithNonDistributable(false) means. 😄

I was suggesting the opposite actually, but I'm open to being convinced.

The point I was trying to make is that if we do decide only true-able, I'd rather see:

```
foo(bar, remote.WithNondistributable)
```

than

```
foo(bar, remote.WithNondistributable())
```

> `WithNonDistributable`

I'm gonna be super pedantic here but I think lowercasing the `D` matches the [media type](https://github.com/google/go-containerregistry/blob/2276eac05fbee1de1315d16ec4b9346f3350c804/pkg/v1/types/types.go#L27) better, if we go with this, so:

`WithNondistributable` is my preference.

ImJasonH: I defer to Jon, `remote.WithNondistributable` (no `()`) sgtm

jonjohnsonjr (approved):
LGTM, thanks!

PR #936, opened 2021-02-08
Delete layout blob

deitch:
We already can WriteBlob, so we should be able to RemoveBlob.

### DIFF
--- a/pkg/v1/layout/write.go
+++ b/pkg/v1/layout/write.go
@@ -264,6 +264,19 @@ func (l Path) writeLayer(layer v1.Layer) error {
 	return l.WriteBlob(d, r)
 }
 
+// RemoveBlob removes a file from the blobs directory in the Path
+// at blobs/{hash.Algorithm}/{hash.Hex}
+// It does *not* remove any reference to it from other manifests or indexes, or
+// from the root index.json.
+func (l Path) RemoveBlob(hash v1.Hash) error {
+	dir := l.path("blobs", hash.Algorithm)
+	err := os.Remove(filepath.Join(dir, hash.Hex))
+	if err != nil && !os.IsNotExist(err) {
+		return err
+	}
+	return nil
+}
+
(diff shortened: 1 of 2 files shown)


pkg/v1/layout/write.go
```diff
@@ -264,6 +264,15 @@ func (l Path) writeLayer(layer v1.Layer) error {
 	return l.WriteBlob(d, r)
 }
 
+// RemoveBlob removes a file from the blobs directory in the Path
+// at blobs/{hash.Algorithm}/{hash.Hex}
+// It does *not* remove any reference to it from other manifests or indexes, or
+// from the root index.json.
+func (l Path) RemoveBlob(hash v1.Hash) error {
+	dir := l.path("blobs", hash.Algorithm)
+	return os.RemoveAll(filepath.Join(dir, hash.Hex))
```
jonjohnsonjr, line 273: I think we just want `Remove` not `RemoveAll`?

jonjohnsonjr (changes requested):
Looks fine beyond my one nit. Happy to merge if you add a unit test.

pkg/v1/layout/write.go
deitch: Because I don't actually want the action of removing it; I want it not to be there. `RemoveAll` handles it not being there, while `Remove` errors if the file doesn't exist. That is easy enough to handle with `os.IsNotExist`; I was just lazy.

I can fix it.

deitch: OK, I changed this part.

deitch:
Test in place too

pkg/v1/layout/write.go
jonjohnsonjr: Ah, that wasn't super clear at a glance. Explicitly checking for not found is easier for me to understand 👍

jonjohnsonjr (approved):

jonjohnsonjr:
Thanks!

PR #922, opened 2021-01-26
bump k8s to v0.19.7 and drop travis

dprotaso:

PR #943, opened 2021-02-22
go: update references to use go 1.15

cpanato:
In the GH actions, it is already using go 1.15.x was missing to update go in two places.

/assign @jonjohnsonjr

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -19,7 +19,7 @@ jobs:
         name: Set up Go
         uses: actions/setup-go@v1
         with:
-          go-version: 1.14
+          go-version: 1.15
       -
         name: Run GoReleaser
         uses: goreleaser/goreleaser-action@v2
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/google/go-containerregistry
 
-go 1.14
+go 1.15
 
 require (
 	cloud.google.com/go v0.57.0 // indirect


ImJasonH:
Does it make sense to include 1.16 as well?

cpanato:
if all agree I can update all the references to be 1.16

jonjohnsonjr:
Let's stick with 1.15 for now? We used to have a matrix of tests in travis, but we've lost that. If we could test both 1.15 and 1.16, I'd be happy.

cpanato:
@jonjohnsonjr i can look and implement the CI GO test matrix using GitHub actions

jonjohnsonjr:
That would be phenomenal -- we dropped travis in https://github.com/google/go-containerregistry/pull/922

cpanato:
gently ping @jonjohnsonjr

jonjohnsonjr:
I'm not 100% clear on what the go version really  _means_ in `go.mod`. Is it the minimum version at which the module will build? When are you supposed to update it? It's hard to google for this stuff :/

cpanato:
me too, but I use that to set my env to the appropriate go version

jonjohnsonjr:
I found https://github.com/golang/go/issues/30791 which seems to agree but I haven't read the whole thing.

Do we know the minimum go version? I'm reluctant to bump this unless we really need to, because it might break folks on go1.14 for no reason...

cpanato:
I dont have strong opinions. feel free to close this PR if you want

cpanato:
@jonjohnsonjr I will close this PR, if we decided to update the references I can make another one

jonjohnsonjr:
Thanks - I think we'll want to bump eventually, but I don't have any kind of policy for compatibility with go versions in mind quite yet.

jonjohnsonjr:
Actually... I'd be happy to bump the version used in the release to 1.16, given that there are some performance improvements in the compiler from 1.14 -> 1.15 and 1.15 -> 1.16.

My only issue is with the go.mod thing, I think.

PR #944, opened 2021-02-22
implement catalog tests

cpanato:
as a continuation of in intention of the work here: https://github.com/google/go-containerregistry/pull/915

This PR implements the `Catalog` tests

Feedbacks are welcome

/assign @jonjohnsonjr

### DIFF
--- a/pkg/crane/crane_test.go
+++ b/pkg/crane/crane_test.go
@@ -34,7 +34,6 @@ import (
 	"github.com/google/go-containerregistry/pkg/v1/remote"
 )
 
-// TODO(jonjohnsonjr): Test crane.Catalog behavior.
 // TODO(jonjohnsonjr): Test crane.Copy failures.
 func TestCraneRegistry(t *testing.T) {
 	// Set up a fake registry.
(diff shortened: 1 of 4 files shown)


pkg/crane/crane_test.go
```diff
@@ -195,6 +194,20 @@ func TestCraneRegistry(t *testing.T) {
...
+	// List Catalog
+	repos, err := crane.Catalog(u.Host)
+	if err != nil {
+		t.Fatal(err)
+	}
+	if len(repos) != 2 {
+		t.Fatalf("wanted 2 repos, got %d", len(repos))
+	}
+
+	_, err = crane.Catalog("registry.honk")
```
jonjohnsonjr, line 207: I believe this will actually hit the internet, which I'd like to avoid. This might be covered by [these tests](https://github.com/google/go-containerregistry/blob/60c7f4d7cd5b5c98fc8113183a91a5b8b374170a/pkg/crane/crane_test.go#L420-L421) already?

cpanato: removed, PTAL

cpanato:
gently ping @jonjohnsonjr

jonjohnsonjr (approved):

jonjohnsonjr:
Thanks!

PR #960, opened 2021-03-10
Implement annotation-based rebase hints

imjasonh:
Adds `crane.Rebase` helper function for higher-level functionality including annotation-based hints

With this change, `crane rebase <image>` without args will attempt to determine the old and new base from (WIP placeholder) annotations, perform the rebase, and push a new image back to `<image>` (`--rebased` flag will override). The rebased image will have base-by-digest and base-by-tag annotations added for future rebase operations.

- `--original` if provided overrides the positional arg
- `--old_base` and `--new_base` override any detected annotations
- `--rebased` (deprecated, now `--tag` to match other surfaces) overrides pushing over the original image
- `base-digest` and `base-tag` annotations are populated from `--new_base` by digest+tag (or previously detected `base-tag`)

~~This is WIP while we sort out whether the annotation will be a reserved annotation in the OCI spec. If the code looks good, I'll take a stab at updating rebase.md as well (it could probably use it anyway 😅 ).~~


pkg/v1/mutate/rebase.go
```diff
@@ -142,3 +143,16 @@ func createAddendums(startHistory, startLayer int, history []v1.History, layers
 
 	return adds
 }
+
+func Manifest(base v1.Image, m *v1.Manifest) (v1.Image, error) {
```
jonjohnsonjr, line 147: I think what you _really_ want is `mutate.Annotations` or something? Allowing you to set the whole manifest introduces some weirdness around invariant violations (`mutate.ConfigFile` does too, but I'd like to minimize the amount of it).

jonjohnsonjr: (Ideally we could make this work for `Image`, `ImageIndex`, and `Layer`, but I'm fine with three variants of that to make the type system nonsense a little less silly.)

ImJasonH: Done, with `mutate.Annotations` (it only supports `v1.Image`, maybe we should expand that?)

pkg/crane/rebase.go
```diff
@@ -0,0 +1,110 @@
...
+
+// Rebase parses the references and uses them to perform a rebase.
+//
+// If oldBase or newBase are "", Rebase attempts to derive them using
+// annotations in the original image. If those annotations are not found,
+// Rebase returns an error.
+//
+// If rebasing is successful, base image annotations are set on the resulting
+// image to facilitate implicit rebasing next time.
+func Rebase(orig, oldBase, newBase string, opt ...Option) (v1.Image, error) {
```
jonjohnsonjr, line 40: Given that this returns a `v1.Image`, I think `orig` should also be a `v1.Image`.

Alternatively, this could operate on just refs, so that we could make it work for index as well.

imjasonh: I'd also be fine just moving all of `pkg/crane.Rebase` back into cmd/crane/cmd/rebase.go for now, at least until we hammer out index rebasing in #1086, then see how we want `pkg/crane.Rebase` to look and work.

cmd/crane/cmd/rebase.go
```diff
@@ -29,48 +35,184 @@ func NewCmdRebase(options *[]crane.Option) *cobra.Command {
...
+			} else if len(args) != 0 || args[0] != "" {
+				return fmt.Errorf("cannot use --original with positional argument")
+			}
+
+			// Parse and fetch the original image or index.
+			origRef, err := name.ParseReference(orig)
+			if err != nil {
+				return fmt.Errorf("parsing tag %q: %v", orig, err)
+			}
+			origDesc, err := remote.Get(origRef)
```
jonjohnsonjr, line 51: Using `remote.Get` here breaks flags because we're not passing down `*options...`. 

I would use `crane.Head` to make sure it's not an index and then `crane.Pull` it otherwise. We waste a request but it doesn't really matter if this is fast. If you want to go fast, maybe add `crane.Get`.

imjasonh: Done. Now using `crane.Head` and `crane.Pull`, but parsing the reference in order to determine whether it was a digest (and to modify it to set a tag instead) still requires calls to `name.ParseReference`. 🤔

imjasonh: --insecure and --platform support now stolen from #1104

cmd/crane/cmd/rebase.go
```diff
@@ -29,48 +35,176 @@ func NewCmdRebase(options *[]crane.Option) *cobra.Command {
...
-	rebaseCmd.Flags().StringVarP(&rebased, "rebased", "", "", "Tag to apply to rebased image")
-
-	rebaseCmd.MarkFlagRequired("original")
-	rebaseCmd.MarkFlagRequired("old_base")
-	rebaseCmd.MarkFlagRequired("new_base")
-	rebaseCmd.MarkFlagRequired("rebased")
+	rebaseCmd.Flags().StringVar(&orig, "original", "", "Original image to rebase; use positional arg instead")
+	rebaseCmd.Flags().StringVar(&oldBase, "old_base", "", "Old base image to remove")
+	rebaseCmd.Flags().StringVar(&newBase, "new_base", "", "New base image to insert")
+	rebaseCmd.Flags().StringVarP(&rebased, "tag", "t", "", "Tag to apply to rebased image")
```
jonjohnsonjr, line 137: You should keep the `rebased` flag for backward compatibility.

cmd/crane/cmd/rebase.go
```diff
@@ -29,48 +35,177 @@ func NewCmdRebase(options *[]crane.Option) *cobra.Command {
...
+			if rebasedDigest == origDigest {
+				logs.Warn.Println("rebasing was no-op")
+			}
+
+			logs.Progress.Println("pushing rebased image as", rebased)
+			if _, ok := r.(name.Digest); ok {
+				digest, err := rebasedImg.Digest()
+				if err != nil {
+					log.Fatalf("digesting new image: %v", err)
+				}
```
jonjohnsonjr, line 123: You compute this as `rebasedDigest` already.

cmd/crane/cmd/rebase.go
```diff
@@ -29,48 +35,177 @@ func NewCmdRebase(options *[]crane.Option) *cobra.Command {
...
+
+	oldBaseImg, err := crane.Pull(oldBase, opt...)
+	if err != nil {
+		return nil, err
+	}
+
+	// NB: if newBase is an index, we need to grab the index's digest to
+	// annotate the resulting image, even though we pull the
+	// platform-specific image to rebase.
+	newBaseDigest, err := crane.Digest(newBase, opt...)
```
jonjohnsonjr, line 194: With `--platform`, we pass `crane.WithPlatform`, which will cause `crane.Digest` to pull through the index and resolve the image digest: https://github.com/google/go-containerregistry/blob/bea59b93e0bf9f2ac673defd6d8f2a8e162937b9/pkg/crane/digest.go#L22

Using `crane.Head` will ignore the platform, so I think you want to use that here.

This is not super intuitive.

cmd/crane/doc/crane_rebase.md
```diff
@@ -12,8 +12,9 @@ crane rebase [flags]
   -h, --help              help for rebase
       --new_base string   New base image to insert
       --old_base string   Old base image to remove
-      --original string   Original image to rebase
-      --rebased string    Tag to apply to rebased image
+      --original string   Original image to rebase; use positional arg instead
```
jonjohnsonjr, line 15: Use same `(DEPRECATED: use positional arg)` format as `--rebased`?

cmd/crane/cmd/rebase.go
```diff
@@ -29,48 +35,177 @@ func NewCmdRebase(options *[]crane.Option) *cobra.Command {
...
 			}
 
-			newBaseImg, err := crane.Pull(newBase, *options...)
+			desc, err := crane.Head(orig, *options...)
 			if err != nil {
-				return fmt.Errorf("pulling %s: %v", newBase, err)
+				log.Fatalf("checking %s: %v", orig, err)
+			}
+			if !cmd.Parent().PersistentFlags().Changed("platform") && desc.MediaType.IsIndex() {
+				log.Fatalf("flattening an index is not yet supported")
```
jonjohnsonjr, line 71: s/flattening/rebasing/

cmd/crane/rebase_test.sh
```diff
@@ -0,0 +1,62 @@
...
+  --annotation org.opencontainers.image.base.name=localhost:1338/base \
+  --annotation org.opencontainers.image.base.digest=$(./crane digest localhost:1338/base))
+
+# Update localhost:1338/base containing b.txt
+echo b > ${tmp}/b.txt
+new_base=$(./crane append -f <(tar -f - -c ${tmp}) -t localhost:1338/base)
+rm ${tmp}/b.txt
+
+# Rebase using annotations
+rebased=$(./crane rebase ${orig})
```
jonjohnsonjr, line 43: This is amazing 👍

pkg/v1/mutate/rebase.go
```diff
@@ -46,6 +46,7 @@ func Rebase(orig, oldBase, newBase v1.Image) (v1.Image, error) {
 			return nil, fmt.Errorf("failed to get digest of layer %d of %q: %v", i, orig, err)
 		}
 		if oldLayerDigest != origLayerDigest {
+			// TODO: this is a bad error message...
```
jonjohnsonjr, line 49: What should we do instead?

cmd/crane/cmd/rebase.go
imjasonh: Thanks for catching this. I agree it's not super great. 🤷

cmd/crane/rebase_test.sh
imjasonh: Thanks. I still need to hook it up into CI tests...

pkg/v1/mutate/rebase.go
imjasonh: Meh, it's fine. I don't have a better alternative in mind, so I've just removed the comment.

jonjohnsonjr:
> This is WIP while we sort out whether the annotation will be a reserved annotation in the OCI spec. If the code looks good, I'll take a stab at updating rebase.md as well (it could probably use it anyway 😅 ).

Wanna do that?

imjasonh:
> Wanna do that? If not, ready for final review?

Done, anything else you'd like to see?

jonjohnsonjr (approved):

PR #961, opened 2021-03-11
Add MustParseReference

dekkagaijin:
Makes things less annoying for testing and `const`ish initialization scenarios

<3

### DIFF
--- a/pkg/name/ref.go
+++ b/pkg/name/ref.go
@@ -47,3 +47,12 @@ func ParseReference(s string, opts ...Option) (Reference, error) {
 	return nil, NewErrBadName("could not parse reference: " + s)
 
 }
+
+// MustParseReference behaves like ParseReference, but panics instead of returning an error.
+func MustParseReference(s string, opts ...Option) Reference {
+	ref, err := ParseReference(s, opts...)
+	if err != nil {
+		panic(err)
+	}
+	return ref
+}
--- a/pkg/name/ref_test.go
+++ b/pkg/name/ref_test.go
@@ -116,3 +116,38 @@ func TestParseReference(t *testing.T) {
 		}
 	}
 }
+
+func TestMustParseReference(t *testing.T) {
+	goodWeakValidationNames := append(goodWeakValidationTagNames, goodWeakValidationDigestNames...)
+	for _, name := range goodWeakValidationNames {
+		func() {
+			defer func() {
+				if err := recover(); err != nil {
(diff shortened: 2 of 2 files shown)


ImJasonH:
I kinda like this! Some random thoughts:

- If it's only intended for testing (and I think it should be) do we want to put it in a subpackage like `pkg/name/nametest`?
  - (I wish Go had a concept for testonly packages like Bazel does...)
- Do we want to only support `ParseReference` as a first step? Users could still do:

```
var tag = nametest.MustParseReference("obviously/valid").(v1.Tag)
```
if they expect the result to be a `name.Tag`.

dekkagaijin:
> If it's only intended for testing (and I think it should be) do we want to put it in a subpackage like pkg/name/nametest?

It's a super common pattern in test, but it's also normal for instances like
```
var (
   img = name.MustParseXXX("intentional.ly/pull:mutablere-release-tag")
   ...
)
```

> Do we want to only support ParseReference as a first step?

Not too opinionated, but is there a reason not to achieve API parity?

dekkagaijin:
https://golang.org/pkg/regexp/#MustCompile is my muse

ImJasonH:
> Not too opinionated, but is there a reason not to achieve API parity?

Baby steps, mostly. It's easy to add stuff, and nearly impossible to remove it.

We do this in our own tests quite a lot, e.g., [`mustNewTag`](https://github.com/google/go-containerregistry/blob/2f4c6f06b18713c67a11f19b413de705a5338309/pkg/v1/remote/write_test.go#L49). It's possible people could want this in non-test scenarios, but I think it's worth nudging them toward thinking about not using it outside of tests.

If we decide we want to move this to `pkg/name`, we can do that relatively easily down the line.

> https://golang.org/pkg/regexp/#MustCompile is my muse

There's also the https://golang.org/pkg/text/template/#Must form, e.g.:

```
func Must(name.Reference, error) name.Reference
```

```
var tag    = name.Must(name.ParseReference("blah")).(v1.Tag)
var digest = name.Must(name.ParseReference("blah@sha256:...")).(v1.Digest)
```

Or the slightly more terrible idea:

```
var img = remote.Must(remote.Image(name.Must(name.ParseReference("blah"))))
```

dekkagaijin:
Pruned the PR down to `MustParseReference`

dekkagaijin:
gonna close and reopen to try and trigger the presubmits again

PR #964, opened 2021-03-12
MustParseReference only accepts consts

### LINKED ISSUE #963: Make MustParseReference only accept consts
The [recently added `MustParseReference`](https://github.com/google/go-containerregistry/issue/962) takes a string, and returns its successfully parsed `name.Reference` if it's successfully parsed, and panics otherwise. This is mainly useful for tests, where a value is known/expected to be valid at authoring time.

What if `MustParseReference` only accepted const strings, to make it clearer that it's not intended to accept user-defined values.

```
type stringConst string

func MustParseReference(s stringConst) Reference {
  ...
}
```

Since the `stringConst` type isn't exported, external callers can't make one themselves. Go's type coercion is able to automatically convert a string-typed const to the correct type.

[Playground link](https://play.golang.org/p/cYcrTjpdxAI)

This would prevent people from using `MustParseReference` with dynamically-constructed strings, and wouldn't stop them from using it in test setups like `const ref = name.MustParseReference("valid")`.

ImJasonH:
Fixes #963

### DIFF
--- a/pkg/name/ref.go
+++ b/pkg/name/ref.go
@@ -48,9 +48,27 @@ func ParseReference(s string, opts ...Option) (Reference, error) {
 
 }
 
-// MustParseReference behaves like ParseReference, but panics instead of returning an error.
-func MustParseReference(s string, opts ...Option) Reference {
-	ref, err := ParseReference(s, opts...)
+type stringConst string
+
+// MustParseReference behaves like ParseReference, but panics instead of
+// returning an error. It's intended for use in tests, or when a value is
+// expected to be valid at code authoring time.
+//
+// To discourage its use in scenarios where the value is not known at code
+// authoring time, it must be passed a string constant:
+//
+//    const str = "valid/string"
+//    MustParseReference(str)
+//    MustParseReference("another/valid/string")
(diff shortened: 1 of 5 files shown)


jonjohnsonjr (approved):
lgtm

dekkagaijin (approved):
lgtm, very sneaky sir

pkg/name/ref.go
```diff
@@ -48,9 +48,27 @@ func ParseReference(s string, opts ...Option) (Reference, error) {
...
+type stringConst string
+
+// MustParseReference behaves like ParseReference, but panics instead of
+// returning an error. It's intended for use in tests, or when a value is
+// expected to be valid at code authoring time.
+//
+// To discourage its use in scenarios where the value is not known at code
+// authoring time, it must be passed a string constant:
+//
+//    MustParseReference("valid/string")
```
dekkagaijin, line 60: Just because `prefix+testCase` is common enough, how about:

```
const validStr = "valid/string"
MustParseReference(validStr)
MustParseReference("a/second/"+validStr)
MustParseReference("yet/another/valid/string")
```

Issue #952, opened 2021-02-25
Add `WithProgress` option for `remote.Write`


tydra-wang:
As the title.

DennisDenuto:
+1

I would find this also useful for `remote.MultiWrite`

DennisDenuto:
Would it be correct to assume that `WithProgress` would work very similar to how it works for `tarball.WriteToFile`? 

https://github.com/google/go-containerregistry/blob/77421d5c5906d1a567ed0193787cac3db8b4a4e4/pkg/v1/progress.go#L21

i.e. `MultiWrite` would send progress information to the `v1.Update` channel?

ImJasonH:
Yeah, I'd like to have them behave similarly. The difference with `MultiWrite` is that there are multiple things having progress made, does it make sense for the update to include information about which things are progressing how?

MultiWrite already doesn't support streaming layers, so we can know ahead of time how many total bytes we anticipate writing. In theory we'd like to support streaming layers though, in which case the `Total` might not be known ahead-of-time.

I think we've also talked about it taking a chan of things to write, so the update's `Total` could increase as new things were requested. I don't know if that breaks any assumptions.

DennisDenuto:
Thanks that is pretty helpful!

I would like to volunteer to work on this (This Friday will probably be the day I work on it). 

wrt MultiWrite and streaming layers: 
I will calculate the total bytes ahead-of-time before any blobs are uploaded (when / if streaming layers are supported, the implementation can be changed)

ImJasonH:
Sounds good @DennisDenuto, thanks!

Let us know if you have any questions or problems, we'd be happy to help.

jonjohnsonjr:
> The difference with MultiWrite is that there are multiple things having progress made, does it make sense for the update to include information about which things are progressing how?

The equivalent in [docker land](https://github.com/moby/moby/blob/40d9e2aff130b42ba0f83d5238b9b53184c8ab3b/pkg/progress/progress.go#L9) partitions upload progress by ID. I think that's usually layers.

I'm honestly not sure what to do here. A lot of the time, I want a single progress bar, and that's an easy model to program against... but folks might want to display more complex things than a single bar (e.g. per-layer progress), but making that change actually seems nontrivial, and I'm not sure how much it makes sense for MultiWrite (there can be hundreds of layers)...

> I think we've also talked about it taking a chan of things to write, so the update's Total could increase as new things were requested. I don't know if that breaks any assumptions.

[...]

PR #967, opened 2021-03-15
Support WithProgress for remote Writes

### LINKED ISSUE #852: TestMultiWrite_Deep is slow
tl;dr

```
=== RUN   TestMultiWrite_Deep
--- PASS: TestMultiWrite_Deep (38.07s)
```

More context in https://github.com/google/go-containerregistry/pull/851#issuecomment-736115162

ImJasonH:
Fixes https://github.com/google/go-containerregistry/issues/952

Shoutout to @DennisDenuto who paired with me on this 🎉 

- supported in `remote.WriteLayer`, `remote.Write`, `remote.WriteIndex`, `remote.MultiWrite`
- given update chan is closed when write completes
- an error is sent along the chan if there are any non-temporary errors uploading
- if a layer already exists or is mounted, progress updates immediately to account for that
- if layer upload fails and is retried, progress goes backward and goes back up

https://github.com/google/go-containerregistry/commit/d7fe93870768a18d524d5379b5d97480ab7715b0 demonstrates how easy it is to consume this channel, by hacking in an upload progress bar for `crane cp` (not part of this PR)

Fixes #852 as well as a bonus ⏩

### DIFF
--- a/pkg/v1/remote/multi_write.go
+++ b/pkg/v1/remote/multi_write.go
@@ -33,7 +33,7 @@ import (
 // Current limitations:
 // - All refs must share the same repository.
 // - Images cannot consist of stream.Layers.
-func MultiWrite(m map[name.Reference]Taggable, options ...Option) error {
+func MultiWrite(m map[name.Reference]Taggable, options ...Option) (rerr error) {
 	// Determine the repository being pushed to; if asked to push to
 	// multiple repositories, give up.
 	var repo, zero name.Repository
@@ -86,9 +86,49 @@ func MultiWrite(m map[name.Reference]Taggable, options ...Option) error {
 		return err
 	}
 	w := writer{
-		repo:    repo,
-		client:  &http.Client{Transport: tr},
-		context: o.context,
+		repo:       repo,
+		client:     &http.Client{Transport: tr},
+		context:    o.context,
+		updates:    o.updates,
+		lastUpdate: &v1.Update{},
+	}
+
+	// Collect the total size of blobs and manifests we're about to write.
+	if o.updates != nil {
+		defer close(o.updates)
+		defer func() { sendError(o.updates, rerr) }()
+		for _, b := range blobs {
+			size, err := b.Size()
+			if err != nil {
+				return err
+			}
+			w.lastUpdate.Total += size
+		}
+		countManifest := func(t Taggable) error {
+			b, err := t.RawManifest()
+			if err != nil {
+				return err
+			}
+			w.lastUpdate.Total += int64(len(b))
+			return nil
+		}
+		for _, i := range images {
+			if err := countManifest(i); err != nil {
+				return err
+			}
+		}
+		for _, nm := range newManifests {
+			for _, i := range nm {
+				if err := countManifest(i); err != nil {
+					return err
+				}
+			}
+		}
+		for _, i := range indexes {
+			if err := countManifest(i); err != nil {
+				return err
+			}
+		}
 	}
 
 	// Upload individual blobs and collect any errors.
@@ -160,8 +200,8 @@ func MultiWrite(m map[name.Reference]Taggable, options ...Option) error {
 	}
 	// Push originally requested index manifests, which might depend on
 	// newly discovered manifests.
-	return commitMany(indexes)
 
+	return commitMany(indexes)
 }
 
 // addIndexBlobs adds blobs to the set of blobs we intend to upload, and
--- a/pkg/v1/remote/multi_write_test.go
+++ b/pkg/v1/remote/multi_write_test.go
@@ -156,7 +156,7 @@ func TestMultiWrite_Deep(t *testing.T) {
 	if err != nil {
 		t.Fatal("random.Image:", err)
 	}
-	for i := 0; i < 10; i++ {
+	for i := 0; i < 4; i++ {
 		idx = mutate.AppendManifests(idx, mutate.IndexAddendum{Add: idx})
 	}
 
--- a/pkg/v1/remote/options.go
+++ b/pkg/v1/remote/options.go
@@ -37,6 +37,7 @@ type options struct {
 	jobs                           int
 	userAgent                      string
 	allowNondistributableArtifacts bool
+	updates                        chan<- v1.Update
 }
 
 var defaultPlatform = v1.Platform{
@@ -183,4 +184,15 @@ func WithUserAgent(ua string) Option {
 func WithNondistributable(o *options) error {
 	o.allowNondistributableArtifacts = true
 	return nil
+}
+
+// WithProgress takes a channel that will receive progress updates as bytes are written.
+//
+// Sending updates to an unbuffered channel will block writes, so callers
+// should provide a buffered channel to avoid potential deadlocks.
+func WithProgress(updates chan<- v1.Update) Option {
+	return func(o *options) error {
+		o.updates = updates
+		return nil
+	}
 }
--- a/pkg/v1/remote/progress_test.go
+++ b/pkg/v1/remote/progress_test.go
@@ -1,0 +1,412 @@
+// Copyright 2021 Google LLC All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package remote
+
+import (
+	"errors"
+	"fmt"
+	"net/http"
+	"net/http/httptest"
+	"net/url"
+	"strings"
+	"testing"
+
+	"github.com/google/go-cmp/cmp"
+	"github.com/google/go-containerregistry/pkg/name"
+	"github.com/google/go-containerregistry/pkg/registry"
+	v1 "github.com/google/go-containerregistry/pkg/v1"
+	"github.com/google/go-containerregistry/pkg/v1/empty"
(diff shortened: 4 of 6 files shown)


jonjohnsonjr (approved):

deitch:
I was just hunting for something like this (especially after having put together the tarball one at #675 :-) . I was stumped, as I did not see it in godoc or pkg.go.dev, and was going to start trying to think about doing it, when I found this PR. Apparently godoc only takes the latest semver release (if one exists). 

Does this send [v1.Update](https://pkg.go.dev/github.com/google/go-containerregistry/pkg/v1#Update)  on the channel:

```go
type Update struct {
	Total    int64
	Complete int64
	Error    error
}
```

so it has just those fields? Does it calculate all of the bytes before sending, across all of the manifests and layers (e.g. `remote.Write()` or `remote.WriteIndex()` (it looks like it from skimming the source), so `Total` really is the total target, and `Complete` is the amount sent?

I also saw discussion about how docker does its `push`, where it has a series of updates per blob sent (layers, config, etc.). Are we interested in going down that path?

I came across this with a `remote.WriteIndex()` that errors out once in a while, and it isn't clear how far it got or why. Having progress reports would go a long way.

ImJasonH:
> I was just hunting for something like this (especially after having put together the tarball one at #675 :-) . I was stumped, as I did not see it in godoc or pkg.go.dev, and was going to start trying to think about doing it, when I found this PR. Apparently godoc only takes the latest semver release (if one exists).

I'm actually planning on doing a ggcr release today! In the meantime, https://pkg.go.dev/github.com/google/go-containerregistry@main/pkg/v1/remote should work.

> 
> Does this send [v1.Update](https://pkg.go.dev/github.com/google/go-containerregistry/pkg/v1#Update) on the channel:
> 
> ```go
> type Update struct {
> 	Total    int64
> 	Complete int64
> 	Error    error
> }
> ```
> 
> so it has just those fields? Does it calculate all of the bytes before sending, across all of the manifests and layers (e.g. `remote.Write()` or `remote.WriteIndex()` (it looks like it from skimming the source), so `Total` really is the total target, and `Complete` is the amount sent?

That's correct. It sends `v1.Update`, having pre-calculated the `Total`, taking into account duplicate blobs and manifests.

> 
> I also saw discussion about how docker does its `push`, where it has a series of updates per blob sent (layers, config, etc.). Are we interested in going down that path?

I think that's something we could investigate and implement, yes. We could accomplish this by adding fields to the `v1.Update` type, but we haven't really looked into it much yet.

deitch:
OK, I will start with that. I have been very bogged down (in a positive sense) with various deliverables the last two weeks, otherwise I would take a stab at it. If I can at some point, I will.

deitch:
This question is _slightly_ off topic, but triggered here. How do I get a `v1.Layer` from a file on disk? The reason I ask here is, I would normally do:

```go
                        f, err := os.Open(somefile)
                        defer f.Close()
                        layer = stream.NewLayer(f)
```

But that then gives you a `*stream.Layer`, which cannot be used with `WithProgress()`. Is there a way to get a layer off an arbitrary local file that is compatible with this?

ImJasonH:
https://pkg.go.dev/github.com/google/go-containerregistry@main/pkg/v1/tarball#LayerFromFile is the best we have right now.

Ideally updates would be able to handle streaming layers by increasing the total as bytes are streamed, but this hasn't been done yet.

deitch:
True, but in this case, I actually _know_ the total size (it is just a file), so no reason not to have proper updates.

Would that actually work? Or only with a tarball?

ImJasonH:
Would it make sense to have some `stream.NewLayerWithSize` that asserted the size? 🤔

deitch:
That certainly would make it easier to get `WithProgress()` working. I would think that the ability to just get a layer from a file - i.e. something that has a known size - would make sense. Then again, that is precisely what `NewLayerWithSize` does. 

At some point, if/when we decide to have per-layer progress reports, we will need the size for each layer as we push it up, won't we? This problem becomes the same there.

Although, as I think about it, when we push an `Image` or `Index`, we have the sizes of each blob from the manifest parent, so `WithSize` actually should cover that use case.

ImJasonH:
A `stream.Layer` doesn't know its size or digest by default today, and so a `v1.Image` that contains a `stream.Layer` can't calculate its `Manifest` or `Digest` or basically anything until that stream has been consumed and its digest and size calculated (these are streamily calculated while the stream is being consumed).

Care is taken so that mutations are lazily applied, so you can `mutate.Append` a `stream.Layer` and `remote.Write` it, which will read and push the blob before committing the just-recently-calculatable manifest. It's complex and brittle and honestly kind of a headache for not much gain in a lot of cases.

The use case for `stream.Layer` is in being able to _generate_ layer contents and stream them to a registry (etc.), not in being able to stream bytes from disk, where size and digest are pretty easily calculatable. `tarball.LayerFromFile` assumes the file contents are already tarred, but if that's not the case you can either tar it and write it to back disk temporarily, or tar+stream it through [`tarball.LayerFromOpener`](https://pkg.go.dev/github.com/google/go-containerregistry@main/pkg/v1/tarball#LayerFromOpener), or just implement [`v1.Layer`](https://pkg.go.dev/github.com/google/go-containerregistry@main/pkg/v1#Layer) however else you want. `stream.Layer` is probably not going to be a good fit if you already know the size and digest, at least not in its current form.

deitch:
> The use case for stream.Layer is in being able to generate layer contents and stream them to a registry (etc.), not in being able to stream bytes from disk, where size and digest are pretty easily calculatable. tarball.LayerFromFile assumes the file contents are already tarred, but if that's not the case you can either tar it and write it to back disk temporarily, or tar+stream it through tarball.LayerFromOpener, or just implement v1.Layer however else you want. stream.Layer is probably not going to be a good fit if you already know the size and digest, at least not in its current form.

So thinking about the bigger picture, we either need a concept of presetting size and digest on `stream.Layer`, or having a different kind of layer that represents one whose digest and size are known.

What do `layout.Image` and `remote.Image` and `tarball.Image` and such do, when returning a `v1.Layer`?

ImJasonH:
`remote.Image` never returns a `stream.Layer`, it only returns a [`remote.remoteLayer`](https://github.com/google/go-containerregistry/blob/0976a278d3ebefb47fdbf5a2ff2b3e8a405c9e89/pkg/v1/remote/layer.go#L28) which calls `fetchBlob` lazily when you request the blob contents. `layout.Image` and `tarball.Image` already assume they has the full manifest (as bytes or struct), so it also doesn't have to deal with `stream.Layer`s directly.

Really the only time you'd have to deal with them is when you `mutate.AppendLayer` a `stream.Layer` on top of them, in which case you're actually dealing with the [`mutate.image`](https://github.com/google/go-containerregistry/blob/0976a278d3ebefb47fdbf5a2ff2b3e8a405c9e89/pkg/v1/mutate/image.go#L29), which handles all the lazy calculation for you, and fails if an appended `stream.Layer` hasn't been computed yet.

`stream.Layer` is intended for cases where the contents are generated on-the-fly, so size and digest are unknowable. If you already know those things, you probably want something less annoying to deal with, such as a `tarball.Layer` or some other `v1.Layer` implementation built for your purpose.

jonjohnsonjr:
> remote.Image never returns a stream.Layer, it only returns a remote.remoteLayer which calls fetchBlob lazily when you request the blob contents.

This is a slight oversimplification. If you've accessed a layer via `remote.Image`, we return a `remoteImageLayer`, which cheaply computes the digest, size, diffid, and mediaType by looking at the image manifest or config file.

If you access a layer via `remote.Layer`, we return a `remoteLayer`, which can get the size via a HEAD request, knows the digest because it was an input, and has the mediaType hardcoded (best we can do, really).

@deitch do you have this layer inside a layout? Or is it just a file? Where do you have this metadata stored? We'd probably want to implement _something_ that knows how to read/parse this metadata and can return a layer with all of this pre-populated, rather than you having to supply each piece individually.

ImJasonH:
> This is a slight oversimplification. If you've accessed a layer via remote.Image, we return a remoteImageLayer, which cheaply computes the digest, size, diffid, and mediaType by looking at the image manifest or config file.

Thanks for this factcheck, you're right. In either case, `stream.Layer` is not involved when you fetch an image using `pkg/remote`.

jonjohnsonjr:
Some maybe useful example of similar situations...

### cosign uploading signatures:

https://github.com/sigstore/cosign/blob/7fee630550dc8939dcbdf4f68898944844794569/pkg/cosign/remote.go#L86-L120

It's a very basic static `v1.Layer` implementation that just returns some bytes and computes its metadata.

It does this itself because in this case we _don't_ want to compress the layer, which `stream` and `tarball` assume you do.

### rules_docker translating from bazel's representation to v1.Layer:

https://github.com/bazelbuild/rules_docker/blob/8c3a8110a0c519929a7e79c39ac345a0f8c74d04/container/go/pkg/compat/image.go#L35-L115

`loadHashes` reads the digest and diffid from rules_docker's internal representation of these images, and `Size`/`UncompressedSize` just `Stat` the relevant file.

### crane uploading a layer from a file:

https://github.com/google/go-containerregistry/blob/37467b5ec0276b05494c8adbd6f43938e1e5030d/pkg/crane/append.go#L42-L52

If it's a streaming thing, we use `stream.Layer`, otherwise `tarball.Layer`. This _doesn't_ have any metadata associated with it, so we just fall back to `v1.Layer` implementations that will compute all of this for you.

deitch:
Ah all of that makes sense and clarifies a lot. I didn't think about the implications when I had a file, so I just wrapped it all with a `stream.Layer`. Clearly, that wasn't the best option.

> do you have this layer inside a layout? Or is it just a file?

Both use cases, actually. I am guessing that you are hinting that in a layout, I can just use the `v1/layout` properties to get a layer? Or even a whole image?

What about for just a file?

deitch:
As for the progress, I really would like to figure out a way to get `WithProgress` to report writing by blob and manifest (similar to `docker push`). I have had a number of cases where the write hangs, or sometimes returns an error, when I am writing an index with 3 manifests underneath it, each of which may have several decent-sized layers. I have no idea which one is causing the error. I imagine reporting by blob would help.

PR #971, opened 2021-03-20
crane digest: fallback to GET when HEAD fails

ImJasonH:
### Before

```
$ crane digest registry.access.redhat.com/ubi8/ubi-minimal
2021/03/20 10:19:50 No matching credentials were found for "registry.access.redhat.com/ubi8/ubi-minimal", falling back on anonymous
2021/03/20 10:19:51 computing digest: too many parts in hash: 
```

_My, what a very unhelpful error message that is!_ 🤔 

### After

```
$ go run ./cmd/crane digest registry.access.redhat.com/ubi8/ubi-minimal
2021/03/20 10:20:10 No matching credentials were found for "registry.access.redhat.com/ubi8/ubi-minimal", falling back on anonymous
2021/03/20 10:20:10 HEAD request failed, falling back on GET: response did not include Docker-Content-Digest header
2021/03/20 10:20:10 No matching credentials were found for "registry.access.redhat.com/ubi8/ubi-minimal", falling back on anonymous
sha256:fdfb0770bff33e0f97d78583efd68b546a19d0a4b0ac23eef25ef261bca3e975
```

This should also make `crane digest` work against registries that don't support `HEAD` at all, or that don't include all the necessary descriptor information in response headers, though neglecting `Docker-Content-Digest` is probably by far the most common omission.

If this lgty in general I can add some tests.

### DIFF
--- a/pkg/crane/digest.go
+++ b/pkg/crane/digest.go
@@ -14,6 +14,8 @@
 
 package crane
 
+import "github.com/google/go-containerregistry/pkg/logs"
+
 // Digest returns the sha256 hash of the remote image at ref.
 func Digest(ref string, opt ...Option) (string, error) {
 	o := makeOptions(opt...)
@@ -39,7 +41,12 @@ func Digest(ref string, opt ...Option) (string, error) {
 	}
 	desc, err := head(ref, opt...)
 	if err != nil {
-		return "", err
+		logs.Warn.Printf("HEAD request failed, falling back on GET: %v", err)
+		rdesc, err := getManifest(ref, opt...)
+		if err != nil {
+			return "", err
+		}
+		return rdesc.Digest.String(), nil
 	}
 	return desc.Digest.String(), nil
 }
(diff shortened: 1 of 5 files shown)


jonjohnsonjr:
LGTM in general. One thing I'm slightly worried about is the headManifest errors not making sense out of context. WDYT about prepending them with `HEAD <url>: `?

ImJasonH:
> LGTM in general. One thing I'm slightly worried about is the headManifest errors not making sense out of context. WDYT about prepending them with `HEAD <url>: `?

Good call, done.

Added tests.

jonjohnsonjr:
One final nit is that the reason this error was particularly confusing is here: https://github.com/google/go-containerregistry/blob/4a92f6cd028daf021b9f29b0b2bc9f10b422e2fb/pkg/v1/hash.go#L90

In particular, "too many" is definitely wrong for `NewHash("")`.

Can we make that slightly better?

ImJasonH:
> Can we make that slightly better?

Updated to `return fmt.Errorf("cannot parse hash: %q", unquoted)`

jonjohnsonjr (approved):

PR #972, opened 2021-03-21
Allow progress also when saving to docker daemon

afbjorklund:
The initial implementation only added progress to tarball

Fortunately the daemon Write uses a temporary tarball file

### DIFF
--- a/pkg/v1/daemon/write.go
+++ b/pkg/v1/daemon/write.go
@@ -54,7 +54,7 @@ func Tag(src, dest name.Tag) error {
 }
 
 // Write saves the image into the daemon as the given tag.
-func Write(tag name.Tag, img v1.Image) (string, error) {
+func Write(tag name.Tag, img v1.Image, opts ...tarball.WriteOption) (string, error) {
 	cli, err := GetImageLoader()
 	if err != nil {
 		return "", err
@@ -62,7 +62,7 @@ func Write(tag name.Tag, img v1.Image) (string, error) {
 
 	pr, pw := io.Pipe()
 	go func() {
-		pw.CloseWithError(tarball.Write(tag, img, pw))
+		pw.CloseWithError(tarball.Write(tag, img, pw, opts...))
 	}()
 
 	// write the image in docker save format first, then load it


pkg/v1/daemon/write.go
```diff
@@ -54,15 +54,15 @@ func Tag(src, dest name.Tag) error {
 }
 
 // Write saves the image into the daemon as the given tag.
-func Write(tag name.Tag, img v1.Image) (string, error) {
+func Write(tag name.Tag, img v1.Image, opts ...tarball.WriteOption) (string, error) {
```
ImJasonH, line 57: It's a bit weird for `daemon.Write` to take a `tarball.Options`, it leaks a bit of the underlying abstraction.

Would it make sense to define a daemon Option for progress and convert it internally into the tarball Option?

afbjorklund: I think that remote.Write has the same problem, due to WriteOptions being added on the wrong level

afbjorklund: But the right thing to do would probably be to add a daemon.WithProgress, instead of tarball.WithProgress

ImJasonH: Yeah exactly. https://github.com/google/go-containerregistry/pull/967 defines a new `remote.WithProgress`. In the case of `daemon`, it might end up using `tarball.WithProgress` underneath, but it feels weird to have that leak into its API.

There isn't a `daemon.WriteOption` type yet, but I think we could add one, and a `daemon.WithProgress` that implements it, that converts to a `tarball.WriteOption` internally.

@jonjohnsonjr wdyt?

afbjorklund: Otherwise it seems to be working, of course we might end up using tarball anyway for Podman support.

afbjorklund:
Unfortunately using the progress seems to have broken the reference, so now the digest is "gone" again...

```
$ docker images --format "{{.Repository}}:{{.Tag}}@{{.Digest}}" | grep kicbase
gcr.io/k8s-minikube/kicbase:v0.0.18@<none>
```

We had patched go-containerregistry to add the digest, but it seems to be conflicting with this new feature.

afbjorklund:
It seems like this internal "temporary archive" also breaks the digest, which is reset to `nil`...

This means that `daemon.Write` behaves differently from `docker pull` (which does update it)

pkg/v1/daemon/write.go
jonjohnsonjr: I've stomped on this PR a little bit by [making everything just `daemon.Option`](https://github.com/google/go-containerregistry/pull/989) (sorry), but it shouldn't be too hard to do this now by just translating the `daemon.WithProgress` into a `tarball.WithProgress`.

afbjorklund: I had to fork everything anyway, to make a `podman` matching the `daemon` in the API.

PR #975, opened 2021-03-28
allow multiple images pull

michelpromonet:
Hi,

Please find this PR that intend to pull multiple images and save it to a tar file that could be loaded by docker load.

It is a quick & dirty try, let me know if you think this feature extension is interesting or if there is an other way to do this with the cli.

Best Regards,
Michel.

### DIFF
--- a/pkg/crane/pull.go
+++ b/pkg/crane/pull.go
@@ -44,27 +44,31 @@ func Pull(src string, opt ...Option) (v1.Image, error) {
 }
 
 // Save writes the v1.Image img as a tarball at path with tag src.
-func Save(img v1.Image, src, path string) error {
-	ref, err := name.ParseReference(src)
-	if err != nil {
-		return fmt.Errorf("parsing ref %q: %v", src, err)
-	}
+func Save(imgMap map[string]v1.Image, path string) error {
+	tagToImage := map[name.Tag]v1.Image{}
 
-	// WriteToFile wants a tag to write to the tarball, but we might have
-	// been given a digest.
-	// If the original ref was a tag, use that. Otherwise, if it was a
-	// digest, tag the image with :i-was-a-digest instead.
-	tag, ok := ref.(name.Tag)
-	if !ok {
-		d, ok := ref.(name.Digest)
-		if !ok {
-			return fmt.Errorf("ref wasn't a tag or digest")
+	for src, img := range imgMap {
+		ref, err := name.ParseReference(src)
+		if err != nil {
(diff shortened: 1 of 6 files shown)


michelpromonet:
@googlebot I signed it!

jonjohnsonjr:
Thanks for the PR!

> It is a quick & dirty try, let me know if you think this feature extension is interesting or if there is an other way to do this with the cli.

Definitely interesting and something I'd like to support. CLI UX looks good to me, but I'm a little reluctant to modify the signature for `crane.Save` etc.

Do you mind instead adding three new functions?

* `crane.MultiSave`
* `crane.MultiSaveLegacy`
* `crane.MultiSaveOCI`

Then we can rewrite `Save{Legacy,OCI}` as trivial implementations that just call into `MultiSave...` without breaking anything that depends on the `Save` functions today.

pkg/crane/pull.go
```diff
@@ -91,18 +110,31 @@ func SaveLegacy(img v1.Image, src, path string) error {
...
+	var err error
+	var p layout.Path
+	for _, img := range imgMap {
+		p, err = layout.FromPath(path)
+		if err != nil {
+			p, err = layout.Write(path, empty.Index)
+			if err != nil {
+				return err
+			}
+		}
```
jonjohnsonjr, line 133: We should be able to just do this once outside of the loop.

pkg/crane/pull.go
```diff
@@ -91,18 +110,31 @@ func SaveLegacy(img v1.Image, src, path string) error {
...
+		err = p.AppendImage(img)
```
jonjohnsonjr, line 134: I'd just:

```
if err := p.AppendImage(img); err != nil {
  return err
}
```

pkg/crane/pull.go
```diff
@@ -91,18 +110,31 @@ func SaveLegacy(img v1.Image, src, path string) error {
...
 		if err != nil {
 			return err
 		}
 	}
-	return p.AppendImage(img)
+	return err
```
jonjohnsonjr, line 139: With above suggestions, if we make it out of the loop, just `return nil`.

jonjohnsonjr (commented):
LGTM beyond a couple small nits.

Can you also add godoc comments to the Multi funcs?

michelpromonet:
Hi jonjohnsonjr,

Please find update adding comment on MultiWrite method and rework on MultiWriteOCI.

Best Regards,
Michel.

jonjohnsonjr (approved):
LGTM, thanks!

PR #979, opened 2021-04-08
Allow localhost repositories

### LINKED ISSUE #840: Unable to use "crane ls" with registry with port
```
$ crane ls localhost:5000/foo
2020/11/23 17:52:45 reading tags for localhost:5000: parsing repo "localhost:5000": repository can only contain the runes `abcdefghijklmnopqrstuvwxyz0123456789_-./`: localhost:5000
```

Related to #762

coryrc:
`crane ls localhost:5000/something`
worked for me after this change. Also a `crane cp`.

Fixes #840

### DIFF
--- a/pkg/name/repository.go
+++ b/pkg/name/repository.go
@@ -21,7 +21,7 @@ import (
 
 const (
 	defaultNamespace = "library"
-	repositoryChars  = "abcdefghijklmnopqrstuvwxyz0123456789_-./"
+	repositoryChars  = "abcdefghijklmnopqrstuvwxyz0123456789_-./:"
 	regRepoDelimiter = "/"
 )
 
--- a/pkg/name/registry.go
+++ b/pkg/name/registry.go
@@ -106,6 +106,7 @@ func checkRegistry(name string) error {
 // NewRegistry returns a Registry based on the given name.
 // Strict validation requires explicit, valid RFC 3986 URI authorities to be given.
 func NewRegistry(name string, opts ...Option) (Registry, error) {
+	// Name is like "", "docker.io", or "localhost:5000"
 	opt := makeOptions(opts...)
 	if opt.strict && len(name) == 0 {
 		return Registry{}, NewErrBadName("strict validation requires the registry to be explicitly defined")


coryrc:
@hasheddan was pinged in 840 so pinging him here

jonjohnsonjr:
This scares me a little bit but hopefully our tests are sufficient to make this change safely @dekkagaijin

jonjohnsonjr:
Okay excellent, the tests did fail.

coryrc:
```
-- FAIL: TestRawManifestDigests (0.02s)
    --- FAIL: TestRawManifestDigests/normal_pull,_by_digest (0.00s)
        image_test.go:166: Unexpected path: /v2/foo/bar:sha256/manifests/b1f8e8e133091bc49ac82bc634314b6ea0883f9d21da0a2bf1d80b31cc6753e4
        image_test.go:189: RawManifest() wrong error: true, want false: Get "http://127.0.0.1:32919/v2/foo/bar:sha256/manifests/b1f8e8e133091bc49ac82bc634314b6ea0883f9d21da0a2bf1d80b31cc6753e4": EOF
```

So just taking the first error above, when I run it w/o my change:
The "Unexpected Path" is `/v2/foo/bar/manifests/sha256:082e0f844d18b00750c38f2d1203dfbdb435eb91a2d92d43ee56f83fa
6ccb739`

so "sha256:" is inserted in the wrong place.

coryrc:
which comes from `manifestPath := fmt.Sprintf("/v2/%s/manifests/%s", expectedRepo, tc.ref)`

so the colon got into `expectedRepo` instead of `tc.ref`, so I will continue looking

pkg/name/repository.go
```diff
@@ -21,7 +21,7 @@ import (
 
 const (
 	defaultNamespace = "library"
-	repositoryChars  = "abcdefghijklmnopqrstuvwxyz0123456789_-./"
+	repositoryChars  = "abcdefghijklmnopqrstuvwxyz0123456789_-./:"
```
hasheddan, line 24: We want the registry to allow `:` not the repository right?

coryrc: Registry delegates the decision making on validity of the "repository" to Repository

jonjohnsonjr:
> `crane ls localhost:5000/something`

This works already, but:

```
crane ls localhost:5000
```

does not, right?

coryrc:
> > `crane ls localhost:5000/something`
> 
> This works already, but:
> 
> ```
> crane ls localhost:5000
> ```
> 
> does not, right?

Correct, but neither does `crane ls gcr.io`

coryrc:
There's zero authoritative way to decide if `crane ls ubuntu:5000` means access port 5000 of host ubuntu or list tag 5000 of the ubuntu image on docker.

I could make a special case for localhost?

jonjohnsonjr:
I don't think that's actually valid, per the distribution spec. Do other tools work with this?

coryrc:
The [spec](https://github.com/opencontainers/distribution-spec/blob/main/spec.md) only seems to care about the HTTP request and not what we do on the command line, but I'm probably misunderstanding your statement.

jonjohnsonjr:
So we need to translate:

```
crane ls localhost:5000
```

Into some http request to list tags (or anything). That http request looks like this:

```
GET /v2/<name>/tags/list
```

And `<name>` is defined as this regex:

```
[a-z0-9]+([._-][a-z0-9]+)*(/[a-z0-9]+([._-][a-z0-9]+)*)*
```

I guess I'm wondering if you can just do something like:

```
docker push localhost:5000:v3
```

And have that work? That would be somewhat equivalent to `crane ls localhost:5000`.

In theory, you may want to pull the tags that come out of that command, but how?

```
docker run localhost:5000:v3
```

Does this work for anything?

coryrc:
`docker run localhost:5000:v3` doesn't make any sense to me; it'd be equivalent to `gcr.io:v3`.

`crane catalog localhost:5000` works though

coryrc:
(btw I'm still tracking down the bug and adding in some unit tests, I think I've isolated the problem to `name.ParseReference()`)

jonjohnsonjr:
I think I understand the confusion -- this isn't about listing tags, you're expected `crane ls` to also list repositories, similarly to how `gcrane ls` works (or even normal `ls`).

We could print a warning like:

```
fmt.Printf("did you mean: crane catalog %s", src)
```

Or even just default to running `crane catalog` if we can't parse the input as a repository but _can_ parse the input as a registry?

jonjohnsonjr:
I've also run into a similar situation where I do something like:

```
crane ls example.com/repo:tag
```

Where I obviously mean to do:

```
crane ls example.com/repo
```

But I'm not sure how many of these "do what I mean" niceties we could add before things get unwieldy...

coryrc:
> I've also run into a similar situation where I do something like:
> 
> ```
> crane ls example.com/repo:tag
> ```
> 
> Where I obviously mean to do:
> 
> ```
> crane ls example.com/repo
> ```
> 
> But I'm not sure how many of these "do what I mean" niceties we could add before things get unwieldy...

I assumed `crane ls example.com/repo:tag` works because you could use it as a "exists" check; I wouldn't want to change it.

coryrc:
> I think I understand the confusion -- this isn't about listing tags, you're expecting `crane ls` to also list repositories, similarly to how `gcrane ls` works (or even normal `ls`).
> 
> We could print a warning like:
> 
> ```
> fmt.Printf("did you mean: crane catalog %s", src)
> ```
> 
> Or even just default to running `crane catalog` if we can't parse the input as a repository but _can_ parse the input as a registry?

I'm really just trying to use crane as a library and copy from a registry server on `localhost:5000`, I was just demonstrating I did at least some basic manual tests before submitting :). I don't have feelings one way or the other on how the CLI works.

jonjohnsonjr:
> I assumed crane ls example.com/repo:tag works because you could use it as a "exists" check; I wouldn't want to change it.

It actually fails to parse:

```
$ crane ls example.com/repo:tag
2021/04/08 12:52:22 reading tags for example.com/repo:tag: parsing repo "example.com/repo:tag": repository can only contain the runes `abcdefghijklmnopqrstuvwxyz0123456789_-./`: repo:tag
```

You'd want to use `crane manifest` or `crane digest` for an existence check.

> I'm really just trying to use crane as a library and copy from a registry server on `localhost:5000`

So unfortunately there's not really a standard way to enumerate repositories and images in a registry. Catalog + tag listing is one way that works for some registries, but only works for tags. GCR does its own thing, which `gcrane` exposes.

I've tried to push on a standard for this, but it seems like a lost cause: https://github.com/opencontainers/distribution-spec/issues/222#issuecomment-769508237

I _think_ what you want to do, for copying from some registry that supports catalog (don't do this for GCR), is roughly this:

```go
srcRegistry := "localhost:5000"
dstRegistry: "localhost:5001"

catalog, err := crane.Catalog(srcRegistry)
if err != nil {
  return err
}

for _, repo := range catalog {
  tags, err := crane.ListTags(repo)
  if err != nil {
     return err
  }
  for _, tag := range tags {
    src := fmt.Sprintf("%s/%s:%s", srcRegistry, repo, tag)
    dst := fmt.Sprintf("%s/%s:%s", dstRegistry, repo, tag)
    if err := crane.Copy(src, dst); err != nil {
      return err
    }
  }
}
```

The `crane.Catalog` response may or may not include the hostname, so you might have to munge some strings a bit...

This is the best you can do in a "generic" way, unfortunately, and a lot of registries don't support catalog.

coryrc:
Oh, I did get confused and the copy worked and when I went to verify with the CLI it broke and I wandered off without realizing the copy was successful. D'oh! I can still finish this though, I think I've found the bug and am building up the test cases to prove it.

coryrc:
(I still think there's a different subtle bug I'm working on fixing re: that test case, unless I misunderstood the format)

jonjohnsonjr:
> I still think there's a different subtle bug I'm working on fixing re: that test case, unless I misunderstood the format

FWIW you are probably right, and I'd like to rework this package entirely, but haven't had the time... though as long as you're not blocked, I'm happy.

PR #980, opened 2021-04-12
Add remote.Digest() to upload a manifest via digest

deitch:
As discussed with @jonjohnsonjr . We already have [remote.Tag](https://pkg.go.dev/github.com/google/go-containerregistry/pkg/v1/remote#Tag) which uploads a manifest and references it with a tag. However, there also is the ability in the API to upload a manifest and just reference it via its digest, just like any other blob. Eventually, the manifest should be referenced by another manifest or a tag, but those are independent.

### DIFF
--- a/pkg/v1/remote/write.go
+++ b/pkg/v1/remote/write.go
@@ -849,11 +849,26 @@ func WriteLayer(repo name.Repository, layer v1.Layer, options ...Option) (rerr e
 // should ensure that all blobs or manifests that are referenced by t exist
 // in the target registry.
 func Tag(tag name.Tag, t Taggable, options ...Option) error {
-	o, err := makeOptions(tag.Context(), options...)
+	return Put(tag, t, options...)
+}
+
+// Put adds a manifest from the given Taggable via PUT /v1/.../manifest/<ref>
+//
+// Notable implementations of Taggable are v1.Image, v1.ImageIndex, and
+// remote.Descriptor.
+//
+// If t implements MediaType, we will use that for the Content-Type, otherwise
+// we will default to types.DockerManifestSchema2.
+//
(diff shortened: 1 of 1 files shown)


pkg/v1/remote/write.go
```diff
@@ -849,27 +849,48 @@ func WriteLayer(repo name.Repository, layer v1.Layer, options ...Option) (rerr e
...
+// Notable implementations of Taggable are v1.Image, v1.ImageIndex, and
+// remote.Descriptor.
+//
+// If t implements MediaType, we will use that for the Content-Type, otherwise
+// we will default to types.DockerManifestSchema2.
+//
+// Digest does not attempt to write anything other than the manifest, so callers
+// should ensure that all blobs or manifests that are referenced by t exist
+// in the target registry.
+func Digest(digest name.Digest, t Taggable, options ...Option) error {
```
jonjohnsonjr, line 866: I'd lean towards just `remote.Put` for this, I think. The existing `remote.Tag` is using "tag" as a verb here, not as a noun, so calling this `Digest` doesn't make a ton of sense.

`remote.Put` would also mirror `remote.Get`, `remote.Head`, and `remote.Delete`.

It should take a `name.Reference` as well, IMO (basically just exposing `manifestReference` as `Put`).

jonjohnsonjr (changes requested):

pkg/v1/remote/write.go
deitch: Done. Take a look.

jonjohnsonjr (approved):

jonjohnsonjr:
LGTM, a test would be cool, but just rebasing also works :)

deitch:
Rebased.

jonjohnsonjr:
Thanks!

PR #987, opened 2021-04-20
Add support for symbol link inside docker tar file

yongxiu:
docker tar file containing multiple images will have symbol link for layer files to save space, this will read from the target file instead of the symbol link file


pkg/v1/tarball/image.go
```diff
@@ -216,6 +218,10 @@ func extractFileFromTar(opener Opener, filePath string) (io.ReadCloser, error) {
 			return nil, err
 		}
 		if hdr.Name == filePath {
+			if hdr.Typeflag == tar.TypeSymlink || hdr.Typeflag == tar.TypeLink {
+				currentDir := filepath.Dir(filePath)
```
jonjohnsonjr, line 222: I have one concern about using `filepath.Dir` here. I'm not familiar with e.g. windows, but would this work when running on a windows machine if the tarball has normal path separators?

I think it's fine based on [this](https://golang.org/src/os/path_windows.go#L12) but I'd love some confirmation.

jonjohnsonjr (commented):
Overall lgtm other than that one question. Can you add a test as well?

There are some test tarball archives under https://github.com/google/go-containerregistry/tree/main/pkg/v1/tarball/testdata

pkg/v1/tarball/image.go
yongxiu: I just tested in Windows, works same.

yongxiu:
> Overall lgtm other than that one question. Can you add a test as well?
> 
> There are some test tarball archives under https://github.com/google/go-containerregistry/tree/main/pkg/v1/tarball/testdata

Added, please take another look, thanks.

jonjohnsonjr (approved):

jonjohnsonjr:
Thanks!

PR #988, opened 2021-04-20
Use cobra's RunE whenever possible

ImJasonH:


cmd/crane/cmd/append.go
```diff
@@ -33,35 +33,35 @@ func NewCmdAppend(options *[]crane.Option) *cobra.Command {
...
-					log.Fatalf("writing output %q: %v", outFile, err)
-				}
-			} else {
-				if err := crane.Push(img, newTag, *options...); err != nil {
-					log.Fatalf("pushing image %s: %v", newTag, err)
+					return fmt.Errorf("writing output %q: %v", outFile, err)
 				}
 			}
+
+			if err := crane.Push(img, newTag, *options...); err != nil {
```
jonjohnsonjr, line 61: This is a behavioral change.

Before, `-o` would only write to disk, not push. Now we always push.

cmd/crane/cmd/delete.go
```diff
@@ -27,11 +27,12 @@ func NewCmdDelete(options *[]crane.Option) *cobra.Command {
 		Use:   "delete IMAGE",
 		Short: "Delete an image reference from its registry",
 		Args:  cobra.ExactArgs(1),
-		Run: func(_ *cobra.Command, args []string) {
+		RunE: func(_ *cobra.Command, args []string) error {
 			ref := args[0]
 			if err := crane.Delete(ref, *options...); err != nil {
-				log.Fatalf("deleting %s: %v", ref, err)
+				return fmt.Errorf("deleting %s: %v", ref, err)
```
jonjohnsonjr, line 33: I'd drop this to be consistent with other commands, maybe?

cmd/crane/cmd/export.go
```diff
@@ -33,23 +33,24 @@ func NewCmdExport(options *[]crane.Option) *cobra.Command {
...
 
 			img, err := crane.Pull(src, *options...)
 			if err != nil {
-				log.Fatal(err)
+				return err
 			}
 
 			if err := crane.Export(img, f); err != nil {
-				log.Fatalf("exporting %s: %v", src, err)
+				return fmt.Errorf("exporting %s: %v", src, err)
```
jonjohnsonjr, line 51: Maybe wrap the pull err instead of the export.

cmd/crane/cmd/pull.go
```diff
@@ -32,13 +31,13 @@ func NewCmdPull(options *[]crane.Option) *cobra.Command {
...
 		Args:  cobra.MinimumNArgs(2),
-		Run: func(_ *cobra.Command, args []string) {
+		RunE: func(_ *cobra.Command, args []string) error {
 			imageMap := map[string]v1.Image{}
 			srcList, path := args[:len(args)-1], args[len(args)-1]
 			for _, src := range srcList {
 				img, err := crane.Pull(src, *options...)
 				if err != nil {
-					log.Fatal(err)
+					return err
```
jonjohnsonjr, line 40: Maybe wrap this

cmd/crane/cmd/push.go
```diff
@@ -27,16 +27,17 @@ func NewCmdPush(options *[]crane.Option) *cobra.Command {
...
 			path, tag := args[0], args[1]
 			img, err := crane.Load(path)
 			if err != nil {
-				log.Fatalf("loading %s as tarball: %v", path, err)
+				return fmt.Errorf("loading %s as tarball: %v", path, err)
 			}
 
 			if err := crane.Push(img, tag, *options...); err != nil {
-				log.Fatalf("pushing %s: %v", tag, err)
+				return fmt.Errorf("pushing %s: %v", tag, err)
```
jonjohnsonjr, line 38: drop this wrapping

cmd/crane/cmd/validate.go
```diff
@@ -33,7 +32,7 @@ func NewCmdValidate(options *[]crane.Option) *cobra.Command {
 		Use:   "validate",
 		Short: "Validate that an image is well-formed",
 		Args:  cobra.ExactArgs(0),
-		Run: func(_ *cobra.Command, args []string) {
+		RunE: func(_ *cobra.Command, args []string) error {
```
jonjohnsonjr, line 35: I think I want to print FAIL for all errors but exit 1 if any of them fail: maybe collect which formats failed and return that as an error at the end of the loop?

cmd/crane/cmd/append.go
ImJasonH: Good catch, thanks.

cmd/crane/cmd/validate.go
ImJasonH: Can we do this in another PR? I was a bit surprised that this accepted both `--tarball` and `--remote` in the same invocation.

jonjohnsonjr (approved):

PR #990, opened 2021-04-22
Tag crane/gcrane images with Git tag

### LINKED ISSUE #755: Add tagged versions to gcr.io
Currently, there is only latest tag available on gcr.io registry for crane. Can you add tagged versions to gcr.io for crane ? 

It would be great to retrieve docker image with a fixed version for more determinism.

Thanks

ImJasonH:
Fixes #755 

When the tag name is set, this will produce images like:

- `gcr.io/go-containerregistry/crane:vX.Y.Z`
- `gcr.io/go-containerregistry/crane:debug-vX.Y.Z`

When the tag name is not set (building an untagged commit), it will produce images as normal (e.g., `gcr.io/go-containerregistry/crane:abcdef` and `:latest`).

Also update the `ko` version used to build images, and fail if using an unset env var.

### DIFF
--- a/cloudbuild.yaml
+++ b/cloudbuild.yaml
@@ -4,8 +4,9 @@ steps:
   args:
   - -c
   - |
-    set -ex
+    set -eux
 
+    export GOROOT=/usr/local/go
     export GO111MODULE=off
     export KO_DOCKER_REPO="gcr.io/$PROJECT_ID"
     export GOFLAGS="-ldflags=-X=github.com/google/go-containerregistry/cmd/crane/cmd.Version=$COMMIT_SHA"
@@ -17,19 +18,23 @@ steps:
     ln -s $$PWD $$shadow || stat $$shadow
 
     # Install ko from release.
-    curl -L -o ko.tar.gz https://github.com/google/ko/releases/download/v0.6.0/ko_0.6.0_Linux_i386.tar.gz
+    curl -L -o ko.tar.gz https://github.com/google/ko/releases/download/v0.8.2/ko_0.8.2_Linux_i386.tar.gz
     tar xvfz ko.tar.gz
     chmod +x ko
     alias ko=./ko
 
     # Use the ko binary to build the crane and gcrane builder images.
-    ko publish --platform=all -B github.com/google/go-containerregistry/cmd/crane -t latest -t "$COMMIT_SHA"
-    ko publish --platform=all -B github.com/google/go-containerregistry/cmd/gcrane -t latest -t "$COMMIT_SHA"
+    ko publish --platform=all -B github.com/google/go-containerregistry/cmd/crane  -t latest -t "$COMMIT_SHA" -t "$TAG_NAME"
+    ko publish --platform=all -B github.com/google/go-containerregistry/cmd/gcrane -t latest -t "$COMMIT_SHA" -t "$TAG_NAME"
 
     # Use the ko binary to build the crane and gcrane builder *debug* images.
-    export KO_CONFIG_PATH=/workspace/.ko/debug/
-    ko publish --platform=all -B github.com/google/go-containerregistry/cmd/crane -t debug
-    ko publish --platform=all -B github.com/google/go-containerregistry/cmd/gcrane -t debug
+    export KO_CONFIG_PATH=./.ko/debug/
+    ko publish --platform=all -B github.com/google/go-containerregistry/cmd/crane  -t "debug"
+    ko publish --platform=all -B github.com/google/go-containerregistry/cmd/gcrane -t "debug"
+
+    # Tag-specific debug images are pushed to gcr.io/go-containerregistry/{g}crane/debug:...
+    KO_DOCKER_REPO=gcr.io/$PROJECT_ID/crane/debug  ko publish --platform=all --bare github.com/google/go-containerregistry/cmd/crane  -t latest -t "$COMMIT_SHA" -t "$TAG_NAME"
+    KO_DOCKER_REPO=gcr.io/$PROJECT_ID/gcrane/debug ko publish --platform=all --bare github.com/google/go-containerregistry/cmd/gcrane -t latest -t "$COMMIT_SHA" -t "$TAG_NAME"
 
 # Use the crane builder to get the digest for crane and gcrane.
 - name: gcr.io/$PROJECT_ID/crane
@@ -37,3 +42,4 @@ steps:
 
 - name: gcr.io/$PROJECT_ID/crane
   args: ['digest', 'gcr.io/$PROJECT_ID/gcrane']
+
--- a/cmd/crane/README.md
+++ b/cmd/crane/README.md
@@ -49,11 +49,13 @@ $ docker run --rm gcr.io/go-containerregistry/crane ls ubuntu
 12.10
 ```
 
-And it's also available with a shell, which uses the `debug` tag
+And it's also available with a shell, at the `:debug` tag:
 
 ```sh
 docker run --rm -it --entrypoint "/busybox/sh" gcr.io/go-containerregistry/crane:debug
 ```
+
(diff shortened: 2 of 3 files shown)


cloudbuild.yaml
```diff
@@ -17,23 +17,28 @@ steps:
...
     # Install ko from release.
-    curl -L -o ko.tar.gz https://github.com/google/ko/releases/download/v0.6.0/ko_0.6.0_Linux_i386.tar.gz
+    curl -L -o ko.tar.gz https://github.com/google/ko/releases/download/v0.8.2/ko_0.8.2_Linux_i386.tar.gz
     tar xvfz ko.tar.gz
     chmod +x ko
     alias ko=./ko
 
+    debug_tag=""
+    if [ "${TAG_NAME}" != "" ]; then
+      debug_tag=debug-${TAG_NAME}
```
jonjohnsonjr, line 27: I think `${TAG_NAME}-debug` sorts a bit better and matches what `kaniko` is doing now.

cloudbuild.yaml
```diff
@@ -17,23 +17,28 @@ steps:
...
-    ko publish --platform=all -B github.com/google/go-containerregistry/cmd/crane -t latest -t "$COMMIT_SHA"
-    ko publish --platform=all -B github.com/google/go-containerregistry/cmd/gcrane -t latest -t "$COMMIT_SHA"
+    ko publish --platform=all -B github.com/google/go-containerregistry/cmd/crane -t latest -t "$COMMIT_SHA" -t "$TAG_NAME"
+    ko publish --platform=all -B github.com/google/go-containerregistry/cmd/gcrane -t latest -t "$COMMIT_SHA" -t "$TAG_NAME"
 
     # Use the ko binary to build the crane and gcrane builder *debug* images.
-    export KO_CONFIG_PATH=/workspace/.ko/debug/
-    ko publish --platform=all -B github.com/google/go-containerregistry/cmd/crane -t debug
-    ko publish --platform=all -B github.com/google/go-containerregistry/cmd/gcrane -t debug
+    KO_CONFIG_PATH=./.ko/debug/ ko publish --platform=all -B github.com/google/go-containerregistry/cmd/crane -t debug -t "$$debug_tag"
```
jonjohnsonjr, line 35: Instead of adding these new debug tags, can we instead add a new line:

```
KO_DOCKER_REPO=gcr.io/go-containerregistry/crane/debug KO_CONFIG_PATH=./.ko/debug/ ko publish --platform=all --bare github.com/google/go-containerregistry/cmd/crane -t latest -t "$COMMIT_SHA" -t "$TAG_NAME"
```
(same for gcrane)

We can keep `.../crane:debug` for backward compat, but these tags get super messy if you start having multiple dimensions within a single tag.

cloudbuild.yaml
jonjohnsonjr: See my comment below for an alternative that I prefer, though I'm open to being convinced otherwise.

ImJasonH: Agreed, done.

cloudbuild.yaml
```diff
@@ -17,23 +18,32 @@ steps:
...
-    ko publish --platform=all -B github.com/google/go-containerregistry/cmd/crane -t latest -t "$COMMIT_SHA"
-    ko publish --platform=all -B github.com/google/go-containerregistry/cmd/gcrane -t latest -t "$COMMIT_SHA"
+    ko publish --platform=all -B github.com/google/go-containerregistry/cmd/crane  -t latest -t "$COMMIT_SHA" -t "$TAG_NAME"
+    ko publish --platform=all -B github.com/google/go-containerregistry/cmd/gcrane -t latest -t "$COMMIT_SHA" -t "$TAG_NAME"
 
     # Use the ko binary to build the crane and gcrane builder *debug* images.
-    export KO_CONFIG_PATH=/workspace/.ko/debug/
-    ko publish --platform=all -B github.com/google/go-containerregistry/cmd/crane -t debug
-    ko publish --platform=all -B github.com/google/go-containerregistry/cmd/gcrane -t debug
+    # These are pushed to gcr.io/go-containerregistry/{g}crane/debug:...
```
jonjohnsonjr, line 36: I think we should continue to publish `gcr.io/go-containerregistry/[g]crane:debug` as well that mirrors `latest`.

cloudbuild.yaml
```diff
@@ -17,23 +18,33 @@ steps:
...
-    ko publish --platform=all -B github.com/google/go-containerregistry/cmd/gcrane -t latest -t "$COMMIT_SHA"
+    ko publish --platform=all -B github.com/google/go-containerregistry/cmd/crane  -t latest -t "$COMMIT_SHA" -t "$TAG_NAME"
+    ko publish --platform=all -B github.com/google/go-containerregistry/cmd/gcrane -t latest -t "$COMMIT_SHA" -t "$TAG_NAME"
 
     # Use the ko binary to build the crane and gcrane builder *debug* images.
-    export KO_CONFIG_PATH=/workspace/.ko/debug/
-    ko publish --platform=all -B github.com/google/go-containerregistry/cmd/crane -t debug
-    ko publish --platform=all -B github.com/google/go-containerregistry/cmd/gcrane -t debug
+    export KO_CONFIG_PATH=./.ko/debug/
+    ko publish --platform=all -B github.com/google/go-containerregistry/cmd/crane  -t latest -t "$COMMIT_SHA" -t "$$debug_tag"
```
jonjohnsonjr, line 37: Should only have t-t debug` for these two lines.

cloudbuild.yaml
```diff
@@ -17,23 +18,33 @@ steps:
     ln -s $$PWD $$shadow || stat $$shadow
 
     # Install ko from release.
-    curl -L -o ko.tar.gz https://github.com/google/ko/releases/download/v0.6.0/ko_0.6.0_Linux_i386.tar.gz
+    curl -L -o ko.tar.gz https://github.com/google/ko/releases/download/v0.8.2/ko_0.8.2_Linux_i386.tar.gz
     tar xvfz ko.tar.gz
     chmod +x ko
     alias ko=./ko
 
+    debug_tag=""
```
jonjohnsonjr, line 26: We can drop `debug_tag` entirely.

cloudbuild.yaml
```diff
@@ -17,23 +18,28 @@ steps:
...
-    ko publish --platform=all -B github.com/google/go-containerregistry/cmd/gcrane -t latest -t "$COMMIT_SHA"
+    ko publish --platform=all -B github.com/google/go-containerregistry/cmd/crane  -t latest -t "$COMMIT_SHA" -t "$TAG_NAME"
+    ko publish --platform=all -B github.com/google/go-containerregistry/cmd/gcrane -t latest -t "$COMMIT_SHA" -t "$TAG_NAME"
 
     # Use the ko binary to build the crane and gcrane builder *debug* images.
-    export KO_CONFIG_PATH=/workspace/.ko/debug/
-    ko publish --platform=all -B github.com/google/go-containerregistry/cmd/crane -t debug
-    ko publish --platform=all -B github.com/google/go-containerregistry/cmd/gcrane -t debug
+    export KO_CONFIG_PATH=./.ko/debug/
+    ko publish --platform=all -B github.com/google/go-containerregistry/cmd/crane  -t latest -t "$COMMIT_SHA" -t "debug"
```
jonjohnsonjr, line 32: These two lines can't have `-t latest` and `-t "$COMMIT_SHA"` because they will overwrite the previous two lines with the debug versions.

cmd/crane/README.md
```diff
@@ -49,12 +49,14 @@ $ docker run --rm gcr.io/go-containerregistry/crane ls ubuntu
 12.10
 ```
 
-And it's also available with a shell, which uses the `debug` tag
+And it's also available with a shell, at with the `:debug` tag:
```
jonjohnsonjr, line 52: `at with`

jonjohnsonjr (approved):

max-wittig:
Thank you very much!

PR #994, opened 2021-04-23
Add presubmit workflow

jonjohnsonjr:
Fix all the things that it caught :(


.github/workflows/presubmit.yaml
```diff
@@ -0,0 +1,48 @@
...
+
+name: Presubmit
+
+on:
+
+  push:
+    branches: [ 'main', 'master' ]
+
+  pull_request:
+    branches: [ 'main', 'master', 'release-*' ]
```
ImJasonH, line 23: We don't actually need `master` and `release-*` right?

ImJasonH (approved):

.github/workflows/presubmit.yaml
jonjohnsonjr: If we need to patch release branches, we'll want that (I think).

All our other workflows work against `master`, and this has been useful for me. I keep my fork's default branch as `master` to easily (and lazily) differentiate it from this repo's `main` branch, so I don't accidentally push the wrong thing ever.

Issue #912, opened 2021-01-12
Plumb a signal.NotifyContext through crane/gcrane


jonjohnsonjr:
https://tip.golang.org/pkg/os/signal/#NotifyContext

I think we currently do nothing useful on Ctrl-C etc, which we should.

This requires go 1.16.

PR #995, opened 2021-04-23
Plumb context through crane and gcrane

jonjohnsonjr:
Fixes https://github.com/google/go-containerregistry/issues/912

Also, use signal.NotifyContext for cancelling the ctx.

Also, don't log errors twice now that we're using RunE.

### DIFF
--- a/cmd/gcrane/main.go
+++ b/cmd/gcrane/main.go
@@ -15,10 +15,12 @@
 package main
 
 import (
+	"context"
 	"os"
 
 	"github.com/google/go-containerregistry/cmd/crane/cmd"
 	gcmd "github.com/google/go-containerregistry/cmd/gcrane/cmd"
+	"github.com/google/go-containerregistry/internal/signal"
 	"github.com/google/go-containerregistry/pkg/crane"
 	"github.com/google/go-containerregistry/pkg/gcrane"
 	"github.com/google/go-containerregistry/pkg/logs"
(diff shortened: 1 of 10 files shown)


pkg/crane/catalog.go
```diff
@@ -29,5 +29,5 @@ func Catalog(src string, opt ...Option) (res []string, err error) {
 		return nil, err
 	}
 
-	return remote.Catalog(context.TODO(), reg, o.remote...)
+	return remote.Catalog(context.Background(), reg, o.remote...)
```
ImJasonH, line 32: Not plumbing context through here?

cmd/gcrane/main.go
```diff
@@ -61,8 +62,9 @@ func main() {
 		root.AddCommand(cmd)
 	}
 
-	if err := root.Execute(); err != nil {
-		fmt.Println(err)
+	ctx, cancel := signal.NotifyContext(context.Background(), os.Interrupt)
```
ImJasonH, line 65: defer cancel () instead?

jonjohnsonjr: Doesn't `os.Exit(1)` pre-empt it?

pkg/crane/catalog.go
jonjohnsonjr: The `remote.WithContext` in `o.remote...` will override it.

imjasonh (approved):

cmd/gcrane/main.go
jonjohnsonjr: ended up doing both instead

imjasonh (approved):

Issue #997, opened 2021-04-24
The checksum on some archives have changed


uhthomas:
Hi,

It looks like the SHA256 sums of some archives have changed.

See run fail: https://github.com/uhthomas/automata/runs/2427916713

Excerpt:

```
WARNING: Download from https://api.github.com/repos/google/go-containerregistry/tarball/8a2841911ffee4f6892ca0083e89752fb46c48dd failed: class com.google.devtools.build.lib.bazel.repository.downloader.UnrecoverableHttpException Checksum was cadb09cb5bcbe00688c73d716d1c9e774d6e4959abec4c425a1b995faf33e964 but wanted 60b9a600affa5667bd444019a4e218b7752d8500cfa923c1ac54ce2f88f773e2
```

This has broken projects like `rules_docker` and requires any upstreams to explicitly import and correct the package.

For example https://github.com/uhthomas/automata/commit/0d68e65e617740c10f321f1937201951b0169dd4.

As I understand, this is may be related to https://github.com/google/ko/issues/315, although I am not sure in this case.

afbjorklund:
The git attributes are causing a lot of breakage in downstream projects, we have also seen podman and cri-o

ImJasonH:
This has all the same hallmarks as https://github.com/google/ko/issues/315, I believe it's the same issue.

Changing the SHA seems to have fixed the issue (please correct me if I'm wrong), and while it shouldn't happen again AIUI we can ensure it doesn't by updating `rules_docker` to depend on a newer version of go-containerregistry that doesn't depend on the gitVersion shenanigans at all. I see [they're on 0.1.4](https://github.com/bazelbuild/rules_docker/blob/f929d80c5a4363994968248d87a892b1c2ef61d4/repositories/go_repositories.bzl#L40), which is pretty old.

ImJasonH:
https://github.com/bazelbuild/rules_docker/pull/1829

steeve:
Repeating what has been said elsewhere:
Github branch/commit tarballs are dynamically generated. Don't depend on them having a stable SHA256. If the upstream rule is not distributing built artefacts, you'll need to use git itself or something stabler.

afbjorklund:
Fortunately the git significant digits don't check _that_ often, so most just suffers the occasional breakage... :shrug: 

Nobody distributes distfiles anymore, and there are not many build systems supporting git bundles either.

uhthomas:
[...]

PR #1001, opened 2021-04-26
Update k8schain's k8s dependency to 0.21.0

imjasonh:
https://github.com/google/go-containerregistry/issues/997#issuecomment-827181676

I edited `go.mod` to bump the versions, then ran `go mod download` as recommended by `go build ./...`, then `go mod tidy`.


ImJasonH:
It looks like the nondeterminism is still present in v1.21.0 :( 

- https://github.com/kubernetes/kubernetes/blob/v1.21.0/staging/src/k8s.io/component-base/version/base.go#L59
- https://github.com/kubernetes/kubernetes/blob/v1.21.0/staging/src/k8s.io/client-go/pkg/version/base.go#L59

jonjohnsonjr:
Do we need to bump `github.com/vdemeester/k8s-pkg-credentialprovider`?

cc @vdemeester

vdemeester:
@jonjohnsonjr we probably should indeed

imjasonh:
> @jonjohnsonjr we probably should indeed

https://github.com/vdemeester/k8s-pkg-credentialprovider/pull/10

imjasonh:
vdemeester/k8s-pkg-credentialprovider#10 is updated

jonjohnsonjr:
Why do we need all the replace directives?

vdemeester:
> Why do we need all the replace directives?

Yeah agreed, I think we should try to get rid of the `replace` as much as possible.

vdemeester:
I think I need to fix something of k8s-pkg-credentialprovider though

vdemeester:
> I think I need to fix something of k8s-pkg-credentialprovider though

It is done, @imjasonh updated to 1.21.0-1 👼🏼

jonjohnsonjr (approved):

Issue #1006, opened 2021-04-28
schema 1 copies always use default keychain


jonjohnsonjr:
This is not ideal for `gcrane cp`, as schema 1 copies use different credentials than every other media type, which is very confusing.

PR #1007, opened 2021-04-28
Propagate remote options through schema 1 copies

jonjohnsonjr:
The schema 1 copying code was very hacked together, but remote.Put
simplifies things a bit. This extends some of the crane/gcrane options
down through to legacy.CopySchema1.

Fixes https://github.com/google/go-containerregistry/issues/1006

### DIFF
--- a/internal/legacy/copy.go
+++ b/internal/legacy/copy.go
@@ -18,20 +18,14 @@ package legacy
 import (
 	"bytes"
 	"encoding/json"
-	"fmt"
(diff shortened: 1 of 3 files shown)


jonjohnsonjr:
Thanks @deitch your change to add https://github.com/google/go-containerregistry/pull/980 helped a lot and got rid of a TODO 😄

imjasonh (approved):

PR #1008, opened 2021-04-28
Update goreleaser config to include arm version

tianon:
This is shamelessly yanked from the default `name_template:` value: https://goreleaser.com/customization/archive/ (which it appears was adjusted here just to remove project version from the filenames).

I decided not to include the 32bit MIPS bit too because honestly, I'm not sure why folks would be doing 32bit MIPS.

image

(This should result in filenames like `go-containerregistry_Linux_armv6.tar.gz` instead of `go-containerregistry_Linux_arm.tar.gz`, since `goarm:` upstream defaults to `[6]`.)

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -60,7 +60,7 @@ archives:
     windows: Windows
     386: i386
     amd64: x86_64
-  name_template: "{{ .ProjectName }}_{{ .Os }}_{{ .Arch }}"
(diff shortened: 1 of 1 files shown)


jonjohnsonjr (approved):

jonjohnsonjr:
Thanks!

tianon:
You know me, gotta stay On Brand :smile: :+1:

PR #1013, opened 2021-05-10
Add --fast flag to crane validate

### LINKED ISSUE #838: how do I check if a layout contains all of the blobs?

jonjohnsonjr:
Fixes #838 

Add validate.Fast option
Add Exists to remote layers
Add Exists to layout
Add Exists to partial

Simplify some of our typechecking logic with unwrap().

Stop double-wrapping layoutImage layers.

### DIFF
--- a/cmd/crane/cmd/validate.go
+++ b/cmd/crane/cmd/validate.go
@@ -26,7 +26,10 @@ import (
 
 // NewCmdValidate creates a new cobra.Command for the validate subcommand.
 func NewCmdValidate(options *[]crane.Option) *cobra.Command {
-	var tarballPath, remoteRef string
+	var (
+		tarballPath, remoteRef string
+		fast                   bool
+	)
 
 	validateCmd := &cobra.Command{
 		Use:   "validate",
@@ -45,7 +48,11 @@ func NewCmdValidate(options *[]crane.Option) *cobra.Command {
 					return fmt.Errorf("failed to read image %s: %v", flag, err)
 				}
 
-				if err := validate.Image(img); err != nil {
+				opt := []validate.Option{}
+				if fast {
+					opt = append(opt, validate.Fast)
+				}
+				if err := validate.Image(img, opt...); err != nil {
 					fmt.Printf("FAIL: %s: %v\n", flag, err)
 					return err
 				} else {
@@ -57,6 +64,7 @@ func NewCmdValidate(options *[]crane.Option) *cobra.Command {
 	}
 	validateCmd.Flags().StringVar(&tarballPath, "tarball", "", "Path to tarball to validate")
 	validateCmd.Flags().StringVar(&remoteRef, "remote", "", "Name of remote image to validate")
+	validateCmd.Flags().BoolVar(&fast, "fast", false, "Skip downloading/digesting layers")
 
 	return validateCmd
 }
--- a/pkg/v1/remote/image.go
+++ b/pkg/v1/remote/image.go
@@ -223,12 +223,7 @@ func (rl *remoteImageLayer) Descriptor() (*v1.Descriptor, error) {
 
 // See partial.Exists.
 func (rl *remoteImageLayer) Exists() (bool, error) {
-	resp, err := rl.ri.headBlob(rl.digest)
-	if err != nil {
-		return false, err
-	}
-	defer resp.Body.Close()
-	return true, nil
+	return rl.ri.blobExists(rl.digest)
 }
 
 // LayerByDigest implements partial.CompressedLayer
--- a/pkg/v1/remote/image.go
+++ b/pkg/v1/remote/image.go
@@ -221,6 +221,16 @@ func (rl *remoteImageLayer) Descriptor() (*v1.Descriptor, error) {
 	return partial.BlobDescriptor(rl, rl.digest)
 }
 
+// See partial.Exists.
+func (rl *remoteImageLayer) Exists() (bool, error) {
+	resp, err := rl.ri.headBlob(rl.digest)
+	if err != nil {
+		return false, err
+	}
+	defer resp.Body.Close()
+	return true, nil
+}
+
 // LayerByDigest implements partial.CompressedLayer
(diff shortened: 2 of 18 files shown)


pkg/v1/validate/image.go
```diff
@@ -249,3 +256,18 @@ func validateManifest(img v1.Image) error {
 
 	return nil
 }
+
+func statLayers(layers []v1.Layer) error {
+	errs := []string{}
+	for _, layer := range layers {
+		if _, err := partial.Stat(layer); err != nil {
```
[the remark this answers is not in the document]
jonjohnsonjr: Good idea.

jonjohnsonjr:
I'm not thrilled about this still, but the `unwrap` thing makes me happy.

I almost want to implement this across the board -- something like:

```go
type Unwrapper interface {
  Unwrap() interface{}
}
```

This is kind of like [go 1.13 error stuff](https://golang.org/src/errors/wrap.go) -- maybe we'd want `partial.Is` and `partial.As` as well?

We could have `MountableLayer` implement this: https://github.com/google/go-containerregistry/blob/6a2e26179bccf5652d8570069779f2f1189d2bec/pkg/v1/remote/mount.go#L31-L35

Instead of adding pass-through methods like we're doing now. We're doing weird stuff to get Descriptor to work:

```
$ rg -g '!*_test.go' "Descriptor\(\)"
v1/tarball/layer.go
45:func (l *layer) Descriptor() (*v1.Descriptor, error) {

v1/tarball/image.go
251:func (fl *foreignUncompressedLayer) Descriptor() (*v1.Descriptor, error) {

v1/remote/mount.go
33:func (ml *MountableLayer) Descriptor() (*v1.Descriptor, error) {
93:func (mi *mountableImage) Descriptor() (*v1.Descriptor, error) {

v1/remote/index.go
110:func (r *remoteIndex) Descriptor() (*v1.Descriptor, error) {

v1/remote/image.go
118:func (r *remoteImage) Descriptor() (*v1.Descriptor, error) {
220:func (rl *remoteImageLayer) Descriptor() (*v1.Descriptor, error) {

v1/partial/with.go
284:    Descriptor() (*v1.Descriptor, error)
301:            return wd.Descriptor()
306:            return wd.Descriptor()
```

This might be a nice interim until we can rip out the partial stuff altogether?

pkg/v1/remote/image.go
```diff
@@ -221,6 +221,16 @@ func (rl *remoteImageLayer) Descriptor() (*v1.Descriptor, error) {
 	return partial.BlobDescriptor(rl, rl.digest)
 }
 
+// See partial.Exists.
+func (rl *remoteImageLayer) Exists() (bool, error) {
+	resp, err := rl.ri.headBlob(rl.digest)
```
imjasonh, line 226: If the method returns `(bool, error)` I would expect there to be some possibility of returning `false, nil` -- otherwise, the bool is just equivalent `err != nil`, and doesn't provide any other information.

Could this `return false, nil` if `headBlob` returns a 404? (Indicating that it _successfully_ determined that the blob _doesn't exist_)

imjasonh (approved):

cmd/crane/cmd/validate.go
```diff
@@ -45,7 +48,11 @@ func NewCmdValidate(options *[]crane.Option) *cobra.Command {
 					return fmt.Errorf("failed to read image %s: %v", flag, err)
 				}
 
-				if err := validate.Image(img); err != nil {
+				opt := []validate.Option{}
+				if fast {
+					opt = append(opt, validate.Fast)
+				}
+				if err := validate.Image(img, opt...); err != nil {
 					fmt.Printf("FAIL: %s: %v\n", flag, err)
```
squee1945, line 56: Will this result in a different exit code?

jonjohnsonjr: Ah yeah I neglected to follow up on https://github.com/google/go-containerregistry/pull/988#discussion_r617088024

Pulled this out into https://github.com/google/go-containerregistry/pull/1014

PR #1020, opened 2021-05-18
Fix install instructions for crane

samj1912:
When running the specified command you get - 
```
$ go install github.com/google/go-containerregistry/cmd/crane                                                                                                                                                                  ✘ 127

go install: version is required when current directory is not in a module
	Try 'go install github.com/google/go-containerregistry/cmd/crane@latest' to install the latest version
```

### DIFF
--- a/cmd/crane/README.md
+++ b/cmd/crane/README.md
@@ -14,7 +14,7 @@ Download [latest release](https://github.com/google/go-containerregistry/release
 Install manually:
 
 ```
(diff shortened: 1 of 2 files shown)


samj1912:
@googlebot I signed it!

imjasonh (approved):
We should also update this for https://github.com/google/go-containerregistry/tree/main/cmd/gcrane but I'm fine doing that in another PR.

PR #1021, opened 2021-05-18
Add auth options to gcrane

jonjohnsonjr:
This allows overriding the use of gcrane.Keychain

### DIFF
--- a/pkg/gcrane/options.go
+++ b/pkg/gcrane/options.go
@@ -18,6 +18,7 @@ import (
 	"context"
 	"runtime"
 
+	"github.com/google/go-containerregistry/pkg/authn"
 	"github.com/google/go-containerregistry/pkg/crane"
 	"github.com/google/go-containerregistry/pkg/v1/google"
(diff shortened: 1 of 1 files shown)


pkg/gcrane/options.go
```diff
@@ -81,3 +82,30 @@ func WithContext(ctx context.Context) Option {
...
 }
+
+// WithKeychain is a functional option for overriding the default
+// authenticator for remote operations, using an authn.Keychain to find
+// credentials.
+//
+// By default, gcrane will use gcrane.Keychain.
+func WithKeychain(keys authn.Keychain) Option {
+	return func(o *options) {
+		// Replace the default keychain at position 0.
```
imjasonh, line 93: This feels gross, and brittle. Can we iterate options until we find the one to replace instead?

jonjohnsonjr: I don't think it's possible :/

We had some discussion in https://github.com/google/go-containerregistry/pull/653 but I never came up with a better solution.

imjasonh (approved):

PR #1028, opened 2021-06-01
use go1.16's module aware install vs tools.go

dprotaso:
`go install module@version` will install pinned versions of executables. Use this approach vs vendoring as it doesn't pollute our `go.mod` file and `vendor` directory


imjasonh (approved):
Nice.

jonjohnsonjr:
Bless you.

Issue #1032, opened 2021-06-03
A function just load  manifest  from tar file


hhstu:
I want to  support  load more than one image for  a tar file, so i need  a function  load original manifest from tar file

hhstu:
you can see https://github.com/google/go-containerregistry/pull/1033

imjasonh:
I believe you should be able to accomplish this already, with existing methods:

```go
img, err := tarball.ImageFromPath(path, tag)
m, err := img.Manifest()
```

Loading the image from a tarball doesn't read or load any layers until those layers are accessed. It only reads and parses the manifest file, the same as your proposed `LoadManifest` does.

If this is not what you're seeing, please let me know. Keeping this behavior consistent with other packages (e.g., `remote`) is a priority, to reduce the size of the API surface.

jonjohnsonjr:
> the same as your proposed LoadManifest does.

Not quite. They want a [`tarball.Manifest`](https://github.com/google/go-containerregistry/blob/3bfab55f3bd99686225d0805e4b3b2f755d088f3/pkg/v1/tarball/image.go#L118) that describes the contents of the tarball.

We have a method of writing multiple artifacts to a tarball, but no way to query the tarball for what's in it.

imjasonh:
Gotcha, sorry for the misunderstanding. 👍

hhstu:
It`s work to push  more than one image for a tar file. 


``` go 

package image_mgr

import (
	"fmt"
	"github.com/beego/beego/v2/core/logs"
	"github.com/google/go-containerregistry/pkg/name"
	v1 "github.com/google/go-containerregistry/pkg/v1"
	"github.com/google/go-containerregistry/pkg/v1/remote"
	"github.com/google/go-containerregistry/pkg/v1/tarball"
	"github.com/pkg/errors"
	"os"
	"path"
	"strings"
)


func(registry *ImageStore)PushImages(file string) error {
	manifests, err := tarball.LoadManifest(file); if err != nil {
		return errors.Wrap(err, "解析镜像 manifest 失败")
	}
	for _, manifest := range manifests{
		for _, repoTag := range manifest.RepoTags {
			if err := registry.PushImage("", "", repoTag, file); err != nil {
				return err
			}
		}
	}
	return nil
}

// PushImage can rename image
func(registry *ImageStore)PushImage(imageName, imageTag, repoTag, file string) error {

	tag, err := name.NewTag(repoTag); if err != nil {
[...]

PR #1033, opened 2021-06-03
load manifest form tar file

hhstu:
Fixes https://github.com/google/go-containerregistry/issues/1032

### DIFF
--- a/pkg/v1/tarball/image.go
+++ b/pkg/v1/tarball/image.go
@@ -117,6 +117,22 @@ type Descriptor struct {
 // Manifest represents the manifests of all images as the `manifest.json` file in a `docker save` tarball.
 type Manifest []Descriptor
 
+// LoadManifest load manifest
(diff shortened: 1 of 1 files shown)


pkg/v1/tarball/image.go
```diff
@@ -117,6 +117,22 @@ type Descriptor struct {
 // Manifest represents the manifests of all images as the `manifest.json` file in a `docker save` tarball.
 type Manifest []Descriptor
 
+// LoadManifest load manifest
+func LoadManifest(path string) (Manifest, error) {
```
jonjohnsonjr, line 121: I agree we need this but I don't want the parameter to be a `string`, so that you can use this for in-memory tarballs. Can you make this take an `Opener`?

jonjohnsonjr:
Thanks! Can you squash these commits and make sure they are both authored as `hhstu` for the CLA check?

jonjohnsonjr (approved):

PR #1036, opened 2021-06-03
Update dependencies

### LINKED ISSUE #916: Docker version upgrade?

imjasonh:
```
go get -u ./...
go mod vendor
./hack/update-codegen.sh
```

This updates our containerd dep to v1.5.2, which contains https://github.com/containerd/containerd/commit/363f2c392c58d54f7e4d1e38bac1dc3a71283335 to only check CPU arch on-demand the first time it's needed, instead of at init-time.

Fixes #916


imjasonh:
Needed to add a lock around `defaultKeychain.Resolve` to avoid a data race, seemingly newly coming from https://github.com/docker/cli/pull/3025 (Thanks to @jonjohnsonjr for digging that up)

jonjohnsonjr:
How would you feel about `sync.Once`ing the `Load`?

imjasonh:
> How would you feel about `sync.Once`ing the `Load`?

I'm not opposed in theory, but I might want to make that as a separate change. The mutex is a quick hack to avoid the data race (which seems relatively innocuous anyway).

Skipping the load each time is probably good for performance, especially for short-lived CLI use cases, but something that uses ggcr in a long-running, potentially multi-tenant scenario might unknowingly depend on the load each time, and could be surprised in a bad way if we optimized it.

jonjohnsonjr (approved):

Issue #1038, opened 2021-06-03
Panic when retrying MultiWrite with the same regv1.Update channel


DennisDenuto:
# Description

We have been enjoying having the ability to display a progress bar using the `regv1.Update` channel in `MultiWrite` and `Write`

However, we have run into a panic when 'retrying' a `MultiWrite` using the same `regv1.Update` channel. (Because the 'writer' has already closed the channel)

The simple solution would be to not use the same channel when 'retrying' a `Write`. However due to abstractions that we have in place (a registry struct that holds state necessary to make a series of go-containerregistry calls with opts) this becomes tricky.

# Use case
1X reader of the channel
1X writer of the channel, invoked multiple times in a retry loop

```
util.Retry(func() error {
	return regremote.MultiWrite(imageOrIndexesToUpload, r.opts)
})
```

# Suggestions:
Writer *never* closes the channel, instead the onus is on the 'reader' of the channel / creator of the channel. 
or
Provide an option to toggle whether the 'reader' or 'writer' closes the channel

imjasonh:
I think it's a bit cleaner to have the method close the channel, it lets you do something like:

```
ch := make(chan v1.Update)
go remote.MultiWrite(..., remote.WithProgress(ch))
for u := range ch {
  fmt.Println(u)
}
```

If the caller was responsible for closing it'd only be possible to consume updates in a goroutine and close it yourself manually.

If retries is what you're after, we already bake retries into MultiWrite (I think...), and could expose an Option to configure that differently.

Out of curiosity, what failures are you seeing that MultiWrite's retries aren't successfully hiding from you?

jonjohnsonjr:
Seems like it would be better if we weren't dealing with channels directly but some different api/interface.

DennisDenuto:
Thanks for the replies!

> Out of curiosity, what failures are you seeing that MultiWrite's retries aren't successfully hiding from you?

From what we can tell, MultiWrite doesn't retry during a `commitManifest`. Same goes with other functions (i.e. WriteTag)

We want to retry on *every* http request to a registry (except on certain http errors such as auth related). 

> could expose an Option to configure that differently

That would be nice, and hopefully remove our need to retry around ggcr functions. And if *that* was true, having the writer close the channel would be a non-issue for us at least.

 > If the caller was responsible for closing it'd only be possible to consume updates in a goroutine and close it yourself manually.

[...]

PR #1041, opened 2021-06-07
Add retry when commiting a manifest

DennisDenuto:
- only retry on certain 'temporary' http status codes
- add tests to ensure retries do not occur on 'non-temp' http status
codes. And to make sure progress bar still works when retrying

Attempts to fix: https://github.com/google/go-containerregistry/issues/1038

Note:
Running `hack/presubmit.sh` resulted in the following:

```
~/workspace/go-containerregistry/pkg/authn/k8schain ~/workspace/go-containerregistry ~/workspace/go-containerregistry
go: github.com/google/go-containerregistry@v0.5.2-0.20210601193515-0ffa4a5c8691 requires
        cloud.google.com/go@v0.83.0: missing go.sum entry; to add it:
        go mod download cloud.google.com/go
~/workspace/go-containerregistry ~/workspace/go-containerregistry
```

This seems to me unrelated to the change in the PR... tried to go mod download  `cloud.google.com/go` but it still complains.

Authored-by: Dennis Leon <leonde@vmware.com>

### DIFF
--- a/pkg/v1/remote/write.go
+++ b/pkg/v1/remote/write.go
@@ -407,6 +407,24 @@ func (w *writer) incrProgress(written int64) {
 	}
 }
 
+var shouldRetry retry.Predicate = func(err error) bool {
+	// Various failure modes here, as we're often reading from and writing to
+	// the network.
+	if retry.IsTemporary(err) || errors.Is(err, io.ErrUnexpectedEOF) || errors.Is(err, syscall.EPIPE) {
+		logs.Warn.Printf("retrying %v", err)
+		return true
+	}
+	return false
+}
+
+// Try this three times, waiting 1s after first failure, 3s after second.
+var backoff = retry.Backoff{
+	Duration: 1.0 * time.Second,
+	Factor:   3.0,
+	Jitter:   0.1,
+	Steps:    3,
+}
+
 // uploadOne performs a complete upload of a single layer.
 func (w *writer) uploadOne(l v1.Layer) error {
 	var from, mount string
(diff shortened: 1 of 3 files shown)


jonjohnsonjr:
>         cloud.google.com/go@v0.83.0: missing go.sum entry; to add it:

@imjasonh any idea?

pkg/v1/remote/write.go
```diff
@@ -616,34 +616,54 @@ func unpackTaggable(t Taggable) ([]byte, *v1.Descriptor, error) {
...
+		// The image was successfully pushed!
+		logs.Progress.Printf("%v: digest: %v size: %d", ref, desc.Digest, desc.Size)
+		w.incrProgress(int64(len(raw)))
+		return nil
 	}
-	defer resp.Body.Close()
 
-	if err := transport.CheckError(resp, http.StatusOK, http.StatusCreated, http.StatusAccepted); err != nil {
-		return err
+	shouldRetry := func(err error) bool {
```
jonjohnsonjr, line 650: Let's pull these out so we can share them with `uploadOne`?

DennisDenuto: Good idea. I extracted it as a private variable scoped to the package. Lemme know if you would prefer it scoped down further to the `writer` as a method receiver.

pkg/v1/remote/write.go
```diff
@@ -616,34 +616,44 @@ func unpackTaggable(t Taggable) ([]byte, *v1.Descriptor, error) {
...
+		logs.Progress.Printf("%v: digest: %v size: %d", ref, desc.Digest, desc.Size)
+		w.incrProgress(int64(len(raw)))
+		return nil
 	}
 
-	// The image was successfully pushed!
-	logs.Progress.Printf("%v: digest: %v size: %d", ref, desc.Digest, desc.Size)
-	w.incrProgress(int64(len(raw)))
-	return nil
+	backoff := retry.Backoff{
```
jonjohnsonjr, line 650: The `backoff` is identical too, I'd pull that out as well.

DennisDenuto: done!

jonjohnsonjr (approved):

PR #1044, opened 2021-06-10
Verify size in verify.ReadCloser

imjasonh:
This prevents cases where a descriptor says it describes a 100-byte blob, but then serves 1TB blob, which could lead to resource exhaustion. Instead, we'll only read as many bytes as the descriptor claims there should be, and fail if the digest doesn't match after reading that many. This also verifies the content isn't shorter than described.

This adds a HEAD request to `remote.Layer` to fetch the size of the blob before reading and verifying it. If that's not acceptable we could work around it.

### DIFF
--- a/pkg/v1/remote/descriptor.go
+++ b/pkg/v1/remote/descriptor.go
@@ -334,10 +334,7 @@ func (f *fetcher) headManifest(ref name.Reference, acceptable []types.MediaType)
 	if lh == "" {
 		return nil, fmt.Errorf("HEAD %s: response did not include Content-Length header", u.String())
 	}
-	size, err := strconv.ParseInt(lh, 10, 64)
-	if err != nil {
-		return nil, err
-	}
+	size := resp.ContentLength
 
 	dh := resp.Header.Get("Docker-Content-Digest")
 	if dh == "" {
--- a/pkg/v1/remote/descriptor.go
+++ b/pkg/v1/remote/descriptor.go
@@ -380,7 +380,17 @@ func (f *fetcher) fetchBlob(ctx context.Context, h v1.Hash) (io.ReadCloser, erro
 		return nil, err
 	}
 
-	return verify.ReadCloser(resp.Body, h)
+	// Verify up to the content-length header value.
+	lh := resp.Header.Get("Content-Length")
+	if lh == "" {
+		return nil, fmt.Errorf("GET %s: response did not include Content-Length header", u.String())
+	}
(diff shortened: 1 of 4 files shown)


imjasonh:
> I'm tempted to have these be separate things because in some cases we don't know the size but we do know the digest (e.g. diffids and schema 1).

That's reasonable. It doesn't look like we use `verify` at all with either of those though, so maybe we should update `verify.ReadCloser` to take size into account, then add `verify.ReadCloserNoSize` (naming tbd) when we need it. That might be something we could use with `remote.Layer` for instance.

jonjohnsonjr:
> That might be something we could use with remote.Layer for instance.

Might still be useful to use the `Content-Length` for that, but also the response body probably does that itself... we should maybe be confirming `Size` == `Content-Length` _and_ doing a `LimitReader`?

imjasonh:
I decided to verify using `Content-Length` in `fetchBlob` so that `remote.Layer` doesn't have to do a HEAD. This made the diff smaller.

pkg/v1/remote/descriptor.go
```diff
@@ -380,7 +380,17 @@ func (f *fetcher) fetchBlob(ctx context.Context, h v1.Hash) (io.ReadCloser, erro
 		return nil, err
 	}
 
-	return verify.ReadCloser(resp.Body, h)
+	// Verify up to the content-length header value.
+	lh := resp.Header.Get("Content-Length")
```
jonjohnsonjr, line 384: `resp.ContentLength` will do this for you.

jonjohnsonjr (commented):
Do we want to validate config blob size as well?

https://github.com/google/go-containerregistry/blob/4c244d6cf55d8d1a4b7287b26651e99ae29b56cb/pkg/v1/remote/image.go#L103

imjasonh:
> Do we want to validate config blob size as well?
> 
> https://github.com/google/go-containerregistry/blob/4c244d6cf55d8d1a4b7287b26651e99ae29b56cb/pkg/v1/remote/image.go#L103

That'll be validated against the Content-Length at least. I'm not sure what to do if the descriptor size disagrees with the content-length header. Nothing would catch that with this version, but nothing would have caught it before either AFAIK.

pkg/v1/remote/descriptor.go
jonjohnsonjr: Check `resp.ContentLength != -1` here and above, and don't `ParseInt` it below.

jonjohnsonjr (approved):

PR #1045, opened 2021-06-10
k8s chain versioning is tons of fun

dprotaso:
- use K8s 1.20 libs
- use K8s 1.21 libs


jonjohnsonjr:
We should fix this race @imjasonh

jonjohnsonjr (approved):

Issue #1042, opened 2021-06-09
Restructure k8schain imports


jonjohnsonjr:
We do side-effectual imports in k8schain to pull in cloud-specific credential providers. We currently have build tags to disable importing specific things in k8schain. This seems backwards.

Callers should be responsible for importing these magic packages, not k8schain, which would make the dependency graph a little less stupid.

jonjohnsonjr:
@dprotaso @hasheddan WDYT

dprotaso:
Yeah I agree - this could be coupled with adoption of the new k8s cloud provider libs

PR #1049, opened 2021-06-15
Create a k8schain directly from pull secrets

scothis:
The k8schain package creates a new keychain by resolving Kubernets
secrets from service accounts and other image pull secrets specified by
name. The k8s client gets the full secret content for these named
resources.

Sometimes it's desirable to pull the content of a secret from a cache,
or to watch a resource for changes in a controller. In this situation
it would be helpful to create a k8schain directly from resolved secrets,
without making calls to the k8s api server.

This change refactors the existing k8schain.New() method to separate the
resolution of secrets from the creation of the keychain. This is only a
refactoring of existing behavior to shortcut the k8s api lookups. There
is no change in existing behavior for the keychain.

Added:

    k8schain.NewFromPullSecrets(context.Context, []corev1.Secret) (authn.Keychain, error)

Signed-off-by: Scott Andrews <andrewssc@vmware.com>

### DIFF
--- a/pkg/authn/k8schain/k8schain.go
+++ b/pkg/authn/k8schain/k8schain.go
@@ -23,7 +23,7 @@ import (
 	"github.com/google/go-containerregistry/pkg/name"
 	credentialprovider "github.com/vdemeester/k8s-pkg-credentialprovider"
 	credentialprovidersecrets "github.com/vdemeester/k8s-pkg-credentialprovider/secrets"
-	v1 "k8s.io/api/core/v1"
+	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/rest"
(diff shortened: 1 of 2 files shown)


pkg/authn/k8schain/k8schain.go
```diff
@@ -132,6 +121,23 @@ func NewNoClient(ctx context.Context) (authn.Keychain, error) {
...
 
+// NewFromPullSecrets returns a new authn.Keychain suitable for resolving image references as
+// scoped by the pull secrets.
+func NewFromPullSecrets(ctx context.Context, pullSecrets []corev1.Secret) (authn.Keychain, error) {
+	once.Do(func() {
+		keyring = credentialprovider.NewDockerKeyring()
+	})
+
+	// Extend the default keyring with the pull secrets.
+	kr, err := credentialprovidersecrets.MakeDockerKeyring(pullSecrets, keyring)
```
jonjohnsonjr, line 132: You're okay with the default keyring here being the magical `NewDockerKeyring()` thing?

I'm starting to lean away from that (https://github.com/google/go-containerregistry/issues/1042) because it's caused problems (https://github.com/knative/serving/issues/10259).

scothis: My goal for this PR was to not change the current behavior. I agree with the spirit of #1042

jonjohnsonjr (approved):
Thanks!

jonjohnsonjr:
Ah sorry, can you fix the fmt error?

scothis:
@jonjohnsonjr done

Issue #1024, opened 2021-05-24
Make --insecure flag's behavior equal to  docker --insecure-registry flag.


ninehills:
Docker `--insecure-registry` flag can ignore TLS certificate validate error. 

This will be compatible with a private docker registry with a self sign certificate.

Dirty workaroud: https://github.com/ninehills/go-containerregistry/commit/c2d589a1f9b722149e552c8ea8caefd11e32ee3b

jonjohnsonjr:
This is easy enough for the `crane` CLI -- we can just add a `crane.WithTransport` option based on the flag, around [here](https://github.com/google/go-containerregistry/blob/0233fcda5d536d4548d66e24829000e64a3e6059/cmd/crane/cmd/root.go#L72).

Omar007:
Reference for the behaviour Docker implements can be found here: https://docs.docker.com/registry/insecure/

> With insecure registries enabled, Docker goes through the following steps:
> - First, try using HTTPS.
> - If HTTPS is available but the certificate is invalid, ignore the error about the certificate.
> - If HTTPS is not available, fall back to HTTP.

PR #1054, opened 2021-06-16
crane: ignore TLS certificate validation when using --insecure flag

cpanato:
when using the `--insecure` flag ignore the TLS validation.
Please let me know if this is ok or what are your thoughts

/assign @jonjohnsonjr 

Fixes: https://github.com/google/go-containerregistry/issues/1024

### DIFF
--- a/cmd/crane/cmd/root.go
+++ b/cmd/crane/cmd/root.go
@@ -45,7 +45,7 @@ func New(use, short string, options []crane.Option) *cobra.Command {
 		Run:               func(cmd *cobra.Command, _ []string) { cmd.Usage() },
 		DisableAutoGenTag: true,
 		SilenceUsage:      true,
-		PersistentPreRunE: func(cmd *cobra.Command, args []string) error {
+		PersistentPreRun: func(cmd *cobra.Command, args []string) {
 			options = append(options, crane.WithContext(cmd.Context()))
 			// TODO(jonjohnsonjr): crane.Verbose option?
 			if verbose {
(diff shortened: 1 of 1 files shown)


cmd/crane/cmd/root.go
```diff
@@ -67,12 +68,26 @@ func New(use, short string, options []crane.Option) *cobra.Command {
 			cf, err := config.Load(os.Getenv("DOCKER_CONFIG"))
 			if err != nil {
 				logs.Debug.Printf("failed to read config file: %v", err)
+				return err
```
jonjohnsonjr, line 71: We don't want this to be fatal.

cmd/crane/cmd/root.go
```diff
@@ -67,12 +68,26 @@ func New(use, short string, options []crane.Option) *cobra.Command {
...
 			} else if len(cf.HTTPHeaders) != 0 {
 				options = append(options, crane.WithTransport(&headerTransport{
-					inner:       http.DefaultTransport,
+					inner: &http.Transport{
+						TLSClientConfig: &tls.Config{InsecureSkipVerify: insecure},
```
jonjohnsonjr, line 75: I'd dedupe this a bit, something like:

```
transport := http.DefaultTransport.(*http.Transport).Clone()
transport.TLSClientConfig = &tls.Config{InsecureSkipVerify: insecure}

...
} else if len(cf.HTTPHeaders) != 0 {
  transport = &headerTransport{transport, cf.HTTPHeaders}
}

options = append(options, crane.WithTransport(transport))

Comment (cpanato) [reply cmd/crane/cmd/root.go]: I applied some changes if I try your suggestion or something related I get this error
`cannot use &(headerTransport literal) (value of type *headerTransport) as *http.Transport value in assignment`

with my limited knowledge, I don't know how to solve it.

Comment (cpanato) [reply cmd/crane/cmd/root.go]: thanks for your review and feedback

File: cmd/crane/cmd/root.go [hunk 3]
```diff
@@ -64,30 +64,23 @@ func New(use, short string, options []crane.Option) *cobra.Command {
...
-				return nil
-			}
 
-			if insecure {
-				options = append(options, crane.WithTransport(&headerTransport{
-					inner: &http.Transport{
-						TLSClientConfig: &tls.Config{InsecureSkipVerify: insecure},
-					},
-				}))
+				return
```
jonjohnsonjr, line 80: Drop this return, otherwise lgtm.

cmd/crane/cmd/root.go
jonjohnsonjr: > cannot use &(headerTransport literal) (value of type *headerTransport) as *http.Transport value in assignment

Ah, yeah i was just using some shorthand, what you did instead works 👍

cmd/crane/cmd/root.go
cpanato: done, thanks again! PTAL

jonjohnsonjr (approved):
Thanks!

Issue #281, opened 2018-10-04
RFC: crane mutate to modify image metadata


ImJasonH:
@squee1945 mentioned it might be useful to have a tool that can change the entrypoint in an image in a registry in one command. Remotely adding/removing envs and labels seems useful too, and would be done nearly exactly the same way.

### Strawman usage

```
$ crane mutate gcr.io/my-project/my-image:my-tag \
    --entrypoint=foo \
    --env=FOO=bar \
    --env=BAR=baz \
    --env=BAZ= \ # unset
    --label=some.label.key=some.label.value \
    --label=another.key=another.value \
    --label=unset.this.key
<digest of newly modified image>
```

This would fetch the config for the image, update entrypoint/env/labels, and push a new manifest with the original image's layers and the modified config. No layers would be copied, and the new image would be pushed to the original image's repository.

From there, you can tag the resulting digest with `crane cp <digest> <tag-you-want>` (sidenote: maybe `crane tag` would be a useful alias?)

### Alternatives considered

We could have separate commands that would each do one operation, e.g.:

```
$ crane mutate env gcr.io/my-project/my-image:my-tag FOO bar
<digest>
$ crane mutate env <prev-digest> BAR baz
<another-digest>
$ crane mutate env <prev-digest> BAR # unset
<yet-another-digest>
```

You get the idea. Since the output is the new digest, you could theoretically chain them:

```
$ crane mutate env $(crane mutate env original FOO bar) BAR baz
```

But that's gross, and also ends up sending multiple updates needlessly. So let's not do that.

jonjohnsonjr:
I want it still.

ImJasonH:
How's this for a fairly minimal first stab:

```
crane mutate gcr.io/my/image
  [--label foo=bar] \
  [--annotate foo=bar] \
  [--entrypoint blah] \
  [--tag gcr.io/my/image:mutated]
```

- if `--tag` is omitted, it tags over the original image
- `--label` and `--annotation` must include a `=`, the key is `strings.SplitN(lbl, "=", 2)[0]`, the value is the rest
- error if no modifications are requested

For `--label` and `--entrypoint`, if `gcr.io/my/image` is a manifest list, the mutation is applied for all images in the manifest list (don't support nested manifest lists for now).

wdyt?

jonjohnsonjr:
[...]

PR #1056, opened 2021-06-17
added support for overriding manifest annotations

developer-guy:

Hello, this PR will add overriding manifest annotations support to crane's `mutate` command, it will work like the following picture:

Screen Shot 2021-06-18 at 01 06 08

### DIFF
--- a/cmd/crane/cmd/mutate.go
+++ b/cmd/crane/cmd/mutate.go
@@ -30,6 +30,7 @@ func NewCmdMutate(options *[]crane.Option) *cobra.Command {
 	var lbls []string
 	var entrypoint string
 	var newRef string
+	var anntns []string
 
 	mutateCmd := &cobra.Command{
 		Use:   "mutate",
@@ -38,6 +39,17 @@ func NewCmdMutate(options *[]crane.Option) *cobra.Command {
 		Run: func(_ *cobra.Command, args []string) {
 			// Pull image and get config.
 			ref := args[0]
+
+			if len(anntns) != 0 {
+				desc, err := crane.Head(ref, *options...)
+				if err != nil {
+					log.Fatalf("checking %s: %v", ref, err)
+				}
+				if desc.MediaType.IsIndex() {
+					log.Fatalf("mutating annotations on an index is not yet supported")
+				}
+			}
+
 			img, err := crane.Pull(ref, *options...)
 			if err != nil {
 				log.Fatalf("pulling %s: %v", ref, err)
@@ -52,16 +64,19 @@ func NewCmdMutate(options *[]crane.Option) *cobra.Command {
 			if cfg.Config.Labels == nil {
 				cfg.Config.Labels = map[string]string{}
 			}
-			labels := map[string]string{}
-			for _, l := range lbls {
-				parts := strings.SplitN(l, "=", 2)
-				if len(parts) == 1 {
-					log.Fatalf("parsing label %q, not enough parts", l)
-				}
-				labels[parts[0]] = parts[1]
+
+			labels, err := splitKeyVals(lbls)
+			if err != nil {
+				log.Fatal(err)
 			}
+
 			for k, v := range labels {
 				cfg.Config.Labels[k] = v
+			}
+
+			annotations, err := splitKeyVals(anntns)
+			if err != nil {
+				log.Fatal(err)
 			}
 
 			// Set entrypoint.
@@ -74,6 +89,12 @@ func NewCmdMutate(options *[]crane.Option) *cobra.Command {
 			img, err = mutate.Config(img, cfg.Config)
 			if err != nil {
 				log.Fatalf("mutating config: %v", err)
+			}
+
+			// Mutate and write image.
+			img, err = mutate.Annotations(img, annotations)
+			if err != nil {
+				log.Fatalf("mutating annotations: %v", err)
 			}
 
 			// If the new ref isn't provided, write over the original image.
@@ -100,8 +121,22 @@ func NewCmdMutate(options *[]crane.Option) *cobra.Command {
 			fmt.Println(r.Context().Digest(digest.String()))
 		},
 	}
+	mutateCmd.Flags().StringSliceVarP(&anntns, "annotation", "a", nil, "New annotations to add")
 	mutateCmd.Flags().StringSliceVarP(&lbls, "label", "l", nil, "New labels to add")
 	mutateCmd.Flags().StringVar(&entrypoint, "entrypoint", "", "New entrypoing to set")
 	mutateCmd.Flags().StringVarP(&newRef, "tag", "t", "", "New tag to apply to mutated image. If not provided, push by digest to the original image repository.")
 	return mutateCmd
+}
+
+// splitKeyVals splits key value pairs which is in form hello=world
+func splitKeyVals(kvPairs []string) (map[string]string, error) {
+	m := map[string]string{}
+	for _, l := range kvPairs {
+		parts := strings.SplitN(l, "=", 2)
+		if len(parts) == 1 {
+			return nil, fmt.Errorf("parsing label %q, not enough parts", l)
+		}
+		m[parts[0]] = parts[1]
+	}
+	return m, nil
 }
--- a/pkg/v1/mutate/mutate.go
+++ b/pkg/v1/mutate/mutate.go
@@ -113,6 +113,16 @@ func Config(base v1.Image, cfg v1.Config) (v1.Image, error) {
 	return ConfigFile(base, cf)
 }
 
+// Annotations mutates the provided v1.Image to have the provided annotations
+func Annotations(base v1.Image, annotations map[string]string) (v1.Image, error) {
+	image := &image{
+		base:        base,
+		annotations: annotations,
+	}
+
+	return image, nil
+}
+
 // ConfigFile mutates the provided v1.Image to have the provided v1.ConfigFile
 func ConfigFile(base v1.Image, cfg *v1.ConfigFile) (v1.Image, error) {
 	m, err := base.Manifest()
--- a/pkg/v1/mutate/mutate_test.go
+++ b/pkg/v1/mutate/mutate_test.go
@@ -266,6 +266,27 @@ func TestMutateConfig(t *testing.T) {
 	}
 }
 
+func TestAnnotations(t *testing.T) {
+	source := sourceImage(t)
+
+	newAnnotations := map[string]string{
+		"im.the.first.annotation": "hello world",
+	}
+
+	result, err := mutate.Annotations(source, newAnnotations)
+	if err != nil {
+		t.Fatalf("failed to mutate annotations: %v", err)
+	}
+
+	if configDigestsAreEqual(t, source, result) {
+		t.Errorf("mutating the manifest annotations MUST mutate the config digest")
+	}
+
+	if err := validate.Image(result); err != nil {
+		t.Errorf("validate.Image() = %v", err)
+	}
+}
+
 func TestMutateCreatedAt(t *testing.T) {
 	source := sourceImage(t)
 	want := time.Now().Add(-2 * time.Minute)
--- a/cmd/crane/doc/crane_mutate.md
+++ b/cmd/crane/doc/crane_mutate.md
@@ -9,10 +9,11 @@ crane mutate [flags]
 ### Options
 
 ```
-      --entrypoint string   New entrypoing to set
-  -h, --help                help for mutate
-  -l, --label strings       New labels to add
-  -t, --tag string          New tag to apply to mutated image. If not provided, push by digest to the original image repository.
+  -a, --annotation strings   New annotations to add
+      --entrypoint string    New entrypoing to set
+  -h, --help                 help for mutate
+  -l, --label strings        New labels to add
+  -t, --tag string           New tag to apply to mutated image. If not provided, push by digest to the original image repository.
 ```
 
 ### Options inherited from parent commands
--- a/pkg/v1/mutate/image.go
+++ b/pkg/v1/mutate/image.go
@@ -30,12 +30,13 @@ type image struct {
 	base v1.Image
 	adds []Addendum
 
-	computed   bool
-	configFile *v1.ConfigFile
-	manifest   *v1.Manifest
-	mediaType  *types.MediaType
-	diffIDMap  map[v1.Hash]v1.Layer
-	digestMap  map[v1.Hash]v1.Layer
+	computed    bool
+	configFile  *v1.ConfigFile
+	manifest    *v1.Manifest
+	annotations map[string]string
+	mediaType   *types.MediaType
+	diffIDMap   map[v1.Hash]v1.Layer
+	digestMap   map[v1.Hash]v1.Layer
 }
 
 var _ v1.Image = (*image)(nil)
@@ -136,6 +137,16 @@ func (i *image) compute() error {
 			manifest.MediaType = ""
 		} else if strings.Contains(string(*i.mediaType), types.DockerVendorPrefix) {
 			manifest.MediaType = *i.mediaType
+		}
+	}
+
+	if i.annotations != nil {
+		if manifest.Annotations == nil {
+			manifest.Annotations = map[string]string{}
+		}
+
+		for k, v := range i.annotations {
+			manifest.Annotations[k] = v
 		}
 	}
 


cmd/crane/cmd/mutate.go
```diff
@@ -100,8 +110,21 @@ func NewCmdMutate(options *[]crane.Option) *cobra.Command {
...
 		},
 	}
+	mutateCmd.Flags().StringSliceVarP(&anntns, "annotation", "a", nil, "New annotations to add")
 	mutateCmd.Flags().StringSliceVarP(&lbls, "label", "l", nil, "New labels to add")
 	mutateCmd.Flags().StringVar(&entrypoint, "entrypoint", "", "New entrypoing to set")
 	mutateCmd.Flags().StringVarP(&newRef, "tag", "t", "", "New tag to apply to mutated image. If not provided, push by digest to the original image repository.")
 	return mutateCmd
 }
+
+func splitKeyValuePairsInFormOfAnEqualSign(kvPairs []string) (map[string]string, error) {
```
imjasonh, line 120: This is a bit wordy. How about just `splitKeyVals` and a doc comment?

pkg/v1/mutate/mutate.go
```diff
@@ -113,6 +113,29 @@ func Config(base v1.Image, cfg v1.Config) (v1.Image, error) {
 	return ConfigFile(base, cf)
 }
 
+// ManifestAnnotations mutates the provided v1.Image to have the provided annotations
+func ManifestAnnotations(base v1.Image, annotations map[string]string) (v1.Image, error) {
+	m, err := base.Manifest()
+	if err != nil {
+		return nil, err
+	}
```
imjasonh, line 121: Can we `m.DeepCopy` here instead, as early as possible and before any changes?

pkg/v1/mutate/mutate_test.go
```diff
@@ -266,6 +266,31 @@ func TestMutateConfig(t *testing.T) {
...
+func TestMutateManifestAnnotations(t *testing.T) {
+	source := sourceImage(t)
+
+	newAnnotations := map[string]string{
+		"im.the.first.annotation": "hello world",
+	}
+
+	result, err := mutate.ManifestAnnotations(source, newAnnotations)
+	if err != nil {
+		t.Fatalf("failed to mutate a config: %v", err)
```
imjasonh, line 278: "failed to mutate annotations"

pkg/v1/mutate/mutate_test.go
```diff
@@ -266,6 +266,31 @@ func TestMutateConfig(t *testing.T) {
...
+	}
+
+	if manifestsAreEqual(t, source, result) {
```
imjasonh, line 281: Does this indicate that the source was mutated in the process? I would expect source to be unmodified and the modified result to be returned.

pkg/v1/mutate/mutate_test.go
```diff
@@ -266,6 +266,31 @@ func TestMutateConfig(t *testing.T) {
...
+		t.Error("mutating the manifest annotations MUST mutate the manifest")
+	}
+
+	if configDigestsAreEqual(t, source, result) {
+		t.Errorf("mutating the manifest annotations MUST mutate the config digest")
```
imjasonh, line 286: This actually should prove that the config isn't modified while modifying annotations.

imjasonh (commented):
Thanks for this change! 🎉

jonjohnsonjr:
I'll add some context from https://github.com/google/go-containerregistry/issues/281 we talked about how this would work for both image and manifest lists. I don't think you need to solve this immediately, but being able to annotate just the top-level manifest vs recursing and annotating every manifest is an interesting problem.

pkg/v1/mutate/mutate.go
```diff
@@ -113,6 +113,29 @@ func Config(base v1.Image, cfg v1.Config) (v1.Image, error) {
 	return ConfigFile(base, cf)
 }
 
+// ManifestAnnotations mutates the provided v1.Image to have the provided annotations
+func ManifestAnnotations(base v1.Image, annotations map[string]string) (v1.Image, error) {
```
jonjohnsonjr, line 117: I'd just call this `Annotations` because most of the other functions assume an image as well. If we want to extend this to other things, we can follow the `IndexMediaType` pattern and just do `IndexAnnotations`.

Ideally, this function would take a Describable, but then it's hard to know what to return :(

pkg/v1/mutate/mutate_test.go
```diff
@@ -266,6 +266,31 @@ func TestMutateConfig(t *testing.T) {
 	}
 }
 
+func TestMutateManifestAnnotations(t *testing.T) {
```
jonjohnsonjr, line 269: Just call this `TestAnnotations`.

pkg/v1/mutate/mutate.go
```diff
@@ -113,6 +113,29 @@ func Config(base v1.Image, cfg v1.Config) (v1.Image, error) {
...
+	m, err := base.Manifest()
+	if err != nil {
+		return nil, err
+	}
+
+	if len(m.Annotations) < 1 {
```
jonjohnsonjr, line 123: Check for `nil` instead?

pkg/v1/mutate/mutate.go
```diff
@@ -113,6 +113,29 @@ func Config(base v1.Image, cfg v1.Config) (v1.Image, error) {
...
+	for k, v := range annotations {
+		m.Annotations[k] = v
+	}
+
+	image := &image{
+		base:     base,
+		manifest: m.DeepCopy(),
+	}
+
+	return image, err
```
jonjohnsonjr, line 136: `return image, nil`

pkg/v1/mutate/mutate_test.go
developer-guy: yes you are right, thank you so much for the feedbacks 🙏

pkg/v1/mutate/mutate_test.go
developer-guy: thank you 🤦

pkg/v1/mutate/mutate.go
developer-guy: done, thank you so much 🙏

pkg/v1/mutate/mutate.go
developer-guy: yes, it should be like this way, it is my bad, thank you 🙏

pkg/v1/mutate/mutate.go
developer-guy: sorry again my bad 🤦

pkg/v1/mutate/mutate_test.go
developer-guy: done, thank you 🙏

pkg/v1/mutate/mutate.go
developer-guy: thank you so much for this review, but I tried to do that but it didn't work for me, I updated the code with the following content, what did I do wrong?
```go
mn := m.DeepCopy()

if mn.Annotations == nil {
	mn.Annotations = map[string]string{}
}

for k, v := range annotations {
	mn.Annotations[k] = v
}

image := &image{
	base:     base,
	manifest: mn,
}
```

cmd/crane/cmd/mutate.go
developer-guy: done, thank you so much 🙏

pkg/v1/mutate/mutate.go
```diff
@@ -113,6 +113,29 @@ func Config(base v1.Image, cfg v1.Config) (v1.Image, error) {
 	return ConfigFile(base, cf)
 }
 
+// IndexAnnotations mutates the provided v1.Image to have the provided annotations
+func IndexAnnotations(base v1.Image, annotations map[string]string) (v1.Image, error) {
```
jonjohnsonjr, line 117: Sorry, my comment wasn't clear, just call this `Annotations`

developer-guy: done.

jonjohnsonjr:
Oh yeah, you'll want to do this lazily. That `manifest` field in `image` is the final result of computing everything, so we need to move this logic into `compute()`:

```
$ git diff
diff --git a/pkg/v1/mutate/image.go b/pkg/v1/mutate/image.go
index 14b9c70a..23635f2b 100644
--- a/pkg/v1/mutate/image.go
+++ b/pkg/v1/mutate/image.go
@@ -30,12 +30,13 @@ type image struct {
        base v1.Image
        adds []Addendum
 
-       computed   bool
-       configFile *v1.ConfigFile
-       manifest   *v1.Manifest
-       mediaType  *types.MediaType
-       diffIDMap  map[v1.Hash]v1.Layer
-       digestMap  map[v1.Hash]v1.Layer
+       computed    bool
+       configFile  *v1.ConfigFile
+       manifest    *v1.Manifest
+       annotations map[string]string
+       mediaType   *types.MediaType
+       diffIDMap   map[v1.Hash]v1.Layer
+       digestMap   map[v1.Hash]v1.Layer
 }
 
 var _ v1.Image = (*image)(nil)
@@ -138,6 +139,15 @@ func (i *image) compute() error {
                        manifest.MediaType = *i.mediaType
                }
        }
+       if i.annotations != nil {
+               if manifest.Annotations == nil {
+                       manifest.Annotations = map[string]string{}
+               }
+
+               for k, v := range i.annotations {
+                       manifest.Annotations[k] = v
+               }
+       }
 
        i.configFile = configFile
        i.manifest = manifest
diff --git a/pkg/v1/mutate/mutate.go b/pkg/v1/mutate/mutate.go
index 7615b323..70df4e46 100644
--- a/pkg/v1/mutate/mutate.go
+++ b/pkg/v1/mutate/mutate.go
@@ -115,22 +115,9 @@ func Config(base v1.Image, cfg v1.Config) (v1.Image, error) {
 
 // Annotations mutates the provided v1.Image to have the provided annotations
 func Annotations(base v1.Image, annotations map[string]string) (v1.Image, error) {
-       m, err := base.Manifest()
-       if err != nil {
-               return nil, err
-       }
-
-       if m.Annotations == nil {
-               m.Annotations = map[string]string{}
-       }
-
-       for k, v := range annotations {
-               m.Annotations[k] = v
-       }
-
        image := &image{
-               base:     base,
-               manifest: m.DeepCopy(),
+               base:        base,
+               annotations: annotations,
        }
 
        return image, nil
```

One thing I'm a little worried about is that if you run this command on a manifest list, I think it will turn that manifest list into an image, because we just call `crane.Pull`. We don't want to do that...

Give me one second, I want to implement something that will let us check if something is an image.

jonjohnsonjr:
> Give me one second, I want to implement something that will let us check if something is an image.

https://github.com/google/go-containerregistry/pull/1057

developer-guy:
> Oh yeah, you'll want to do this lazily. That `manifest` field in `image` is the final result of computing everything, so we need to move this logic into `compute()`:
> 
> ```
> $ git diff
> diff --git a/pkg/v1/mutate/image.go b/pkg/v1/mutate/image.go
> index 14b9c70a..23635f2b 100644
> --- a/pkg/v1/mutate/image.go
> +++ b/pkg/v1/mutate/image.go
> @@ -30,12 +30,13 @@ type image struct {
>         base v1.Image
>         adds []Addendum
>  
> -       computed   bool
> -       configFile *v1.ConfigFile
> -       manifest   *v1.Manifest
> -       mediaType  *types.MediaType
> -       diffIDMap  map[v1.Hash]v1.Layer
> -       digestMap  map[v1.Hash]v1.Layer
> +       computed    bool
> +       configFile  *v1.ConfigFile
> +       manifest    *v1.Manifest
> +       annotations map[string]string
> +       mediaType   *types.MediaType
> +       diffIDMap   map[v1.Hash]v1.Layer
> +       digestMap   map[v1.Hash]v1.Layer
>  }
>  
>  var _ v1.Image = (*image)(nil)
> @@ -138,6 +139,15 @@ func (i *image) compute() error {
>                         manifest.MediaType = *i.mediaType
>                 }
>         }
> +       if i.annotations != nil {
> +               if manifest.Annotations == nil {
> +                       manifest.Annotations = map[string]string{}
> +               }
> +
> +               for k, v := range i.annotations {
> +                       manifest.Annotations[k] = v
> +               }
> +       }
>  
>         i.configFile = configFile
>         i.manifest = manifest
> diff --git a/pkg/v1/mutate/mutate.go b/pkg/v1/mutate/mutate.go
> index 7615b323..70df4e46 100644
> --- a/pkg/v1/mutate/mutate.go
> +++ b/pkg/v1/mutate/mutate.go
> @@ -115,22 +115,9 @@ func Config(base v1.Image, cfg v1.Config) (v1.Image, error) {
>  
>  // Annotations mutates the provided v1.Image to have the provided annotations
>  func Annotations(base v1.Image, annotations map[string]string) (v1.Image, error) {
> -       m, err := base.Manifest()
> -       if err != nil {
> -               return nil, err
> -       }
> -
> -       if m.Annotations == nil {
> -               m.Annotations = map[string]string{}
> -       }
> -
> -       for k, v := range annotations {
> -               m.Annotations[k] = v
> -       }
> -
>         image := &image{
> -               base:     base,
> -               manifest: m.DeepCopy(),
> +               base:        base,
> +               annotations: annotations,
>         }
>  
>         return image, nil
> ```
> 
> One thing I'm a little worried about is that if you run this command on a manifest list, I think it will turn that manifest list into an image, because we just call `crane.Pull`. We don't want to do that...
> 
> Give me one second, I want to implement something that will let us check if something is an image.

thank you so much for this, I really appreciate it, I updated the PR according to your feedback. 🙋🏻‍♂️😋

developer-guy:
> > Give me one second, I want to implement something that will let us check if something is an image.
> 
> #1057

where should we do this validation, would you like to help me if you don't mind? 🙏

jonjohnsonjr:
> where should we do this validation

So at the beginning of the command, before we `crane.Pull`, we want to check to make sure we're not trying to mutate annotations on an index, because that won't work... Something like:

```go
if len(anntns) != 0 {
  desc, err := crane.Head(ref, *options...)
  if err != nil {
    log.Fatalf("checking %s: %v", ref, err)
  }
  if desc.MediaType.IsIndex() {
    log.Fatalf("mutating annotations on an index is not yet supported")
  }
}
```

developer-guy:
> > where should we do this validation
> 
> So at the beginning of the command, before we `crane.Pull`, we want to check to make sure we're not trying to mutate annotations on an index, because that won't work... Something like:
> 
> ```go
> if len(anntns) != 0 {
>   desc, err := crane.Head(ref, *options...)
>   if err != nil {
>     log.Fatalf("checking %s: %v", ref, err)
>   }
>   if desc.MediaType.IsIndex() {
>     log.Fatalf("mutating annotations on an index is not yet supported")
>   }
> }
> ```

done @jonjohnsonjr in the commit #de9902f.

jonjohnsonjr:
> can you please tell me why we are doing this, and what is the index thing actually, thank you

Sure, let's look at `ubuntu:latest` on DockerHub:

https://explore.ggcr.dev/?image=ubuntu@sha256:aba80b77e27148d99c034a987e7da3a287ed455390352663418c0f2ed40417fe

This is a multi-platform image, AKA an [image index](https://github.com/opencontainers/image-spec/blob/master/image-index.md), which I often shorten to just "index". Per the spec, an image index can have annotations, so it would be reasonable to expect calling `crane mutate -a "hello=world" ubuntu` to work and add an annotation to the multi-platform image.

However, this will not work! It would instead add the annotation to one of it's children -- the linux/amd64 image:

https://explore.ggcr.dev/?image=index.docker.io/library/ubuntu@sha256:376209074d481dca0a9cf4282710cd30a9e7ff402dea8261acdaaf57a18971dd&mt=application/vnd.docker.distribution.manifest.v2+json

Why? Because `crane.Pull` returns a `v1.Image`, which means it has to pick one of the child images to pull. There's a `--platform` flag and `crane.WithPlatform` option for specifying which platform you should pick.

After annotating the _child_ image, it would update the `ubuntu:latest` tag to point to that _child_ instead of the multi-platform image, which would essentially break any other platforms trying to pull it.

We want to avoid that last part, so if we encounter something that isn't an image, we'll just abort. In the future, we could add something like `mutate.IndexAnnotations` or something to support annotating multi-platform images. This is awful because we bolted on multi-platform stuff after the fact, so it doesn't work that well with the type system. We've had some discussion around how to make this better (https://github.com/google/go-containerregistry/pull/835) but I think we'll need to redesign everything and bump to a v2 to improve things, really :/

jonjohnsonjr:
Looks like the linter is mad, need to run `gofmt`.

developer-guy:
> we'll just abort

> > can you please tell me why we are doing this, and what is the index thing actually, thank you
> 
> Sure, let's look at `ubuntu:latest` on DockerHub:
> 
> https://explore.ggcr.dev/?image=ubuntu@sha256:aba80b77e27148d99c034a987e7da3a287ed455390352663418c0f2ed40417fe
> 
> This is a multi-platform image, AKA an [image index](https://github.com/opencontainers/image-spec/blob/master/image-index.md), which I often shorten to just "index". Per the spec, an image index can have annotations, so it would be reasonable to expect calling `crane mutate -a "hello=world" ubuntu` to work and add an annotation to the multi-platform image.
> 
> However, this will not work! It would instead add the annotation to one of it's children -- the linux/amd64 image:
> 
> https://explore.ggcr.dev/?image=index.docker.io/library/ubuntu@sha256:376209074d481dca0a9cf4282710cd30a9e7ff402dea8261acdaaf57a18971dd&mt=application/vnd.docker.distribution.manifest.v2+json
> 
> Why? Because `crane.Pull` returns a `v1.Image`, which means it has to pick one of the child images to pull. There's a `--platform` flag and `crane.WithPlatform` option for specifying which platform you should pick.
> 
> After annotating the _child_ image, it would update the `ubuntu:latest` tag to point to that _child_ instead of the multi-platform image, which would essentially break any other platforms trying to pull it.
> 
> We want to avoid that last part, so if we encounter something that isn't an image, we'll just abort. In the future, we could add something like `mutate.IndexAnnotations` or something to support annotating multi-platform images. This is awful because we bolted on multi-platform stuff after the fact, so it doesn't work that well with the type system. We've had some discussion around how to make this better (#835) but I think we'll need to redesign everything and bump to a v2 to improve things, really :/

Omg, thank you so much for this detailed explanation, I got it now 👍

developer-guy:
> Looks like the linter is mad, need to run `gofmt`.

fixed, thank you so much 👍

jonjohnsonjr (approved):

jonjohnsonjr:
Thanks!

PR #1063, opened 2021-06-27
cmd/mutate: ability to remove lbls and anntns

### LINKED ISSUE #1062: proposal: crane: add the ability to remove labels and annotations through mutate command
**TL;DR:** I propose adding a new flag called `--remove` _(`-r` for shorthand)_ for `mutate` command to remove _existing_ label and annotations.

**Proposal**
If we pass `--remove` flag to `mutate` command, the actual flow will be changed. Currently, we [check](https://github.com/google/go-containerregistry/blob/ce35c99b3fafc7b6ea9eb192933ae2730fe360d0/cmd/crane/cmd/mutate.go#L133) if a given _value_ is in `key=value` format. If we want to remove a _label_ or an _annotation_, we really do not want to pass its _value_, since each key is unique.

**Implementation**
At a minimum, `--remove` flag will impact:
* We can not pass a value to this flag
* It's just a _bool_ value, if passed, we will in _removeMode_
* This flag traverses all the given _label_s and _annotation_s

**Example**

Some happy paths are the following commands:

* Initially, mutate with some key and values:
```
$ crane mutate --label foo=bar --label bar=baz --annotation qux=quux $IMAGE
```

* Removing the _existing_ label with a given _key_
```
$ crane mutate --remove --label foo $IMAGE # OK
```

* Removing the _existing_ label and annotation with given _keys_
```
$ crane mutate --remove --label foo --annotation qux $IMAGE # OK
```

* Removing the _multiple-existing_ labels with given _keys_
```
$ crane mutate --remove --label foo --label bar $IMAGE # OK
```

* Removing the _non-existing_ label with a given _key_
```
$ crane mutate --remove --label qux $IMAGE # OH!
```

* Removing the _non-existing_ label with a given _key=value_
```
$ crane mutate --remove --label qux=quux $IMAGE # Not allowed!
```

* Removing the _existing_ label with a given _key=value_
```
$ crane mutate --remove --label bar=baz $IMAGE # Not allowed!
```

**P.S:** All the given instruction steps for the `--label` _(stated in the examples)_, will accurately the same for the `--annotation` flag.

**Additional Notes**

If you ask that why not this should be in the `delete` command, I am not so sure about that idea since we do not "apply" any _mutation_ operation there. It is just for _deletion_ for the given image.

Dentrax:
Fixes #1062

Signed-off-by: Furkan <furkan.turkal@trendyol.com>

### DIFF
--- a/cmd/crane/cmd/mutate.go
+++ b/cmd/crane/cmd/mutate.go
@@ -31,6 +31,8 @@ func NewCmdMutate(options *[]crane.Option) *cobra.Command {
 	var entrypoint string
 	var newRef string
 	var anntns []string
+	var unLbls []string
+	var unAnntns []string
 
 	mutateCmd := &cobra.Command{
 		Use:   "mutate",
@@ -74,6 +76,12 @@ func NewCmdMutate(options *[]crane.Option) *cobra.Command {
 				cfg.Config.Labels[k] = v
 			}
 
+			if len(unLbls) > 0 {
+				for _, l := range unLbls {
+					delete(cfg.Config.Labels, l)
+				}
+			}
+
 			annotations, err := splitKeyVals(anntns)
 			if err != nil {
 				log.Fatal(err)
@@ -92,6 +100,10 @@ func NewCmdMutate(options *[]crane.Option) *cobra.Command {
 			}
 
 			img = mutate.Annotations(img, annotations)
+
+			if len(unAnntns) > 0 {
+				img = mutate.UnAnnotations(img, unAnntns)
+			}
 
 			// If the new ref isn't provided, write over the original image.
 			// If that ref was provided by digest (e.g., output from
@@ -119,7 +131,9 @@ func NewCmdMutate(options *[]crane.Option) *cobra.Command {
 	}
 	mutateCmd.Flags().StringSliceVarP(&anntns, "annotation", "a", nil, "New annotations to add")
 	mutateCmd.Flags().StringSliceVarP(&lbls, "label", "l", nil, "New labels to add")
-	mutateCmd.Flags().StringVar(&entrypoint, "entrypoint", "", "New entrypoing to set")
+	mutateCmd.Flags().StringVar(&entrypoint, "entrypoint", "", "New entrypoint to set")
+	mutateCmd.Flags().StringSliceVar(&unLbls, "unlabel", nil, "Labels to remove")
+	mutateCmd.Flags().StringSliceVar(&unAnntns, "unannotate", nil, "Annotations to remove")
 	mutateCmd.Flags().StringVarP(&newRef, "tag", "t", "", "New tag to apply to mutated image. If not provided, push by digest to the original image repository.")
 	return mutateCmd
 }
--- a/pkg/v1/mutate/mutate.go
+++ b/pkg/v1/mutate/mutate.go
@@ -121,6 +121,14 @@ func Annotations(base v1.Image, annotations map[string]string) v1.Image {
 	}
 }
 
+// UnAnnotations removes annotations the provided v1.Image
+func UnAnnotations(base v1.Image, unAnnotations []string) v1.Image {
+	return &image{
+		base:          base,
+		unAnnotations: unAnnotations,
+	}
+}
+
 // ConfigFile mutates the provided v1.Image to have the provided v1.ConfigFile
 func ConfigFile(base v1.Image, cfg *v1.ConfigFile) (v1.Image, error) {
 	m, err := base.Manifest()
--- a/pkg/v1/mutate/image.go
+++ b/pkg/v1/mutate/image.go
@@ -30,13 +30,14 @@ type image struct {
 	base v1.Image
 	adds []Addendum
 
-	computed    bool
-	configFile  *v1.ConfigFile
-	manifest    *v1.Manifest
-	annotations map[string]string
-	mediaType   *types.MediaType
-	diffIDMap   map[v1.Hash]v1.Layer
-	digestMap   map[v1.Hash]v1.Layer
+	computed      bool
+	configFile    *v1.ConfigFile
+	manifest      *v1.Manifest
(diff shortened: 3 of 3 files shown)


Dentrax:
@googlebot I signed it!

cmd/crane/cmd/mutate.go
```diff
@@ -65,16 +66,20 @@ func NewCmdMutate(options *[]crane.Option) *cobra.Command {
...
 
-			labels, err := splitKeyVals(lbls)
+			labels, err := splitKeyVals(lbls, !removeMode)
 			if err != nil {
 				log.Fatal(err)
 			}
 
 			for k, v := range labels {
-				cfg.Config.Labels[k] = v
+				if !removeMode {
```
imjasonh, line 75: I think I'd prefer to phrase this as:

```go
if removeMode {
  delete(cfg.Config.Labels, k)
} else {
  cfg.Config.Labels[k] = v
}
```

cmd/crane/cmd/mutate.go
```diff
@@ -91,7 +96,17 @@ func NewCmdMutate(options *[]crane.Option) *cobra.Command {
...
+			if removeMode {
+				manifest, err := img.Manifest()
+				if err != nil {
+					log.Fatalf("get manifest: %v", err)
+				}
+				for k, _ := range annotations {
+					delete(manifest.Annotations, k)
+				}
+			} else {
+				img = mutate.Annotations(img, annotations)
```
imjasonh, line 108: Should we have some `mutate.RemoveAnnotations` method that we use here? I think it's confusing and inconsistent at least, and a bug at most, that `img.Manifest()` returns a mutable `manifest` that modifies the `img`. When we eventually fix that behavior, we'll want to make sure we don't break this code.

cmd/crane/cmd/mutate.go
```diff
@@ -119,20 +134,28 @@ func NewCmdMutate(options *[]crane.Option) *cobra.Command {
...
+	mutateCmd.Flags().StringVar(&entrypoint, "entrypoint", "", "New entrypoint to set")
+	mutateCmd.Flags().BoolVarP(&removeMode, "remove", "r", false, "Enable remove mode")
 	mutateCmd.Flags().StringVarP(&newRef, "tag", "t", "", "New tag to apply to mutated image. If not provided, push by digest to the original image repository.")
 	return mutateCmd
 }
 
 // splitKeyVals splits key value pairs which is in form hello=world
-func splitKeyVals(kvPairs []string) (map[string]string, error) {
+// if requireSplit set true, only gathers keys otherwise.
+func splitKeyVals(kvPairs []string, enforceSplit bool) (map[string]string, error) {
```
imjasonh, line 145: Instead of repurposing this method, and making it do two things depending on `enforceSplit`, maybe let's just have a separate method that translates a `[]string` -> `map[string]string` (or `map[string]struct{}`, even).

imjasonh (commented):
Most of this feedback is a bit less relevant in light of Jon's advice to support `--unlabel` and `--unannotate`, but some of it might still be useful after incorporating that feedback.

Dentrax:
Hey @imjasonh! I couldn't find a time to get to it due to our national holiday. I updated the code according to your reviews! I also added `RemoveAnnotations()` function into `mutate` package. :)

Commands now should be work as expected:

1. Mutate the image:
```bash
$ go run . mutate --label foo=bar --annotation baz=qux furkanturkal/busybox:0.2.0`
```

2. Check the mutations:
```bash
$ crane manifest furkanturkal/busybox:0.2.0 | jq
...
"annotations": {
  "baz": "qux"
}

$ crane config furkanturkal/busybox:0.2.0 | jq
...
"Labels": {
  "foo": "bar"
}
```

3. Remote both _label_s and _annotation_s:
```bash
$ go run . mutate --unlabel foo --unannotate baz furkanturkal/busybox:0.2.0
# check again if both removed
```

**P.S:** Minor flag name inconsistency has occurred: `--unannotate` and `--annotation`. Should we re-name one of them? (`tation` and `tate`)

---

In the `RemoveAnnotations` function, I had to traverse _annotations_ because I could not figure it out how to make it work using `match.Matcher` instead of _for-loop_. I was try to add a new field into `image` struct, as `index` struct did. And in the `compute()` function, I was trying to figure it out using this `matcher` function to remove (filter) annotations. I tried to call this new function by passing `img, match.Annotation("foo", "bar")` params.

```go
	if i.remove != nil {
		var cleanedManifests []v1.Descriptor
		for _, m := range manifests {
			if !i.remove(m) {
				cleanedManifests = append(cleanedManifests, m)
			}
		}
		manifests = cleanedManifests
	}
```

But somehow, `img.Manifest().Annotations` returned `nil` (which caused by `if !i.remove(m)` never returned _true_) ; but still, i was able to see the `foo: bar` value in the manifest: `$ crane manifest furkanturkal/busybox:0.2.0 | jq`. Maybe, I missed some important point. 🤷‍♂️

pkg/v1/mutate/mutate.go
```diff
@@ -101,6 +101,23 @@ func RemoveManifests(base v1.ImageIndex, matcher match.Matcher) v1.ImageIndex {
 	}
 }
 
+// RemoveAnnotations removes annotations the provided v1.Image
+func RemoveAnnotations(base v1.Image, annotations []string) (v1.Image, error) {
```
jonjohnsonjr, line 105: Don't return an error here, we can do this lazily. We'll need to store this list of annotations to delete in image{} (maybe call it `unannotations` and call `delete()` around here: https://github.com/google/go-containerregistry/blob/eca1cd86affe459c22a2c7a818769ea70d13496e/pkg/v1/mutate/image.go#L157

I also wonder if we want to give this the same treatment as https://github.com/google/go-containerregistry/pull/1082

Dentrax: That makes sense! Now, it's much cleaner. 👍

PR #1075, opened 2021-07-12
Handle multiple www-authenticate headers

seba-ban:
It might happen there are multiple `www-authenticate` headers returned from a registry, e.g. `Negotiate` and `Basic`. In the current code, in such situation the first challenge would be picked with no further checks, which would result eventually in `unrecognized challenge` error failing the whole build, even though `Basic` challenge could be used instead. `pickFromMultipleChallenges` function was added to loop through the challenges in search for a supported one.

### DIFF
--- a/pkg/v1/remote/transport/ping.go
+++ b/pkg/v1/remote/transport/ping.go
@@ -130,19 +130,19 @@ func ping(ctx context.Context, reg name.Registry, t http.RoundTripper) (*pingRes
 
 func pickFromMultipleChallenges(challenges []authchallenge.Challenge) authchallenge.Challenge {
 
- 	// It might happen there are multiple www-authenticate headers, e.g. `Negotiate` and `Basic`.
- 	// Picking simply the first one could result eventually in `unrecognized challenge` error,
+	// It might happen there are multiple www-authenticate headers, e.g. `Negotiate` and `Basic`.
+	// Picking simply the first one could result eventually in `unrecognized challenge` error,
 	// that's why we're looping through the challenges in search for one that can be handled.
- 	allowedSchemes := []string{"basic", "bearer"}
+	allowedSchemes := []string{"basic", "bearer"}
 
- 	for _, wac := range challenges {
- 		currentScheme := strings.ToLower(wac.Scheme)
- 		for _, allowed := range allowedSchemes {
- 			if allowed == currentScheme {
- 				return wac
- 			}
- 		}
- 	}
+	for _, wac := range challenges {
+		currentScheme := strings.ToLower(wac.Scheme)
+		for _, allowed := range allowedSchemes {
+			if allowed == currentScheme {
+				return wac
+			}
+		}
+	}
 
- 	return challenges[0]
- }
+	return challenges[0]
+}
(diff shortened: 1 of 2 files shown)


imjasonh (commented):
This seems like a code path that would benefit from having a unit test.

Is there a real-world scenario where you're seeing responses including multiple headers like this?

jonjohnsonjr:
Seems related to kerberos? https://stackoverflow.com/questions/4265975/authentication-issues-with-www-authenticate-negotiate

Reminds me of https://github.com/bazelbuild/rules_k8s/issues/141

seba-ban:
Thanks for the comment. I'll try to add a unit test later this week.

So, yeah, it can be related to Kerberos. In fact, the real-world scenario I'm aware of is using [SPNEGO NGINX module](https://github.com/stnoonan/spnego-http-auth-nginx-module) with basic authentication fallback - in such configuration two www-authenticate headers will be sent: `Negotiate` and `Basic`.

jonjohnsonjr:
I don't know much about kerberos, but would it make sense to attempt to support it? I assume not, but it sounds like it would be neat :)

seba-ban:
To be honest, I don't know much about Kerberos myself either. :( I'm also not sure how common it is to have a Kerberos (Negotiate) only registry, but I guess not that common. I don't think it's supported by Docker either [(here?)](https://github.com/docker/cli/blob/5f07d7d5a12423c0bc1fb507f4d006ad0cdfef42/vendor/github.com/containerd/containerd/remotes/docker/auth/parse.go#L98).

seba-ban:
Unit test added.

pkg/v1/remote/transport/ping.go
```diff
@@ -108,8 +108,8 @@ func ping(ctx context.Context, reg name.Registry, t http.RoundTripper) (*pingRes
 			}, nil
 		case http.StatusUnauthorized:
 			if challenges := authchallenge.ResponseChallenges(resp); len(challenges) != 0 {
-				// If we hit more than one, I'm not even sure what to do.
-				wac := challenges[0]
+				// If we hit more than one, let's try to find one that we know how to handle.
+				wac := pickFromMultipleChallenges(challenges)
```
jonjohnsonjr, line 112: What happens if we don't know how to handle any of the challenges?

I feel like maybe we should filter out unrecognized challenges and make sure `len(filtered) != 0`?

seba-ban: If there's neither basic nor bearer challenge, the last challenge will be returned. This will lead to `unrecognized challenge` error [here](https://github.com/google/go-containerregistry/blob/main/pkg/v1/remote/transport/transport.go#L101).
I think I would leave checking if challenge can be handled outside ping.go. Here, we only have some preference as to which challenge to choose, but ultimately we can return one that's not basic nor bearer.

jonjohnsonjr: I don't think it will return the last challenge because the `wac` in `for _, wac := range challenges` will shadow the outer `wac` variable. Instead, it would return a zero-value Challenge.

Maybe if we don't encounter a useful challenge, we just return the first one? (line 148 currently)

seba-ban: Ah, right, my bad, it should be `for _, wac = range challenges` so `wac` is not shadowed... But ok, I will change it so it returns the first challenge if neither basic nor bearer are found.

seba-ban: Thanks for noticing that!

jonjohnsonjr:
Can you fix the gofmt issue?

seba-ban:
Sorry for late response. I've run gofmt on ping.go.

jonjohnsonjr (approved):
Thanks for the PR!

PR #1076, opened 2021-07-12
Use Data field when fetching in remote

jonjohnsonjr:


pkg/v1/remote/image.go
```diff
@@ -100,6 +101,11 @@ func (r *remoteImage) RawConfigFile() ([]byte, error) {
 		return nil, err
 	}
 
+	if m.Config.Data != nil && int64(len(m.Config.Data)) == m.Config.Size {
```
imjasonh, line 104: If len(data) != size, should we fail? That definitely sounds like something is wrong.

The proposed spec says "The decoded data ... SHOULD be verified against the digest and size fields"

imjasonh (commented):
The addition of a possibly-gigantic `data` elements in manifest contents makes me want a `remote.WithMaxManifestSize` option (and maybe some sane large default) to prevent resource exhaustion when fetching manifests.

Clients can throw a `io.LimitReader` around `layer.Compressed()`, but they can't currently easily prevent a huge manifest from being fetched and read.

pkg/v1/remote/image.go
jonjohnsonjr: Yeah I probably should have marked this as a draft, just playing around right now.

jonjohnsonjr:
> The addition of a possibly-gigantic data elements in manifest contents makes me want a remote.WithMaxManifestSize option (and maybe some sane large default) to prevent resource exhaustion when fetching manifests.

I think we'd need to rewrite a ton of code if we want to start going down this path. I'd propose we punt on that for now and rethink how this would work in a v2 world.

imjasonh (approved):
Love it, ship it.

Issue #1061, opened 2021-06-26
Error: failed to copy image: response did not include Content-Length header


gaurav517:
`gcrane copy` throws `Error: failed to copy image: response did not include Content-Length header` for last image layer.

Details:
```
gcrane copy -v myregistry/mygroup/myimage:mytag asia.gcr.io/myproject/external/myregistry/mygroup/myimage:mytag
2021/06/26 08:41:45 Copying from myregistry/mygroup/myimage:mytag to asia.gcr.io/myproject/external/myregistry/mygroup/myimage:mytag
2021/06/26 08:41:45 --> GET https://myregistry/v2/
2021/06/26 08:41:45 GET /v2/ HTTP/1.1
Host: myregistry
User-Agent: gcrane/ce35c99b3fafc7b6ea9eb192933ae2730fe360d0 go-containerregistry
Accept-Encoding: gzip


2021/06/26 08:41:47 <-- 401 https://myregistry/v2/ (1.499409766s)
2021/06/26 08:41:47 HTTP/1.1 401 UNAUTHORIZED
Content-Length: 4
Cneonction: close
Content-Type: text/html; charset=utf-8
Date: Sat, 26 Jun 2021 08:41:47 GMT
Docker-Distribution-Api-Version: registry/2.0
Server: nginx/1.12.1
Strict-Transport-Security: max-age=0
Www-Authenticate: Bearer realm="https://myregistry/v2/auth",service="myregistry"
X-Served-By: 10.48.19.32:8080

true
2021/06/26 08:41:47 --> GET https://myregistry/v2/auth?scope=repository%3Amygroup%2Fmyimage%3Apull&service=myregistry [body redacted: basic token response contains credentials]
2021/06/26 08:41:47 GET /v2/auth?scope=repository%3Amygroup%2Fmyimage%3Apull&service=myregistry HTTP/1.1
Host: myregistry
User-Agent: gcrane/ce35c99b3fafc7b6ea9eb192933ae2730fe360d0 go-containerregistry
Accept-Encoding: gzip


2021/06/26 08:41:47 <-- 200 https://myregistry/v2/auth?scope=repository%3Amygroup%2Fmyimage%3Apull&service=myregistry (302.189116ms) [body redacted: basic token response contains credentials]
2021/06/26 08:41:47 HTTP/1.1 200 OK
Content-Length: 869
Cache-Control: no-cache, no-store, must-revalidate
Cneonction: close
Content-Type: application/json
[...]

PR #1080, opened 2021-07-13
Only verify size if we can

jonjohnsonjr:
Sometimes Content-Length isn't set, and sometimes we don't know what it
should be. Do our best regardless.

Fixes https://github.com/google/go-containerregistry/issues/1061

### DIFF
--- a/internal/verify/verify.go
+++ b/internal/verify/verify.go
@@ -26,6 +26,7 @@ import (
 	v1 "github.com/google/go-containerregistry/pkg/v1"
 )
 
+// SizeUnknown is a sentinel value to indicate that the expected size is not known.
 const SizeUnknown = -1
 
 type verifyReader struct {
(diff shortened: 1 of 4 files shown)


imjasonh:
How do you feel about `const SizeUnknown = -1`

jonjohnsonjr:
> How do you feel about const SizeUnknown = -1

I love it, done.

internal/verify/verify.go
```diff
@@ -26,6 +26,8 @@ import (
 	v1 "github.com/google/go-containerregistry/pkg/v1"
 )
 
+const SizeUnknown = -1
```
imjasonh, line 29:
```suggestion
// SizeUnknown is a sentinel value to indicate that the expected size is not known.
const SizeUnknown = -1
```

jonjohnsonjr:
PTAL

imjasonh (approved):

PR #1086, opened 2021-07-23
Add mutate.RebaseIndex

imjasonh:
This method rebases all images in an ImageIndex on corresponding
platform-specific images in a new base ImageIndex.

If the new base image includes a different set of platform-specific
images than the old base image, the intersection of the two is used to
produce the rebased image index. If the intersection is the empty set,
rebasing fails.


pkg/v1/mutate/rebase.go
```diff
@@ -146,175 +145,54 @@ func createAddendums(startHistory, startLayer int, history []v1.History, layers
...
 	mf, err := idx.IndexManifest()
 	if err != nil {
 		return nil, err
 	}
-
-	d, err := descByPlatform(*mf, &p)
-	if err != nil {
-		return nil, err
+	for _, desc := range mf.Manifests {
+		if reflect.DeepEqual(desc.Platform, plat) {
```
jonjohnsonjr, line 193: https://pkg.go.dev/github.com/google/go-containerregistry@v0.5.1/pkg/v1#Platform.Equals

pkg/v1/mutate/rebase_test.go
```diff
@@ -284,9 +285,7 @@ func imageByPlatform(t *testing.T, idx v1.ImageIndex, p v1.Platform) v1.Image {
 		t.Fatal(err)
 	}
 	for _, d := range mf.Manifests {
-		if d.Platform.OS == p.OS &&
-			d.Platform.Architecture == p.Architecture &&
-			d.Platform.Variant == p.Variant {
+		if reflect.DeepEqual(*d.Platform, p) {
```
jonjohnsonjr, line 288: https://pkg.go.dev/github.com/google/go-containerregistry@v0.5.1/pkg/v1#Platform.Equals

imjasonh: Nice.

imjasonh:
One concern I have with this current approach is that we're doing some pretty deep iterations, and it might get out of hand if there are lots of images in all three indexes.

pseudocode:
```
for img in orig:
  for desc in old:
    if desc.plat == img.plat:
      for desc in old:  # remote.index's childByHash via idx.Image(desc.digest)
        if desc.h == old.img.digest:
          oldBaseImg = image(desc)
  for desc in new:
    if desc.plat == img.plat:
      for desc in new:  # remote.index's childByHash via idx.Image(desc.digest)
        if desc.h == new.img.digest:
          newBaseImg = image(desc)
  rebase(img, oldBaseImg, newBaseImg)
```

If platforms could be map keys 🙄 this would be a lot simpler.

jonjohnsonjr:
> we're doing some pretty deep iterations

Should be fine as long as we're caching stuff appropriately. READ_LATENCY * 3n is going to be much much bigger than n^3 for iterating over lists, I would imagine.

imjasonh:
> > we're doing some pretty deep iterations
> 
> Should be fine as long as we're caching stuff appropriately. READ_LATENCY * 3n is going to be much much bigger than n^3 for iterating over lists, I would imagine.

Yeah you're right. I was mainly worried about some jerk (let's be honest it's going to be me) giving this some indexes of 100k images, but what kind of monster would do that (again, it's me).

PR #1088, opened 2021-07-26
Add recipe to calculate image size

imjasonh:
```
$ crane manifest gcr.io/buildpacks/builder:v1 | jq '.config.size + ([.layers[].size] | add)'
238034322
$ crane manifest gcr.io/buildpacks/builder:v1 | jq '.config.size + ([.layers[].size] | add)' | numfmt --to=iec
228M
```


jonjohnsonjr:
This won't work for an index, how should we handle that?

jonjohnsonjr (approved):

PR #1090, opened 2021-07-28
Straighten out remote.List{WithContext}

imjasonh:
`ListWithContext` is kinda gross when we also have `List(..., remote.WithContext(ctx)` -- let's deprecate it.

edit: the rest is no longer relevant, but useful for context:
~This led a test to fail because the /v2/ request fails on a cancelled context, which I think is good because we shouldn't ignore the provided context for that ping request. In fixing that, I wanted to make the test check for whether the error `errors.Is(context.Canceled)`, but that was stymied by `transport.NewTransportWithContext` losing that information when it does it's little multi-error thing.~

```
    list_test.go:115: error was not context.Canceled, got Get "https://127.0.0.1:60957/v2/": context canceled; Get "http://127.0.0.1:60957/v2/": context canceled
```
Gross.

So we have a couple options:
1. change the test to just search the error message for the string "context canceled" and keep the http fallback multierror logic from before (probably safest)
2. ~don't return multiple errors, maybe `if logs.Enabled(logs.Debug) { logs.Debug.Printf("failed https, falling back to http") }`~
3. ~drop this whole thing, `ListWithContext` is fine the way it is, and ignoring context to do the ping is fine.~

### DIFF
--- a/pkg/v1/remote/list.go
+++ b/pkg/v1/remote/list.go
@@ -31,14 +31,16 @@ type tags struct {
 	Tags []string `json:"tags"`
 }
 
-// List wraps ListWithContext using the background context.
-func List(repo name.Repository, options ...Option) ([]string, error) {
-	return ListWithContext(context.Background(), repo, options...)
+// ListWithContext calls List with the given context.
+//
+// Deprecated: Use List and WithContext. This will be removed in a future release.
+func ListWithContext(ctx context.Context, repo name.Repository, options ...Option) ([]string, error) {
+	return List(repo, append(options, WithContext(ctx))...)
 }
 
-// ListWithContext calls /tags/list for the given repository, returning the list of tags
+// List calls /tags/list for the given repository, returning the list of tags
 // in the "tags" property.
-func ListWithContext(ctx context.Context, repo name.Repository, options ...Option) ([]string, error) {
+func List(repo name.Repository, options ...Option) ([]string, error) {
 	o, err := makeOptions(repo, options...)
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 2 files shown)


pkg/v1/remote/list.go
```diff
@@ -31,14 +31,16 @@ type tags struct {
...
-
 // ListWithContext calls /tags/list for the given repository, returning the list of tags
 // in the "tags" property.
+//
+// Deprecated: Use List and WithContext. This will be removed in a future release.
 func ListWithContext(ctx context.Context, repo name.Repository, options ...Option) ([]string, error) {
+	return List(repo, append(options, WithContext(ctx))...)
+}
+
+// List wraps ListWithContext using the background context.
```
jonjohnsonjr, line 42: Maybe update this.

pkg/v1/remote/transport/ping.go
```diff
@@ -80,17 +80,19 @@ func ping(ctx context.Context, reg name.Registry, t http.RoundTripper) (*pingRes
...
 		req, err := http.NewRequest(http.MethodGet, url, nil)
 		if err != nil {
 			return nil, err
 		}
 		resp, err := client.Do(req.WithContext(ctx))
 		if err != nil {
-			errs = append(errs, err.Error())
 			// Potentially retry with http.
-			continue
+			if i < len(schemes)-1 {
```
jonjohnsonjr, line 92: I think we want the dumb multi-err thing (maybe in a less dumb way) because it's really confusing otherwise.

pkg/v1/remote/list_test.go
```diff
@@ -110,8 +111,8 @@ func TestCancelledList(t *testing.T) {
 	}
 
 	_, err = ListWithContext(ctx, repo)
-	if want, got := context.Canceled, err; got != want {
-		t.Errorf("wanted %v got %v", want, got)
+	if err == nil || !strings.Contains(err.Error(), "context canceled") {
+		t.Errorf(`unexpected error; want "context canceled", got %v`, err)
```
jonjohnsonjr, line 115: Let's have the transport.New stuff just double-wrap its fallback errors so that we can use the magic go 1.13 error stuff.

jonjohnsonjr (approved):

PR #1093, opened 2021-07-29
Add static.NewLayer

imjasonh:
Mainly copied from [cosign](https://github.com/sigstore/cosign/blob/0fdfaa93446aefba72af08a43d2833dc5f5e2d08/pkg/cosign/remote/remote.go#L198), with basic tests, and a `sync.Once` around `Digest()`.

### DIFF
--- a/pkg/v1/static/layer.go
+++ b/pkg/v1/static/layer.go
@@ -25,6 +25,8 @@ import (
 )
 
 // NewLayer returns a layer containing the given bytes, with the given mediaType.
+//
+// Contents will not be compressed.
 func NewLayer(b []byte, mt types.MediaType) v1.Layer {
 	return &staticLayer{b: b, mt: mt}
 }
--- a/pkg/v1/static/layer.go
+++ b/pkg/v1/static/layer.go
@@ -1,0 +1,67 @@
+// Copyright 2018 Google LLC All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package static
+
+import (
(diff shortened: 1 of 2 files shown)


pkg/v1/static/layer.go
```diff
@@ -0,0 +1,67 @@
...
+	})
+	return l.h, err
+}
+
+func (l *staticLayer) DiffID() (v1.Hash, error) {
+	return l.Digest()
+}
+
+func (l *staticLayer) Compressed() (io.ReadCloser, error) {
+	return ioutil.NopCloser(bytes.NewReader(l.b)), nil
```
jonjohnsonjr, line 54: Maybe want to document on `NewLayer` that we don't do any compression?

pkg/v1/static/static_test.go
```diff
@@ -0,0 +1,84 @@
...
+	"github.com/google/go-containerregistry/pkg/v1/types"
+	"github.com/google/go-containerregistry/pkg/v1/validate"
+)
+
+func TestNewLayer(t *testing.T) {
+	b := []byte(strings.Repeat(".", 10))
+	l := NewLayer(b, types.OCILayer)
+
+	// This does basically nothing.
+	if err := validate.Layer(l, validate.Fast); err != nil {
```
jonjohnsonjr, line 32: Why fast? You can drop half this test if you do standard.

pkg/v1/static/static_test.go
```diff
@@ -0,0 +1,84 @@
...
+		t.Fatal(err)
+	}
+	h2, err := l.DiffID()
+	if err != nil {
+		t.Fatal(err)
+	}
+	if h != h2 {
+		t.Errorf("Digest != DiffID; digest is %v, diffid is %v", h, h2)
+	}
+	// sha256("..........")
```
jonjohnsonjr, line 48: Drop this comment?

pkg/v1/static/static_test.go
imjasonh:
```
    static_test.go:33: gzip: invalid header
```

jonjohnsonjr: Of course :(

jonjohnsonjr: We should maybe fix that to actually look at media types (for +gzip)

imjasonh: Do you want to do that? I'm fine just manually validating in this test for now.

jonjohnsonjr (approved):

PR #1097, opened 2021-08-02
add mutating annotations section to the recipes of the crane

developer-guy:


imjasonh (commented):
Thanks for suggesting this. This functionality is already pretty well covered in the generated docs: https://github.com/google/go-containerregistry/blob/main/cmd/crane/doc/crane_mutate.md

And in `crane mutate --help`, which comes from the same place.

This recipes.md doc is mainly to document things you can do with crane and other tools (jq, diff, etc.) or multiple nested invocations of crane in the case of mutate+append.

developer-guy:
Thanks for the explanation @imjasonh, I'm closing this 🙋🏻‍♂️

PR #1098, opened 2021-08-03
Add --set-base-image-annotations flag to crane append

imjasonh:
This flag instructs crane append to annotate the resulting image with
information about the base image (name and current digest). This flag is
false by default.

### DIFF
--- a/cmd/crane/cmd/append.go
+++ b/cmd/crane/cmd/append.go
@@ -63,7 +63,7 @@ func NewCmdAppend(options *[]crane.Option) *cobra.Command {
 				}
 				var baseName string
 				if _, ok := ref.(name.Tag); ok {
-					baseName = ref.String()
+					baseName = ref.Name()
 				}
 
 				baseDigest, err := base.Digest()
--- a/cmd/crane/cmd/append.go
+++ b/cmd/crane/cmd/append.go
@@ -22,6 +22,8 @@ import (
 	"github.com/google/go-containerregistry/pkg/name"
 	v1 "github.com/google/go-containerregistry/pkg/v1"
 	"github.com/google/go-containerregistry/pkg/v1/empty"
+	"github.com/google/go-containerregistry/pkg/v1/mutate"
+	specsv1 "github.com/opencontainers/image-spec/specs-go/v1"
 	"github.com/spf13/cobra"
 )
 
@@ -29,6 +31,7 @@ import (
 func NewCmdAppend(options *[]crane.Option) *cobra.Command {
 	var baseRef, newTag, outFile string
 	var newLayers []string
+	var annotate bool
 
 	appendCmd := &cobra.Command{
 		Use:   "append",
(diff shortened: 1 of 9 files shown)


cmd/crane/cmd/append.go
```diff
@@ -78,6 +97,7 @@ func NewCmdAppend(options *[]crane.Option) *cobra.Command {
 	appendCmd.Flags().StringVarP(&newTag, "new_tag", "t", "", "Tag to apply to resulting image")
 	appendCmd.Flags().StringSliceVarP(&newLayers, "new_layer", "f", []string{}, "Path to tarball to append to image")
 	appendCmd.Flags().StringVarP(&outFile, "output", "o", "", "Path to new tarball of resulting image")
+	appendCmd.Flags().BoolVar(&annotate, "annotate", false, "If true, annotate the resulting image as being based on the base image")
```
jonjohnsonjr, line 100: I know I suggested this flag name, but I'm a little worried this will get confused with `crane mutate`'s `annotation` flag... wdyt?

cmd/crane/cmd/append.go
```diff
@@ -53,6 +56,22 @@ func NewCmdAppend(options *[]crane.Option) *cobra.Command {
...
+				if err != nil {
+					return fmt.Errorf("parsing ref %q: %v", baseRef, err)
+				}
+
+				baseDigest, err := base.Digest()
+				if err != nil {
+					return err
+				}
+				img = mutate.Annotations(img, map[string]string{
+					specsv1.AnnotationBaseImageName:   ref.String(),
```
jonjohnsonjr, line 70: `baseRef` might be a digest, we don't necessarily have a tag here, right?

cmd/crane/cmd/append.go
imjasonh: `crane append -b ubuntu --set-base-image-annotations` ?

I don't mind being verbose.

cmd/crane/cmd/append.go
imjasonh: Yeah, what should we do in that case? (Technically `ko` has this problem too, you could do `defaultBaseImage: gcr.io/distroless/static@sha256:...`)

Given `crane append -b ubuntu@sha256:abc`, should we:
a.) just not set the `base.image.name` annotation, callers can maybe find value in the `base.image.digest` annotation by itself?
b.) not set either annotation, you don't get base image info 🤷 
c.) only set `base.image.name` to the fully-specified ref with digest, since this can still be useful for signaling base image info, even if it means rebasing/rebuilding won't be useful.

(we could maybe add this expected behavior to the spec)

jonjohnsonjr: If we don't have a tag, I think we should drop the `base.image.name` annotation and just set the digest annotation.

That's annoying for situations where you want to pull by digest (e.g. `ko`), but we could hack around that a little bit. If base is specified with both tag and digest as:

```
foo:v1@sha256:abc
```

Then we could know that `name=foo:v1` and `digest=sha256:abc`.

If we lean into that convention, we could even have `ko` log a warning or something if your base image is out of date before even pushing anything.

jonjohnsonjr (approved):

PR #1102, opened 2021-08-11
feat: ability to set page size for tags list and catalog calls

alexey-igrychev:
I recently found that the GitHub Container Registry doesn't support pagination, so I'm limited to 1000 tags when using the library (rel https://github.community/t/pagination-when-listing-image-tags-does-not-work/194586).

I've added an option that allows disabling the default limit of 1000 tags when requesting image tags.

### DIFF
--- a/pkg/v1/remote/options.go
+++ b/pkg/v1/remote/options.go
@@ -38,6 +38,7 @@ type options struct {
 	userAgent                      string
 	allowNondistributableArtifacts bool
 	updates                        chan<- v1.Update
+	pageSize                       int
 }
 
 var defaultPlatform = v1.Platform{
@@ -45,7 +46,13 @@ var defaultPlatform = v1.Platform{
 	OS:           "linux",
 }
 
-const defaultJobs = 4
+const (
+	defaultJobs = 4
+
+	// ECR returns an error if n > 1000:
+	// https://github.com/google/go-containerregistry/issues/1091
+	defaultPageSize = 1000
+)
 
 func makeOptions(target authn.Resource, opts ...Option) (*options, error) {
 	o := &options{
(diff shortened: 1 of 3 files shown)


pkg/v1/remote/options.go
```diff
@@ -193,3 +194,11 @@ func WithProgress(updates chan<- v1.Update) Option {
 		return nil
 	}
 }
+
+// WithNoTagsLimit disables request tags limit when listing image tags.
+//
+// The default behavior is a limit of 1000 tags.
+func WithNoTagsLimit(o *options) error {
```
jonjohnsonjr, line 201: Probably want something like "WithPageSize", where anything <= 0 indicates no size. Then we can have 1000 as the default page size.

alexey-igrychev: Yes, I agree; your approach covers more cases.
I have reworked the logic and added support for the catalog call.

pkg/v1/remote/options.go
```diff
@@ -38,6 +38,7 @@ type options struct {
 	userAgent                      string
 	allowNondistributableArtifacts bool
 	updates                        chan<- v1.Update
+	pageSize                       int
```
jonjohnsonjr, line 41: Set this to 1000 in `makeOptions` below, then we can drop the `else` clause in `list.go` and `catalog.go`.

If a user doesn't set page size, it defaults to 1000. 
If a user sets page size <= 0, we omit `n` altogether.
If a user sets page size > 0, we'll use that.

pkg/v1/remote/options.go
```diff
@@ -193,3 +194,13 @@ func WithProgress(updates chan<- v1.Update) Option {
 		return nil
 	}
 }
+
+// WithPageSize sets the given size as the value of parameter 'n' in the request.
+//
+// The default page size is 1000 for list (/tags/list) and catalog (/_catalog) calls.
+func WithPageSize(size int) Option {
```
jonjohnsonjr, line 201:
```
// To omit the `n` parameter entirely, use WithPageSize(0).

jonjohnsonjr:
> I apologise for my inattention

No problem, there's a lot of indirection going on, it's easy to get lost in it 😄

jonjohnsonjr (approved):
Thanks!

Issue #1109, opened 2021-08-20
Support crane push tarBall contains more than one image?


dashjay:
when I push a tarball with only one image in it, crane works very well, but when I push a tarball contains more than one image,
error throw out.
```
crane push /tmp/bundle.tar docker-registry.com
Error: loading /tmp/bundle.tar as tarball: tarball must contain only a single image to be used with tarball.Image
exit status 1
```
Instead, pull can work for more than one image with `crane pull image:tag image2:tag ... /tmp/bundle.tar`

I don't know if there is any other considerations that block this.

jonjohnsonjr:
We're missing some library functionality that would make this a lot easier to implement.

Right now we have [`tarball.Write`](https://github.com/google/go-containerregistry/blob/de8aff85c9090d37ed6fe436d696f3a820e5c767/pkg/v1/tarball/write.go#L68) and its dual [`tarball.Image`](https://github.com/google/go-containerregistry/blob/de8aff85c9090d37ed6fe436d696f3a820e5c767/pkg/v1/tarball/image.go#L90).

We also have [`tarball.MultiWrite`](https://github.com/google/go-containerregistry/blob/de8aff85c9090d37ed6fe436d696f3a820e5c767/pkg/v1/tarball/write.go#L77) that writes multiple images into a tarball, but we _don't_ have anything that reads multiple images from a tarball.

I think I'd like to see something like `tarball.Images` that returns a `map[name.Reference]v1.Image`.

You _can_ hack around this with [`tarball.LoadManifest`](https://github.com/google/go-containerregistry/blob/de8aff85c9090d37ed6fe436d696f3a820e5c767/pkg/v1/tarball/image.go#L74) but that's not nearly as nice.

dashjay:
Due to urgent needs, I wrote one version with registry v2 http last week 😄 . Now I am trying to rewrite it with internal pkg and struct, maybe I will make pr soon if I made it.

afbjorklund:
We had a lot of historical problems, when using `podman` to load tarballs wtih multiple images.

These should all be solved now (podman3), but decided to stick with one image per tarball....
(the workaround was to pre-load all the images, and then save the entire `/var` as a snapshot)

If I recall correctly, there was no support in OCI for having multiple images per archive ?

jonjohnsonjr:
> If I recall correctly, there was no support in OCI for having multiple images per archive ?

[...]

PR #1110, opened 2021-08-22
Feat(crane) support push tarball contains more than one image

dashjay:
[Issue here: Support crane push tarBall contains more than one image?](https://github.com/google/go-containerregistry/issues/1109)
This is my first contribution to the open source repo. Bear with me 🥺 .

- [x] add command mpush for pushing tarball contains more than one image
- [x] merge command mpush and command push

.....Please don't hesitate to ask any questions that may exist

### DIFF
--- a/cmd/crane/cmd/push.go
+++ b/cmd/crane/cmd/push.go
@@ -41,7 +41,7 @@ func NewCmdPush(options *[]crane.Option) *cobra.Command {
 			}
 			images, err := crane.LoadMulti(path, *options...)
 			if err != nil {
-				return fmt.Errorf("loading %s as tarball: %#v", path, err)
+				return fmt.Errorf("loading %s as tarball: %v", path, err)
 			}
 			return crane.MultiPush(images, target, concurrent, *options...)
 		},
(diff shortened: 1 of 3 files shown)


dashjay:
@googlebot I signed it!

dashjay:
@googlebot I fixed it.

dashjay:
I want to know why this function [remote.MultiWrite](https://github.com/google/go-containerregistry/blob/de8aff85c9090d37ed6fe436d696f3a820e5c767/pkg/v1/remote/multi_write.go#L44) says that all refs must share the same repository. I can not understand it... so I define a func named WriteMulti

dashjay:
`crane push TARBALL image:tag` can push a tarball contain one image to specific tag. but now it can just only change the registry. like this:
```
[registry]/[repository]:[tag] -> [new-registry]/[repository]:[tag]
```

you can not change the repository and tag...

jonjohnsonjr:
Hey sorry I missed this PR. The idea in general is good. I don't think we need to add a new method to `remote` -- we could just iterate over the map in `crane` and push each individually. We can optimize that later.

Want to start with just a PR to add the `MultiImage` function to the `tarball` package? I have some feedback but don't want to have to churn too much until we're happy with that function.

dashjay:
@jonjohnsonjr I didn't mean any offense, I will do like what you said(remove the new method in remote and just iter the map ~), thank you for review again.
Two weeks ago, I closed it because I found that, there is something wrong with the auth. I can not WriteMulti to registry with Self-signed SSL. I read all code but can not resolve it.

Issue #1113, opened 2021-09-01
Change logic around whether an error is 'Temporary'


DennisDenuto:
Currently, if an [actionable error](https://github.com/distribution/distribution/blob/main/docs/spec/api.md#errors)  is returned by the registry, that will determine whether to retry (and takes precedence over the http status code returned). See https://github.com/google/go-containerregistry/blob/main/pkg/v1/remote/transport/error.go#L91 for the code logic i’m talking about.

We have encountered a scenario where a registry (flakey) sometimes returns a `500` status code error, *and* also returns an actionable error of `UNKNOWN`

```
{"errors":[{"code":"UNKNOWN","message":"internal server error"}]}
```

This results in gcr *not* retrying. (However, retrying in this case *would* have succeeded)

Reading the [distribution spec](https://github.com/distribution/distribution/blob/main/docs/spec/api.md#errors), it sounds like actionable errors should be produced by status codes of `4xx`. 

If the above is true, what are your thoughts on instead retrying based on the http status code (even if an actionable error was provided) if the http status code is *not* `4xx`?

i.e. change the condition to be this:

```
 if (e.StatusCode < http.StatusBadRequest || e.StatusCode >= http.StatusInternalServerError) || len(e.Errors) == 0
```

Or, allowing the consumer of the library the ability to configure the conditions which mark whether an error is ‘Temporary’?

jonjohnsonjr:
> We have encountered a scenario where a registry (flakey) sometimes returns a 500 status code error, and also returns an actionable error of UNKNOWN
> 
> `{"errors":[{"code":"UNKNOWN","message":"internal server error"}]}`

Ahh, now I understand your point, sorry I was confused.

Yes, this is a problem. Let's add `UNKNOWN` to [this list](https://github.com/google/go-containerregistry/blob/7e0ed51a7bb1930bfb6859f319975b578ed04de2/pkg/v1/remote/transport/error.go#L138-L157) as well.

[...]

PR #1115, opened 2021-09-02
Change logic around whether an error is 'Temporary'

DennisDenuto:
Related to: https://github.com/google/go-containerregistry/issues/1113 

Keeping this PR in draft since there are other details to discuss in #1113 

Authored-by: Dennis Leon <leonde@vmware.com>


jonjohnsonjr (approved):

DennisDenuto:
@jonjohnsonjr 

Thanks!! would you happen to have a ballpark estimate on when the next ggcr release will be? 

We really really want this change to help with test flakiness we are experiencing and to include it in our next release.

Issue #627, opened 2019-12-10
getting image ID (sha) takes longer than docker cli


medyagh:
I am trying to find a cheap way to get the image SHA, if I do docker inspect it takes 0.07 s

```
$ time docker inspect --format='{{.Id}}' ubuntu
sha256:775349758637aff77bf85e2ff0597e86e3e859183ef0baba8b3e8fc8d3cba51c

real	0m0.073s
user	0m0.035s
sys	0m0.033s

```

however if I use go-containerregistry it will take 4s,

```
ref, _ := name.ParseReference("ubuntu", name.WeakValidation)
img, _ := daemon.Image(ref)
m, _ := img.Manifest()
```

 is there any cheaper way to get the image ID using go-containerregistry ?

I also tried crane.Digest that still takes 2.1 seconds.
```
fmt.Println(crane.Digest("ubuntu"))
```

Update: I also tried getting it through config but it is still 1.95 seconds

code:
```
	ref, _ := name.ParseReference("ubuntu", name.WeakValidation)
	img, _ := daemon.Image(ref)
	cf, _ := img.ConfigName()
	fmt.Println(cf)
```
time
```
$ time ./example1 
sha256:775349758637aff77bf85e2ff0597e86e3e859183ef0baba8b3e8fc8d3cba51c

real    0m1.954s
```

This would be an important improvment for iterative development on minikube. when users constantly rebuild an image and push.

jonjohnsonjr:
> is there any cheaper way to get the image ID using go-containerregistry ?

It might make sense to use the docker client directly? There's [`ImageInspectWithRaw`](https://godoc.org/github.com/docker/docker/client#Client.ImageInspectWithRaw), which I believe does what you want. We just thinly wrap the docker client library to make access slightly easier. Our `daemon` package isn't particularly optimized, because I mostly use it to load images into the daemon.

The reason this is really slow is that our `daemon.Image` implementation is very naive. We essentially take the output of `docker save` and parse it as a `tarball.Image`, which is really inefficient if all you care about is the Image ID.

Something like this should work:
```go
func getImageID(ctx context.Context, ref string) (string, error) {
	cli, err := client.NewClientWithOpts(client.FromEnv)
	if err != nil {
		return "", err
	}
	cli.NegotiateAPIVersion(ctx)
	img, err := cli.ImageInspectWithRaw(ctx, ref)
	if err != nil {
		return "", err
	}
	return img.ID, nil
}
```

[...]

PR #1121, opened 2021-09-08
feat(daemon): lazy image saving

knqyf263:
Fix https://github.com/google/go-containerregistry/issues/627

A faster way of getting an image ID is really useful for caching. We don't want to call `docker save` if we already have the image information in the cache. For example, in our case, we should skip scanning vulnerabilities of the image when the scan result exists in the cache. The cache key is an image ID, so it is enough to call `docker inspect` to know only the image ID.

This implementation is based on `computed()` as below.
https://github.com/google/go-containerregistry/blob/45aaa6c000ebf2acacad9beef7c82667ec51662c/pkg/v1/mutate/image.go#L51

The downside is that `daemon.Image` doesn't return an error and it might return an error when each method is called (https://github.com/knqyf263/go-containerregistry/commit/881e16e70fdf797d2b68b2e5267639d4dafaca8c).

I probably should add tests, but I'd like to hear your thought on this idea before working on tests 🙇 

I also want to replace `ConfigFile` with this approach in another PR because I think we can generate the config file from the result of `docker inspect` and `docker history`.

### DIFF
--- a/pkg/v1/daemon/image.go
+++ b/pkg/v1/daemon/image.go
@@ -185,5 +185,5 @@ func (i image) LayerByDiffID(h v1.Hash) (v1.Layer, error) {
 	if err := i.initialize(); err != nil {
 		return nil, err
 	}
-	return i.tarballImage.LayerByDigest(h)
+	return i.tarballImage.LayerByDiffID(h)
 }
--- a/pkg/v1/daemon/image.go
+++ b/pkg/v1/daemon/image.go
@@ -24,7 +24,14 @@ import (
 	"github.com/google/go-containerregistry/pkg/name"
 	v1 "github.com/google/go-containerregistry/pkg/v1"
 	"github.com/google/go-containerregistry/pkg/v1/tarball"
+	"github.com/google/go-containerregistry/pkg/v1/types"
 )
+
+type image struct {
+	ref          name.Reference
+	opener       *imageOpener
+	tarballImage v1.Image
+}
 
 type imageOpener struct {
 	ref name.Reference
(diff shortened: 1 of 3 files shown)


pkg/v1/daemon/image.go
```diff
@@ -85,5 +92,98 @@ func Image(ref name.Reference, options ...Option) (v1.Image, error) {
...
+		return nil, err
+	}
+	return i.tarballImage.LayerByDigest(h)
+}
+
+func (i image) LayerByDiffID(h v1.Hash) (v1.Layer, error) {
+	if err := i.initialize(); err != nil {
+		return nil, err
+	}
+	return i.tarballImage.LayerByDigest(h)
```
jonjohnsonjr, line 188: Should be `LayerByDiffID`

jonjohnsonjr:
LGTM in general other than the one bug. I would like to see a test (you should be able to catch most issues by calling `validate.Image`) 👍

pkg/v1/daemon/image.go
knqyf263: Thanks. Fixed https://github.com/google/go-containerregistry/blob/bb5ea5efd13b203c12b6a174e2e9e168cb1032a5/pkg/v1/daemon/image.go

knqyf263:
@jonjohnsonjr Thanks for the review! I fixed the bug and added `validate.Image`. Could you take a look again?

jonjohnsonjr (approved):

jonjohnsonjr:
Thanks!

Issue #1114, opened 2021-09-01
Allow configuring the retry attempts


DennisDenuto:
Currently the number of retries gcr attempts is hardcoded to `3`. See https://github.com/vmware-tanzu/carvel-imgpkg/blob/15b20e52873608cd5d3699d2c4e2ca61d04b3511/vendor/github.com/google/go-containerregistry/pkg/v1/remote/write.go#L423-L428

In environments where the network is super flakey, this number is too low. 

It would be great if the consumer of this library can configure the `backoff` values

Assuming the above seems reasonable, I'd be happy to submit a PR.

tkrishtop:
Hi @DennisDenuto could you please provide an example of how to use your code to configure the retries? 

I'm trying to do this: https://go.dev/play/p/kyxdXxCAxyN

but have the undefined errors
```
$ go run main.go 
# command-line-arguments
./main.go:10:31: undefined: crane.Backoff
./main.go:18:9: undefined: crane.WithRetryBackoff
``` 

But it seems like both `Backoff` and `WithRetryBackoff` are supposed to be exposed

https://github.com/google/go-containerregistry/blob/main/pkg/v1/remote/options.go#L278-L292

```
// Backoff is an alias of retry.Backoff to expose this configuration option to consumers of this lib
type Backoff = retry.Backoff

...

// WithRetryPredicate sets the predicate for retry HTTP operations.
func WithRetryPredicate(predicate retry.Predicate) Option {
	return func(o *options) error {
		o.retryPredicate = predicate
		return nil
	}
}
```

imjasonh:
[`WithRetryPredicate`](https://pkg.go.dev/github.com/google/go-containerregistry/pkg/v1/remote#WithRetryBackoff) is in `pkg/v1/remote`, not `pkg/crane`.

You can set this as a crane option using:

```
	options := []crane.Option{
		Remote: []remote.Option{
			remote.WithRetryBackoff(customRetryBackoff),
		},
	}
```

tkrishtop:
Thank you @imjasonh worked like a magic!

```
package main

import (
	"log"
	"time"

	"github.com/google/go-containerregistry/pkg/crane"
	"github.com/google/go-containerregistry/pkg/v1/remote"
)

func withRemoteOption(ro remote.Option) crane.Option {
	return func(o *crane.Options) {
		o.Remote = append(o.Remote, ro)
	}
}

func main() {
	customRetryBackoff := &remote.Backoff{
[...]

Issue #666, opened 2020-01-30
Requires authentication even for public registries


samos123:
Currently when using a public image from GCR it will fail with trying to authenticate with Google Cloud. However for public images that shouldn't be needed.

Steps to reproduce:
1. Remove any authentication with gcloud. E.g. `mv ~/.config/gcloud{,.bak}` 
2. Use go-containerregistry to pull an image from a public GCR repo
```
crane pull gcr.io/distroless/base-debian10 test.tgz
WARNING: Could not open the configuration file: [/home/stoelinga/.config/gcloud/configurations/config_default].
ERROR: (gcloud.auth.docker-helper) You do not currently have an active account selected.
Please run:

  $ gcloud auth login

to obtain new credentials, or if you have already logged in with a
different account:

  $ gcloud config set account ACCOUNT

to select an already authenticated account to use.
2020/01/29 16:40:05 error getting credentials - err: exit status 1, out: ``
```

Current result:
Failure telling you to authenticate with google cloud

Expected result:
Successfully pull down the image

jonjohnsonjr:
Seems like you've configured a [credential helper](https://github.com/google/go-containerregistry/tree/master/pkg/authn#credential-helpers) in ~/.docker/config.json (or $DOCKER_CONFIG) but it isn't set up properly.

We do eagerly get credentials for a registry, which is the cause of this error. If we deferred fetching creds until we actually got a 401, we could avoid this... which is definitely something we should do.

tejal29:
@samos123, i ended up removing `~/.docker/config.json and `~/.config/gcloud`
```
$mv ~/.config/gcloud ~/Desktop

$ls ~/.config/
configstore	gbackup		hub
$ls ~/.config/gbackup/
exclude		exclude.d	include

 $crane pull gcr.io/distroless/base-debian10 test.tgz
2020/01/29 16:59:55 No matching credentials were found, falling back on anonymous

$ls -alh test.tgz 
-rw-r--r--  1 tejaldesai  primarygroup   8.3M Jan 29 17:02 test.tgz

```

samos123:
interesting then it must be something in Kaniko that we're doing incorrectly in relation to this library since on Travis CI, i did not configure the gcloud config helper but it would still fail with a similar message that authentication with gcloud was needed.

filesnate:
In the kaniko image, there exists the file `/kaniko/.docker/config.json` with the following contents.
```
/kaniko/.docker # more config.json 
{
        "auths": {},
[...]

PR #1122, opened 2021-09-09
Make retrying transport and http errors configurable 

DennisDenuto:
Related to https://github.com/google/go-containerregistry/issues/1114

Adds the following 4 options:
- WithRetryHTTPBackoff
- WithRetryHTTPPredicate
- WithRetryTransportBackoff
- WithRetryTransportPredicate

Went with a finer level API around configuring these (in my head) 2 groups of retry logic (low-level transport errors and higher level HTTP errors).

I was also aiming to keep the 'default' retry logic values the same (if these options are not provided).

### DIFF
--- a/pkg/v1/remote/options.go
+++ b/pkg/v1/remote/options.go
@@ -118,7 +118,7 @@ func makeOptions(target authn.Resource, opts ...Option) (*options, error) {
 		}
 
 		// Wrap the transport in something that can retry network flakes.
-		o.transport = transport.NewRetry(o.transport, transport.WithRetryBackoff(o.retryBackoff), transport.WithRetryPredicate(o.retryPredicate))
+		o.transport = transport.NewRetry(o.transport)
 
 		// Wrap this last to prevent transport.New from double-wrapping.
 		if o.userAgent != "" {
--- a/pkg/v1/remote/options.go
+++ b/pkg/v1/remote/options.go
@@ -17,8 +17,12 @@ package remote
 import (
 	"context"
 	"errors"
+	"io"
 	"net/http"
+	"syscall"
+	"time"
 
+	"github.com/google/go-containerregistry/internal/retry"
 	"github.com/google/go-containerregistry/pkg/authn"
 	"github.com/google/go-containerregistry/pkg/logs"
 	v1 "github.com/google/go-containerregistry/pkg/v1"
@@ -39,11 +43,34 @@ type options struct {
 	allowNondistributableArtifacts bool
 	updates                        chan<- v1.Update
 	pageSize                       int
+	retryBackoff                   Backoff
+	retryPredicate                 retry.Predicate
 }
 
 var defaultPlatform = v1.Platform{
 	Architecture: "amd64",
 	OS:           "linux",
+}
+
+// Backoff is an alias of retry.Backoff to expose this configuration option to consumers of this lib
+type Backoff = retry.Backoff
+
+var defaultRetryPredicate retry.Predicate = func(err error) bool {
+	// Various failure modes here, as we're often reading from and writing to
+	// the network.
+	if retry.IsTemporary(err) || errors.Is(err, io.ErrUnexpectedEOF) || errors.Is(err, syscall.EPIPE) {
+		logs.Warn.Printf("retrying %v", err)
+		return true
+	}
+	return false
+}
+
+// Try this three times, waiting 1s after first failure, 3s after second.
+var defaultRetryBackoff = Backoff{
+	Duration: 1.0 * time.Second,
+	Factor:   3.0,
+	Jitter:   0.1,
+	Steps:    3,
 }
 
 const (
@@ -56,12 +83,14 @@ const (
 
 func makeOptions(target authn.Resource, opts ...Option) (*options, error) {
 	o := &options{
-		auth:      authn.Anonymous,
-		transport: http.DefaultTransport,
-		platform:  defaultPlatform,
-		context:   context.Background(),
-		jobs:      defaultJobs,
-		pageSize:  defaultPageSize,
+		auth:           authn.Anonymous,
+		transport:      http.DefaultTransport,
+		platform:       defaultPlatform,
+		context:        context.Background(),
+		jobs:           defaultJobs,
+		pageSize:       defaultPageSize,
+		retryPredicate: defaultRetryPredicate,
+		retryBackoff:   defaultRetryBackoff,
 	}
 
 	for _, option := range opts {
@@ -87,7 +116,7 @@ func makeOptions(target authn.Resource, opts ...Option) (*options, error) {
 		}
 
 		// Wrap the transport in something that can retry network flakes.
-		o.transport = transport.NewRetry(o.transport)
+		o.transport = transport.NewRetry(o.transport, transport.WithRetryBackoff(o.retryBackoff), transport.WithRetryPredicate(o.retryPredicate))
 
 		// Wrap this last to prevent transport.New from double-wrapping.
 		if o.userAgent != "" {
@@ -211,6 +240,22 @@ func WithProgress(updates chan<- v1.Update) Option {
 func WithPageSize(size int) Option {
 	return func(o *options) error {
 		o.pageSize = size
+		return nil
+	}
+}
+
+// WithRetryBackoff sets the httpBackoff for retry HTTP operations.
+func WithRetryBackoff(backoff Backoff) Option {
+	return func(o *options) error {
+		o.retryBackoff = backoff
+		return nil
+	}
+}
+
+// WithRetryPredicate sets the predicate for retry HTTP operations.
+func WithRetryPredicate(predicate retry.Predicate) Option {
+	return func(o *options) error {
+		o.retryPredicate = predicate
 		return nil
 	}
 }
--- a/pkg/v1/remote/options.go
+++ b/pkg/v1/remote/options.go
@@ -107,7 +107,9 @@ func makeOptions(target authn.Resource, opts ...Option) (*options, error) {
 		o.auth = auth
 	}
 
-	if _, ok := o.transport.(*transport.Transport); !ok {
+	// transport.Wrapper is a signal that consumers are opt-ing into providing their own transport without any additional wrapping.
+	// This is to allow consumers full control over the transports logic, such as providing retry logic.
+	if _, ok := o.transport.(*transport.Wrapper); !ok {
 		// Wrap the transport in something that logs requests and responses.
 		// It's expensive to generate the dumps, so skip it if we're writing
 		// to nothing.
@@ -129,6 +131,8 @@ func makeOptions(target authn.Resource, opts ...Option) (*options, error) {
 
 // WithTransport is a functional option for overriding the default transport
 // for remote operations.
+// If transport.Wrapper is provided, this signals that the consumer does *not* want any further wrapping to occur.
+// i.e. logging, retry and useragent
 //
 // The default transport its http.DefaultTransport.
 func WithTransport(t http.RoundTripper) Option {
--- a/pkg/v1/remote/options.go
+++ b/pkg/v1/remote/options.go
@@ -78,7 +78,7 @@ func makeOptions(target authn.Resource, opts ...Option) (*options, error) {
 		o.auth = auth
 	}
 
-	if _, ok := o.transport.(*transport.Transport); ok {
+	if _, ok := o.transport.(*transport.Transport); !ok {
 		// Wrap the transport in something that logs requests and responses.
 		// It's expensive to generate the dumps, so skip it if we're writing
 		// to nothing.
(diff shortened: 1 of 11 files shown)


jonjohnsonjr:
So I don't love how many options this adds, but I get the need for this to be more configurable.

For the transport pieces, I think I'd like to revisit this: https://github.com/google/go-containerregistry/issues/740

Ideally, you could supply your own transport via `remote.WithTransport` and use `transport.NewRetry` to configure it. The only issue is around double-wrapping because of how we wrap transports by default. One option here is to just not wrap the supplied transport if you pass in a custom transport, but then you'd lose out on some "niceness". If we exposed [this stuff](https://github.com/google/go-containerregistry/blob/8388fde8bcadbc8ddb820b5ac0046e10e6c572e7/pkg/v1/remote/options.go#L81-L94) in a nice way, you could claw that back by opting into it, e.g.:

```go
// Wraps with retries and useragent and debug logging
remote.Image(..., remote.WithTransport(remote.Transport{t}))

// No wrapping of t
remote.Image(..., remote.WithTransport(t))

// Configurable backoff stuff.
remote.Image(..., remote.WithTransport(transport.NewRetry(t, ...)))
```

My only issue with that approach is that it's probably a breaking change :/

Another approach might be exposing more on the errors we return from `transport`, e.g. the request: https://github.com/google/go-containerregistry/blob/8388fde8bcadbc8ddb820b5ac0046e10e6c572e7/pkg/v1/remote/transport/error.go#L52

That would allow you to write a predicate that checked the request method, URL, and status code yourself to determine if you should retry it.

DennisDenuto:
@jonjohnsonjr 

Yeah I think iterating over the API change to make it as minimal as needed makes lots of sense. Both from maintainability and from a usability perspective! With the changes above sounds like a trade-off between a simpler API vs breaking changes to existing APIs.

I played around with the idea of using `WithTransport` as a way to allow a consumer to configure retrying.

> Another approach might be exposing more on the errors we return from transport

I did this by creating a new `RetryError` that wraps the transport error (if it exists), and exposes the status code and request.
I also made it implement the `Temporary` function to make it a compatible with the ['default' transport predicate](https://github.com/google/go-containerregistry/blob/main/pkg/v1/remote/transport/retry.go#L67)

I pushed the changes. I'd love some early feedback to make sure i'm properly understanding your idea to begin with!

This PR comes with breaking changes:
- Using `WithTransport(t)` with a non `remote.Transport` no longer wraps retry, logging etc...
- Retrying with the transport.Retry will return a `RetryError` (even if the `RoundTripper` did not return an error). It always returns a `RetryError` since it contains http data (status code) to allow specifying a `Predicate` to determine whether to retry or not.

Note:

Some tests are failing due to these breaking changes. But I wanted to get feedback before proceeding down this path.

jonjohnsonjr:
I don't think we need `RetryError` if we just expose the existing request field on the `transport.Error` struct. Callers can check if an error is `transport.Error` and use that information. If it's some other kind of error, they can just do that in an "else" rather than using `Inner`.

> Using WithTransport(t) with a non remote.Transport no longer wraps retry, logging etc...

This feels kinda backwards. I think if someone gives us a `transport.Transport`, we should avoid wrapping it at all. We could use that as a signal that the caller knows what they're doing and we should get out of their way.

It would still make sense to me to expose Backoff/Predicate options in `remote` for higher level things that cannot be retried at a request level, e.g. layer uploads. I'm not sure how I would want to structure those hooks, but there are a handful of places that might make sense to inject some kind of retry check.

I'd expect the default predicate to be somewhat conservative, but having the option would let callers do some pretty flexible things like only retry for certain methods or for certain paths.

DennisDenuto:
> if we just expose the existing request field on the transport.Error

hmmm. I don't see how `transport.Error` ever gets returned by a RoundTripper. I *can* however see how it is returned at the http req/resp layer via `CheckError(...)`  

or put another way...

I don't see how a consumer providing their own Retry Transport would ever get an error of type `transport.Error`

Although I see how exposing the `Request` on the `transport.Error` can be useful in conjunction with exposing Backoff/Predicate options in remote for higher level http retries.

> I think if someone gives us a transport.Transport, we should avoid wrapping it at all.

hmm ok, I mis-understood your code snippet above then

```
// Wraps with retries and useragent and debug logging
remote.Image(..., remote.WithTransport(remote.Transport{t}))
```

but thinking about it more, it does make sense that it should *not* wrap, and i think will result in this not being a breaking change. i'll def update the PR with this.

 > It would still make sense to me to expose Backoff/Predicate options in `remote`

I think this helps clarify things a lot for me.

I was trying to have `transport.Retry` provide a way to retry both low level transport errors and http level errors. I updated the PR to put back Backoff/Predicate options in remote :-)

jonjohnsonjr:
> hmm ok, I mis-understood your code snippet above then

Yeah I _didn't_ love my code snippet, but what you've got here is actually a better solution, I think.

> I think this helps clarify things a lot for me.
> 
> I was trying to have transport.Retry provide a way to retry both low level transport errors and http level errors. I updated the PR to put back Backoff/Predicate options in remote :-)

Ah yeah, sorry this wasn't more clear. There were two categories of retry stuff going on, but we kind of lumped them together. Current PR is pretty close to optimal, I think.

pkg/v1/remote/transport/transport.go
```diff
@@ -101,3 +101,14 @@ func NewWithContext(ctx context.Context, reg name.Registry, auth authn.Authentic
 		return nil, fmt.Errorf("unrecognized challenge: %s", pr.challenge)
 	}
 }
+
+// Transport results in *not* wrapping supplied transport with additional logic such as retries, useragent and debug logging
+// Consumers are opt-ing into providing their own transport without any additional wrapping.
+type Transport struct {
```
jonjohnsonjr, line 107: I think all that's left is to make the `transport` package actually return these. 

I don't love the `transport.Transport` name. Maybe `transport.Wrapper` or something?

Another issue is that it's kind of a pain to actually construct a transport outside of the remote package (you need to know where layers came from, where you're pushing, etc). We could make that a lot more ergonomic by introducing a new transport implementation that defers those decisions until they're needed, like containerd (https://github.com/google/go-containerregistry/issues/666#issuecomment-580034491).

pkg/v1/remote/options.go
```diff
@@ -78,19 +107,21 @@ func makeOptions(target authn.Resource, opts ...Option) (*options, error) {
 		o.auth = auth
 	}
 
-	// Wrap the transport in something that logs requests and responses.
-	// It's expensive to generate the dumps, so skip it if we're writing
-	// to nothing.
-	if logs.Enabled(logs.Debug) {
-		o.transport = transport.NewLogger(o.transport)
-	}
+	if _, ok := o.transport.(*transport.Transport); !ok {
```
jonjohnsonjr, line 110: Add a comment here and in `google` describing _why_ we're doing things this way.

Also for the `WithTransport` option, we'll want to describe this workaround.

DennisDenuto: done!

pkg/v1/remote/transport/transport.go
DennisDenuto: Yep! changed it to `transport.Wrapper`

>  make the transport package actually return these.

I wasn't 100% sure what you meant here. I took this to mean `NewWithContext` should return `transport.Wrapper`. I pushed the change.

> Another issue is that it's kind of a pain to actually construct a transport outside of the remote package (you need to know where layers came from, where you're pushing, etc). We could make that a lot more ergonomic by introducing a new transport implementation that defers those decisions until they're needed, like containerd (#666 (comment)).

hmm ok, is it possible to split introducing a new deferring transport implementation into a separate PR? Or do you see this as a blocker to this PR?

jonjohnsonjr: > I wasn't 100% sure what you meant here. I took this to mean NewWithContext should return transport.Wrapper. I pushed the change.

Yep, that's perfect.

> hmm ok, is it possible to split introducing a new deferring transport implementation into a separate PR? Or do you see this as a blocker to this PR?

Not a blocker, just thinking out loud.

pkg/v1/remote/options.go
```diff
@@ -78,26 +107,32 @@ func makeOptions(target authn.Resource, opts ...Option) (*options, error) {
...
 
-	// Wrap the transport in something that logs requests and responses.
-	// It's expensive to generate the dumps, so skip it if we're writing
-	// to nothing.
-	if logs.Enabled(logs.Debug) {
-		o.transport = transport.NewLogger(o.transport)
-	}
+	// transport.Wrapper is a signal that consumers are opt-ing into providing their own transport without any additional wrapping.
+	// This is to allow consumers full control over the transports logic, such as providing retry logic.
+	if _, ok := o.transport.(*transport.Wrapper); !ok {
```
jonjohnsonjr, line 112: This is going to be tricky to maintain -- can you add a test that _doesn't_ get wrapped so we know this works?

DennisDenuto: Finding a good place to test this wasn't a clear call.

Ended up writing a couple tests in the `MultiWrite` test:
-  Asserts that if a `transport.Wrapper` is used, then retrying should *not* occur. 
-  Asserts that if a `transport.Wrapper` is used, then User-Agent should not be wrapped..

DennisDenuto:
@jonjohnsonjr Just checking in to see if theres anything else needed to have this merged in?

jonjohnsonjr:
Sorry I missed your follow-up commits!

One final thing I forgot to mention. If we're dealing with a `transport.Wrapper`, we probably also want to skip all of this:
https://github.com/google/go-containerregistry/blob/8388fde8bcadbc8ddb820b5ac0046e10e6c572e7/pkg/v1/remote/transport/transport.go#L53-L68

We can do a little introspection to determine if the the given transport can be reused by looking at `reg` to see if it's the same as what `t` already pinged. Then we can skip some pinging. For the bearer case, we probably want to merge the existing scopes with the given `scopes` like with: https://github.com/google/go-containerregistry/blob/8388fde8bcadbc8ddb820b5ac0046e10e6c572e7/pkg/v1/remote/transport/bearer.go#L99-L105

DennisDenuto:
Thanks I appreciate the pointers @jonjohnsonjr 

> We can do a little introspection to determine if the the given transport can be reused by looking at reg to see if it's the same as what t already pinged

I did this by saving state ('pingedRegistries') in `transport.Wrapper`

> For the bearer case, we probably want to merge the existing scopes with the given scopes like with:

I figured that the scopes should be tied to the registry that the transport pinged against. Hence scopes being a map on the wrapped transport.

jonjohnsonjr:
Okay so this might make you hate me, but I'm not thrilled with how this turned out after my suggestion (I know... I'm sorry).

Do you mind backing this out to the previous revision? We can just submit because it's useful on its own, then I can revisit the ping optimizations when I have more time to look at it.

Again, sorry, I appreciate your patience.

DennisDenuto:
@jonjohnsonjr  lol no worries. i've reverted that commit

fwiw I have enjoyed working on this PR and we get a ton of value from using this library. so more than happy to be patient and get things done right! 😄

pkg/v1/remote/options.go
```diff
@@ -78,26 +107,32 @@ func makeOptions(target authn.Resource, opts ...Option) (*options, error) {
...
+		// It's expensive to generate the dumps, so skip it if we're writing
+		// to nothing.
+		if logs.Enabled(logs.Debug) {
+			o.transport = transport.NewLogger(o.transport)
+		}
 
-	// Wrap the transport in something that can retry network flakes.
-	o.transport = transport.NewRetry(o.transport)
+		// Wrap the transport in something that can retry network flakes.
+		o.transport = transport.NewRetry(o.transport, transport.WithRetryBackoff(o.retryBackoff), transport.WithRetryPredicate(o.retryPredicate))
```
jonjohnsonjr, line 121: I'm not sure if we want to plumb this down into `transport.NewRetry` -- but maybe you can convince me.

The defaults for that transports are geared towards smaller blips (maybe this makes sense, maybe it doesn't), and I'd expect the `remote` options are more geared towards higher level retries (i.e. at the HTTP level).

pkg/v1/remote/transport/transport.go
```diff
@@ -96,8 +96,19 @@ func NewWithContext(ctx context.Context, reg name.Registry, auth authn.Authentic
...
+		return &Wrapper{bt}, nil
 	default:
 		return nil, fmt.Errorf("unrecognized challenge: %s", pr.challenge)
 	}
 }
+
+// Wrapper results in *not* wrapping supplied transport with additional logic such as retries, useragent and debug logging
+// Consumers are opt-ing into providing their own transport without any additional wrapping.
+type Wrapper struct {
+	Inner http.RoundTripper
```
jonjohnsonjr, line 108: Let's not expose `Inner` quite yet? I think we only want you to be able to produce a `Wrapper` by going through `transport.New` for now. We can relax that later if it makes sense.

DennisDenuto: ok done!

jonjohnsonjr (approved):
Thanks for sticking with this. This looks good to me -- let's merge and let it bake for a bit while I try to figure out how we can get rid of extra pings :)

PR #1124, opened 2021-09-18
Make `mutate` constructs immutable.

mattmoor:
I noticed that `cosign` was taking advantage of a bug in the mutate package
where the `Manifest()` returns a mutable copy of it's internal state, which
allowed them to permanently alter the result of `Manifest()`.  This mutability
violates the philosophy of our abstractions, and should not be allowed, so
this change ensures that each of the methods returning state in the `mutate`
package return a `.DeepCopy()` instead.  In `image` this included the
`configFile`, and `manifest`.  In `index` this included the `indexManifest`.
I have added tests that when the returned copies are mutated, a subsequent
invocation of the getter will return a pristine copy.

Since I'm not completely heartless, I am also adding a new
`mutate.ConfigMediaType` method to enable `cosign` to perform the alteration
they want, now that the trick they were using will no longer work.

Related: https://github.com/sigstore/cosign/issues/718

### DIFF
--- a/pkg/legacy/tarball/write_test.go
+++ b/pkg/legacy/tarball/write_test.go
@@ -382,9 +382,15 @@ func TestMultiWriteMismatchedHistory(t *testing.T) {
 	if err != nil {
 		t.Fatalf("Error getting image config: %v", err)
 	}
+
 	// Set the history such that number of history entries != layers. This
 	// should trigger an error during the image write.
 	cfg.History = make([]v1.History, 1)
+	img, err = mutate.ConfigFile(img, cfg)
+	if err != nil {
+		t.Fatalf("mutate.ConfigFile() = %v", err)
+	}
+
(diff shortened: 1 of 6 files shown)


mattmoor:
Weird, this is a legit failure that repros for me locally, but not against `main`, so we may have our own reliance on this bug.

mattmoor:
Ah, we do, but it's a test bug 🤣 
https://github.com/google/go-containerregistry/blob/8388fde8bcadbc8ddb820b5ac0046e10e6c572e7/pkg/legacy/tarball/write_test.go#L302-L303

dlorenc:
> Since I'm not completely heartless, I am also adding a new

All hail not heartless Matt

imjasonh (approved):

imjasonh:
Thanks for fixing this!

PR #1126, opened 2021-09-21
feat(daemon): avoid multiple initialization

knqyf263:
## Description
For example, if `LayerByDiffID` of `daemon.Image` is called twice in parallel, `initialize` might be called twice at the same time. Then, `tarball.Image` might be called twice. It is not efficient, so this PR locks `initialize()` before checking `i.tarballImage` for nil so that it will not be initialized twice.

Please let me know if you prefer `sync.Once`. Note that we need to keep `err` in the `image` struct like `imageOpener`.
https://github.com/google/go-containerregistry/blob/c3aee497b158a4c0e4da6584668637ac6b0b13a3/pkg/v1/daemon/image.go#L46

## Related PR
https://github.com/google/go-containerregistry/pull/1121

### DIFF
--- a/pkg/v1/daemon/image.go
+++ b/pkg/v1/daemon/image.go
@@ -28,10 +28,12 @@ import (
 )
 
 type image struct {
-	mu           sync.Mutex
 	ref          name.Reference
 	opener       *imageOpener
 	tarballImage v1.Image
+
+	once sync.Once
+	err  error
 }
 
 type imageOpener struct {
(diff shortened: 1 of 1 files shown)


pkg/v1/daemon/image.go
```diff
@@ -99,6 +100,9 @@ func Image(ref name.Reference, options ...Option) (v1.Image, error) {
 }
 
 func (i *image) initialize() error {
+	i.mu.Lock()
```
imjasonh, line 103: Instead of a mutex, could we use a [`sync.Once`](https://pkg.go.dev/sync#Once)? This would call `tarball.Image` only the first time `initialize` is called, taking into account multiple concurrent calls.

knqyf263: Sure. As described in the description, we need to keep the err. Is it ok with you?

>Note that we need to keep err in the image struct like imageOpener.

knqyf263: Fixed https://github.com/google/go-containerregistry/pull/1126/commits/3eac11ecf1a82f386ecf187e58c802eb8b27a3f1

imjasonh (approved):
Nice. Thanks for doing this!

edit: I'll wait for @jonjohnsonjr to approve+merge in case he has other comments.

jonjohnsonjr (approved):

PR #1128, opened 2021-09-21
Don't reuse errgroups, propagate contexts better

jonjohnsonjr:
When you Wait() on an errgroup, the underlying context gets cancelled,
so calling Wait() twice will break if you correctly propagate contexts.
We weren't propagating contexts correctly, so this kind of worked
before, but if we actually want to leverage contexts better we should
fix this :)


imjasonh (approved):
Nice.

PR #1129, opened 2021-09-21
Handle remote.WriteManifest manifests concurrently

jonjohnsonjr:
This speeds up writing an image index quite a bit in the happy path.

Also propagate jobs through crane options


jonjohnsonjr:
So the thing of some interest here is that the "job pool" isn't shared between different functions, so recursive index can exceed the number of jobs by quite a bit... I'm not sure how I would want to plumb that around.

PR #1158, opened 2021-10-26
Make the blob storage in pkg/registry pluggable.

mattmoor:
With this change folks can use `registry.WithBlobHandler(...)` to substitute a custom blob handling layer.

### DIFF
--- a/pkg/registry/blobs.go
+++ b/pkg/registry/blobs.go
@@ -18,6 +18,7 @@ import (
 	"bytes"
 	"crypto/sha256"
 	"encoding/hex"
+	"errors"
 	"fmt"
 	"io"
 	"io/ioutil"
@@ -59,13 +60,13 @@ type blobs struct {
 
 // BlobHandler is the interface for the storage layer underneath this registry.
 type BlobHandler interface {
-	// Stat returns the size of the blob whose hash is specified, and true,
-	// if it exists. If not, it returns (0, false).
-	Stat(repo name.Repository, h v1.Hash) (int, bool)
+	// Stat returns the size of the blob whose hash is specified,
+	// if it exists. If not, it returns (0, error).
+	Stat(repo name.Repository, h v1.Hash) (int64, error)
 
 	// Get returns true and a reader for consuming the blob specified with the hash,
-	// if it exists.  It now, it returns (nil, false).
-	Get(repo name.Repository, h v1.Hash) (io.ReadCloser, bool)
+	// if it exists.  It now, it returns (nil, error).
+	Get(repo name.Repository, h v1.Hash) (io.ReadCloser, error)
 
 	// Store stores the stream of content with the given hash, or returns the error
 	// encountered doing so.
@@ -110,12 +111,18 @@ func (b *blobs) handle(resp http.ResponseWriter, req *http.Request) *regError {
 			}
 		}
 
-		sz, ok := b.bh.Stat(repo, h)
-		if !ok {
+		sz, err := b.bh.Stat(repo, h)
+		if errors.Is(err, ErrBlobNotFound) {
 			return &regError{
 				Status:  http.StatusNotFound,
 				Code:    "BLOB_UNKNOWN",
-				Message: "Unknown blob",
+				Message: err.Error(),
+			}
+		} else if err != nil {
+			return &regError{
+				Status:  http.StatusInternalServerError,
+				Code:    "BLOB_UNKNOWN",
+				Message: err.Error(),
 			}
 		}
 
@@ -142,21 +149,33 @@ func (b *blobs) handle(resp http.ResponseWriter, req *http.Request) *regError {
 			}
 		}
 
-		sz, ok := b.bh.Stat(repo, h)
-		if !ok {
+		sz, err := b.bh.Stat(repo, h)
+		if errors.Is(err, ErrBlobNotFound) {
 			return &regError{
 				Status:  http.StatusNotFound,
 				Code:    "BLOB_UNKNOWN",
-				Message: "Unknown blob",
+				Message: err.Error(),
+			}
+		} else if err != nil {
+			return &regError{
+				Status:  http.StatusInternalServerError,
+				Code:    "BLOB_UNKNOWN",
+				Message: err.Error(),
 			}
 		}
 
-		b, ok := b.bh.Get(repo, h)
-		if !ok {
+		b, err := b.bh.Get(repo, h)
+		if errors.Is(err, ErrBlobNotFound) {
 			return &regError{
 				Status:  http.StatusNotFound,
 				Code:    "BLOB_UNKNOWN",
-				Message: "Unknown blob",
+				Message: err.Error(),
+			}
+		} else if err != nil {
+			return &regError{
+				Status:  http.StatusInternalServerError,
+				Code:    "BLOB_UNKNOWN",
+				Message: err.Error(),
 			}
 		}
 		defer b.Close()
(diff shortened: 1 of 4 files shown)


pkg/registry/blobs.go
```diff
@@ -102,10 +151,20 @@ func (b *blobs) handle(resp http.ResponseWriter, req *http.Request) *regError {
 			}
 		}
 
-		resp.Header().Set("Content-Length", fmt.Sprint(len(b)))
+		b, ok := b.bh.Get(repo, h)
```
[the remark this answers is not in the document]
mattmoor: I think there are a bunch of additional checks we could add, so I'm sorta reluctant to start pulling that particular thread in this PR.  Another notable bit is that we don't check the possible err return value on `io.Copy()` below, which at one point was the source of a fun bug in the Docker client (around digest verification!).

I think my personal bias would be towards leaning on internal consistency and checking things on their way in (vs. on the way out), but I'd be happy to iterate with you on follow up PRs here.

mattmoor:
@imjasonh I'll optimistically start to look at what we might be able to do with manifests tomorrow.

mattmoor:
@jonjohnsonjr PTAL

pkg/registry/blobs.go
```diff
@@ -20,11 +20,15 @@ import (
...
 	"fmt"
 	"io"
+	"io/ioutil"
 	"math/rand"
 	"net/http"
 	"path"
 	"strings"
 	"sync"
+
+	"github.com/google/go-containerregistry/pkg/name"
```
jonjohnsonjr, line 30: I've been trying to keep this from having any dependencies outside the stdlib.

pkg/registry/blobs.go
```diff
@@ -46,11 +50,26 @@ func isBlob(req *http.Request) bool {
...
 	lock    sync.Mutex
+
+	bh BlobHandler
+}
+
+// BlobHandler is the interface for the storage layer underneath this registry.
+type BlobHandler interface {
+	// Stat returns the size of the blob whose hash is specified, and true,
+	// if it exists. If not, it returns (0, false).
+	Stat(repo name.Repository, h v1.Hash) (int, bool)
```
jonjohnsonjr, line 64:
```suggestion
	Stat(repo name.Repository, h v1.Hash) (int64, bool)
```

pkg/registry/blobs.go
```diff
@@ -46,11 +50,26 @@ func isBlob(req *http.Request) bool {
...
-	contents map[string][]byte
 	// Each upload gets a unique id that writes occur to until finalized.
 	uploads map[string][]byte
 	lock    sync.Mutex
+
+	bh BlobHandler
+}
+
+// BlobHandler is the interface for the storage layer underneath this registry.
+type BlobHandler interface {
```
jonjohnsonjr, line 61: If `Store` can return an error, shouldn't `Stat` and `Get`?

pkg/registry/blobs.go
mattmoor: Too late: https://github.com/google/go-containerregistry/blob/0dfbb568512eae093cfeaefe620aabec76a6681b/pkg/registry/manifest.go#L31-L32

mattmoor: Generally, I agree with this goal for *implementations* of the interfaces, but it seems wasteful to reimplement validation that exists in the `name` and similar packages.

jonjohnsonjr: https://github.com/google/go-containerregistry/pull/798#discussion_r515407375

imjasonh: Jon beat me to it 🙈

pkg/registry/blobs.go
mattmoor: done

pkg/registry/blobs.go
mattmoor: done

pkg/registry/blobs.go
mattmoor: My comment stands.  It's foolish to reimplement the validation we have in `name` and other places that let us avoid dealing with stringly typed data, and I don't see the risk of a dependency cycle for these superficial things because they should absolutely NOT need a full-blown registry implementation to test interfaces.

I could absolutely get behind a `depcheck` style test that this never depend on `pkg/v1/remote` in particular.

pkg/registry/blobs.go
```diff
@@ -74,38 +94,96 @@ func (b *blobs) handle(resp http.ResponseWriter, req *http.Request) *regError {
...
-		if !ok {
+		h, err := v1.NewHash(target)
+		if err != nil {
+			return &regError{
+				Status:  http.StatusBadRequest,
+				Code:    "NAME_INVALID",
+				Message: err.Error(),
+			}
+		}
+		repo, err := name.NewRepository(req.URL.Host + path.Join(elem[1:len(elem)-2]...))
```
imjasonh, line 105: (Resurrecting this because I'm sad we couldn't make it happen)

Aside from the question of whether we need to depend on `pkg/name` here, I wonder if `req.URL.Host` is adding anything besides letting the string pass through `NewRepository`. Really all blob storers want is the subrepo path.

I hear the concern about duplicating validation logic though. Would it be reasonable to export a method from `pkg/name` to validate a subrepo path just so `pkg/registry` could use it?

mattmoor: > I'm sad we couldn't make it happen

Me too, but alas I've unblocked myself (and it is glorious 😉 ).


> I hear the concern about duplicating validation logic though

We have a bigger issue here.  This repo is now apparently governed by unwritten rules in one person's head, which also have had no rationale articulated anywhere that was arrived at by consensus of the stakeholders in this library.

I've clearly demonstrated a willingness to send PRs here (and elsewhere) to fix things, move things forward, and I'm happy to engage in productive discussions about them, but waiting a week to get a Nack based on unwritten rules (**that are already violated with what is checked in**), is frankly both aggravating and unsustainable.

jonjohnsonjr: https://github.com/google/go-containerregistry/blob/a0c4bd256482b8522065d5f6cf966281ef270680/pkg/registry/registry.go#L15-L23

mattmoor: In order for this to disqualify this change, "low dependency" must be interpreted to exclude *what it currently imports*.  🙃

mattmoor: Also, I'd recommend a README: https://github.com/google/go-containerregistry#philosophy

imjasonh: > We have a bigger issue here. This repo is now apparently governed by unwritten rules in one person's head, which also have had no rationale articulated anywhere that was arrived at by consensus of the stakeholders in this library.

Hoping to address this in #1166 

Taking the issue of dependency philosophy aside, I think I'm still convinced that the dependency on `pkg/name` is unnecessary, or at least having a full `name.Repository` in the interface is. If not for validation of the repository path, you don't actually need a full `name.Repository`, and could just get by with `Get(string path, h v1.Hash)`, and not add any new dependencies. Path validation itself is [fairly simple](https://github.com/google/go-containerregistry/blob/v0.6.0/pkg/name/check.go#L35), and could be moved to `internal/something` and shared by `pkg/name` and `pkg/registry`.

mattmoor: You are assuming details of the blob layer by stating that it can only care about path.  Every single cloud vendor registry uses host to route storage requests, Amazon even gives each customer their own hostname per region.  I doubt this is how you wrote kontain.me, but…. 

Hopefully you get my point.

imjasonh: Ah I see what you're saying. The same code depending on pkg/registry with some blob storage impl, running at `foo.registry.example` and `bar.registry.example`, might decide to store those blobs differently based on the host.

Can we document this expectation in the doc comments for `BlobHandler`? It wasn't immediately clear to me (because I'm a big dumb dummy) why the host was included, except to make path validation easier.

mattmoor: > Can we document this expectation in the doc comments for BlobHandler

Absolutely.

pkg/registry/blobs.go
```diff
@@ -46,11 +51,26 @@ func isBlob(req *http.Request) bool {
...
+
+// BlobHandler is the interface for the storage layer underneath this registry.
+type BlobHandler interface {
+	// Stat returns the size of the blob whose hash is specified,
+	// if it exists. If not, it returns (0, error).
+	Stat(repo name.Repository, h v1.Hash) (int64, error)
+
+	// Get returns true and a reader for consuming the blob specified with the hash,
+	// if it exists.  It now, it returns (nil, error).
+	Get(repo name.Repository, h v1.Hash) (io.ReadCloser, error)
```
imjasonh, line 69: Something occurred to me about this API, thinking through how I could use it with kontain.me -- this API would require kontain.me to proxy and serve blobs from GCS, instead of HTTP redirecting like I do today, which is a lot faster and cheaper than proxying through Cloud Run.

This might just mean that `pkg/registry` isn't going to be a good fit for kontain.me, and that's fine, but I wanted to surface that here since I think it's a pretty nice use case that isn't supported by this proposal.

PR #1162, opened 2021-11-02
Enable golangci-lint

imjasonh:
This previously wasn't linting anything, despite it being in our (copied from Knative forever ago) [Code Style check](https://github.com/google/go-containerregistry/runs/4083873412?check_suite_focus=true), because the repo didn't have a `.golangci.yaml`.

I've disabled most linters that either didn't run for me locally, or had a bunch of false positives. If we feel like fixing these later we can.

cc @mattmoor thanks for catching this.

### DIFF
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -1,0 +1,15 @@
+run:
+  timeout: 5m
+
+  skip-dirs:
+    - internal
+    - pkg/registry
+
+linters:
+  enable:
+    - asciicheck
+    - depguard
+    - stylecheck
+    - unconvert
+  disable:
+    - errcheck
--- a/pkg/v1/daemon/image.go
+++ b/pkg/v1/daemon/image.go
@@ -118,7 +118,7 @@ func (i *image) Layers() ([]v1.Layer, error) {
 	return i.tarballImage.Layers()
 }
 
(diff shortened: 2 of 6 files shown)


jonjohnsonjr (approved):

.golangci.yaml
```diff
@@ -0,0 +1,15 @@
+run:
+  timeout: 5m
+
+  skip-dirs:
+    - internal
+    - pkg/registry
+
+linters:
```
mattmoor, line 8: We're running:
```
linters:
  enable:
  - asciicheck
  - deadcode
  - depguard
  - errcheck
  - errorlint
  - gofmt
  - goimports
  - gosec
  - gocritic
  - importas
  - prealloc
  - revive
  - misspell
  - stylecheck
  - tparallel
  - unconvert
  - unparam
  - whitespace
```

imjasonh:
```
$ golangci-lint run | grep pkg | wc -l
      58
```

We'll get there. Thanks for sharing this!

mattmoor: 57 of those are probably errcheck on `io.Copy` in `pkg/registry` 😉

imjasonh:
```
skip-dirs:
    - internal
    - pkg/registry
```

😎

PR #1163, opened 2021-11-04
Give the ping context a timeout.

mattmoor:
Debugging why requests to an insecure registry seem to hang, I noticed that there are 3+ requests to the https endpoint before it falls back to http (at that point my logs truncate because the test was over).  The time between the request being issued and completing in the roundtripper was ~30s each.

This change gives each ping a generous 5s timeout to complete each request.

### DIFF
--- a/pkg/v1/remote/transport/ping.go
+++ b/pkg/v1/remote/transport/ping.go
@@ -22,6 +22,7 @@ import (
 	"io/ioutil"
 	"net/http"
 	"strings"
+	"time"
 
 	authchallenge "github.com/docker/distribution/registry/client/auth/challenge"
 	"github.com/google/go-containerregistry/pkg/name"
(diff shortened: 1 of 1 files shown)


mattmoor:
I'm testing another mechanism for this, where I pass a custom http.Transport (based on the default transport) with a shorter dial timeout:
```
remote.WithTransport(&http.Transport{
	Proxy: http.ProxyFromEnvironment,
	DialContext: (&net.Dialer{
		Timeout:   5 * time.Second,
		KeepAlive: 30 * time.Second,
	}).DialContext,
	ForceAttemptHTTP2:     true,
	MaxIdleConns:          100,
	IdleConnTimeout:       90 * time.Second,
	TLSHandshakeTimeout:   10 * time.Second,
	ExpectContinueTimeout: 1 * time.Second,
})
```

If this works, we may want to consider something like this as a better default transport for GGCR since the experience with a repeated dial timeout on the https ping retried 5(!!!) times is miserable.

I haven't confirmed, but I believe I'm hitting the same thing publishing stuff with cosign as well and I'm sure all the other tools will be affected (ko, buildpacks, kaniko).

I'll report back as to whether this actually works, and if so we can discuss which change we'd rather have.

mattmoor:
The custom transport works.  Given our built-in retries, I'd be comfortable moving the direction of the transport, but I'm open to other options as well.

jonjohnsonjr (approved):
I'm ok to just merge this until we can find a better solution.

PR #1164, opened 2021-11-04
Enable some more golangci-lint checks, fix findings

imjasonh:
There's still quite a bit of errchecks to chase down, but this is an
improvement at least.

god what a mind-numbing chore.

### DIFF
--- a/pkg/gcrane/copy.go
+++ b/pkg/gcrane/copy.go
@@ -16,6 +16,7 @@ package gcrane
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"net/http"
 	"strings"
@@ -91,12 +92,12 @@ type copier struct {
 func newCopier(src, dst string, o *options) (*copier, error) {
 	srcRepo, err := name.NewRepository(src)
 	if err != nil {
-		return nil, fmt.Errorf("parsing repo %q: %v", src, err)
+		return nil, fmt.Errorf("parsing repo %q: %w", src, err)
 	}
 
 	dstRepo, err := name.NewRepository(dst)
 	if err != nil {
-		return nil, fmt.Errorf("parsing repo %q: %v", dst, err)
+		return nil, fmt.Errorf("parsing repo %q: %w", dst, err)
 	}
 
 	// A queue of size 2*jobs should keep each goroutine busy.
@@ -121,7 +122,7 @@ func recursiveCopy(ctx context.Context, src, dst string, o *options) error {
 				tags, err = google.List(repo, o.google...)
 				return err
 			}); err != nil {
-				return fmt.Errorf("failed List for repo %s: %v", repo, err)
+				return fmt.Errorf("failed List for repo %s: %w", repo, err)
 			}
 		}
 
@@ -129,7 +130,7 @@ func recursiveCopy(ctx context.Context, src, dst string, o *options) error {
 		if err := backoffErrors(GCRBackoff(), func() error {
 			return c.copyRepo(ctx, repo, tags)
 		}); err != nil {
-			return fmt.Errorf("failed to copy repo %q: %v", repo, err)
+			return fmt.Errorf("failed to copy repo %q: %w", repo, err)
 		}
 
 		return nil
@@ -139,7 +140,7 @@ func recursiveCopy(ctx context.Context, src, dst string, o *options) error {
 	g.Go(func() error {
 		defer close(c.tasks)
 		if err := google.Walk(c.srcRepo, walkFn, o.google...); err != nil {
-			return fmt.Errorf("failed to Walk: %v", err)
+			return fmt.Errorf("failed to Walk: %w", err)
 		}
 		return nil
 	})
@@ -153,7 +154,7 @@ func recursiveCopy(ctx context.Context, src, dst string, o *options) error {
 				if err := backoffErrors(GCRBackoff(), func() error {
 					return c.copyImages(ctx, task)
 				}); err != nil {
-					return fmt.Errorf("failed to copy %q: %v", task.digest, err)
+					return fmt.Errorf("failed to copy %q: %w", task.digest, err)
 				}
 			}
 			return nil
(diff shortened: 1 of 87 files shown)


pkg/gcrane/copy.go
```diff
@@ -267,8 +267,8 @@ func hasStatusCode(err error, code int) bool {
 	if err == nil {
 		return false
 	}
-	if err, ok := err.(*transport.Error); ok {
-		if err.StatusCode == code {
+	if terr, ok := err.(*transport.Error); ok { //nolint: errorlint
+		if terr.StatusCode == code {
```
mattmoor, line 271: Hmm, why don't you want `errors.As` here?

pkg/name/errors.go
```diff
@@ -32,6 +32,6 @@ func NewErrBadName(fmtStr string, args ...interface{}) *ErrBadName {
 
 // IsErrBadName returns true if the given error is an ErrBadName.
 func IsErrBadName(err error) bool {
-	_, ok := err.(*ErrBadName)
+	_, ok := err.(*ErrBadName) //nolint: errorlint
```
mattmoor, line 35: Why not `errors.Is`?

mattmoor: Or deprecate in favor of `errors.Is`?

pkg/v1/google/keychain.go
```diff
@@ -77,5 +77,5 @@ func resolve() (authn.Authenticator, error) {
 		return auth, nil
 	}
 
-	return nil, fmt.Errorf("failed to create token source from env: %v or gcloud: %v", envErr, gErr)
+	return nil, fmt.Errorf("failed to create token source from env: %v or gcloud: %v", envErr, gErr) //nolint: errorlint
```
mattmoor, line 80: I am curious why you chose to not wrap here?

pkg/v1/remote/descriptor_test.go
```diff
@@ -68,7 +68,7 @@ func TestGetSchema1(t *testing.T) {
 	want := `unsupported MediaType: "application/vnd.docker.distribution.manifest.v1+prettyjws", see https://github.com/google/go-containerregistry/issues/377`
 	// Should fail based on media type.
 	if _, err := desc.Image(); err != nil {
-		if _, ok := err.(*ErrSchema1); !ok {
+		if _, ok := err.(*ErrSchema1); !ok { //nolint: errorlint
```
mattmoor, line 71: another `.As` (I care less about tests)

pkg/v1/remote/multi_write.go
```diff
@@ -99,7 +99,7 @@ func MultiWrite(m map[name.Reference]Taggable, options ...Option) (rerr error) {
 	// Collect the total size of blobs and manifests we're about to write.
 	if o.updates != nil {
 		defer close(o.updates)
-		defer func() { sendError(o.updates, rerr) }()
+		defer func() { _ = sendError(o.updates, rerr) }()
```
mattmoor, line 102: I'm curious if this complains with:
```suggestion
		defer sendError(o.updates, rerr)
```

mattmoor (approved):
I stopped commenting on them, but there were a bunch of `// nolint errorlint` that could be `errors.As`, and a license presubmit check, but otherwise this LGTM

pkg/v1/google/keychain.go
mattmoor: Ah the long line hid that.  I don't especially care since this seems like a relatively niche part of the library.

pkg/v1/remote/multi_write.go
imjasonh: I gave up and disabled `errcheck` entirely when I saw how much more drudgery it would take to address them all. This is at least a signal to readers that the error is unchecked.

pkg/name/errors.go
```diff
@@ -32,6 +35,6 @@ func NewErrBadName(fmtStr string, args ...interface{}) *ErrBadName {
 
 // IsErrBadName returns true if the given error is an ErrBadName.
 func IsErrBadName(err error) bool {
-	_, ok := err.(*ErrBadName)
-	return ok
+	var berr *ErrBadName
+	return errors.As(err, &berr)
```
mattmoor, line 39: I believe this can just be `errors.Is`

pkg/v1/remote/descriptor_test.go
```diff
@@ -68,7 +69,8 @@ func TestGetSchema1(t *testing.T) {
 	want := `unsupported MediaType: "application/vnd.docker.distribution.manifest.v1+prettyjws", see https://github.com/google/go-containerregistry/issues/377`
 	// Should fail based on media type.
 	if _, err := desc.Image(); err != nil {
-		if _, ok := err.(*ErrSchema1); !ok {
+		var s1err ErrSchema1
+		if errors.Is(err, &s1err) {
```
mattmoor, line 73:
```suggestion
		if errors.Is(err, &ErrSchema1{}) {
```

pkg/name/errors.go
imjasonh: I tried a bunch of different ways and couldn't get it to work.

It turns out this method is never used outside of tests in `pkg/name`, so I just removed it and inlined it (with `errors.As` for now) into the two places in tests where it was used. In theory someone is using it today, since it was exported. If we want I can put it back and deprecate it.

I also removed `NewErrBadName` which isn't used outside of `pkg/name`, and should not be needed outside of `pkg/name`.

pkg/name/errors.go
```diff
@@ -28,13 +27,7 @@ func (e *ErrBadName) Error() string {
 	return e.info
 }
 
-// NewErrBadName returns a ErrBadName which returns the given formatted string from Error().
-func NewErrBadName(fmtStr string, args ...interface{}) *ErrBadName {
+// newErrBadName returns a ErrBadName which returns the given formatted string from Error().
+func newErrBadName(fmtStr string, args ...interface{}) *ErrBadName {
```
jonjohnsonjr, line 31: This is technically breaking but do we care?

imjasonh: I think this was a mistake to export, and can't see a situation where an external user would want to create new `ErrBadName`s.

I feel somewhat less sure about removing `IsErrBadName`, since someone might have used that.

PR #1165, opened 2021-11-04
Define a new `remote.DefaultTransport`.

mattmoor:
_The first commit here reverts https://github.com/google/go-containerregistry/pull/1163 which went in before I could change it to the transport-based option._

Previously we used `http.DefaultTransport` (on which this is based),
but this uses a default dial timeout of 30s, which when we (by default)
wrap things in 5x retries can lead to ~150s delays simply pinging
an http-based registry.

If folks encounter issues with this via the library they can restore
the current behavior with:
```go
remote.WithTransport(http.DefaultTransport)
```

If folks encounter issues with this via `crane` they can restore the
current behavior with:
```
--dial-timeout 30s
```

### DIFF
--- a/cmd/crane/cmd/root.go
+++ b/cmd/crane/cmd/root.go
@@ -15,13 +15,16 @@ package cmd
 import (
 	"crypto/tls"
 	"fmt"
+	"net"
 	"net/http"
 	"os"
 	"path/filepath"
+	"time"
 
 	"github.com/docker/cli/cli/config"
 	"github.com/google/go-containerregistry/pkg/crane"
 	"github.com/google/go-containerregistry/pkg/logs"
+	"github.com/google/go-containerregistry/pkg/v1/remote"
 	"github.com/spf13/cobra"
 )
 
@@ -37,6 +40,7 @@ var Root = New(use, short, []crane.Option{})
 func New(use, short string, options []crane.Option) *cobra.Command {
 	verbose := false
 	insecure := false
+	dialTimeout := 5 * time.Second
 	platform := &platformValue{}
 	var osVersion string
 
@@ -69,7 +73,11 @@ func New(use, short string, options []crane.Option) *cobra.Command {
 
 			options = append(options, crane.WithPlatform(platform.platform))
 
-			transport := http.DefaultTransport.(*http.Transport).Clone()
+			transport := remote.DefaultTransport.Clone()
+			transport.DialContext = (&net.Dialer{
+				Timeout:   dialTimeout,
+				KeepAlive: 30 * time.Second,
+			}).DialContext
 			transport.TLSClientConfig = &tls.Config{InsecureSkipVerify: insecure}
 
 			var rt http.RoundTripper = transport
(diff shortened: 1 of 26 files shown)


mattmoor:
@googlebot I fixed it.

mattmoor:
@googlebot I fixed it.

mattmoor:
@googlebot I fixed it.

mattmoor:
image

omg I hate you CLA bot

mattmoor:
Switched to the old email 🙃

mattmoor:
Fixed the gcrane test that assumed `http.DefaultTransport` was the default transport to use `WithTransport`.

pkg/gcrane/copy_test.go
```diff
@@ -175,26 +175,27 @@ func TestCopy(t *testing.T) {
 	defer s.Close()
 
 	// Make sure we don't actually talk to XCR.
-	http.DefaultTransport = s.Client().Transport
+	opt := remote.WithTransport(s.Client().Transport)
```
imjasonh, line 178: Rather than take steps to not accidentally talk to xcr.io, how about using registry.example.com which will definitely never exist.

imjasonh (commented):
This looks okay to me. I'm not super in love with making crane users think about the dial timeout flag, but I guess it's okay to let them get back to previous behavior if they want it.

imjasonh:
Friendly ping. If we do a release soon do we feel like we need to get this included in it?

mattmoor:
Rebasing now, then I'll add a commit for the `xcr.io` thing.

pkg/gcrane/copy_test.go
mattmoor: Ok this is done.

cmd/crane/cmd/root.go
```diff
@@ -115,6 +123,7 @@ func New(use, short string, options []crane.Option) *cobra.Command {
 
 	root.PersistentFlags().BoolVarP(&verbose, "verbose", "v", false, "Enable debug logs")
 	root.PersistentFlags().BoolVar(&insecure, "insecure", false, "Allow image references to be fetched without TLS")
+	root.PersistentFlags().DurationVar(&dialTimeout, "dial-timeout", dialTimeout, "Modify the dial timeout used to contact the registry.")
```
imjasonh, line 126: Sorry for the runaround, but I think we can probably just remove this flag until we find out we need it. If you'd prefer I'm happy to make this change myself after we merge this, before we cut another release. Let me know.

(Total aside, I'm starting to hate these global flags. `crane login --insecure --platform=windows/arm64` does nothing, and sounds unnecessarily scary.)

mattmoor: I'm happy to kill it.  Gimme a few minutes to tweak things

mattmoor: Ok RFAL

jonjohnsonjr: > (Total aside, I'm starting to hate these global flags. crane login --insecure --platform=windows/arm64 does nothing, and sounds unnecessarily scary.)

Ack, we should find a way to make these not-quite-global.

jonjohnsonjr (approved):

imjasonh (approved):

PR #1172, opened 2021-11-09
Pass Options when recursively calling writeIndex

rvmiller89:
Noticed while pushing nested indexes, we lose authentication options
when attempting to upload a nested index with `WriteIndex`.

This PR ensures options are maintained and passed to recursive calls
to `writeIndex`.


imjasonh (approved):
Good catch! Thanks for this contribution 👍

PR #1173, opened 2021-11-10
Adding OS version to Crane for better Windows support.

phillipsj:
Windows images present some unique challenges and one of those challenges is being able to pull a specific OS version. This especially comes into play when working with a manifest like the k8s.gcr.io/pause:3.6 which doesn't offer tags specific to Windows OS version. 

Here is an example use case:

```
crane pull --platform windows/amd64 --osversion 10.0.17763.2114 k8s.gcr.io/pause:3.6 windows-1809.tar
```

Without the version specified Windows Server 2022 would have been pulled.

### DIFF
--- a/cmd/crane/cmd/root.go
+++ b/cmd/crane/cmd/root.go
@@ -38,6 +38,7 @@ func New(use, short string, options []crane.Option) *cobra.Command {
 	verbose := false
 	insecure := false
 	platform := &platformValue{}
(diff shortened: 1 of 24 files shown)


imjasonh (approved):
Thanks for this contribution!

nickgerace (approved):
🚀 

(just giving a teammate review :) )

PR #1178, opened 2021-11-14
update crane mutate annotation/label args to allow commas in label values

### LINKED ISSUE #1177: Crane mutate for label does not like commas
It looks like there is an issue adding commas to labels, e.g.,

```bash
$ crane mutate ghcr.io/vsoch/spack-crane-ubuntu-20.04:latest --label org.spack.compilers="clang@10.0.0,gcc@9.3.0"
2021/11/14 11:09:48 parsing label "gcc@9.3.0", not enough parts
```
It looks like the third issue in the boxes here https://github.com/spf13/viper/issues/380#issuecomment-561278080 - the comma is being parsed as a delimiter! I'll open a PR to follow up with a suggested fix.

vsoch:
With the current argument type, a common in a label value (a reasonable thing to have) will be split and parsed as a separate label, and the client errors. This commit updates the flag to StringToStringVarP so we start with a map[string]string off the bat, and do not need to do special parsing beyond checking for empty strings.

I also added some additional notes to the crane mutate doc page for points that weren't clear to me - 1. that the image needed to exist in a registry and would be changed there, and 2. a full example for using the command.

This will close #1177 

Signed-off-by: vsoch <vsoch@users.noreply.github.com>


vsoch:
How are the client docs updated? I found the hack/presubmit.sh to validate but am not sure how to re-generate the docs for the updated client.

imjasonh:
> How are the client docs updated? I found the hack/presubmit.sh to validate but am not sure how to re-generate the docs for the updated client.

./hack/update-codegen.sh

Specifically the last line: https://github.com/google/go-containerregistry/blob/main/hack/update-codegen.sh#L52

vsoch:
Ahh I literally just ran that! Asked too soon :P

cmd/crane/cmd/mutate.go
```diff
@@ -118,22 +117,21 @@ func NewCmdMutate(options *[]crane.Option) *cobra.Command {
...
+	mutateCmd.Flags().StringToStringVarP(&lbls, "label", "l", nil, "New labels to add")
+	mutateCmd.Flags().StringVar(&entrypoint, "entrypoint", "", "New entrypoing to set")
 	mutateCmd.Flags().StringVarP(&newRef, "tag", "t", "", "New tag to apply to mutated image. If not provided, push by digest to the original image repository.")
 	return mutateCmd
 }
 
-// splitKeyVals splits key value pairs which is in form hello=world
-func splitKeyVals(kvPairs []string) (map[string]string, error) {
+// validateKeyVals ensures no values are empty
+func validateKeyVals(kvPairs map[string]string) (map[string]string, error) {
```
imjasonh, line 128: This could just return `error`, and not bother creating a copy of the input kv map.

vsoch: All set! I like that much better - we can use "labels" and "annotations" off the bat.

imjasonh (approved):
Thank you! 👍

vsoch:
uhoh looks like we've hit some pull limit?
image

The presubmit.sh worked locally!

imjasonh:
> uhoh looks like we've hit some pull limit? image
> 
> The presubmit.sh worked locally!

Weird. That doesn't look like the rate limit error message, just general slowness. Retrying, let's hope 🤞

vsoch:
Thank you!

Issue #1111, opened 2021-08-23
Crane operations on Windows containers


imjasonh:
Layers in Windows containers have an...odd... format, as discovered when [building Windows support for `ko`](https://github.com/google/ko/pull/374). It might be helpful to package this knowledge into ggcr and/or `crane`, so fewer folks have to deal with it.

For example: layers must have a `Hives/` top-level directory (which can be empty), and all file contents must be in the `Files/` top-level directory. `/` is still accepted as a file separator. There's a mysterious `PAXHeader` value that's needed for any tar entries that need to be executable.

Off the top of my head, `crane append` could Windows-ify the layer contents if `--platform=windows/*`. We might want to make Windowsification some kind of ggcr/`crane` method or option.

jonjohnsonjr:
> We might want to make Windowsification some kind of ggcr/crane method or option.

I like this idea a lot.

imjasonh:
`func Windowsify(in v1.Layer) v1.Layer` returns a layer that reads from the `in` layer and inserts and entry for `Hives/`, and prepends `Files/` to all other paths, and sets the `PAXHeader` value.

If `Digest`/`DiffID`/`Size` are called on it, it could generate and buffer the layer contents instead of failing like `stream.Layer` does.

jonjohnsonjr:
> If Digest/DiffID/Size are called on it, it could generate and buffer the layer contents instead of failing like stream.Layer does.

I think we'd want both?

adamrehn:
@imjasonh would you like any assistance with this? Now that the OS version selection support added by @phillipsj in #1173 has been merged, this appears to be the biggest remaining blocker standing in the way of using `crane append` and `crane mutate` to act as an equivalent to `ko publish` for non-Go applications in automation pipelines running under Kubernetes.

(Motivating example: being able to build a Windows executable and package it into a container image inside a Tekton task, without needing to resort to using a [Windows HostProcess container](https://kubernetes.io/docs/tasks/configure-pod-container/create-hostprocess-pod/) so hcsshim can communicate with the underlying host node's HCS.)

phillipsj:
I have been working on doing this very thing already following along with what KO is doing. Happy to assist with this too.

imjasonh:
[...]

PR #1179, opened 2021-11-15
Windowsify layers when `crane append`ing

imjasonh:
Also:
- support layer windowsification in `crane.Append` if the base image is a Windows image
- add an e2e test that this works on Linux and Windows

After talking with Jon, we decided to keep the Windowsification package `internal` for now, and only expose it via base image detection logic in `crane append` and `crane.Append` for now. This gives us more flexibility to break the API and behavior as we learn more about use cases and Windows images in general.

There are currently some blocking TODOs:
- [ ] ~~mutation should be lazy~~ -- since it's `internal`, it's more okay to punt on this
- [ ] ~~don't set PAXRecords unnecessarily (and understand when/why it's necessary)~~ -- since it's `internal`, it's more okay to punt on this
- [x] unit tests

#1111 

With this change I was able to:

```
go run ./cmd/crane append \
    --platform=windows/amd64 \
    -b mcr.microsoft.com/windows/nanoserver:1809 \
    -t gcr.io/imjasonh/windows \
    -f files.tar
```

...which produced this image: `gcr.io/imjasonh/windows@sha256:1d488c869cf3522b0316976fd3d328fbb3785264eb956b21909953a4cfed266a`

cc @phillipsj


imjasonh:
FYI: I'm also checking that `mutate.Windows` works with `ko` in https://github.com/google/ko/compare/main...imjasonh:crane-windows to make sure that this doesn't regress over there.

phillipsj:
I will be doing some testing on the PR in the morning. LGTM so far.

phillipsj:
So I just created a .NET 6.0 and placed it in a container using the following command.

```
crane append --platform=windows/amd64 -f app.tar -t phillipsj/craneapp:0.4.0 -b mcr.microsoft.com/dotnet/aspnet:6.0-nanoserver-1809

docker run -it -p 5000:80 phillipsj/craneapp:0.4.0 dotnet craneapp.dll
````
And it works:
image

Now I have been running mutate to change the entrypoint to be set to the command and that actually fails. That seems like a different issue that isn't directly related to creating the image.

**Comment by jonjohnsonjr** [comment]
> Now I have been running mutate to change the entrypoint to be set to the command and that actually fails.

How is it failing?

**Comment by phillipsj** [comment]
Here is the error I get after mutating the entrypoint and trying to run on a host.

```
docker: Error response from daemon: container 4f479dba6adc18da6995ec93c2d728c571882a6186ecc5c77dbe960d47b6d806 encountered an error during hcsshim::System::CreateProcess: failure in a Windows system call: The system cannot find the file specified. (0x2)
[Event Detail:  Provider: 00000000-0000-0000-0000-000000000000]
[Event Detail:  Provider: 00000000-0000-0000-0000-000000000000]
[Event Detail: onecore\vm\compute\management\orchestration\vmhostedcontainer\processmanagement.cpp(173)\vmcomputeagent.exe!00007FF73F4FEFBD: (caller: 00007FF73F4A4E97) Exception(2) tid(3c8) 80070002 The system cannot find the file specified.
    CallContext:[\Bridge_ProcessMessage\VmHostedContainer_ExecuteProcess]
 Provider: 00000000-0000-0000-0000-000000000000].
```

imjasonh:
> Here is the error I get after mutating the entrypoint and trying to run on a host.
> 
> ```
> docker: Error response from daemon: container 4f479dba6adc18da6995ec93c2d728c571882a6186ecc5c77dbe960d47b6d806 encountered an error during hcsshim::System::CreateProcess: failure in a Windows system call: The system cannot find the file specified. (0x2)
> [Event Detail:  Provider: 00000000-0000-0000-0000-000000000000]
> [Event Detail:  Provider: 00000000-0000-0000-0000-000000000000]
> [Event Detail: onecore\vm\compute\management\orchestration\vmhostedcontainer\processmanagement.cpp(173)\vmcomputeagent.exe!00007FF73F4FEFBD: (caller: 00007FF73F4A4E97) Exception(2) tid(3c8) 80070002 The system cannot find the file specified.
>     CallContext:[\Bridge_ProcessMessage\VmHostedContainer_ExecuteProcess]
>  Provider: 00000000-0000-0000-0000-000000000000].
> ```

Well I definitely have no idea what any of that means. 😅 

I'm inclined to forge ahead with this PR though, and start another with an e2e that does `crane append` and `crane mutate --entrypoint` and see what we can fix about it.

phillipsj:
Yeah I think that is a good idea to separate out those two concerns. It
works if you pass the command.

jonjohnsonjr (approved):
This is sufficiently punted onto future me's plate, LGTM.

PR #1181, opened 2021-11-16
Check for Podman's auth.json in DefaultKeychain

### LINKED ISSUE #1145: authn: Also consult ~/.config/containers/auth.json for auth config
See https://github.com/containers/podman/issues/11923 and https://github.com/tektoncd/operator/issues/448

Currently [`auth.DefaultKeychain`](https://pkg.go.dev/github.com/google/go-containerregistry/pkg/authn#DefaultKeychain) only consults ~/.docker/config.json, and fails if it's not present.

Podman and other tools also consult ~/.config/containers/auth.json (mainly to avoid the "docker" name, AFAICT), so Podman users won't be able to easily use tools built using go-containerregistry.

`DefaultKeychain` could consult ~/.config/containers/auth.json if ~/.docker/config.json is not found. AFAIK they have the same expected structure.

imjasonh:
Fixes #1145 

This changes `authn.DefaultKeychain` to always check Podman's preferred auth config location (`${XDG_RUNTIME_DIR}/containers/auth.json`) for auth configuration first, and only if that file doesn't exist, fallback to using Docker's `config.Load` and `${DOCKER_CONFIG}` to look for the user's registry auth config.

This matches Podman's [documented behavior](https://docs.podman.io/en/latest/markdown/podman-login.1.html):

> The default path for reading and writing credentials is ${XDG_RUNTIME_DIR}/containers/auth.json. Podman will use existing credentials if the user does not pass in a username. Podman will first search for the username and password in the ${XDG_RUNTIME_DIR}/containers/auth.json, if they are not valid, Podman will then use any existing credentials found in $HOME/.docker/config.json.

This may result in a subtle behavior change for users who have auth configured in both locations. Before this change, the Podman config would be ignored completely; after this change, it will be preferred over Docker's preferred location.

### DIFF
--- a/pkg/authn/keychain.go
+++ b/pkg/authn/keychain.go
@@ -15,10 +15,13 @@
 package authn
 
 import (
+	"errors"
 	"os"
+	"path/filepath"
 	"sync"
 
 	"github.com/docker/cli/cli/config"
+	"github.com/docker/cli/cli/config/configfile"
 	"github.com/docker/cli/cli/config/types"
 	"github.com/google/go-containerregistry/pkg/name"
 )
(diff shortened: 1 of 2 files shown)


jonjohnsonjr:
> This may result in a subtle behavior change for users who have auth configured in both locations. Before this change, the Podman config would be ignored completely; after this change, it will be preferred over Docker's preferred location.

I think for this reason I'd prefer falling back to podman config rather than falling back to docker config. Any reason to do this order?

imjasonh:
It was a bit more difficult because `config.Load` doesn't tell us if the file doesn't exist, but I just did an `os.Stat` on `${DOCKER_CONFIG}/config.json` first and that seems to work.

jonjohnsonjr (approved):

PR #1198, opened 2021-12-06
Allow uppercase repository names with `name.WeakValidation` (default)

mattmoor:
We recently discovered that `docker push` implicitly lowercases repository names, which isn't supported by GGCR.  This adds it to the list of magic :sparkles: that we do under `name.WeakValidation` to align with what docker supports.

Related: https://github.com/actions/starter-workflows/issues/1293

### DIFF
--- a/pkg/name/digest_test.go
+++ b/pkg/name/digest_test.go
@@ -38,6 +38,7 @@ var goodStrictValidationTagDigestNames = []string{
 var goodWeakValidationDigestNames = []string{
 	"namespace/pathcomponent/image@" + validDigest,
 	"library/ubuntu@" + validDigest,
+	"gcr.io/CROSSplane/PROVIDER-gcp@" + validDigest,
 }
 
 var goodWeakValidationTagDigestNames = []string{
--- a/pkg/name/ref_test.go
+++ b/pkg/name/ref_test.go
@@ -26,11 +26,13 @@ var (
 		"crossplane/provider-gcp:v0.14.0",
 		"ubuntu",
 		"gcr.io/crossplane/provider-gcp:latest",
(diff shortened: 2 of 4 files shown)


mattmoor:
Failing example: https://github.com/bishal-pdMSFT/docker-app/runs/4434089694?check_suite_focus=true#step:7:187

pkg/name/digest_test.go
```diff
@@ -38,6 +38,7 @@ var goodStrictValidationTagDigestNames = []string{
 var goodWeakValidationDigestNames = []string{
 	"namespace/pathcomponent/image@" + validDigest,
 	"library/ubuntu@" + validDigest,
+	"gcr.io/CROSSplane/PROVIDER-gcp@" + validDigest,
```
imjasonh, line 41: Do we care to transform and test `GCR.io` as well?

jonjohnsonjr (approved):

pkg/name/digest_test.go
jonjohnsonjr: Given that domain names are case-insensitive, I could be convinced that "we don't need to" but also "we can if we want".

mattmoor: DNS names are case insensitive, so technically we could, and this would be fine.  I guess I'd be more interested in what the Docker behavior here is, and matching that.  I am in a meeting now and don't have a docker setup, so if you are able to quickly tell me what it does I'll patch after I get out.  👍

imjasonh (approved):

mattmoor:
@imjasonh found that this is actually the `metadata-action` doing stuff, so perhaps we don't actually want this.

PR #1199, opened 2021-12-07
Added env vars and optional layers to mutate cmd

ehmm:


cmd/crane/cmd/mutate.go
```diff
@@ -78,6 +87,14 @@ func NewCmdMutate(options *[]crane.Option) *cobra.Command {
 				return err
 			}
 
+			// set envvars if specified - split to a function as it is more
+			// complex then labels because in the config the key value pairs
+			// are just strings with a = in them
```
imjasonh, line 92: Thanks for this context, and I agree, but I'm not sure the comment is useful outside the context of this code review.
```suggestion
			// set envvars if specified.
```

cmd/crane/cmd/mutate.go
```diff
@@ -78,6 +87,14 @@ func NewCmdMutate(options *[]crane.Option) *cobra.Command {
...
+			err = setEnvVars(&cfg.Config, envVars)
+			if err != nil {
+				return err
+			}
```
imjasonh, line 96: nit:

```suggestion
			if err := setEnvVars(&cfg.Config, envVars); err != nil {
				return err
			}
```

(This makes it clearer that the `err` isn't going to escape this scope)

imjasonh (commented):
We'll need to regen docs again, but overall this looks great. Much simpler than the whole `build` command, and pretty much just as powerful, especially with the `<(tar ...)` trick documented in recipes.md.

Thanks for doing this!

ehmm:
Glad this fits better.
I've fixed the nits. Let me know what else is needed to gets this upstreamed

Thanks

imjasonh:
Need to run `./hack/update-codegen.sh` to generate docs for the newly introduced flags. Other than that, lgtm 🚀

cmd/crane/cmd/mutate.go
```diff
@@ -124,9 +138,11 @@ func NewCmdMutate(options *[]crane.Option) *cobra.Command {
 	}
 	mutateCmd.Flags().StringToStringVarP(&annotations, "annotation", "a", nil, "New annotations to add")
 	mutateCmd.Flags().StringToStringVarP(&labels, "label", "l", nil, "New labels to add")
+	mutateCmd.Flags().StringToStringVarP(&envVars, "env", "e", nil, "New envvar to add")
 	mutateCmd.Flags().StringSliceVar(&entrypoint, "entrypoint", nil, "New entrypoint to set")
 	mutateCmd.Flags().StringSliceVar(&cmd, "cmd", nil, "New cmd to set")
 	mutateCmd.Flags().StringVarP(&newRef, "tag", "t", "", "New tag to apply to mutated image. If not provided, push by digest to the original image repository.")
+	mutateCmd.Flags().StringSliceVarP(&newLayers, "new_layer", "f", []string{}, "Path to tarball to append to image")
```
jonjohnsonjr, line 145: I'm tempted to change this flag to just `--append` and not include a short flag.

(Within the context of the `append` command, `--new_layer` and `-f` make sense, but not so much in a generic `mutate` command.)

imjasonh (approved):

cmd/crane/cmd/mutate.go
```diff
@@ -139,3 +155,23 @@ func validateKeyVals(kvPairs map[string]string) error {
...
 	return nil
 }
+
+// setEnvVars override envvars in a config
+func setEnvVars(cfg *v1.Config, envVars map[string]string) error {
+	existing := map[string]string{}
+	for _, kv := range cfg.Env {
+		split := strings.SplitN(kv, "=", 2)
+		if len(split) != 2 {
+			return fmt.Errorf("Invalid key value pair in config: %s", kv)
```
imjasonh, line 165:
```suggestion
			return fmt.Errorf("invalid key value pair in config: %s", kv)
```

cmd/crane/cmd/mutate.go
```diff
@@ -139,3 +155,23 @@ func validateKeyVals(kvPairs map[string]string) error {
 	}
 	return nil
 }
+
+// setEnvVars override envvars in a config
+func setEnvVars(cfg *v1.Config, envVars map[string]string) error {
```
jonjohnsonjr, line 160: This implementation will shuffle the order of environment variables due to golang's random map iteration order. To be safe, I think we want to preserve the existing order.

Also, I believe linux environment variable keys are case sensitive, but windows keys are case insensitive.

ehmm: I can sort alphabetically -in order to keep stable - but I don't think preserving is worth the extra code do you?
Re windows - as  crane and OCI support multi-arch what would you prefer?
I was thinking as anyway this is a util - we can validate that all new ones are upper-case (the convention on Linux and enforced on Windows) - and roll with that - or we only convert/filter if there is a Windows  image.

WYDT?

ehmm: @jonjohnsonjr waiting to see regarding case sensitivity

jonjohnsonjr: > I can sort alphabetically -in order to keep stable - but I don't think preserving is worth the extra code do you?

I'd like to keep the original order if we can. The approach I'd take is to iterate over the original list, replace any value if it's in the new `envVars` map and remove that from `envVars`. If `envVars` isn't not empty at the end, append any remaining values.

I don't care about the order of _new_ environment variables that much. I mostly want to avoid disturbing the order of existing environment variables.

> or we only convert/filter if there is a Windows image

I think I'd prefer checking for windows (like we do in [append](https://github.com/google/go-containerregistry/blob/c75901cce6272aea6bbcd9b517ba688d63a413f5/pkg/crane/append.go#L45)) and do case-insensitive comparison.

For windows, I'd uppercase all new env keys in the map so we can just do something like:

```
old := split[0]
if windows {
  old = strings.ToUpper(old)
}
if v, ok := envVars[old]; ok {
  split[1] = v 
}
```

ehmm: Done

cmd/crane/cmd/mutate.go
```diff
@@ -157,21 +157,30 @@ func validateKeyVals(kvPairs map[string]string) error {
...
+	for _, old := range cfg.Config.Env {
+		split := strings.SplitN(old, "=", 2)
 		if len(split) != 2 {
-			return fmt.Errorf("invalid key value pair in config: %s", kv)
+			return fmt.Errorf("invalid key value pair in config: %s", old)
 		}
-		existing[split[0]] = split[1]
+		oldKey := split[0]
+		// keep order so override if specified again
+		if  v, ok := envVars[oldKey]; ok {
```
jonjohnsonjr, line 169:
```suggestion
		if v, ok := envVars[oldKey]; ok {
```

jonjohnsonjr: LGTM otherwise.

ehmm: Fixed

cmd/crane/cmd/mutate.go
```diff
@@ -139,3 +155,32 @@ func validateKeyVals(kvPairs map[string]string) error {
...
+		}
+		// keep order so override if specified again
+		oldKey := split[0]
+		if v, ok := envVars[oldKey]; ok {
+			newEnv = append(newEnv, fmt.Sprintf("%s=%s", oldKey, v))
+			delete(envVars, oldKey)
+		} else {
+			newEnv = append(newEnv, old)
+		}
+
```
jonjohnsonjr, line 175:
```suggestion
```

jonjohnsonjr (approved):

Issue #1150, opened 2021-10-18
Q: easier way to grab/read a layer?


06kellyjac:
Related #1011

If I want to read provenance from an `.att` image pushed by tekton chains

I currently have to do the following:

```
$ crane pull my/image:some-tag tmp.tar.gz
$ tar xvf tmp.tar.gz
sha256:19648729e40a4e798b7a61d30901430d5808a53b0621be36d7f6f9a2fd0ec540
e2307fd561d470611ad337485f0c53b7b8b79473005508788a6c5a9d40faf563.tar.gz
manifest.json
$ mv e2307fd561d470611ad337485f0c53b7b8b79473005508788a6c5a9d40faf563.tar.gz my-provenance.json
$ cat my-provenance.json | jq
```

I can shorten this and write a script etc but is there a better way to just grab `e2307fd561d470611ad337485f0c53b7b8b79473005508788a6c5a9d40faf563.tar.gz` (or the contents) in the first place?

---

This is a function I've written

```bash
# takes an image
getAtt() { crane pull "$1" /dev/stdout | tar xfO /dev/stdin "$(crane manifest "$1" | jq -r .layers[0].digest | cut -d':' -f2).tar.gz"; }
```

PR #1203, opened 2021-12-14
Make crane export more ergonomic

jonjohnsonjr:
Default second arg to stdout so I don't have to type "-".

For images with a single layer, avoid calling mutate.Extract to flatten
the filesystem, instead we just write the uncompressed bytes to stdout.
In order to make that work, partial.Compressed now actually detects
whether the payload is gzipped or not when calling Uncompressed instead of just
assuming all blobs are compressed (and failing).

Fixes https://github.com/google/go-containerregistry/issues/1150


imjasonh (approved):
Nice.

PR #1205, opened 2021-12-14
Add --working_dir to mutate cmd

0x5A6I:


imjasonh (approved):

cmd/crane/cmd/mutate.go
```diff
@@ -183,3 +189,4 @@ func setEnvVars(cfg *v1.ConfigFile, envVars map[string]string) error {
 	cfg.Config.Env = newEnv
 	return nil
 }
+
```
imjasonh, line 192:
```suggestion
```

cmd/crane/cmd/mutate.go
```diff
@@ -143,6 +148,7 @@ func NewCmdMutate(options *[]crane.Option) *cobra.Command {
 	mutateCmd.Flags().StringSliceVar(&cmd, "cmd", nil, "New cmd to set")
 	mutateCmd.Flags().StringVarP(&newRef, "tag", "t", "", "New tag to apply to mutated image. If not provided, push by digest to the original image repository.")
 	mutateCmd.Flags().StringSliceVar(&newLayers, "append", []string{}, "Path to tarball to append to image")
+	mutateCmd.Flags().StringVar(&workingDir, "working_dir", "", "New WorkingDir to set")
```
jonjohnsonjr, line 151: I'd make this flag `--workdir` to match Dockerfile's `WORKDIR` mostly because I don't like underscores in flags... @imjasonh any strong opinions?

imjasonh: I hate all the possible options roughly equally.

I wish Cobra supported flag aliases better, I'd want all of `--workingdir`, `--working_dir`, `--working-dir`, `--workdir`, etc., to map to the same value.

Given the options I think I actually prefer `--working_dir` over `--workdir`, since that's what it's called in the config, and `WORKDIR` seems to only be a Dockerfileism.

0x5A6I: Changed to --working_dir

cmd/crane/cmd/mutate.go
```diff
@@ -143,6 +152,9 @@ func NewCmdMutate(options *[]crane.Option) *cobra.Command {
...
+	mutateCmd.Flags().BoolVar(&argsEscaped, "args_escaped", false, "[true|false] Command is already escaped (Windows only)")
```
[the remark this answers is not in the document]
jonjohnsonjr: What does this do? https://github.com/opencontainers/image-spec/pull/829

Is there a way we can just not have it?

PR #1206, opened 2021-12-15
Support specifying osversion in the --platform flag

imjasonh:
This removes the --osversion flag, and lets users specify the osversion
in the --platform flag itself, separated by a ":"

For example:

    crane manifest golang:1.17 --platform=windows/amd64:10.0.17763.2366

This matches the proposed behavior in https://github.com/google/ko/pull/536

cc @phillipsj -- is this format acceptable for your uses? `ko` has a stronger need to be able to specify the osversion in the platform, since it allows specifying multiple values, but I'd really like to keep them aligned if we can.


phillipsj:
Yep, that works for me.

jonjohnsonjr (approved):

Issue #1116, opened 2021-09-05
Add support for OCI layouts in `crane pull/push`


mattmoor:
Tools like `kaniko` and `ko` support writing OCI layouts, but i was surprised to find that `crane push` doesn't seem to support pushing those: https://github.com/google/go-containerregistry/blob/7e0ed51a7bb1930bfb6859f319975b578ed04de2/pkg/crane/push.go#L33-L44

Along the same lines, `kaniko` has a [`warmer` image](https://github.com/GoogleContainerTools/kaniko/blob/master/README.md#caching-base-images) that can be used to cache base images on disk (not sure if oci-layout!), so there's definitely a `crane pull` analog.

---

Being able to efficiently separate fetch/publish from what these tools do will become increasingly critical for hermetic builds where folks want to:
1. pull
2. build (shut off network)
3. push

mattmoor:
cc @jonjohnsonjr @imjasonh

Issue #726, opened 2020-05-30
`crane push` doesn't support OCI image tarballs


jspawar:
One of the first actions `crane push` undertakes is to obtain the image manifest as such: https://github.com/google/go-containerregistry/blob/a873a21aff2360d401b48410e63edbdc3c97bc8c/pkg/v1/tarball/image.go#L155

However, `manifest.json` is not defined in the OCI image-layout spec and this command should instead be opening up `index.json` and following its contents to eventually obtain the manifest in the image tarball to be pushed per the OCI spec: https://github.com/opencontainers/image-spec/blob/master/image-layout.md#oci-image-layout-specification

I then discovered you all have a helpful readme for that package which states:

> This package produces tarballs that can consumed via `docker load`

So I see this implementation is indeed Docker-specific and wondering if y'all are open to accepting changes which would allow `crane push` to support pushing OCI image tarballs? Perhaps behind a flag to the push command using your `layout` package implementation? 

Don't know if there are any issues with pushing to certain registries such as GCR or Docker Hub (i.e. they don't support pushing/pulling OCI images), so would be good to know that as well as any other context I might be missing if you don't mind. 

Thank you!

ImJasonH:
I'd definitely be in support of adding this. What CLI flags were you thinking exactly?

jspawar:
@ImJasonH perhaps something simple on the push command like `--layout` which would accept some enum value to indicate what type of tarball layout your input is in? 

E.g. `crane push --layout docker` or `crane push --layout oci`

Not really picky and can even leave Docker as the default layout to mitigate introducing a breaking change.

ImJasonH:
A new `--layout` flag which defaults to `docker` for compatibility sgtm, @jonjohnsonjr any objection?

jonjohnsonjr:
That seems reasonable to me, though I'd probably go with `--format` over `--layout` since [we already have that for pull](https://github.com/google/go-containerregistry/blob/2a1a46d929f8e396080340669edf5456a67493b5/cmd/crane/cmd/pull.go#L65).

jspawar:
@jonjohnsonjr @ImJasonH Great! Also like the sound of `--format` over `--layout`.

Would you all prefer a contribution for this or is there some sort of public backlog y'all will add this into?

jonjohnsonjr:
We'll take any contributions we can get 😄

t0rr3sp3dr0:
Any updates on this?

PR #1208, opened 2021-12-16
crane push: Support OCI layout

jonjohnsonjr:
Detect if the first argument is a directory. If so, attempt to read it
as an OCI image layout and push to the destination if it contains a
single image.

Fixes https://github.com/google/go-containerregistry/issues/726
Fixes https://github.com/google/go-containerregistry/issues/1116


mattmoor:
nice!  Will TAL.

cc @dlorenc @priyawadhwa

cmd/crane/cmd/push.go
```diff
@@ -15,26 +15,74 @@
...
+		return nil, fmt.Errorf("loading %s as OCI layout: %w", path, err)
+	}
+	m, err := l.IndexManifest()
+	if err != nil {
+		return nil, err
+	}
+	if len(m.Manifests) != 1 {
+		return nil, errors.New("pushing multiple images from layout not yet supported (PRs welcome)")
+	}
+	return l.Image(m.Manifests[0].Digest)
```
mattmoor, line 74: Haven't you marked this as `Fixes: #{the issue covering pushing layouts}`?

Maybe we should leave one open

cmd/crane/cmd/push.go
```diff
@@ -15,26 +15,74 @@
...
+}
+
+func isDir(path string) (bool, error) {
+	file, err := os.Open(path)
+	if err != nil {
+		return false, err
+	}
+	defer file.Close()
+	fi, err := file.Stat()
```
mattmoor, line 83: Why not `os.Stat()`?

mattmoor (commented):
Can you add a little roundtrip e2e test where we `crane pull` -> `crane push` -> `crane pull` and check for diffs?

For extra credit, I'd love to check that it works with https://github.com/GoogleContainerTools/kaniko/blob/master/README.md#--oci-layout-path

cmd/crane/cmd/push.go
jonjohnsonjr: nice

cmd/crane/cmd/push.go
jonjohnsonjr: There are ~3 ways I can imagine people wanting to use this, let's chat.

mattmoor: Did we resolve the next steps here?  I'm fine with landing this part as-is, assuming it's good enough for `kaniko`'s `--oci-layout`.  However, I'd like a clear path for how we support the same for `ko`.

mattmoor:
I think the main thing I'd want to see to land this is a roundtrip test, but otherwise it LGTM.  I'd even be fine if we plan to change the roundtrip test as the semantics of how we handle single images evolves later, I mostly want the confidence that things WAI.

mattmoor:
Alternately (or additionally), a test that runs `kaniko --oci-layout` -> `crane push` would be 🤩

cmd/crane/cmd/push.go
```diff
@@ -35,18 +38,31 @@ func NewCmdPush(options *[]crane.Option) *cobra.Command {
...
+			if idx, ok := img.(v1.ImageIndex); ok {
+				// TODO(generics): Make crane.Push support index.
+				o := crane.GetOptions(*options...)
+				ref, err := name.ParseReference(tag, o.Name...)
+				if err != nil {
+					return err
+				}
+				return remote.WriteIndex(ref, idx, o.Remote...)
+			}
+			return fmt.Errorf("cannot push type (%T) to registry", img)
```
mattmoor, line 58: I'd consider a type switch with a default failure case, I think it might be slightly clearer.

cmd/crane/cmd/push.go
```diff
@@ -35,18 +38,31 @@ func NewCmdPush(options *[]crane.Option) *cobra.Command {
...
 		},
 	}
+	cmd.Flags().BoolVar(&index, "index", false, "Push the collection of images as a single index")
```
mattmoor, line 61: My interpretation of this text is that the fallback behavior is to push N images instead, which is not what the code does, so perhaps some clarity is called for there?

cmd/crane/cmd/push.go
jonjohnsonjr: Done.

mattmoor (approved):

imjasonh (approved):

PR #1209, opened 2021-12-16
Pluggable blob storage for pkg/registry

imjasonh:
Features:
- supports redirects if the blob handler hosts blobs elsewhere.
- supports automatically verifying digests/sizes if possible, transparent to the blob handler (if they want it to be).
- repo is passed to `blobHandler` methods as `string` and not `name.Repository`, but they should be easily parsed by implementations if they want that.

This is still hidden in unexported interfaces and always uses memHandler for now, while we iterate on the shape we want these things to take.

Longer term, we should also extend this to manifest and upload storage, using similar interfaces.


pkg/registry/blobs.go
```diff
@@ -44,10 +47,90 @@ func isBlob(req *http.Request) bool {
 		elem[len(elem)-2] == "uploads")
 }
 
+// blobHandler represents a blob storage backend.
+//
+// For all methods, repo is a string that can be passed to
+// pkg/name.NewRepository to validate the repository.
+type blobHandler interface {
```
jonjohnsonjr, line 54: These should take a `context.Context`.

pkg/registry/blobs.go
```diff
@@ -44,10 +47,90 @@ func isBlob(req *http.Request) bool {
...
+	// Stat returns the size of the blob, or ErrNotFound if the blob wasn't found.
+	Stat(repo string, h v1.Hash) (int64, error)
```
jonjohnsonjr, line 56: If we want to extend this interface to manifests, it would be great if `Stat` returned a media type as well.

Similarly, if we take a `string` instead of a `v1.Hash` this would work for manifests.

pkg/registry/blobs.go
```diff
@@ -44,10 +47,90 @@ func isBlob(req *http.Request) bool {
...
+	// as the contents are read, and an error will be returned if these
+	// don't match. Implementations should return that error, or a wrapper
+	// around that error, to return the correct error when these don't match.
+	Put(repo string, h v1.Hash, rc io.ReadCloser) error
+}
+
+// RedirectError represents a signal that the blob handler doesn't have the blob
+// contents, but that those contents are at another location which registry
+// clients should redirect to.
+type RedirectError struct {
```
jonjohnsonjr, line 73: Don't expose this yet.

pkg/registry/blobs.go
```diff
@@ -44,10 +47,90 @@ func isBlob(req *http.Request) bool {
...
+	Location string
+
+	// Code is the HTTP redirect status code to return to clients.
+	Code int
+}
+
+func (e RedirectError) Error() string { return fmt.Sprintf("Redirecting (%d): %s", e.Code, e.Location) }
+
+// ErrNotFound represents an error locating the blob.
+var ErrNotFound = errors.New("not found")
```
jonjohnsonjr, line 84: Don't expose this yet.

pkg/registry/blobs.go
imjasonh: Ack. We'll cross that bridge when we get to it, uploads and manifests are next on the list.

pkg/registry/blobs.go
```diff
@@ -44,10 +48,90 @@ func isBlob(req *http.Request) bool {
...
+	// as the contents are read, and an error will be returned if these
+	// don't match. Implementations should return that error, or a wrapper
+	// around that error, to return the correct error when these don't match.
+	Put(ctx context.Context, repo string, h v1.Hash, rc io.ReadCloser) error
+}
+
+// redirectError represents a signal that the blob handler doesn't have the blob
+// contents, but that those contents are at another location which registry
+// clients should redirect to.
+type redirectError struct {
```
mattmoor, line 74: I kinda regret letting Scott use errors for non-error cases like this in `genreconciler`, but that ship sailed there.  Perhaps we can have the methods return a `Result` type or something, which can be an error, but don't have to be an error.

pkg/registry/blobs.go
```diff
@@ -44,10 +48,90 @@ func isBlob(req *http.Request) bool {
 		elem[len(elem)-2] == "uploads")
 }
 
+// blobHandler represents a blob storage backend.
+//
+// For all methods, repo is a string that can be passed to
+// pkg/name.NewRepository to validate the repository.
+type blobHandler interface {
```
mattmoor, line 55: I believe this must be public to be useful.

I think one of the things I realized in my fork was that I wanted the default implementation to be public as well, so that I could compose my own BlobHandler with the default to layer functionality on.

pkg/registry/blobs.go
```diff
@@ -220,21 +352,43 @@ func (b *blobs) handle(resp http.ResponseWriter, req *http.Request) *regError {
...
+		size := int64(verify.SizeUnknown)
+		if req.ContentLength > 0 {
+			size = int64(len(b.uploads[target])) + req.ContentLength
+		}
+
+		vrc, err := verify.ReadCloser(in, size, h)
+		if err != nil {
+			return errTODO(err.Error())
+		}
+		defer vrc.Close()
```
mattmoor, line 377: Don't you have to check the error here for the verify.ReadCloser to be meaningful at all?

pkg/registry/blobs.go
mattmoor: Here's what I have in my fork:
```go
// BlobHandler is the interface for the storage layer underneath this registry.
type BlobHandler interface {
	// Stat returns the size of the blob whose hash is specified,
	// if it exists. If not, it returns (0, error).
	Stat(repo name.Repository, h v1.Hash) (int64, error)

	// Get returns true and a reader for consuming the blob specified with the hash,
	// if it exists.  It now, it returns (nil, error).
	Get(repo name.Repository, h v1.Hash) (io.ReadCloser, error)

	// Store stores the stream of content with the given hash, or returns the error
	// encountered doing so.
	Store(repo name.Repository, h v1.Hash, content io.ReadCloser) error
}
```

I think plumbing `context.Context` through makes sense, and am on board with that.

For completeness, this is what I have for manifests:
```go
// ManifestHandler is the interface for the metadata storage layer underneath
// this registry.
type ManifestHandler interface {
	// ListRepositories enumerates up to some count of repositories starting at
	// a specified offset.  This is used to implement the "catalog" handler.
	ListRepositories(offset, count int) ([]name.Repository, error)

	// GetRepository fetches a handler for interacting with a collection of
	// manifests rooted under a particular repository.
	GetRepository(repo name.Repository) (RepositoryHandler, error)
}

// RepositoryHandler is the interface for accessing manifest data under a
// particular repository.
type RepositoryHandler interface {
	// GetDigest fetches the raw bytes of the manifest, and its media type,
	// or returns an error indicating why it cannot.
	GetDigest(v1.Hash) ([]byte, types.MediaType, error)
	// GetTag fetches the raw bytes of the manifest currently labeled by the
	// provided tag, and its media type, or returns an error indicating why
	// it cannot.
	GetTag(string) ([]byte, types.MediaType, error)

	// DeleteDigest removes the manifest with the given hash, it returns an
	// error when the digest does not exist.
	DeleteDigest(v1.Hash) error
	// DeleteTag removes the tag with the given name, the digest is left untouched,
	// it returns an error when the digest does not exist.
	DeleteTag(string) error

	// PutDigest adds the provided manifest content, with the associated media type
	// to the store under the provided hash name.  It returns an error if this cannot
	// be completed.
	PutDigest(v1.Hash, []byte, types.MediaType) error
	// PutTag applies the provided tag to the given hash.  It returns an error if
	// this is not possible.
	PutTag(string, v1.Hash) error

	// ListTags enumerates up to some count of tags starting at
	// a specified offset.  This is used to implement the "tags" handler.
	ListTags(offset, count int) ([]string, error)
}
```

pkg/registry/blobs.go
imjasonh: `verify.ReadCloser` verifies when `inner.Read` hits and `EOF`, not on close.

pkg/registry/blobs.go
imjasonh: Yeah the intention is that we add support for manifests and uploads and then export it when we agree we like the shape. This gives us maximum flexibility to change things willy nilly.

pkg/registry/blobs.go
imjasonh: I think I can justify this as a valid error scenario: "oops, I don't have it, but I know who does"

I could collapse this with `errNotFound`, so the same error type expresses "oops not found[, but I know who does]" if `Location` is set.

pkg/registry/blobs.go
```diff
@@ -72,40 +156,78 @@ func (b *blobs) handle(resp http.ResponseWriter, req *http.Request) *regError {
...
+		if err != nil {
+			return &regError{
+				Status:  http.StatusBadRequest,
+				Code:    "NAME_INVALID",
+				Message: "invalid digest",
+			}
+		}
+
+		size, err := b.blobHandler.Stat(req.Context(), repo, h)
+		if errors.Is(err, errNotFound) {
```
imjasonh, line 173: It occurs to me this should handle redirect signals too.

pkg/registry/blobs.go
mattmoor: At one point this is exactly how Docker checked digests and there was a bug that turned out to be an unchecked error on (drumroll) `io.Copy` 😬 

I haven't looked carefully to know if this has that problem, but there are lots of unchecked `io.Copy` calls in this file.

pkg/registry/blobs.go
mattmoor: 4xx and 5xx are clear error cases, but 3xx feels like a reach.

I'm not gonna die on this hill, it just feels like a funny API.

pkg/registry/blobs.go
imjasonh: Checking those errors exposed a bug in `verify.ReadCloser` when the size isn't known ahead of time. Fixed, with test that fails without the change. 🐛 🔨

pkg/registry/blobs.go
imjasonh: I don't disagree, it's weird. Luckily everything's still internal so we can change it to something we like better if we find it.

imjasonh:
Friendly ping, I have upload handling in the next PR which I'll send out after this one merges.

pkg/registry/blobs.go
```diff
@@ -72,43 +160,126 @@ func (b *blobs) handle(resp http.ResponseWriter, req *http.Request) *regError {
...
+				return regErrBlobUnknown
+			} else if err != nil {
+				var rerr redirectError
+				if errors.As(err, &rerr) {
+					http.Redirect(resp, req, rerr.Location, rerr.Code)
+					return nil
+				}
+				return regErrInternal(err)
+			}
+		} else {
```
mattmoor, line 189: Alternately `Stat()` could be an optional method on `BlobHandler` that folks can use to implement a more efficient stat operation and this could be a type assertion.

pkg/registry/blobs.go
```diff
@@ -72,43 +160,126 @@ func (b *blobs) handle(resp http.ResponseWriter, req *http.Request) *regError {
...
+					http.Redirect(resp, req, rerr.Location, rerr.Code)
+					return nil
+				}
+				return regErrInternal(err)
+			}
+			defer rc.Close()
+			size, err = io.Copy(ioutil.Discard, rc)
+			if err != nil {
+				return regErrInternal(err)
+			}
```
mattmoor, line 205: Or we could require `Stat`, but provide a `partial.Blah` like method that "completes" the interface we expect.

pkg/registry/blobs.go
imjasonh: Done.

pkg/registry/blobs.go
imjasonh: What we have now is effectively the same, without a new partial type. It might be worth adding to make implementors' lives easier, but I'll put that off for now until these interfaces have settled some more.

mattmoor (approved):

PR #1211, opened 2021-12-16
Add crane serve to run a local registry

imjasonh:
TODO for future changes:
- `--readonly`
- `--oci-layout-path`


jonjohnsonjr (commented):
I'm tempted to make in-memory just not an option at all and always requiring a path for an OCI layout... wdyt?

imjasonh:
> I'm tempted to make in-memory just not an option at all and always requiring a path for an OCI layout... wdyt?

Good call. I'll put this on hold until we have a layout-backed `pkg/registry` and pick it back up.

PR #1215, opened 2021-12-21
crane pull: support pulling index to OCI Layout

jonjohnsonjr:
Prior to this, we'd also resolve an index to an image when using crane
pull. For OCI Layouts, we can actually pull the whole index, so when
--format=oci and --platform is unset, just do that.

### DIFF
--- a/cmd/crane/cmd/pull.go
+++ b/cmd/crane/cmd/pull.go
@@ -18,8 +18,11 @@ import (
 	"fmt"
 
 	"github.com/google/go-containerregistry/pkg/crane"
+	"github.com/google/go-containerregistry/pkg/name"
 	v1 "github.com/google/go-containerregistry/pkg/v1"
 	"github.com/google/go-containerregistry/pkg/v1/cache"
+	"github.com/google/go-containerregistry/pkg/v1/layout"
+	"github.com/google/go-containerregistry/pkg/v1/remote"
 	"github.com/spf13/cobra"
 )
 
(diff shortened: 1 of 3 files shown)


cmd/crane/cmd/pull.go
```diff
@@ -33,16 +36,38 @@ func NewCmdPull(options *[]crane.Option) *cobra.Command {
 		Args:  cobra.MinimumNArgs(2),
 		RunE: func(_ *cobra.Command, args []string) error {
 			imageMap := map[string]v1.Image{}
+			indexMap := map[string]v1.ImageIndex{}
 			srcList, path := args[:len(args)-1], args[len(args)-1]
 			for _, src := range srcList {
-				img, err := crane.Pull(src, *options...)
+				o := crane.GetOptions(*options...)
+				ref, err := name.ParseReference(src, o.Name...)
```
mattmoor, line 43: Can we annotate the entries with where they come from?

cmd/crane/cmd/pull.go
```diff
@@ -33,16 +36,38 @@ func NewCmdPull(options *[]crane.Option) *cobra.Command {
...
+
+				rmt, err := remote.Get(ref, o.Remote...)
+				if err != nil {
+					return err
+				}
+
+				// If we're writing an index to a layout and --platform hasn't been set,
+				// pull the entire index, not just a child image.
+				if format == "oci" && rmt.MediaType.IsIndex() && o.Platform == nil {
+					idx, err := rmt.ImageIndex()
```
mattmoor, line 56: I'd just add an e2e test like this: https://github.com/google/go-containerregistry/blob/2874338840a65b73444563e6bd3540dd2f9271b0/.github/workflows/e2e.yaml#L63-L65

... but use `gcr.io/distroless/static`.  I'd include a variant with `--platform` as well.

imjasonh (approved):

mattmoor (approved):

PR #1216, opened 2021-12-23
Pluggable registry storage for uploads

imjasonh:
Continuing from #1158 

This adds interfaces and an in-memory implementation for storing partial uploads and committing them to finished blobs.

### DIFF
--- a/pkg/registry/blobs.go
+++ b/pkg/registry/blobs.go
@@ -87,22 +87,26 @@ type uploadHandler interface {
 	// upload contents, and returns the new total size.
 	AppendUpload(ctx context.Context, uploadID string, rc io.ReadCloser) (int64, error)
 
-	// FinishUpload appends the contents of the ReadCloser to the curent
-	// upload contents, and returns the total contents and their size.
-	FinishUpload(ctx context.Context, uploadID string, rc io.ReadCloser) (io.ReadCloser, int64, error)
+	// GetUpload returns the contents and size of the upload.
+	//
+	// If an implementation implements FinalizeUpload (below), then
+	// GetUpload will not be called, and can simply return an error.
+	GetUpload(ctx context.Context, uploadID string) (io.ReadCloser, int64, error)
+
+	// DeleteUpload deletes the upload contents.
+	DeleteUpload(ctx context.Context, uploadID string) error
 }
 
 // uploadFinalizeHandler is an extension interface representing upload storage
-// that can finalize contents into blob storage.
+// that can finalize contents directly into blob storage.
 type uploadFinalizeHandler interface {
-	// FinalizeUpload appends the contents of the ReadCloser to the current
-	// uplaod contents, checks the total contents digest matches, and
-	// finalizes it into blob storage.
+	// FinalizeUpload copies the upload contents into blob storage and
+	// deletes it from upload storage.
 	//
 	// Implementations that implement this method are responsible for
 	// verifying the given hash matches the total contents before
 	// persisting to blob storage.
-	FinalizeUpload(ctx context.Context, uploadID string, rc io.ReadCloser, h v1.Hash) error
+	FinalizeUpload(ctx context.Context, uploadID string, h v1.Hash) error
 }
 
 // redirectError represents a signal that the blob handler doesn't have the blob
@@ -136,6 +140,7 @@ func (m *memHandler) Stat(_ context.Context, _ string, h v1.Hash) (int64, error)
 	}
 	return int64(len(b)), nil
 }
+
 func (m *memHandler) Get(_ context.Context, _ string, h v1.Hash) (io.ReadCloser, error) {
 	m.lock.Lock()
 	defer m.lock.Unlock()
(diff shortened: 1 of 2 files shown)


pkg/registry/blobs.go
```diff
@@ -76,6 +76,35 @@ type blobPutHandler interface {
...
+	// StatUpload returns the current size of the given upload.
+	StatUpload(ctx context.Context, uploadID string) (int64, error)
+
+	// AppendUpload appends the contents of the ReadCloser to the current
+	// upload contents, and returns the new total size.
+	AppendUpload(ctx context.Context, uploadID string, rc io.ReadCloser) (int64, error)
+
+	// FinishUpload appends the contents of the ReadCloser to the curent
+	// upload contents, and returns the total contents and their size.
+	FinishUpload(ctx context.Context, uploadID string, rc io.ReadCloser) (io.ReadCloser, int64, error)
```
imjasonh, line 92: These names bother me. I think I want `GetUpload(uploadID) (io.ReadCloser, error)`, which naive upload finalization will call after `Append`ing the new bits, and then `DeleteUpload(uploadID) error` before `PutBlob` with the full upload contents.

Implementations that support `FinalizeUpload` can just have `GetUpload` and `DeleteUpload` always return an error since that's what `FinalizeUpload` is expected to do.

pkg/registry/blobs.go
```diff
@@ -76,6 +76,35 @@ type blobPutHandler interface {
 	Put(ctx context.Context, repo string, h v1.Hash, rc io.ReadCloser) error
 }
 
+// uploadHandler represents a minimal upload storage backend, capable of
+// appending streamed upload contents and finally returning them to be stored
+// in blob storage.
+type uploadHandler interface {
```
mattmoor, line 82: I'm not entirely sure I follow why this is separate from the `Put` interface above that enables making things writeable.  Nor why the "Finalize" stuff below is separate.

pkg/registry/blobs.go
```diff
@@ -126,20 +155,82 @@ func (m *memHandler) Put(_ context.Context, _ string, h v1.Hash, rc io.ReadClose
 	if err != nil {
 		return err
 	}
-	m.m[h.String()] = all
+	m.blobs[h.String()] = all
+	return nil
+}
+func (m *memHandler) StatUpload(_ context.Context, uploadID string) (int64, error) {
```
mattmoor, line 161: nit: space btw functions 🙏

pkg/registry/blobs.go
mattmoor: One variant I was thinking about was surfacing an upload interface for each of the possible upload styles:
 * POST/PATCH/PUT (most common)
 * POST/PUT
 * PUT

These each have names we could use.  I haven't fully thought this through, so mostly thinking out loud.

imjasonh: > I'm not entirely sure I follow why this is separate from the `Put` interface above that enables making things writeable.

A registry may only support some upload protocols and not others, and the way they'd tell us which they support is which interfaces they implement.

> Nor why the "Finalize" stuff below is separate.

Finalize is meant to be a shortcut if you have your own way of finalizing an upload into the final blob. In memHandler we don't re-read the whole blob to move it to a finalized blob, we just move it to the blob map.

I'm open to naming them better definitely, I'll try some things out this week.

mattmoor: > A registry may only support some upload protocols and not others

This is sort of what I was suggesting by re-orienting the interfaces around those upload protocols.  I believe each has a name, e.g. "monolithic upload".  Did that make sense?

> Finalize is meant to be a shortcut if you have your own way of finalizing an upload into the final blob

That makes sense.  I assume you are using this downstream for something.  Maybe just reflect this info into the comment block, if you haven't already?

imjasonh: I've thought about this a bit more and I don't know if "which upload protocols I support" matches very well to "what operations my backing store can support". At least not well enough to make them the one and only knob.

I may have an in-memory impl that still shouldn't accept pushes (even though `memHandler` supports it), and I might want to express those differently. OTOH, you might have a storage impl that doesn't support pushes, and just bringing that should be enough to tell us not to accept pushes.

I'm thinking that might look like:

```golang
h := registry.New(
    registry.WithStorage(layoutStore),
    registry.DisablePush,  // or just registry.DisableChunkedUpload
)
```

or

```golang
h := registry.New(
    registry.WithStorage(readOnlyLayoutStorage),
)
```

(these would be equivalent)

An alternative would be to provide a `ReadOnlyer` that takes whatever impl it's given and shadows the implementation of `PutBlob`, `*Upload`, etc.  -- that feels a bit like a too-cute use of Go's type system, and ultimately less expressive than just making the caller tell us whether to disable pushes.

This way, storage implementations get some control over what registry operations are allowed, based on what they support, but so do callers setting up the registry.

All of this might just indicate we're being too cute with the type system in general, and I should refocus on exactly the two impls we expect to see: in-memory and layout-on-disk, both of which support all these methods, including the shortcutting upload finalizer.

pkg/registry/blobs.go
```diff
@@ -76,6 +76,39 @@ type blobPutHandler interface {
...
+// uploadHandler represents a minimal upload storage backend, capable of
+// appending streamed upload contents and finally returning them to be stored
+// in blob storage.
+type uploadHandler interface {
+	// StatUpload returns the current size of the given upload.
+	StatUpload(ctx context.Context, uploadID string) (int64, error)
+
+	// AppendUpload appends the contents of the ReadCloser to the current
+	// upload contents, and returns the new total size.
+	AppendUpload(ctx context.Context, uploadID string, rc io.ReadCloser) (int64, error)
```
jonjohnsonjr, line 88: Can we [ab]use `io.WriterAt` to express this instead? Feels like it maps pretty well to chunked uploads.

PR #1225, opened 2022-01-01
RFC: pkg/authn/amazon 

imjasonh:
Opening this as a draft to gather feedback.

Today, `k8schain` provides support for implicit auth for ~~a number of popular clouds~~ GCR, ECR and ACR, via `k8s.io/kubernetes/pkg/credentialprovider` (actually, [Vincent's fork of that](https://github.com/vdemeester/k8s-pkg-credentialprovider)) -- see [`k8schain.NewNoClient`](https://pkg.go.dev/github.com/google/go-containerregistry/pkg/authn/k8schain#NewNoClient):

> What remains is an interesting sweet spot: this variant can serve as a credential provider for all of the major public clouds, but in library form (vs. an executable you exec).

Not requiring separate cred helper executables has been nice, but the large k8s dependency and maintenance burden has been less nice. If a user just wants implicit ECR auth, they for some reason need to import a bunch of K8s deps, including some nominally-maintained forked code, that also pulls in GCR and ACR implicit auth. It's a huge dependency bomb.

I'd like to propose moving k8schain toward _just_ being responsible for explicit Kubernetes Secrets-based auth in `imagePullSecrets`, and remove the implicit behavior over time.

Instead, we can provide `authn.Keychain` implementations for popular clouds, like we do for GCR today, based on their credential helper codebases. In this case, `pkg/authn/amazon` is heavily based on [code in `docker-credential-ecr-login`](https://github.com/awslabs/amazon-ecr-credential-helper/blob/f15886b5ccdc9802e373e5b33fb38feab3241b5d/ecr-login/ecr.go#L45). Because this adds a ~77k dependency, `pkg/authn/amazon` is in a separate Go module, like `k8schain`.

(We could similarly borrow code from ACR's cred helper, [here](https://github.com/chrismellard/docker-credential-acr-env/blob/09e2b5a8ac86/pkg/credhelper/helper.go#L45), and others as needs arise.)

This means a client that might consider auth between GCR and ECR, and possibly with creds in Kubernetes Secrets, could use:

```golang
kc := authn.NewMultiKeychain(google.Keychain, amazon.Keychain, k8schain.NewInCluster, ...)
```

This change also moves `pkg/v1/google`'s auth methods to `pkg/authn/google` and deprecates the `pkg/v1/google` equivalents. This is not in its own module, but could be, after we remove the `pkg/v1/google` flavors, to make depending on `golang.org/x/oauth2`, which transitively depends on the relatively-large `cloud.google.com/go`, more opt-in for users importing ggcr.

wdyt @jonjohnsonjr @mattmoor


mattmoor:
My $0.02 tl;dr is: +1.

k8schain was always a “cheap and cheerful” way of achieving this when (as Googlers) we couldn’t touch AWS stuff with a 10ft pole.

imjasonh:
If we go with this, I think we should split K8s secret auth into `pkg/authn/kubernetes` and have `k8schain` depend on the cloud-specific authns and that, and be deprecated.

mattmoor:
That sounds like a good plan to me.  NewNoClient is something I use a lot and we could totally shed the K8s deps 🤩

imjasonh:
Another option which might give us a bit more flexibility, and a bit less maintenance burden, would be to just provide an adapter from [Docker's `credentials.Helper` interface](https://pkg.go.dev/github.com/docker/docker-credential-helpers/credentials#Helper) -- which [`ecr-login`](https://pkg.go.dev/github.com/awslabs/amazon-ecr-credential-helper/ecr-login#ECRHelper) and [`acr-env`](https://pkg.go.dev/github.com/chrismellard/docker-credential-acr-env@v0.0.0-20210203204924-09e2b5a8ac86/pkg/credhelper#ACRCredHelper) implement -- to [our `authn.Keychain` interface](https://pkg.go.dev/github.com/google/go-containerregistry/pkg/authn#Keychain).

Inside `pkg/authn`, we could add:

```golang
// Subset of:
// https://pkg.go.dev/github.com/docker/docker-credential-helpers/credentials#Helper
type Helper interface {
        Get(serverURL string) (string, string, error)
}

func NewFromHelper(h Helper) Keychain { return wrapper{h} }

type wrapper struct{ h Helper }

func (w wrapper) Resolve(r Resource) (Authenticator, error) {
        u, p, err := w.h.Get(r.String())
        if err != nil {
                return Anonymous, nil
        }
        return FromConfig(AuthConfig{Username: u, Password: p}), nil
}
```

Then users could pass whatever `credential.Helper` impl they want:

```golang
kc := authn.NewMultiKeychain(
    authn.DefaultKeychain,
    google.Keychain,
    authn.NewFromHelper(ecr.ECRHelper{ClientFactory: api.DefaultClientFactory{}}),
    authn.NewFromHelper(acr.ACRCredHelper{}),
    ...
)
```

This is equivalent to having `pkg/authn/amazon` wrap that `ecr.ECRHelper{...}` code, but doesn't require us to maintain a module to provide it -- we'd just document that users who want to statically link the ECR cred helper logic can do so as described above, by directly depending on upstream `ecr-login` code.

mattmoor:
I like it, but I'd probably vote for both.

We should still have an `amazon.Keychain`, but that's just sugar around this.  Could we make `google.Keychain` do the same, or are there reasons not to?

imjasonh:
> I like it, but I'd probably vote for both.
>
> We should still have an `amazon.Keychain`, but that's just sugar around this. 

I was worried you'd say that 😆 

> Could we make `google.Keychain` do the same, or are there reasons not to?

`docker-credential-gcr` is a bit more to set up, but it does implement the necessary interface: 

https://github.com/GoogleCloudPlatform/docker-credential-gcr/blob/6093d30b51d725877bc6971aa6700153c1a364f1/credhelper/helper.go#L51

I don't know that `google.Keychain` is that much of a burden to us, that it warrants deleting it in favor of wrapping the cred helper's code. But having the adapter would let people try it out if they want it, to generate feedback.

imjasonh:
https://github.com/google/go-containerregistry/pull/1227 added the credential helper adapter, and documents how to use it by depending on ECR/ACR cred helpers yourself.

https://github.com/docker-credential-magic/docker-credential-magic/pull/41 adds `pkg/amazon` and `pkg/azure` to formally wrap those in a nice package, outside of ggcr.

PR #1226, opened 2022-01-01
Use Temp File For `layout.{Write,Append,Replace}` Image/Index Methods

ben-krieger:
This pull request was prompted by https://github.com/linuxkit/linuxkit/pull/3748.

Linuxkit, which uses an oci-layout cache, suffers the need for frequent cache clears, because interrupted downloads of layers results in images which fail `validate.Image` but cannot be repaired with a subsequent `(layout.Path).ReplaceImage`. This is due to the underlying call to `(layout.Path).WriteBlob` which does not overwrite existing files, even if they are smaller than the size the manifest claims they should be.

I found these comments in `pkg/v1/layout/write.go`:

```
// TODO: A streaming version of WriteBlob so we don't have to know the hash
// before we write it.

// TODO: For streaming layers we should write to a tmp file then Rename to the
// final digest.
```

I believe that this PR addresses both issues and will resolve the issue in Linuxkit, as well. However, I have no need for calling `WriteLayer` as library code, so I'm not opposed to making it unexported, as long as `ReplaceImage` makes use of its behaviors of overwriting undersized layers and/or writing blobs to a temporary location and renaming. 

---

Changes:

`WriteLayer` is an alternative to `WriteBlob` which can be used with streaming layers which have not yet calculated a digest or size. A temporary file is first written and then renamed when the reader has been fully consumed. Also, existing layers which do not have the correct size (and thus invalid hashes, but the hash is not checked) are overwritten.

`{Append,Replace,Write}{Image,Layer}` functions utilize `WriteLayer` instead of `WriteBlob` to correctly overwrite undersized layers which were previously ignored without error. Undersized layers should no longer occur, since `WriteLayer` writes to a temporary file first, but when `WriteBlob` was used, these were common due to incomplete downloads.

Improve error messages for `validate.Image` and `validate.Index` when layer blob content is truncated.

### DIFF
--- a/pkg/v1/layout/write.go
+++ b/pkg/v1/layout/write.go
@@ -256,8 +256,10 @@ func (l Path) writeBlob(hash v1.Hash, size int64, r io.Reader, renamer func() (v
 	defer w.Close()
 
 	// Write to file and exit if not renaming
-	if _, err := io.Copy(w, r); err != nil || renamer == nil {
+	if n, err := io.Copy(w, r); err != nil || renamer == nil {
 		return err
+	} else if size != -1 && n != size {
+		return fmt.Errorf("expected blob size %d, but only wrote %d", size, n)
 	}
 
 	// Always close file before renaming
--- a/pkg/v1/layout/write.go
+++ b/pkg/v1/layout/write.go
@@ -229,8 +229,7 @@ func (l Path) WriteBlob(hash v1.Hash, r io.ReadCloser) error {
 
 func (l Path) writeBlob(hash v1.Hash, size int64, r io.Reader, renamer func() (v1.Hash, error)) error {
 	if hash.Hex == "" && renamer == nil {
-		// Should we panic, since this is a programming error?
-		return errors.New("writeBlob called an invalid hash and no renamer")
+		panic("writeBlob called an invalid hash and no renamer")
 	}
 
 	dir := l.path("blobs", hash.Algorithm)
--- a/pkg/v1/layout/write.go
+++ b/pkg/v1/layout/write.go
@@ -16,9 +16,9 @@ package layout
 
 import (
 	"bytes"
-	"crypto/rand"
-	"encoding/hex"
 	"encoding/json"
+	"errors"
+	"fmt"
 	"io"
 	"io/ioutil"
 	"os"
@@ -28,6 +28,7 @@ import (
 	"github.com/google/go-containerregistry/pkg/v1/match"
 	"github.com/google/go-containerregistry/pkg/v1/mutate"
 	"github.com/google/go-containerregistry/pkg/v1/partial"
+	"github.com/google/go-containerregistry/pkg/v1/stream"
 	"github.com/google/go-containerregistry/pkg/v1/types"
 	"golang.org/x/sync/errgroup"
 )
@@ -223,10 +224,15 @@ func (l Path) WriteFile(name string, data []byte, perm os.FileMode) error {
 // WriteBlob copies a file to the blobs/ directory in the Path from the given ReadCloser at
 // blobs/{hash.Algorithm}/{hash.Hex}.
 func (l Path) WriteBlob(hash v1.Hash, r io.ReadCloser) error {
-	return l.writeBlob(hash, 0, r, nil)
+	return l.writeBlob(hash, -1, r, nil)
 }
 
-func (l Path) writeBlob(hash v1.Hash, size int64, r io.ReadCloser, renamer func() (v1.Hash, error)) error {
+func (l Path) writeBlob(hash v1.Hash, size int64, r io.Reader, renamer func() (v1.Hash, error)) error {
+	if hash.Hex == "" && renamer == nil {
+		// Should we panic, since this is a programming error?
+		return errors.New("writeBlob called an invalid hash and no renamer")
+	}
+
 	dir := l.path("blobs", hash.Algorithm)
 	if err := os.MkdirAll(dir, os.ModePerm); err != nil && !os.IsExist(err) {
 		return err
@@ -234,72 +240,83 @@ func (l Path) writeBlob(hash v1.Hash, size int64, r io.ReadCloser, renamer func(
 
 	// Check if blob already exists and is the correct size
 	file := filepath.Join(dir, hash.Hex)
-	if s, err := os.Stat(file); err == nil && (s.Size() == size || size == 0) {
+	if s, err := os.Stat(file); err == nil && !s.IsDir() && (s.Size() == size || size == -1) {
 		return nil
 	}
 
-	// If a rename func was provided, use a temporary file suffix
+	// If a renamer func was provided write to a temporary file
+	open := func() (*os.File, error) { return os.Create(file) }
 	if renamer != nil {
-		file += ".tmp"
+		open = func() (*os.File, error) { return ioutil.TempFile(dir, hash.Hex) }
 	}
-	w, err := os.Create(file)
+	w, err := open()
 	if err != nil {
 		return err
 	}
 	defer w.Close()
 
-	// Write to file and optionally rename
+	// Write to file and exit if not renaming
 	if _, err := io.Copy(w, r); err != nil || renamer == nil {
 		return err
 	}
-	finalHash, err := renamer()
-	if err != nil {
-		return err
-	}
+
 	// Always close file before renaming
 	if err := w.Close(); err != nil {
 		return err
 	}
-	return os.Rename(file, l.path("blobs", finalHash.Algorithm, finalHash.Hex))
+
+	// Rename file based on the final hash
+	finalHash, err := renamer()
+	if err != nil {
+		return fmt.Errorf("error getting final digest of layer: %w", err)
+	}
+
+	renamePath := l.path("blobs", finalHash.Algorithm, finalHash.Hex)
+	return os.Rename(w.Name(), renamePath)
 }
 
-// WriteLayer writes the compressed layer to a blob. Unlike WriteBlob it will
+// writeLayer writes the compressed layer to a blob. Unlike WriteBlob it will
 // write to a temporary file (suffixed with .tmp) within the layout until the
 // compressed reader is fully consumed and written to disk. Also unlike
 // WriteBlob, it will not skip writing and exit without error when a blob file
 // exists, but does not have the correct size. (The blob hash is not
 // considered, because it may be expensive to compute.)
-func (l Path) WriteLayer(layer v1.Layer) error {
+func (l Path) writeLayer(layer v1.Layer) error {
 	d, err := layer.Digest()
-	if err != nil {
+	if errors.Is(err, stream.ErrNotComputed) {
 		// Allow digest errors, since streams may not have calculated the hash
-		// yet. Instead, use a random value for the digest and require it to be
+		// yet. Instead, use an empty value, which will be transformed into a
+		// random file name with `ioutil.TempFile` and the final digest will be
 		// calculated after writing to a temp file and before renaming to the
 		// final path.
-		var randHash [32]byte
-		if _, err := rand.Read(randHash[:]); err != nil {
-			return err
-		}
-		d = v1.Hash{Algorithm: "sha256", Hex: hex.EncodeToString(randHash[:])}
+		d = v1.Hash{Algorithm: "sha256", Hex: ""}
+	} else if err != nil {
+		return err
 	}
 
 	s, err := layer.Size()
-	if err != nil {
+	if errors.Is(err, stream.ErrNotComputed) {
 		// Allow size errors, since streams may not have calculated the size
 		// yet. Instead, use zero as a sentinel value meaning that no size
 		// comparison can be done and any sized blob file should be considered
 		// valid and not overwritten.
 		//
 		// TODO: Provide an option to always overwrite blobs.
-		s = 0
+		s = -1
+	} else if err != nil {
+		return err
 	}
 
 	r, err := layer.Compressed()
 	if err != nil {
 		return err
 	}
+	defer r.Close()
 
-	return l.writeBlob(d, s, r, layer.Digest)
+	if err := l.writeBlob(d, s, r, layer.Digest); err != nil {
+		return fmt.Errorf("error writing layer: %w", err)
+	}
+	return nil
 }
 
 // RemoveBlob removes a file from the blobs directory in the Path
@@ -332,7 +349,7 @@ func (l Path) WriteImage(img v1.Image) error {
 	for _, layer := range layers {
 		layer := layer
 		g.Go(func() error {
-			return l.WriteLayer(layer)
+			return l.writeLayer(layer)
 		})
 	}
 	if err := g.Wait(); err != nil {
--- a/pkg/v1/layout/write.go
+++ b/pkg/v1/layout/write.go
@@ -311,7 +311,6 @@ func (l Path) writeLayer(layer v1.Layer) error {
 	if err != nil {
 		return err
 	}
-	defer r.Close()
 
 	if err := l.writeBlob(d, s, r, layer.Digest); err != nil {
 		return fmt.Errorf("error writing layer: %w", err)
--- a/pkg/v1/layout/write.go
+++ b/pkg/v1/layout/write.go
@@ -24,6 +24,7 @@ import (
 	"os"
 	"path/filepath"
 
+	"github.com/google/go-containerregistry/pkg/logs"
 	v1 "github.com/google/go-containerregistry/pkg/v1"
 	"github.com/google/go-containerregistry/pkg/v1/match"
 	"github.com/google/go-containerregistry/pkg/v1/mutate"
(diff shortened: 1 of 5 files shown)


jonjohnsonjr:
Thanks @ben-krieger are you able to sign the CLA?

ben-krieger:
> Thanks @ben-krieger are you able to sign the CLA?

@jonjohnsonjr I'm working on the process with my employer (Intel) now, but it's a holiday, so it might take a day or two. In the meantime, I'll fix the failing test and try to increase code coverage of tests.

pkg/v1/validate/image.go
```diff
@@ -113,12 +114,25 @@ func validateLayers(img v1.Image, opt ...Option) error {
 		return layersExist(layers)
 	}
 
+	cf, err := img.ConfigFile()
```
jonjohnsonjr, line 117: I believe these calls were intentionally after processing layers to make this work with streaming layers. We could call `img.Manifest()` early only if we hit `io.ErrUnexpectedEOF` to generate a better error?

pkg/v1/layout/write.go
```diff
@@ -221,43 +223,83 @@ func (l Path) WriteFile(name string, data []byte, perm os.FileMode) error {
...
 
+	// Check if blob already exists and is the correct size
 	file := filepath.Join(dir, hash.Hex)
-	if _, err := os.Stat(file); err == nil {
-		// Blob already exists, that's fine.
+	if s, err := os.Stat(file); err == nil && (s.Size() == size || size == 0) {
 		return nil
 	}
+
+	// If a rename func was provided, use a temporary file suffix
```
jonjohnsonjr, line 241: Is it possible to use io.TempFile for this?

ben-krieger: Sure, I have no preference.

The reason I had it using the digest (when available) plus a suffix was because it would be easy to watch a directory to see the progress of a large layer downloading - something I've commonly done.

With that in mind, would you like me to use the digest (when available) as part of the pattern - the second param to `ioutil.TempFile`?

jonjohnsonjr: > The reason I had it using the digest (when available) plus a suffix was because it would be easy to watch a directory to see the progress of a large layer downloading - something I've commonly done.

I don't think I have a problem with pointing `TempFile` at the same directory (or a sibling/child directory) if that's useful to you.

> With that in mind, would you like me to use the digest (when available) as part of the pattern - the second param to ioutil.TempFile?

Yeah this is a great idea. We might find that doing some synchronization would be useful here as well (to avoid multiple writeLayer calls downloading the same blob)? We can just TODO that and follow up.

jonjohnsonjr:
> However, I have no need for calling WriteLayer as library code, so I'm not opposed to making it unexported, as long as ReplaceImage makes use of its behaviors of overwriting undersized layers and/or writing blobs to a temporary location and renaming.

Let's keep this unexported, but I'm not opposed to exporting it if/when you need to call it directly.

pkg/v1/validate/image.go
ben-krieger: I don't think that calling `img.ConfigFile()` earlier will result in different behavior or errors, because the result of `img.Layers()`, which was previously called, should always be based on a successfully read/parsed config file for the image.

However, I'm not sure if there's a similar guarantee for the result of `img.Manifest()`, so your suggestion makes sense.

pkg/v1/layout/write.go
ben-krieger: I don't totally follow the synchronization comment. Are there any functions in the layout package that you think should be safe for concurrent use? Please suggest any TODOs and I'll add them.

pkg/v1/layout/write.go
```diff
@@ -221,43 +221,83 @@ func (l Path) WriteFile(name string, data []byte, perm os.FileMode) error {
...
+
+	s, err := layer.Size()
+	if err != nil {
+		// Allow size errors, since streams may not have calculated the size
+		// yet. Instead, use zero as a sentinel value meaning that no size
+		// comparison can be done and any sized blob file should be considered
+		// valid and not overwritten.
+		//
+		// TODO: Provide an option to always overwrite blobs.
+		s = 0
```
jonjohnsonjr, line 292: We've used `-1` as a sentinel value in similar situations: https://github.com/google/go-containerregistry/blob/7adcadef6bc8adf07cc587a5a5a443afe7603d54/internal/verify/verify.go#L32

A zero-sized blob seems maybe reasonable, but a `-1` sized blob doesn't make much sense, so I think it's a better sentinel value.

pkg/v1/stream/layer.go
```diff
@@ -159,13 +159,17 @@ func newCompressedReader(l *Layer) (*compressedReader, error) {
...
-		closer: newMultiCloser(zw, l.blob),
-		pr:     pr,
-		bw:     bw,
-		h:      h,
-		zh:     zh,
-		count:  count,
-		l:      l,
+		// NOTE: Order matters! If zw is closed first, then it will panic,
+		// because io.Copy in the goroutine below will still be copying the
+		// contents of l.blob into it.
```
ben-krieger, line 164: I found this out when calling `Close` on a `stream.Layer` that had not been fully read. I can write a test, but it's getting a bit out of scope for this PR.

And this stops a panic from occurring. It was still confusing to see a `stream.ErrNotConsumed` after calling `Close()` even though it technically makes sense.

jonjohnsonjr: I'm really surprised nobody has caught this yet. Where do we panic with the other order?

ben-krieger: Try this. Revert the change:

```diff
diff --git a/pkg/v1/stream/layer.go b/pkg/v1/stream/layer.go
index 9f42f630..3a5993f7 100644
--- a/pkg/v1/stream/layer.go
+++ b/pkg/v1/stream/layer.go
@@ -162,7 +162,8 @@ func newCompressedReader(l *Layer) (*compressedReader, error) {
 		// NOTE: Order matters! If zw is closed first, then it will panic,
 		// because io.Copy in the goroutine below will still be copying the
 		// contents of l.blob into it.
-		closer: newMultiCloser(l.blob, zw),
+		closer: newMultiCloser(zw, l.blob),
+		// closer: newMultiCloser(l.blob, zw),
 
 		pr:    pr,
 		bw:    bw,
```

Add this test:

```go
func TestCloseBeforeConsume(t *testing.T) {
	// Write small tar to pipe
	pr, pw := io.Pipe()
	tw := tar.NewWriter(pw)
	go func() {
		pw.CloseWithError(func() error {
			body := "test file"
			if err := tw.WriteHeader(&tar.Header{
				Name:     "test.txt",
				Mode:     0600,
				Size:     int64(len(body)),
				Typeflag: tar.TypeReg,
			}); err != nil {
				return err
			}
			if _, err := tw.Write([]byte(body)); err != nil {
				return err
			}
			return tw.Close()
		}())
	}()

	// Create stream layer from tar pipe
	l := NewLayer(pr)
	rc, err := l.Compressed()
	if err != nil {
		t.Fatalf("Compressed: %v", err)
	}

	// Close stream layer before consuming
	if err := rc.Close(); err != nil {
		t.Errorf("Close: %v", err)
	}
}
```

If you run it, you'll see that it panics maybe 15-30% of the time. Below, it panicked on the third run.

```
$ go test -count=20 -run CloseBeforeConsume ./pkg/v1/stream/
--- FAIL: TestCloseBeforeConsume (0.00s)
    layer_test.go:257: Close: io: read/write on closed pipe
--- FAIL: TestCloseBeforeConsume (0.00s)
    layer_test.go:257: Close: io: read/write on closed pipe
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x4ebeb4]

goroutine 14 [running]:
compress/flate.(*Writer).Write(...)
        /home/bkrieger/.gimme/versions/go1.17.2.linux.amd64/src/compress/flate/deflate.go:712
compress/gzip.(*Writer).Write(0xc0000b6840, {0xc0002c0000, 0x200, 0x8000})
        /home/bkrieger/.gimme/versions/go1.17.2.linux.amd64/src/compress/gzip/gzip.go:196 +0x334
io.(*multiWriter).Write(0x40b54d, {0xc0002c0000, 0x200, 0x8000})
        /home/bkrieger/.gimme/versions/go1.17.2.linux.amd64/src/io/multi.go:60 +0x86
io.copyBuffer({0x5d1b00, 0xc0002be000}, {0x5d1a60, 0xc0000100c0}, {0x0, 0x0, 0x0})
        /home/bkrieger/.gimme/versions/go1.17.2.linux.amd64/src/io/io.go:425 +0x204
io.Copy(...)
        /home/bkrieger/.gimme/versions/go1.17.2.linux.amd64/src/io/io.go:382
github.com/google/go-containerregistry/pkg/v1/stream.newCompressedReader.func1()
        /home/bkrieger/go/src/github.com/google/go-containerregistry/pkg/v1/stream/layer.go:176 +0xe7
created by github.com/google/go-containerregistry/pkg/v1/stream.newCompressedReader
        /home/bkrieger/go/src/github.com/google/go-containerregistry/pkg/v1/stream/layer.go:175 +0x575
FAIL    github.com/google/go-containerregistry/pkg/v1/stream    0.007s
FAIL
```

Running the test with the race detector helps explain a bit further.

jonjohnsonjr: Interesting... this looks like it might be a race in the gzip package where concurrently calling Write and Close isn't safe. Nice find.

jonjohnsonjr: https://go.dev/play/p/H9WwjIFCs9C

jonjohnsonjr: Looks like this is working as intended but I disagree: https://github.com/golang/go/issues/18883

IMO Write or Close should return an error and not panic here. I'll send a PR for compress/gzip.

ben-krieger: Something I was wondering as I explored the code: why does `Compressed()` and similar functions return an `io.ReadCloser` rather than an `io.Reader`? And why do functions like `layout.WriteBlob` take an `io.ReadCloser`? It doesn't seem like the `Close()` method is ever or should ever be called by a user of this library.

jonjohnsonjr: Calling `Close()` allows us to clean up e.g. the goroutines that are gzipping stuff and writing into an `io.PipeWriter` or an open file. Otherwise we'd just leak resources sometimes.

pkg/v1/layout/write.go
```diff
@@ -221,43 +224,99 @@ func (l Path) WriteFile(name string, data []byte, perm os.FileMode) error {
...
 	if err != nil {
 		return err
 	}
 	defer w.Close()
 
-	_, err = io.Copy(w, r)
-	return err
-}
+	// Write to file and exit if not renaming
+	if _, err := io.Copy(w, r); err != nil || renamer == nil {
```
jonjohnsonjr, line 259: We could confirm that `size` matches here if we don't throw away `n` and `size != -1`.

pkg/v1/layout/write.go
```diff
@@ -221,43 +224,99 @@ func (l Path) WriteFile(name string, data []byte, perm os.FileMode) error {
...
-		// Blob already exists, that's fine.
+	if s, err := os.Stat(file); err == nil && !s.IsDir() && (s.Size() == size || size == -1) {
 		return nil
 	}
-	w, err := os.Create(file)
+
+	// If a renamer func was provided write to a temporary file
+	open := func() (*os.File, error) { return os.Create(file) }
+	if renamer != nil {
+		open = func() (*os.File, error) { return ioutil.TempFile(dir, hash.Hex) }
```
jonjohnsonjr, line 250: We probably want to delete this tempfile if copying or renaming fails?

pkg/v1/stream/layer.go
ben-krieger: I see. I guess that's true for tarball and remote layers. It's just stream layers that close themselves automatically once fully read. And `layout.WriteBlob` probably should have just taken a reader since cleanup duties are generally deferred to the opener, but now the API is set.

pkg/v1/stream/layer.go
```diff
@@ -159,19 +157,39 @@ func newCompressedReader(l *Layer) (*compressedReader, error) {
...
-		if _, err := io.Copy(io.MultiWriter(h, zw), l.blob); err != nil {
-			pw.CloseWithError(err)
+		_, copyErr := io.Copy(io.MultiWriter(h, zw), l.blob)
+
+		// TODO: Why does flushing zw before closing cause TestStreamVsBuffer
+		// to fail? Is the test correct? Has the behavior of gzip Writer
+		// changed? The docs claim that Close does a flush, but that doesn't
+		// look like the case in 1.17.2.
+		//
+		// flushErr := zw.Flush()
```
ben-krieger, line 175: @jonjohnsonjr Do you have any idea what's going on here? Flushing before closing shouldn't cause a test to fail.

ben-krieger: I'm not sure what to make of this: https://go.dev/play/p/LNvLhKpO09n

ben-krieger: Ah, it looks like the additional 5 bytes code for... nothing. And they're before the footer. So calling flush before close in this case is wasteful. I'll remove this comment.

pkg/v1/layout/write.go
```diff
@@ -221,34 +224,92 @@ func (l Path) WriteFile(name string, data []byte, perm os.FileMode) error {
 // WriteBlob copies a file to the blobs/ directory in the Path from the given ReadCloser at
 // blobs/{hash.Algorithm}/{hash.Hex}.
 func (l Path) WriteBlob(hash v1.Hash, r io.ReadCloser) error {
+	return l.writeBlob(hash, -1, r, nil)
+}
+
+func (l Path) writeBlob(hash v1.Hash, size int64, r io.Reader, renamer func() (v1.Hash, error)) error {
+	if hash.Hex == "" && renamer == nil {
+		// Should we panic, since this is a programming error?
+		return errors.New("writeBlob called an invalid hash and no renamer")
```
ben-krieger, line 233: The stdlib often takes the panic approach in cases like these, but not many libraries take the same approach. What's the style of this repo, @jonjohnsonjr?

jonjohnsonjr: Panic here seems fine.

jonjohnsonjr:
Thanks for fixing all these races, let me do another pass through everything but generally lgtm

ben-krieger:
> Thanks for fixing all these races, let me do another pass through everything but generally lgtm

My pleasure! I didn't really intend for this PR to grow like it did, but it was a fun excursion from the norm to jump into a new codebase and really get into the weeds of concurrent processing of streams.

jonjohnsonjr (approved):
Thanks!

pkg/v1/layout/write.go
```diff
@@ -221,34 +224,91 @@ func (l Path) WriteFile(name string, data []byte, perm os.FileMode) error {
...
+	if renamer != nil {
+		open = func() (*os.File, error) { return ioutil.TempFile(dir, hash.Hex) }
+	}
+	w, err := open()
 	if err != nil {
 		return err
 	}
+	if renamer != nil {
+		// Delete temp file if an error is encountered before renaming
+		defer os.Remove(w.Name())
```
imjasonh, line 257: Can we log an error if removing this tempfile fails for some reason? I can imagine a scenario where files are repeatedly failing to get deleted until we run out of space, and there's no indication something bad is happening until things go boom.

pkg/v1/layout/write.go
```diff
@@ -221,34 +224,91 @@ func (l Path) WriteFile(name string, data []byte, perm os.FileMode) error {
...
-// before we write it.
+	// Always close file before renaming
+	if err := w.Close(); err != nil {
+		return err
+	}
 
-// TODO: For streaming layers we should write to a tmp file then Rename to the
-// final digest.
+	// Rename file based on the final hash
+	finalHash, err := renamer()
```
imjasonh, line 274: Does this need an `if renamer != nil` ?

pkg/v1/layout/write_test.go
```diff
@@ -463,3 +481,246 @@ func TestRemoveBlob(t *testing.T) {
...
+	img, err = l.Image(imgDigest)
+	if err != nil {
+		t.Fatalf("error loading image after writeLayer for validation; (Path).Image = %v", err)
+	}
+	if err := validate.Image(img); err != nil {
+		t.Fatalf("validate.Image() = %v", err)
+	}
+}
+
+func TestOverwriteWithwriteLayer(t *testing.T) {
```
imjasonh, line 548: typo: `WithWriteLayer`

pkg/v1/layout/write.go
ben-krieger: Adding a warn level log. Good call.

We only rely on the deferred `os.Remove` to cleanup if something fails at or before `os.Rename`, so hopefully that's not too common unless the network connection is being constantly interrupted.

pkg/v1/layout/write.go
ben-krieger: No, because the conditional on line 262 causes a return if renamer is nil: `if n, err := io.Copy(w, r); err != nil || renamer == nil {`

I added a comment above to try to explain the line (`// Write to file and exit if not renaming`), but let me know if you have any suggestions on how to make it more clear.

pkg/v1/layout/write_test.go
ben-krieger: Fixing. `sed` fail.

imjasonh (approved):

imjasonh:
Thanks for this contribution! 👍

PR #1234, opened 2022-01-06
Add pkg/authn/kubernetes

imjasonh:
This takes the Kubernetes client-go parts from k8schain, removes the
forked-upstream credentialprovider bits and rewrites them to include
only exactly what we need.

It also reimplements k8schain in terms of this new K8s helper, and of
wrapped cred helpers using authn.NewKeychainFromHelper.

cc @vdemeester


imjasonh:
hack/presubmit.sh failed in CI:

```
# sigs.k8s.io/json/internal/golang/encoding/json
Error: /home/runner/go/pkg/mod/sigs.k8s.io/json@v0.0.0-20211020170558-c049b76a60c6/internal/golang/encoding/json/encode.go:1249:12: sf.IsExported undefined (type reflect.StructField has no field or method IsExported)
Error: /home/runner/go/pkg/mod/sigs.k8s.io/json@v0.0.0-20211020170558-c049b76a60c6/internal/golang/encoding/json/encode.go:1255:18: sf.IsExported undefined (type reflect.StructField has no field or method IsExported)
~/work/go-containerregistry/go-containerregistry ~/work/go-containerregistry/go-containerregistry
Error: Process completed with exit code 2.
```

This is because [`StructField.IsExported` was added in Go 1.17](https://pkg.go.dev/reflect#StructField.IsExported) -- the new module specifies `go 1.17` which presumably picks up a later compatible `sigs.k8s.io/json` than is supported by the Go version used in CI. I bumped the CI version used to Go 1.17, since it's not that recent.

The pkg/authn/k8schain module is still `go 1.14`, maybe we should bump that to 1.17 too while we're at it. Maybe we should bump everything. I don't think it needs to be done in this PR.

pkg/authn/kubernetes/keychain.go
```diff
@@ -0,0 +1,161 @@
...
+func NewFromPullSecrets(ctx context.Context, secrets []corev1.Secret) (authn.Keychain, error) {
+	m := map[string]authn.AuthConfig{}
+	for _, secret := range secrets {
+		if b, exists := secret.Data[corev1.DockerConfigJsonKey]; secret.Type == corev1.SecretTypeDockerConfigJson && exists && len(b) > 0 {
+			var cfg config
+			if err := json.Unmarshal(b, &cfg); err != nil {
+				return nil, err
+			}
+			for k, v := range cfg.Auths {
+				m[k] = v
```
imjasonh, line 115: This (and line 119 below) means that later-specified secrets with matching repo/registry keys will overwrite previous ones. `New` fetches explicitly listed `imagePullSecrets` before searching for those attached to the SA, so those later ones, if there's a conflict, will win.

We have some options:
- don't overwrite -- first matching auth wins
- fetch implicit SA secrets before explicitly listed secrets

In any case, we probably need to spell it out better and guard it with tests.

pkg/authn/k8schain/k8schain.go
```diff
@@ -42,60 +46,35 @@ type Options struct {
...
-var (
-	keyring credentialprovider.DockerKeyring
-	once    sync.Once
-)
-
 // New returns a new authn.Keychain suitable for resolving image references as
 // scoped by the provided Options.  It speaks to Kubernetes through the provided
 // client interface.
+//
+// Deprecated: Use pkg/authn/kubernetes.
```
mattmoor, line 53: It is probably worth separating the conversation about Deprecation from this PR, because I don't see the value in forcing all of the folks that are using `k8schain` to drop it so that we can stop maintaining what is now a very small shim.  We can always add these later if we can agree on a path forward for folks.

To me the main win here is cutting out the need for Vincent's fork, which is HUGE, and I totally appreciate all of the work (and arguing!) to get us to this point.  When the world stops ending, I owe you some beverages 🍻

pkg/authn/kubernetes/go.mod
```diff
@@ -0,0 +1,47 @@
+module github.com/google/go-containerregistry/pkg/authn/kubernetes
```
jonjohnsonjr, line 1: Can this just live in k8schain rather than having yet another submodule?

mattmoor: I suspect (@imjasonh can correct) this was in an effort to deprecate `k8schain`, which I think warrants more discussion.  So if we push out deprecation for another convo, this SGTM.

pkg/authn/k8schain/k8schain.go
imjasonh: Yeah, you're probably right. Done.

pkg/authn/kubernetes/go.mod
imjasonh: I think it's useful to be able to have a K8s-secret-backed Keychain that doesn't also pull in all of k8schain's dependencies, even though they're much smaller and more well behaved now.

pkg/authn/kubernetes/keychain.go
mattmoor: Just to chime in here (I meant to before).  If this matches K8s' behavior, then it SGTM.  It's unfortunate we don't track that anymore, but it felt like that was only a matter of time anyways.

pkg/authn/k8schain/go.mod
```diff
@@ -3,11 +3,16 @@ module github.com/google/go-containerregistry/pkg/authn/k8schain
 go 1.14
 
 require (
-	github.com/google/go-containerregistry v0.5.2-0.20210609162550-f0ce2270b3b4
-	github.com/vdemeester/k8s-pkg-credentialprovider v1.21.0-1
```
vdemeester, line 7: ❤️

vdemeester (commented):
@imjasonh just one question, because we remove the files that have `//go:build !disable_gcp`, are we sure the behavior of the new thing we use instead (some part of `go-containerregistry`) doesn't do what upstream k8s did (querying a url infinitely if it came to not answering) ? 
I am mainly asking to not re-open a bug in tekton (and knative) on openshift 😝. Otherwise looks good and simple 😻

imjasonh:
> @imjasonh just one question, because we remove the files that have `//go:build !disable_gcp`, are we sure the behavior of the new thing we use instead (some part of `go-containerregistry`) doesn't do what upstream k8s did (querying a url infinitely if it came to not answering) ? I am mainly asking to not re-open a bug in tekton (and knative) on openshift 😝. Otherwise looks good and simple 😻

Good question.

First, based on my testing with the AWS keychain imported but not used, it doesn't exhibit the 10-minute mystery wait that was observed in Tekton (https://github.com/tektoncd/pipeline/issues/4087). This isn't a guaranteed fix, because to be fair I never saw the mystery wait anyway, even before Tekton snipped it out. Maybe I wasn't testing the right way? 🤷 

Secondly, a nice thing about `k8schain` effectively becoming a shorthand wrapper for N different keychains is that consumers like Tekton/Knative can just choose their own keychains, to remove some, or reorder them, etc. So in this world, _if_ it's a concern, Tekton/Knative or OpenShift Pipelines/Serverless, could replace `k8schain.NewNoClient()` with the equivalent `authn.NewMultiKeychain(default, k8s, google)` and remove problematic/unnecessary keychains.

Tekton could even jump through some hoops and retain the build tag logic, and have AWS disabled selectively at compile-time. We can discuss that in the PR where we bring this change into Tekton.

vdemeester:
> Secondly, a nice thing about `k8schain` effectively becoming a shorthand wrapper for N different keychains is that consumers like Tekton/Knative can just choose their own keychains, to remove some, or reorder them, etc. So in this world, _if_ it's a concern, Tekton/Knative or OpenShift Pipelines/Serverless, could replace `k8schain.NewNoClient()` with the equivalent `authn.NewMultiKeychain(default, k8s, google)` and remove problematic/unnecessary keychains.

Ah that make complete sense 😬 👍🏼

pkg/authn/kubernetes/keychain.go
imjasonh: Well, it turns out the previous behavior was first-one-wins, which I verified with a test on `main` that I ported to this PR.

mattmoor:
Another alternative would be to simply `go mod -replace` `k8schain` downstream with a shim that includes the helpers you want.  Personally, I have been toying with the idea of doing precisely that (for other reasons), so inlining this shim downstream is also (in some sense) a breaking change 😅

mattmoor (approved):

pkg/authn/k8schain/go.mod
```diff
@@ -3,11 +3,16 @@ module github.com/google/go-containerregistry/pkg/authn/k8schain
...
-	k8s.io/api v0.21.1
-	k8s.io/apimachinery v0.21.1
-	k8s.io/client-go v0.21.1
+	github.com/awslabs/amazon-ecr-credential-helper/ecr-login v0.0.0-20211027214941-f15886b5ccdc
+	github.com/chrismellard/docker-credential-acr-env v0.0.0-20210203204924-09e2b5a8ac86
+	github.com/google/go-containerregistry v0.8.0
+	github.com/google/go-containerregistry/pkg/authn/kubernetes v0.0.0-00010101000000-000000000000
+	k8s.io/api v0.23.1
```
dprotaso, line 10: care to pin this to 0.22.5?

pkg/authn/kubernetes/go.mod
```diff
@@ -0,0 +1,47 @@
...
+
+go 1.17
+
+require (
+	github.com/google/go-containerregistry v0.8.0
+	k8s.io/api v0.23.1
+	k8s.io/apimachinery v0.23.1
+	k8s.io/client-go v0.23.1
```
dprotaso, line 9: likewise care to pin to 0.22.5?

jonjohnsonjr (approved):

hasheddan:
@imjasonh just a heads up, we have previously had issues with the hanging requests in @crossplane and recently had them resurface when folks are using proxies, so definitely happy to see this land. I have some community members running tests of consuming this implementation across a variety of environments (pull through caches, [pulling using IRSA](https://docs.aws.amazon.com/eks/latest/userguide/iam-roles-for-service-accounts.html), etc.) in https://github.com/crossplane/crossplane/pull/2841. Happy to exercise any other scenarios of interest and report back as helpful. Appreciate the efforts here :)

imjasonh:
@hasheddan that's great to hear! More testers on this will definitely be helpful. ❤️

PR #1244, opened 2022-01-14
return transport errors that support errors.Is

dprotaso:
Previously timed out requests could be checked using `errors.Is(err, context. DeadlineExceeded)` 

That changed when multiple errors were returned in an earlier PR. This change introduces Uber's multi-err library that coalesces errors and supports `error.Is`


pkg/v1/remote/transport/ping.go
```diff
@@ -125,7 +125,7 @@ func ping(ctx context.Context, reg name.Registry, t http.RoundTripper) (*pingRes
 			return nil, CheckError(resp, http.StatusOK, http.StatusUnauthorized)
 		}
 	}
-	return nil, errors.New(strings.Join(errs, "; "))
```
imjasonh, line 128: Could we do something like `fmt.Errorf("%s; %w", strings.Join(errs[:len(errs)], ","), errs[len(errs)]` so that `errors.Is` is respected (by the last error), but without taking a new dependency on `multierr`?

dprotaso: Sure

dprotaso: took a hybrid approach

imjasonh (approved):
Nice.

PR #1247, opened 2022-01-19
Bump acr cred helper dep

imjasonh:
This picks up https://github.com/chrismellard/docker-credential-acr-env/pull/2 to fix hanging behavior when checking to see whether the ACR cred helper could be used to auth against a registry.

```
cd pkg/authn/k8schain
go get -u github.com/chrismellard/docker-credential-acr-env@master
go mod tidy && go mod download
cd ../../../cmd/krane
go get -u github.com/google/go-containerregistry/pkg/authn/k8schain@main
go mod tidy && go mod download
cd ../../
./hack/presubmit.sh # passes
```


dprotaso:
/lgtm
cc @jonjohnsonjr @mattmoor

cmd/krane/go.mod
```diff
@@ -10,49 +10,49 @@ replace (
 
 require (
 	github.com/google/go-containerregistry v0.8.1-0.20220110151055-a61fd0a8e2bb
-	github.com/google/go-containerregistry/pkg/authn/k8schain v0.0.0-20211223213658-2874338840a6
+	github.com/google/go-containerregistry/pkg/authn/k8schain v0.0.0-20220114205711-890d5b362eb8
```
dprotaso, line 13: I believe you'll need to update this in a separate PR

dprotaso: Actually - probably not worth doing this here until k8s chain merges

imjasonh: I'm not sure we need to do this in another PR, since it's just `replace`d above. If downstream deps who don't respect the `replace` are broken by this (I think that's what you're describing?), we should figure out a way to capture that in a presubmit to prevent future breakages.

In any case, I can `go get -u` again after this merges. I'm kinda hating all these interdependent modules in this repo 🤦 .

dprotaso: > In any case, I can go get -u again after this merges.

Oh nm - I didn't see the `replace` directive. That's not necessary then. It would be If you want `krane` to be `go install`-able. 

>  I'm kinda hating all these interdependent modules in this repo 🤦 .

Yeah the explicit refs are not fun - I'll open a go issue ;)

hasheddan:
Thank you @imjasonh @dprotaso 🙌🏻

PR #1252, opened 2022-01-22
Add pkg/authn/github.Keychain to authenticate with ghcr.io

### LINKED ISSUE #1251: pkg/authn/github.Keychain
We could add a keychain that uses the value of `$GITHUB_TOKEN`, if available, when asked to authenticate operations against ghcr.io.

This should be a very lightweight package and doesn't need its own go module.

It should be very easy to test this e2e with GitHub Actions.

We can include this in k8schain (increasingly becoming a misnomer) or use its absence as a carrot for consumers of k8schain to switch to just using a MultiKeychain.

imjasonh:
Fixes #1251

### DIFF
--- a/cmd/krane/main.go
+++ b/cmd/krane/main.go
@@ -24,6 +24,7 @@ import (
 	"github.com/google/go-containerregistry/cmd/crane/cmd"
 	"github.com/google/go-containerregistry/internal/signal"
 	"github.com/google/go-containerregistry/pkg/authn"
+	"github.com/google/go-containerregistry/pkg/authn/github"
 	"github.com/google/go-containerregistry/pkg/crane"
 	"github.com/google/go-containerregistry/pkg/logs"
 	"github.com/google/go-containerregistry/pkg/v1/google"
@@ -51,6 +52,7 @@ func main() {
 	keychain := authn.NewMultiKeychain(
 		authn.DefaultKeychain,
 		google.Keychain,
+		github.Keychain,
 		amazonKeychain,
 		azureKeychain,
 	)
--- a/pkg/authn/github/keychain.go
+++ b/pkg/authn/github/keychain.go
@@ -1,0 +1,59 @@
+// Copyright 2022 Google LLC All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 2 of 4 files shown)


pkg/authn/github/keychain.go
```diff
@@ -0,0 +1,49 @@
...
+// Keychain exports an instance of the GitHub Keychain.
+//
+// This keychain matches on requests for ghcr.io and provides the value of the
+// environment variable $GITHUB_TOKEN, if it's set.
+var Keychain authn.Keychain = githubKeychain{}
+
+type githubKeychain struct{}
+
+func (githubKeychain) Resolve(r authn.Resource) (authn.Authenticator, error) {
+	if strings.HasPrefix(r.String(), "ghcr.io") {
```
imjasonh, line 34: Bug here, ghcr.io.mydomain.com would match. Fixing now.

cmd/krane/main.go
```diff
@@ -51,6 +52,7 @@ func main() {
 	keychain := authn.NewMultiKeychain(
 		authn.DefaultKeychain,
 		google.Keychain,
+		github.Keychain,
```
mattmoor, line 55: This was gonna be my feedback.  Now I’m regretting not calling this omicrane after all.

imjasonh: It's only been released once as `krane`, you can change it.

Ooooooor, release it as `github.com/mattmoor/omnicrane` instead... 😈

imjasonh:
@jonjohnsonjr ping

pkg/authn/github/keychain.go
```diff
@@ -0,0 +1,55 @@
...
+	}
+	return authn.Anonymous, nil
+}
+
+type githubAuthenticator struct{ tok string }
+
+func (g githubAuthenticator) Authorization() (*authn.AuthConfig, error) {
+	return &authn.AuthConfig{
+		Username: "unused",
+		Password: g.tok,
```
jonjohnsonjr, line 53: Does this work if you return this as an IdentityToken? It would be nice if we didn't use `unused` as the username, but if they don't support the oauth flow this seems fine.

pkg/authn/github/keychain.go
```diff
@@ -0,0 +1,55 @@
...
+		}
+	}
+	return authn.Anonymous, nil
+}
+
+type githubAuthenticator struct{ tok string }
+
+func (g githubAuthenticator) Authorization() (*authn.AuthConfig, error) {
+	return &authn.AuthConfig{
+		Username: "unused",
```
jonjohnsonjr, line 52: `os.Getenv("GITHUB_TOKEN")` instead?

imjasonh: sgtm, done.

jonjohnsonjr (approved):

PR #1254, opened 2022-01-22
Ensure that layer is closed before renaming file in layout

### LINKED ISSUE #1253: TestStreamingWriteLayer: value not computed until stream is consumed
Saw this flake in CI: https://github.com/google/go-containerregistry/runs/4908636177?check_suite_focus=true#step:4:41

```
--- FAIL: TestStreamingWriteLayer (0.01s)
    write_test.go:531: (Path).AppendImage() = error writing layer: error getting final digest of layer: value not computed until stream is consumed
```

This test was added recently in https://github.com/google/go-containerregistry/pull/1226

@ben-krieger any idea what might cause this to happen? 🤔

ben-krieger:
Fixes #1253

### DIFF
--- a/pkg/v1/layout/write.go
+++ b/pkg/v1/layout/write.go
@@ -228,7 +228,7 @@ func (l Path) WriteBlob(hash v1.Hash, r io.ReadCloser) error {
 	return l.writeBlob(hash, -1, r, nil)
 }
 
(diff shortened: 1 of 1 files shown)


imjasonh (approved):
Nice! Thanks for the quick fix! 🙏

PR #1257, opened 2022-01-23
Add output option to mutate to save to a tar file and not push to a registry (same as append)

ehmm:
Support passing crane mutate an output as a tar. Append has it and it is useful for all kinds of more complex build and push flows


ehmm:
@imjasonh can you please review?

imjasonh (approved):
Seems reasonable to me, curious if @jonjohnsonjr has any feedback.

ehmm:
Pinging @jonjohnsonjr

PR #1260, opened 2022-01-24
Bump deps, add script to make it easier

imjasonh:
Dependabot gets tripped up by our multi-module repo, so instead I wrote a script that should automate it for us in [`hack/bump-deps.sh`](https://github.com/google/go-containerregistry/compare/main...imjasonh:bump-deps?expand=1#diff-fbee68960cb989d4fd5d970791c81de9bc010892c0c3f077c46e215469ae96ea).

If we want, we can hook it up to GitHub Actions to [create PRs for us](https://github.com/marketplace/actions/create-pull-request) and drop dependabot, or we can just use dependabot to notify us when bumps are needed and do it manually, like I've been doing lately. A benefit to this is that it's one big PR instead of individual ones.

lmk what you think.

This PR gets us past a [containerd dependabot alert](https://github.com/google/go-containerregistry/security/dependabot/go.sum/github.com%2Fcontainerd%2Fcontainerd/open) that doesn't affect us, but spams us and downstream users with the alert.


pkg/authn/kubernetes/go.mod
```diff
@@ -4,9 +4,9 @@ go 1.17
 
 require (
 	github.com/google/go-containerregistry v0.8.0
-	k8s.io/api v0.22.5
-	k8s.io/apimachinery v0.22.5
-	k8s.io/client-go v0.22.5
+	k8s.io/api v0.23.2
+	k8s.io/apimachinery v0.23.2
+	k8s.io/client-go v0.23.2
```
mattmoor, line 9: Didn't we explicitly leave these at 0.22?

cc @dprotaso

justaugustus: > Didn't we explicitly leave these at 0.22?

@mattmoor -- There was a request to use 0.22.5, but there wasn't a reason provided:
- https://github.com/google/go-containerregistry/pull/1234#discussion_r780449706
- https://github.com/google/go-containerregistry/pull/1234#discussion_r780449910

Would suggest using 0.23.**3**, unless there's a good reason not to.

mattmoor: tl;dr It's because that's what Knative and Tekton are pinned to.  I'll defer to @dprotaso to explain the cadence.

imjasonh: Couldn't Knative and Tekton solve this with a `replace k8s.io/foo => k8s.io/foo v0.22.5`?

justaugustus: They could, based on my understanding.
I think it's important that the upstream take input from downstreams, but set the pace by keeping deps as up-to-date as possible.

imjasonh:
This is effectively blocking automatic security updates, like https://github.com/google/go-containerregistry/pull/1287, since dependabot can't manage our deps without help.

PR #1261, opened 2022-01-24
Support partial platform matching, implement platform.String and PlatformFromString

imjasonh:


imjasonh:
Decision we came to was to sequester this logic inside ko for now since it already does partial matching, just not partial enough.

I'll split out `Platform.String` and `v1.ParsePlatform` (a much better name) into a separate PR since we think that's useful.

PR #1256, opened 2022-01-22
Fix Azure Container Registry authentication

simongottschlag:
Fixes #1255

I'm not sure if this is the best or preferred way of handling this, another way may be to add a check to `NewKeychainFromHelper()` that checks if the username is `<token>` and adds `IdentityToken` to the `AuthConfig{}`.

Any and all feedback is welcome and if you see that I've missed something obvious and that's why I've done this, please point it out and I'll fix the issue where it's located.

PR #1265, opened 2022-01-25
Fix NewKeychainFromHelper

imjasonh:
- only pass RegistryStr, since helpers don't expect to see repo paths,
  and might not handle them correctly
- if the returned username is "<token>", interpret the returned password
  as an identity token
- add unit tests to cover both of these behaviors

This builds on the excellent work done by @simongottschlag in https://github.com/google/go-containerregistry/pull/1256, and adds unit tests.

### DIFF
--- a/pkg/authn/keychain.go
+++ b/pkg/authn/keychain.go
@@ -154,9 +154,14 @@ func NewKeychainFromHelper(h Helper) Keychain { return wrapper{h} }
 type wrapper struct{ h Helper }
 
 func (w wrapper) Resolve(r Resource) (Authenticator, error) {
-	u, p, err := w.h.Get(r.String())
+	u, p, err := w.h.Get(r.RegistryStr())
 	if err != nil {
 		return Anonymous, nil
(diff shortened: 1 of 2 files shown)


pkg/authn/keychain.go
```diff
@@ -154,9 +154,14 @@ func NewKeychainFromHelper(h Helper) Keychain { return wrapper{h} }
 type wrapper struct{ h Helper }
 
 func (w wrapper) Resolve(r Resource) (Authenticator, error) {
-	u, p, err := w.h.Get(r.String())
+	u, p, err := w.h.Get(r.RegistryStr())
 	if err != nil {
 		return Anonymous, nil
 	}
+	// If the secret being stored is an identity token, the Username should be set to <token>
+	// ref: https://docs.docker.com/engine/reference/commandline/login/#credential-helper-protocol
```
mattmoor, line 162: Thanks for the link!

mattmoor (approved):

PR #1267, opened 2022-01-26
Update k8schain README -- don't recommend modifting authn.DefaultKeyc…

imjasonh:
…hain


imjasonh:
This questionable advice was discovered via https://github.com/kyverno/kyverno/pull/3088 which resulted in ever-growing MultiKeychains, and memory leaks.

If this becomes a problem we might consider deprecating the mutable `authn.DefaultKeychain` and instead preferring a public method that returns an immutable `Keychain` impl.

mattmoor (approved):

PR #1270, opened 2022-01-27
Implement Platform.String and v1.ParsePlatform

imjasonh:


pkg/v1/platform.go
```diff
@@ -28,11 +30,89 @@ type Platform struct {
...
+	if len(p.OSFeatures) != 0 {
+		b.WriteString(" (osfeatures=")
+		b.WriteString(strings.Join(p.OSFeatures, ","))
+		b.WriteString(")")
+	}
+	if len(p.Features) != 0 {
+		b.WriteString(" (features=")
+		b.WriteString(strings.Join(p.Features, ","))
+		b.WriteString(")")
+	}
```
jonjohnsonjr, line 60: I don't love this part but I love the rest.

imjasonh: I thought you might feel that way. I can just drop that, but then roundtripping platforms that include these will become lossy. Almost no platforms in the wild will include these fields, so this codepath will basically never be exercised. I don't know whether that's an argument to allow this grossness to go in, or if it's an argument to just have lossy roundtrips in that case.

The rest of this will be really valuable, so I don't want to hang up too long on it. I'd rather just delete it. But I'll do whatever you think is right.

jonjohnsonjr: I'm upset that my only objection is the syntax but because this is going to become a public interface, I think it actually  matters enough that I'd prefer lossy roundtrips (for now) :/

pkg/v1/platform_test.go
```diff
@@ -21,30 +21,128 @@ import (
...
+		"linux/amd64:1.2.3.4",
+	}} {
+		if got := c.plat.String(); got != c.want {
+			t.Errorf("got %q, want %q", got, c.want)
+		}
+
+		if len(c.plat.OSFeatures) > 0 || len(c.plat.Features) > 0 {
+			// If these values are set, roundtripping back to the
+			// Platform will be lossy, and we expect that.
+			return
```
jonjohnsonjr, line 54:
```suggestion
			continue
```

jonjohnsonjr (approved):

Issue #1273, opened 2022-01-28
FR: crane: export to accept tarball on stdin


abitrolly:
### Describe the bug

I couldn't find a way to [extract a single file](https://github.com/google/go-containerregistry/blob/main/cmd/crane/recipes.md#extract-a-single-file-from-an-image) from a local `podman` image, so I ended up here.

The idea is to save local image from podman/docker, which produces tarball with layers, and then pipe it to `crane` to flatten the layers into plain filesystem.
```
podman image save test - | crane export - - | tar -Oxf - etc/passwd
```

This doesn't work, of course, because `crate export` don't work with local tarballs.

### Expected behavior

Ideally.
```
crane cat local://test /etc/passwd
```
### Additional context

- crane 0.8.0

PR #1274, opened 2022-01-28
crane export: Support reading tarball from a stream

abitrolly:
Implementation for #1273.

The goal is to read from stdin if the first argument is "-".

```
./crane export - ubuntu.tar < test.tar
```

~~I was not able to trace what is wrong with reading from `os.Stdin`, and my Go skills are not that good to understand it from the code.~~

### DIFF
--- a/cmd/crane/cmd/export.go
+++ b/cmd/crane/cmd/export.go
@@ -31,14 +31,14 @@ import (
 func NewCmdExport(options *[]crane.Option) *cobra.Command {
 	return &cobra.Command{
 		Use:   "export IMAGE|- TARBALL",
-		Short: "Export contents of a remote image as a filesystem tarball",
+		Short: "Export filesystem of a container image as a tarball",
 		Example: `  # Write tarball to stdout
   crane export ubuntu -
 
   # Write tarball to file
   crane export ubuntu ubuntu.tar
 
-  # Read image tarball from stdin
+  # Read image from stdin
   crane export - ubuntu.tar`,
 		Args: cobra.RangeArgs(1, 2),
 		RunE: func(_ *cobra.Command, args []string) error {
--- a/cmd/crane/cmd/export.go
+++ b/cmd/crane/cmd/export.go
@@ -17,6 +17,8 @@ package cmd
 import (
 	"fmt"
 	"io"
+	"io/ioutil"
+	"log"
 	"os"
 
 	"github.com/google/go-containerregistry/pkg/crane"
@@ -53,10 +55,18 @@ func NewCmdExport(options *[]crane.Option) *cobra.Command {
 
 			var img v1.Image
 			if src == "-" {
-				img, err = tarball.Image(func() (io.ReadCloser, error) {
-					return io.ReadCloser(os.Stdin), nil
-				}, nil)
-				//img, err = tarball.ImageFromPath("test.tar", nil)
+				tmpfile, err := ioutil.TempFile("", "crane")
+				if err != nil {
+					log.Fatal(err)
+				}
+				defer os.Remove(tmpfile.Name())
+
+				if _, err := io.Copy(tmpfile, os.Stdin); err != nil {
+					log.Fatal(err)
+				}
+				tmpfile.Close()
+
+				img, err = tarball.ImageFromPath(tmpfile.Name(), nil)
 				if err != nil {
 					return fmt.Errorf("reading tarball from stdin: %w", err)
 				}
--- a/cmd/crane/cmd/export.go
+++ b/cmd/crane/cmd/export.go
@@ -19,19 +19,24 @@ import (
 	"os"
 
 	"github.com/google/go-containerregistry/pkg/crane"
+	v1 "github.com/google/go-containerregistry/pkg/v1"
+	"github.com/google/go-containerregistry/pkg/v1/tarball"
 	"github.com/spf13/cobra"
 )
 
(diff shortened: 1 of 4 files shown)


cmd/crane/cmd/export.go
```diff
@@ -45,9 +51,20 @@ func NewCmdExport(options *[]crane.Option) *cobra.Command {
...
 
-			img, err := crane.Pull(src, *options...)
-			if err != nil {
-				return fmt.Errorf("pulling %s: %w", src, err)
+			var img v1.Image
+			if src == "-" {
+				img, err = tarball.Image(func() (io.ReadCloser, error) {
+					return io.ReadCloser(os.Stdin), nil
+				}, nil)
+				//img, err = tarball.ImageFromPath("test.tar", nil)
```
jonjohnsonjr, line 59: Drop this?

jonjohnsonjr (commented):
Does this actually work? We need to read through the tarball multiple times (or buffer it) to effectively export the filesystem.

cmd/crane/cmd/export.go
abitrolly: Just a reminder that there is an alternative API that may work if the file is first saved to disk from `os.Stdin`. I will move it to the issue comment instead.

abitrolly:
This doesn't work correctly. Why the `tar` needs to be read multiple times? It is a stream format, so it should be possible to process it in one go, no?

imjasonh:
One reason is we can't assume the ordering of files in the archive. We go through a few times to [a.) find the `manifest.json` in the archive](https://github.com/google/go-containerregistry/blob/d9bfbcb99e526b2a9417160e209b816e1b1fb6bd/pkg/v1/tarball/image.go#L172), maybe [b.) find the descriptor corresponding to the requested tag (or only image)](https://github.com/google/go-containerregistry/blob/d9bfbcb99e526b2a9417160e209b816e1b1fb6bd/pkg/v1/tarball/image.go#L187), then [c.) find the config blob discovered from (b)](https://github.com/google/go-containerregistry/blob/d9bfbcb99e526b2a9417160e209b816e1b1fb6bd/pkg/v1/tarball/image.go#L192).

We can't guarantee the ordering of these files in the archive, so we don't (currently) attempt to do this in a single read. Since the (c) depends on (b), we'd have to rely on archives being well-ordered to be able to read them in one go, and I don't think tools currently enforce that ordering.

Aside from that, we also [peek at a layer blob to tell if it's compressed](https://github.com/google/go-containerregistry/blob/d9bfbcb99e526b2a9417160e209b816e1b1fb6bd/pkg/v1/tarball/image.go#L159).

There might be other places I'm forgetting about where we assume a tar file is rewindable.

It might be possible to refactor all this so we can read a tar archive without repeatedly restarting the read, but we don't today. Doing so might break reading some images that we support today.

cmd/crane/cmd/export.go
abitrolly: Dropped.

abitrolly:
I chose to write `stdin` to temporary file. On Linux it would be equivalent of writing to memory, because `/tmp` is mouted as `tmpfs`. If one pass tar parser is implemented, well-ordered images can save this memory. Even non-ordered images can benefit from that if the `manifest.json` doesn't come at the end of `tar`.

This is probably doable for this PR, but the change surface might get too vast to handle.

imjasonh (commented):
Not blocking, but any idea why this doesn't work?

```
$ ./crane pull ubuntu - | ./crane export - flat.tar
Error: reading tarball from stdin: file manifest.json not found in tar
```

...but this works:

```
./crane pull ubuntu ubuntu.tar
./crane export - - < ubuntu.tar | tar -Oxf - etc/passwd
```

I was hoping to be able to demonstrate this needless monstrosity:

```
$ ./crane pull ubuntu - | ./crane export - - | tar -Oxf - etc/passwd
Error: reading tarball from stdin: file manifest.json not found in tar
tar: etc/passwd: Not found in archive
tar: Error exit delayed from previous errors.
```

pkg/crane/export.go
```diff
@@ -29,6 +30,7 @@ func Export(img v1.Image, w io.Writer) error {
 	if err != nil {
 		return err
 	}
+	logs.Debug.Printf("layers found: %v", len(layers))
```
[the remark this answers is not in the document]
abitrolly: Why? `--verbose` mode is useful to understand what is going on without recompiling all the stuff. For example, the right  error message could give you the answer to the error below in verbosity mode. If the input image to `export` was empty, or it was wrong image.

```
$ ./crane pull ubuntu - | ./crane export - flat.tar
Error: reading tarball from stdin: file manifest.json not found in tar
```

I suspect adding `-v` will show that the image is empty, but I haven't checked.
```
$ ./crane pull ubuntu - | ./crane -v export - flat.tar
layers found: 0
Error: reading tarball from stdin: file manifest.json not found in tar
```

abitrolly:
> Not blocking, but any idea why this doesn't work?

Because `pull` doesn't implement `os.Stdout` handling when `-` is supplied.
https://github.com/google/go-containerregistry/blob/main/cmd/crane/cmd/pull.go
It is not automatic.

pkg/crane/export.go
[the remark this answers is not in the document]
abitrolly: Checked. The error message fires before the debug message, so it is useless in this case.

imjasonh (approved):

jonjohnsonjr (approved):

PR #1280, opened 2022-02-08
Check docker config auths for repo and registry

imjasonh:
This enables the case where a config's auth section can specify different auth per repository within the same registry.

This seems to be the root cause of https://github.com/GoogleContainerTools/kaniko/issues/687

### DIFF
--- a/pkg/authn/keychain.go
+++ b/pkg/authn/keychain.go
@@ -114,20 +114,27 @@ func (dk *defaultKeychain) Resolve(target Resource) (Authenticator, error) {
 	// See:
 	// https://github.com/google/ko/issues/90
 	// https://github.com/moby/moby/blob/fc01c2b481097a6057bec3cd1ab2d7b4488c50c4/registry/config.go#L397-L404
-	key := target.RegistryStr()
-	if key == name.DefaultRegistry {
-		key = DefaultAuthKey
-	}
+	var cfg, empty types.AuthConfig
+	for _, key := range []string{
+		target.String(),
+		target.RegistryStr(),
+	} {
+		if key == name.DefaultRegistry {
+			key = DefaultAuthKey
+		}
 
-	cfg, err := cf.GetAuthConfig(key)
-	if err != nil {
-		return nil, err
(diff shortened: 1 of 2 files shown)


jonjohnsonjr:
I'm a bit reluctant to diverge from docker's behavior -- do they support anything like this? Does the red hat stuff?

jonjohnsonjr:
Related FR: https://github.com/docker/cli/issues/2928

imjasonh:
Oh, I incorrectly thought this was already how Docker worked 😢 

I'll close this, and point folks up to that FR. Thanks!

jonjohnsonjr:
> wdyt?

I think ideally we could upstream this to docker/cli?

imjasonh:
> I think ideally we could upstream this to docker/cli?

I'm willing to try. I think we might have a better case if we can say Podman, and all tools using `authn.DefaultKeychain` have this behavior. Want me to merge and update all these tools with this behavior? I know Tekton has expressed interest in supporting this before.

pkg/authn/keychain.go
```diff
@@ -114,20 +114,27 @@ func (dk *defaultKeychain) Resolve(target Resource) (Authenticator, error) {
...
+		if key == name.DefaultRegistry {
+			key = DefaultAuthKey
+		}
 
-	cfg, err := cf.GetAuthConfig(key)
-	if err != nil {
-		return nil, err
+		cfg, err = cf.GetAuthConfig(key)
+		if err != nil {
+			return nil, err
```
jonjohnsonjr, line 128: I'm a little bit worried that we'll hit a new error here but it looks like it currently never returns an error so it's probably fine:

https://github.com/docker/cli/blob/25eee83d6b8c475548254b2decc9c8e0490d229c/cli/config/configfile/file.go#L321
https://github.com/docker/cli/blob/25eee83d6b8c475548254b2decc9c8e0490d229c/cli/config/configfile/file.go#L311
https://github.com/docker/cli/blob/25eee83d6b8c475548254b2decc9c8e0490d229c/cli/config/credentials/file_store.go#L32-L47

PR #1285, opened 2022-02-11
LayerFromReader: buffer contents to a temp file instead of in memory

imjasonh:
See https://twitter.com/ariadneconill/status/1491913976795602947

Also mark the method as `Deprecated`, and reimplement our own only usage of it in `pkg/crane.Layer(filemap)` using `tarball.LayerFromOpener` instead.

### DIFF
--- a/pkg/crane/filemap_test.go
+++ b/pkg/crane/filemap_test.go
@@ -16,6 +16,7 @@ package crane_test
 
 import (
 	"archive/tar"
+	"errors"
 	"io"
 	"io/ioutil"
 	"testing"
@@ -85,7 +86,7 @@ func TestLayer(t *testing.T) {
 			saw := map[string]struct{}{}
 			for {
 				th, err := tr.Next()
-				if err == io.EOF {
+				if errors.Is(err, io.EOF) {
 					break
 				}
 				if err != nil {
--- a/pkg/crane/filemap_test.go
+++ b/pkg/crane/filemap_test.go
@@ -15,6 +15,9 @@
 package crane_test
 
 import (
+	"archive/tar"
+	"io"
+	"io/ioutil"
 	"testing"
 
 	"github.com/google/go-containerregistry/pkg/crane"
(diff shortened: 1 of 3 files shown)


pkg/v1/tarball/layer.go
```diff
@@ -249,15 +250,17 @@ func LayerFromOpener(opener Opener, opts ...LayerOption) (v1.Layer, error) {
 }
 
 // LayerFromReader returns a v1.Layer given a io.Reader.
+//
+// The reader's contents are read and buffered to a temp file in the process.
 func LayerFromReader(reader io.Reader, opts ...LayerOption) (v1.Layer, error) {
-	// Buffering due to Opener requiring multiple calls.
-	a, err := ioutil.ReadAll(reader)
+	tmp, err := ioutil.TempFile("", "")
```
jonjohnsonjr, line 256: Are we okay with leaking this file?

jonjohnsonjr: I feel like this is just a misleading interface and we should make people use `LayerFromFile` or `pkg/v1/stream`.

imjasonh: I'd be totally fine with marking it deprecated, moving it to `internal` and only using it for testing.

We only use it ourself for [`crane.Layer(map[string][]byte)`](https://pkg.go.dev/github.com/google/go-containerregistry/pkg/crane#Layer) (discussion in https://github.com/google/go-containerregistry/pull/485), and there's no real reason that needs to use this.

Should we just deprecate it and recommend against its usage, favoring `LayerFromFile`/`LayerFromOpener` instead?

jonjohnsonjr: Yeah I think that's the easiest thing to do for now. `crane.Layer` is trivially fixable to use FromOpener.

We can still keep this change but also add Deprecated? Better to leak a temp file than buffer, maybe?

jonjohnsonjr (approved):

pkg/crane/filemap_test.go
```diff
@@ -69,7 +72,44 @@ func TestLayer(t *testing.T) {
...
+			rc, err := l.Uncompressed()
+			if err != nil {
+				t.Fatalf("Uncompressed: %v", err)
+			}
+			defer rc.Close()
+			tr := tar.NewReader(rc)
+			saw := map[string]struct{}{}
+			for {
+				th, err := tr.Next()
+				if err == io.EOF {
```
jonjohnsonjr, line 88:
```suggestion
				if errors.Is(err, io.EOF) {
```

Issue #1245, opened 2022-01-18
question: does crane pull handle manifest conversion


thesayyn:
I have been playing around with crane pull OCI format. upon inspection of the OCI contents of the image, I realized that crane does not make image config/manifest OCI compliant.  mediaType is still `application/vnd.docker.distribution.manifest.v2+json`/ `application/vnd.docker.container.image.v1+json`

On the other hand, tools such as skopeo handle the conversion automatically and give you OCI with the original manifest stripped down that conforms to OCI spec. 

Perhaps this is the intended behavior, though I want to better understand why would one want to have docker manifests laid out as OCI layout.

If this is not the intended behavior, is there intent to handle this conversion? perhaps I could help.

thesayyn:
this quirk is only present when the source image is in docker v2 format.

jonjohnsonjr:
> Perhaps this is the intended behavior, 

This is intended behavior, as we want to be able to roundtrip images through the OCI Layout. For example, you should be able to `crane pull` an image into a layout, then `crane push` that image to another registry, and the digests should be identical. If we did conversion between OCI and Docker media types, we'd lose the original payload, and the digests would no longer match.

> though I want to better understand why would one want to have docker manifests laid out as OCI layout.

It's a generally useful mechanism for storing content-addressable things on disk. If you look at [the example layout](https://github.com/opencontainers/image-spec/blob/main/image-layout.md#index-example), there's even an [appstream](https://www.freedesktop.org/wiki/Distributions/AppStream/) artifact referenced.

What's your use case for converting to the OCI spec? I could imagine that being something we support, but maybe it would live in [`crane mutate`](https://github.com/google/go-containerregistry/blob/main/cmd/crane/doc/crane_mutate.md)?

thesayyn:
@jonjohnsonjr sorry for pinging. I have been working on a patch to add `--manifest_type` to force conversion when specified. I was able to do it using `mutate.MediaType`,  `mutate.ConfigMediaType`, `mutate.IndexMediaType` and introducing `mutate.Config` and a conversion func (similar to CreatedAt and Canonical) to drop properties that are only present in DockerV2 images.

Would you be interested to see the patch? with this patch, we can use crane to pull-push images with conversion.

jonjohnsonjr:
[...]

PR #1293, opened 2022-02-16
feat: implement oci conversion

thesayyn:
Implements #1245. cc @jonjohnsonjr 

Open to any kind of suggestions. I did my best to keep the patch as small as possible.

I will add tests after i address the changes.

the patch in action: [thesayyn/oci-image:latest-crane](https://hub.docker.com/layers/thesayyn/oci-image/latest-crane/images/sha256-67ad5ad685aaeaa4402d3568f82d00e61cff5ba5f76bc63acea6b151871718c9?context=explore) pulled debian and pushed via crane.


cmd/crane/cmd/pull.go
```diff
@@ -92,7 +105,7 @@ func NewCmdPull(options *[]crane.Option) *cobra.Command {
 				}
 				for ref, idx := range indexMap {
 					anns := map[string]string{
-						"dev.ggcr.image.name": ref,
+						"org.opencontainers.image.ref.name": ref,
```
thesayyn, line 108: This is the expected annotation standardized by image-spec. hence I believe we should follow that. 
also, the other tools follow this convention as well.

cmd/crane/cmd/pull.go
```diff
@@ -106,6 +119,7 @@ func NewCmdPull(options *[]crane.Option) *cobra.Command {
 	}
 	cmd.Flags().StringVarP(&cachePath, "cache_path", "c", "", "Path to cache image layers")
 	cmd.Flags().StringVar(&format, "format", "tarball", fmt.Sprintf("Format in which to save images (%q, %q, or %q)", "tarball", "legacy", "oci"))
+	cmd.Flags().StringVar(&convert, "convert", "oci", fmt.Sprintf("Image spec to convert the pulled image (%q)", "oci"))
```
thesayyn, line 122: It is so weird to have these two options available to us.

pkg/crane/pull.go
```diff
@@ -135,7 +135,7 @@ func MultiSaveOCI(imgMap map[string]v1.Image, path string) error {
 	}
 	for ref, img := range imgMap {
 		anns := map[string]string{
-			"dev.ggcr.image.name": ref,
+			"org.opencontainers.image.ref.name": ref,
```
thesayyn, line 138: same as above.

pkg/v1/mutate/oci.go
```diff
@@ -0,0 +1,113 @@
...
+	}
+
+	manifest := m.DeepCopy()
+
+	for i, layer := range manifest.Layers {
+		switch layer.MediaType {
+		case types.DockerLayer:
+			manifest.Layers[i].MediaType = types.OCILayer
+		case types.DockerUncompressedLayer:
+			manifest.Layers[i].MediaType = types.OCIUncompressedLayer
```
thesayyn, line 54: I did not cover the other cases as they don't have their OCI counterparts. I just let arbitrary descriptors (docker-specific and user-defined ones) pass through. docker hub was pretty happy accepting a mix of these. 

these two left unchanged: `DockerPluginConfig` and `DockerForeignLayer`

imjasonh:
Sorry for the delay in responding to this. Even more apologies if these are topics you've already discussed. 😞 

+1 to changing the OCI annotations to match the recommended/reserved standard.

I'm slightly -1 to having a "just mutate this manifest" method, since it breaks some invariants we have in place otherwise, and we might find out those invariants are load-bearing. I believe it should be possible to achieve the same result just using `mutate.AppendLayers(empty.Image, layers)`, `mutate.ConfigFile`, and `mutate.Annotations`. Is there something else you specifically need from `mutate.ImageManifest`?

I agree that having `--format` and `--convert` is a bit confusing. I wonder if we might just want a single-purpose `crane ocify` command that takes a docker-typed image||index and pushes over it with the OCI-flavored version. This could be used in a bash pipeline to achieve the same thing, without the confusing flags.

I'm not _totally_ convinced even this is a good idea, so if we go down this path I might want us to make it a hidden CLI surface, with docs saying that it might go away in the future.

In fact, as I'm thinking of it, would a command that does just that be sufficient for your use case, even outside of `crane`? I don't believe there's anything happening in this PR that can't be done using existing methods outside of this repo (correct me if I'm wrong), so it should be possible to unblock you just using code you write yourself. I'm sensitive to me becoming a speedbump for you, if you have a need for this sooner than I can review.

Let me know what you think.

thesayyn:
> I'm slightly -1 to having a "just mutate this manifest" method, since it breaks some invariants we have in place otherwise, and we might find out those invariants are load-bearing. I believe it should be possible to achieve the same result just using `mutate.AppendLayers(empty.Image, layers)`, `mutate.ConfigFile`, and `mutate.Annotations`. Is there something else you specifically need from `mutate.ImageManifest`?

I am on the same page with you. I used mutators for everything else except for changing the manifest. the reason i did that is I wanted to preserve the ordering of the layer field. in theory i could have just do removeLayers and appendLayers but i am only interested in some specific layers so I could, in theory, displace a layers order and break the image.  I didn't realize there was a mutate.ImageManifest.



> I wonder if we might just want a single-purpose `crane ocify` command that takes a docker-typed image||index and pushes over it with the OCI-flavored version.

having --format oci and ocify would be even more confusing IMHO :D. maybe we rename the option to be something more descriptive? i believe options should have been `--layout=oci|tar` `--format=oci|auto` . 


> I'm not _totally_ convinced even this is a good idea, so if we go down this path I might want us to make it a hidden CLI surface, with docs saying that it might go away in the future.
I'd like to know more about concerns maybe I am doing something wrong? I didn't think it would be a bad idea to have this given that a few tools already do this and is a well-defined standard and supported by almost every container runtime/registry. 


> In fact, as I'm thinking of it, would a command that does just that be sufficient for your use case, even outside of `crane`?

I feel your pain as a maintainer that you don't feel like landing a feature unless there is a great public interest. I do that myself too. a lot of people really don't care about what mediaTypes their container images have, hell they don't know what it is as long as their deployment works. However, I am under the impression this would useful for others as well and I want to land it here. 

my use case was to rip out any randomness from the base images. I'll try to enumerate my points

1- docker specification has randomness built-in into their images. it is so easy to break reproducibility and docker helps that with properties such as `docker_version` `container`. they are just random and bound to their build environment. an example; building the same image with the same `mtime` etc yields a different hash. 
2- properties such as health check and their friends are not portable and yet still docker is encouraging it.
3- this is the standard that docker should be following but they just do whatever they want as an industry dominator.

also given the size of the change, it wouldn't be such a big maintenance burden. 


> I'm sensitive to me becoming a speedbump for you, if you have a need for this sooner than I can review.

I don't want you to land anything here if you don't feel strongly about it. I am just trying to push crane to a place where you can build images on the disk.  (amend RFC).

Yet still, if you think my arguments are not good enough, then maybe we could hide this option behind an experimental flag so that people depend on it carefully and beware that it might go away if it is good enough?

thesayyn:
you could be anticipating some unforeseen issues with this as well. but if there is anything i can address, i am happy to address them.

thesayyn:
friendly ping - @imjasonh @jonjohnsonjr

thesayyn:
I have addressed the issues above. I'd be happy if someone could take another look. I'd really like reach to a conclusion about whether to accept this or not

thesayyn:
Mixing docker media types definitely should not be prohibited. Maybe we could avoid this case when the image contains layers that don't have OCI counterparts?

hown3d:
This is something that's also interesting (almost neccessary) for https://github.com/google/ko/pull/738.

To provide annotations for the wasi image, OCI format is neccessary and for example quay.io prohibits mixed layers.

thesayyn:
> I'll chat with @jonjohnsonjr about it this week and see if we can make some progress. Thanks for hanging on this long, and thanks for this contribution.

I started working on rules_oci. we kinda need this. This seems to affect other people as well. See https://github.com/project-zot/zot/issues/622

So far the concerns were;

- mixing `mediaTypes`
- losing cosign signatures due to mismatching signatures
- some images can't be converted 100%

thesayyn:
@jonjohnsonjr just had another pass at this. removing changes already landed. skimmed down to just oci conversion.

pkg/v1/mutate/oci.go
```diff
@@ -0,0 +1,162 @@
...
+			if err != nil {
+				return nil, fmt.Errorf("getting layer: %w", err)
+			}
+			layer, err = tarball.LayerFromOpener(func() (io.ReadCloser, error) { return reader, nil }, tarball.WithMediaType(types.OCIUncompressedLayer))
+			if err != nil {
+				return nil, fmt.Errorf("building layer: %w", err)
+			}
+		}
+		newLayers = append(newLayers, layer)
+	}
```
jonjohnsonjr, line 106: I suspect there is a better way to do this, this is rather expensive.

thesayyn: I'd expect that but i wasn't able to find it.

alexeagle: ping @jonjohnsonjr do you have suggestions to make this less expensive? This PR seems stuck.

jonjohnsonjr: This is annoying because of how `partial.Descriptor` works, but I would probaby create a new type, something like:

```go
type layerWithMediaType struct {
  v1.Layer
  mt types.MediaType
}

func (l *layerWithMediaType) Descriptor() (*v1.Descriptor, error) {
  // call partial.Descriptor on wrapped layer and update MediaType
}

func (l *layerWithMediaType) MediaType() (types.MediaType, error) {
  return l.mt, nil
}
```

alexeagle:
Not stale

Issue #1295, opened 2022-02-18
crane: allow appending of complete docker images


tnaroska:
### Describe the bug

I have a use case were I would like to append all layers of image A onto image B to create a new image C. First idea was to use `crane rebase` with old_base set to an empty (scratch) image. However, this results in crane trying to load _scratch_ from the registry and an image not found error.

### To Reproduce
```
crane rebase -t C --new_base B  --old_base scratch A
Error: rebasing image: GET https://index.docker.io/v2/library/scratch/manifests/latest: MANIFEST_UNKNOWN: manifest unknown; unknown tag=latest
```

### Expected behavior

Rebase must not try to resolve `scratch` from a remote registry, but rather treat it as an empty image. A rebase operation with a `--old_base=scratch` then practically becomes an append operation. Since scratch has no content, no layers of the original image are removed and all of originals image then appended to the new base.

### Additional context

- Output of `crane version`
```
$ crane version
0.8.0
```

tnaroska:
Hi @jonjohnsonjr, yes that should work in general. However, as you say it would merge the image into a single layer which I'd like to prevent.
The use case I'm experimenting with is a java application generated with [jib](https://github.com/GoogleContainerTools/jib). Jib cuts the image into separate layers to allow re-use between image builds (for example, dependency jars which hardly change between builds get separated into their own layer). I have a java app that I need to append onto multiple base images while ideally still keeping the advantages of layer re-use.
General idea is to build java side with jib onto a scratch image so I get a an empty image with only the java layers generated by jib. Then use crane append to stitch the layers onto the several base images that I need the app in.

PR #1296, opened 2022-02-18
crane: allow appending of complete docker images

tnaroska:
crane rebase: correctly interpret 'scratch' as reference to the empty image rather than trying to resolve it from a registry.

This effectively allows rebase to work as an append operation, appending all images of a source onto a destination image:
`crane rebase -t C --new_base B  --old_base scratch A`
-> appends all layers of image A onto image B and stores as image C

Fixes https://github.com/google/go-containerregistry/issues/1295

### DIFF
--- a/cmd/crane/cmd/rebase.go
+++ b/cmd/crane/cmd/rebase.go
@@ -23,6 +23,7 @@ import (
 	"github.com/google/go-containerregistry/pkg/logs"
 	"github.com/google/go-containerregistry/pkg/name"
 	v1 "github.com/google/go-containerregistry/pkg/v1"
+	"github.com/google/go-containerregistry/pkg/v1/empty"
 	"github.com/google/go-containerregistry/pkg/v1/mutate"
 	specsv1 "github.com/opencontainers/image-spec/specs-go/v1"
 	"github.com/spf13/cobra"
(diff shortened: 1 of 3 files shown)


cmd/crane/cmd/rebase.go
```diff
@@ -179,9 +180,12 @@ func rebaseImage(orig v1.Image, oldBase, newBase string, opt ...crane.Option) (v
 		return nil, fmt.Errorf("either old base or %q annotation is required", specsv1.AnnotationBaseImageDigest)
```
tnaroska, line 180: `oldBase == ""` will still return an error for backwards compatibility. oldBase needs to be set explicitly to 'scratch' for the new behavior to kick in. Even though this is inconsistent with `crane append --base`

pkg/v1/empty/image.go
```diff
@@ -22,6 +22,9 @@ import (
 	"github.com/google/go-containerregistry/pkg/v1/types"
 )
 
+// docker scratch image name
+const Scratch = "scratch"
```
imjasonh, line 26: I'd like to avoid adding an exported field to this package until we get more signal that it's useful. In the meantime just matching against the string in rebase.go sounds fine to me.

tnaroska:
@imjasonh any recommendation regarding tests? I could add something here https://github.com/google/go-containerregistry/blob/main/cmd/crane/rebase_test.sh

imjasonh:
> @imjasonh any recommendation regarding tests? I could add something here https://github.com/google/go-containerregistry/blob/main/cmd/crane/rebase_test.sh

Yeah that seems like a good place. Thanks!

tnaroska:
@imjasonh https://github.com/google/go-containerregistry/pull/1296/commits/42eab71ffde4c34f66a6a440728f544210083265 adds a test to rebase_test.sh.
If more tests of that form a coming, might want to look into [shunit2](https://github.com/kward/shunit2) or similar. That would allow structuring tests into separate methods within the script and also support setup/teardown steps.

imjasonh:
Thanks for adding the test! It looks like it isn't run in presubmit, fixing in https://github.com/google/go-containerregistry/pull/1304

The change lgtm, I think I might just ask that we have a section in [rebase.md](https://github.com/google/go-containerregistry/blob/main/cmd/crane/rebase.md) that briefly describes what `--old_base=scratch` means. If you feel like adding that, that would be helpful. Otherwise I can.

skirsten:
Hey, what is the status of this?
Just ran into this problem and would love for this to move forward.

As far as I can tell this change looked good so if nobody has time I can create a new PR with the rebased code.

PR #1301, opened 2022-02-21
Take advantage of Chainguard maintained versions of various actions.

mattmoor:
I created these so we could more centrally manage them for folks.  We can snap a release and use `@v1` instead of `@main` (and manage with dependabot) but I wanted to test the waters.

If y'all are interested in this, I can do the same for `ko` next.


imjasonh (approved):
Beautiful red!

mattmoor:
I pinned the actions to commits, and `dependabot` should bump them as they change.

Issue #1330, opened 2022-03-31
feature request: explore usage of sha256-simd


howardjohn:
https://github.com/minio/sha256-simd#drop-in-replacement

Currently when profiling building images, almost 95%+ of cpu time (and the CPU time is ~= wall time) is on sha256 operations. sha256-simd claims to make this much faster, which in turn could make image building faster.

I haven't proven it will actually have real impacts. Opening this to track + get some early feedback on if this is something that would be an acceptable dependency.

imjasonh:
Seems worth experimenting with at least!

This should be the one place we create a sha256 hasher: https://github.com/google/go-containerregistry/blob/7e30746dde2f5a19b609f885dfef4e7b1d1503ad/pkg/v1/hash.go#L78

There also this, which should probably use `v1.Hasher()`: https://github.com/google/go-containerregistry/blob/e7cd6afc43c1dd7f1c5cb4e2dbab3836d9110b24/pkg/registry/manifest.go#L113

We might want to make a new experiment triggered on the presence of an env var, so folks can opt in to it for now. If it's promising we can make it opt-out.

PR #1331, opened 2022-03-31
Add experimental option to use use sha256-simd on CPUs that support it

howardjohn:
Fixes https://github.com/google/go-containerregistry/issues/1330

This library is a small drop-in replacement for `crypto/sha256` that utilizes hardware operations to compute hashes. In many cases (or at least mine, anyhow), sha256 computation is the bulk of time spent building images; this patch gives around 2x improvement in real world cases (I think the bottleneck becomes the `registry:2` image I am using, which could apply similar optimizations probably).

For safety, this is disabled by default.


pkg/v1/hash.go
```diff
@@ -19,12 +19,16 @@ import (
...
 	"fmt"
+	sha256simd "github.com/minio/sha256-simd"
 	"hash"
 	"io"
+	"os"
 	"strconv"
 	"strings"
 )
 
+const HasherSHA256 = "sha256"
```
imjasonh, line 30: Maybe we should even have this be `var HasherSHA256 hash.Hash`, which we may override at `init` time based on the experiment env var? Then callers can use either `v1.Hasher("sha256")` or `v1.HasherSHA256` more or less interchangeably.

imjasonh:
@jonjohnsonjr points out that there's an [open proposal](https://github.com/golang/go/issues/50543) to add this directly to Go (possibly in the 1.19 timeframe 🤞), in which case I'd probably prefer to just wait for that, and avoid a new dependency.

Chasing links around, I found https://github.com/opencontainers/go-digest/issues/71 where we seem to be having the same wait-or-do-it-now questions 😆

PR #1334, opened 2022-04-06
fix(v1/remote): return an error if both auth and keychain are set

### LINKED ISSUE #1332: v1/remote: `WithAuth` and `WithAuthFromKeychain` conflict
### Describe the bug

Setting `WithAuthFromKeychain` and `WithAuth` will cause the explicitly-set `auth.Authenticator` from the latter to be overwritten.

### To Reproduce

```go
myAuth := remote.WithAuth(authn.FromConfig(authn.AuthConfig{
	Username: "estroz",
	Password: "password",
}))
// Assuming you have no local config or config does not contain myAuth's creds,
// this keychain's Resolve() will return authn.Anonymous.
myKeychain := remote.WithAuthFromKeychain(authn.DefaultKeychain)

const srcImage = `docker.io/library/nginx:latest`
srcRef, err := name.ParseReference(srcImage)
if err != nil {
	panic(err)
}

img, err := remote.Image(srcRef, myAuth)
if err != nil {
	panic(err)
}

const dstImage = `docker.io/estroz/nginx:testtag1`
dstRef, err := name.ParseReference(dstImage)
if err != nil {
	panic(err)
}

badOpts := []remote.Option{myKeychain, myAuth}
if err := remote.Write(dstRef, img, badOpts...); err != nil {
	fmt.Println("badopts:", err) // this prints
}

goodOpts := []remote.Option{myAuth}
if err := remote.Write(dstRef, img, goodOpts...); err != nil {
	fmt.Println("goodopts:", err) // this does not print
}
```

### Expected behavior

If auth is explicitly set, use it and either ignore keychain or return an error that both cannot be set.

### Additional context

Relates to #156

- Version: v0.8.0
- Registry: DockerHub (but I image this would result from any authenticated registry usage)

estroz:
Fixes #1332

ping @imjasonh

### DIFF
--- a/pkg/v1/remote/options.go
+++ b/pkg/v1/remote/options.go
@@ -102,7 +102,6 @@ var DefaultTransport = &http.Transport{
 
 func makeOptions(target authn.Resource, opts ...Option) (*options, error) {
 	o := &options{
-		auth:           authn.Anonymous,
 		transport:      DefaultTransport,
 		platform:       defaultPlatform,
 		context:        context.Background(),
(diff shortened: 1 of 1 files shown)


jonjohnsonjr:
I haven't verified myself but I'm a little worried about how this might interact with the crane and gcrane packages, which use a keychain my default. This might break callers who supply `WithAuth`?

estroz:
@jonjohnsonjr I'll investigate and update this PR if I find a reasonable fix.

estroz:
From what I can see there is no way a caller that passes a `crane.WithAuth` and `crane.WithAuthFromKeychain`, either explicitly or implicitly via [`makeOptions`](https://github.com/google/go-containerregistry/blob/v0.8.0/pkg/crane/options.go#L44-L46), can get this error. Both options [override](https://github.com/google/go-containerregistry/blob/v0.8.0/pkg/crane/options.go#L80-L101) the default keychain option. Effectively, the behavior in `remote` this PR corrects was already enforced by crane/gcrane.

jonjohnsonjr (approved):
Thanks!

PR #1337, opened 2022-04-08
Deprecate transport.New

benmoss:
Not sure if there was something @jonjohnsonjr was waiting on to deprecate this

### DIFF
--- a/pkg/v1/remote/check.go
+++ b/pkg/v1/remote/check.go
@@ -24,7 +24,7 @@ func CheckPushPermission(ref name.Reference, kc authn.Keychain, t http.RoundTrip
 	}
 
 	scopes := []string{ref.Scope(transport.PushScope)}
-	tr, err := transport.New(ref.Context().Registry, auth, t, scopes)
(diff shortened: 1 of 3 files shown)


imjasonh (approved):
sgtm

pkg/v1/remote/check.go
```diff
@@ -24,7 +24,7 @@ func CheckPushPermission(ref name.Reference, kc authn.Keychain, t http.RoundTrip
 	}
 
 	scopes := []string{ref.Scope(transport.PushScope)}
-	tr, err := transport.New(ref.Context().Registry, auth, t, scopes)
+	tr, err := transport.NewWithContext(context.Background(), ref.Context().Registry, auth, t, scopes)
```
jonjohnsonjr, line 27: Should be `context.TODO` probably? (We want to remove this func eventually given that TODO, but in case we don't, it's better to track this lack of ctx propagation)

benmoss: You could probably say the same about the `context.Background` on line 40 😄 . Should I change that as well or leave it alone?

PR #1338, opened 2022-04-09
pull: Write image to stdout with "-"

abitrolly:
This allows `crane pull` to stream image to stdout, so that it is possible to stream-extract files from filesystems of remote images without explicitly saving them.
```
✗ ./crane pull ubuntu - | ./crane export - | tar -Oxf - etc/lsb-release
DISTRIB_ID=Ubuntu
DISTRIB_RELEASE=20.04
DISTRIB_CODENAME=focal
DISTRIB_DESCRIPTION="Ubuntu 20.04.4 LTS"  
```
As asked by @imjasonh in https://github.com/google/go-containerregistry/pull/1274#pullrequestreview-895584402

### DIFF
--- a/pkg/v1/tarball/write.go
+++ b/pkg/v1/tarball/write.go
@@ -55,15 +55,9 @@ func MultiWriteToFile(p string, tagToImage map[name.Tag]v1.Image, opts ...WriteO
 // MultiRefWriteToFile writes in the compressed format to a tarball, on disk.
 // This is just syntactic sugar wrapping tarball.MultiRefWrite with a new file.
 func MultiRefWriteToFile(p string, refToImage map[name.Reference]v1.Image, opts ...WriteOption) error {
-	var w io.WriteCloser
-	var err error
-	if p == "-" {
-		w = os.Stdout
-	} else {
-		w, err = os.Create(p)
-		if err != nil {
-			return err
-		}
+	w, err := os.Create(p)
+	if err != nil {
+		return err
 	}
 	defer w.Close()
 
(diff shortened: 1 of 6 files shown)


imjasonh:
Why

```
 ./crane pull ubuntu - | ./crane export - | ...
```

Instead of

```
./crane export ubuntu
```

?

abitrolly:
The real use case is working with local images from `podman`.
```
podman image save 01b01 - | ./crane export - - | tar -Oxf - etc/os-release
```

abitrolly:
But yes, I have no use case for `pull` to stream image tarball to stdout.

pkg/v1/tarball/write.go
```diff
@@ -55,9 +55,15 @@ func MultiWriteToFile(p string, tagToImage map[name.Tag]v1.Image, opts ...WriteO
...
-		return err
+	var w io.WriteCloser
+	var err error
+	if p == "-" {
+		w = os.Stdout
+	} else {
+		w, err = os.Create(p)
+		if err != nil {
+			return err
+		}
```
jonjohnsonjr, line 66: Move this logic into `crane.MultiSave` and have it call `tarball.MultiRefWrite` directly instead of this method.

abitrolly: Called `tarball.MultiWrite` instead of `tarball.MultiRefWrite`, because `crane.MultiSave` construct `tagToImage` instead of `refToImage`.

abitrolly: Replaced `tarball.MultiWrite` in favor of `tarball.MultiRefWrite` after cleaning out the old code.

abitrolly:
Rebased and addressed the review. Checked out that code in `crane.pull` was written before `tarball.MultiRefWrite` was written.  And that code was generating `i-was-a-digest` filler tag to be able to write the `tarball`. I removed that old code, and now the `tarball` is written with `"RepoTags": null` and `i-was-a-digest` is removed.

@jonjohnsonjr PTAL.

/spent 3h

abitrolly:
Squashed some commits.

abitrolly:
@github-actions still waiting for review.

PR #1344, opened 2022-04-11
Make test registry more resilient

joaopapereira:
In this PR there are 3 changes done to the test Registry that is part of ggcr.
1. Enable blob mounting in the Registry
    This initial part will add a new Interface that is used when the Registry is called with a PUT that contains a mount request
3. Allow the users to decide if they want to save to disk the blobs pushed to the registry
    Activated via the Option `registry.DiskBlobStorage` this is a very simple process that just saves the blobs to disk and keeps a map with information on where to locate the blobs
5. Store blob information based on the repository and SHA
    Activated via the Option `registry.SplitBlobsByRepository` this changes the way the blob storage structs store information to ensure that the repository is also taken into account when accessing the blob information

With this PR we keep the original defaults that existed, so it should have no impact on people currently using the registry.

### DIFF
--- a/cmd/registry/main.go
+++ b/cmd/registry/main.go
@@ -10,12 +10,33 @@ import (
 )
 
 var port = flag.Int("port", 1338, "port to run registry on")
+var storage = flag.String("storage", "memory", "Option [memory, disk], default value is memory")
+var diskLocation = flag.String("path", "", "Required when selecting disk storage. Path where the registry will store the blobs")
+var splitRepositories = flag.Bool("split-repositories", false, "Separate blob storage per repository. This feature will ensures that each repository only has access to blobs inside the repository space.")
 
 func main() {
 	flag.Parse()
+
+	var opts []registry.Option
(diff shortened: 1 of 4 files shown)


imjasonh (commented):
So these changes look pretty good to me. If we're adding options to pkg/registry, I think I'd also like to try to pipe them through to cmd/registry as well, and then I can pipe them through to https://github.com/chainguard-dev/actions/tree/main/setup-registry so folks can set them there too. But that could be a separate PR.

I think we had vaguely hoped that disk-backed storage would come in the form of an OCI layout backend for blob and manifest storage, so you could have a local registry collecting pushes, and then take that OCI layout that was built as a result and use it elsewhere.

That of course would depend on having pkg/registry's manifest storage also be swappable, which kinda petered out a while ago. If you're interested in picking up that torch though, that could be awesome. If there's anything I can do to help there (besides provide months-late PR reviews 😓) let me know, I'd be happy to help.

PR #1346, opened 2022-04-13
use k8s keychain first

dprotaso:
This speeds up instances where the credentials are in image pull secrets

### DIFF
--- a/pkg/authn/k8schain/k8schain.go
+++ b/pkg/authn/k8schain/k8schain.go
@@ -46,11 +46,11 @@ func New(ctx context.Context, client kubernetes.Interface, opt Options) (authn.K
 	}
 
 	return authn.NewMultiKeychain(
+		k8s,
 		authn.DefaultKeychain,
 		google.Keychain,
 		amazonKeychain,
 		azureKeychain,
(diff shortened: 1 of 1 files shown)


pkg/authn/k8schain/k8schain.go
```diff
@@ -46,11 +46,11 @@ func New(ctx context.Context, client kubernetes.Interface, opt Options) (authn.K
 	}
 
 	return authn.NewMultiKeychain(
+		k8s,
 		authn.DefaultKeychain,
```
imjasonh, line 50:
```suggestion
		authn.DefaultKeychain,
		k8s,
```

I think we may still want the default keychain first, and it should at least be much quicker than the others, barring any cred helper shenanigans.

WDYT?

dprotaso: I'm ok with that .

Just be aware that if you're testing k8schain (pulling from creds from secrets) and you're hitting a real cluster you'll want to `docker logout` beforehand :)

dprotaso:
ok k8s is first now

imjasonh (approved):

PR #1352, opened 2022-04-20
update go action to always get the latest available and other updates

cpanato:
- update go action to always get the latest available
- update some jobs to use go 1.17
- add go1.18 to be tested as well
- update golangci-lint to 1.45.2


cpanato:
the 1.14.x is now called 1.14, if this change is ok we will need to update the github status check

imjasonh (approved):

PR #1369, opened 2022-05-18
Include header referring to the image ref being pulled

imjasonh:
This can be useful to registry operators to help associate blob pull traffic with the actual image being requested.

I'm opening this mainly to gather input (and to bikeshed the header name). I think this information could be useful to registry operators, both as information that can be usefully aggregated to show how many blobs or bytes were pulled "because" of an image ref, and possibly to inform better rate limiting decisions.

Ideally I'd like to propose something like this to OCI's distribution-spec, as a "clients MAY send ..." section, and try to get that into more and more clients (docker, containers/image, etc), so this is more uniformly reported across the ecosystem. But first I wanted to try it out in here and get feedback.

edit: Instead of `X-Go-Container-Registry-For-Ref`, which is a mouthful, maybe just `OCI-Ref: index.docker.io/library/ubuntu:latest` ?

Example:

```
$ go run ./cmd/crane pull ubuntu /dev/null --verbose
...
2022/05/18 10:15:07 --> GET https://index.docker.io/v2/library/ubuntu/manifests/sha256:aa6c2c047467afc828e77e306041b7fa4a65734fe3449a54aa9c280822b0d87d
2022/05/18 10:15:07 GET /v2/library/ubuntu/manifests/sha256:aa6c2c047467afc828e77e306041b7fa4a65734fe3449a54aa9c280822b0d87d HTTP/1.1
Host: index.docker.io
User-Agent: crane/(devel) go-containerregistry/(devel)
Accept: application/vnd.docker.distribution.manifest.v2+json
Authorization: <redacted>
X-Go-Container-Registry-For-Ref: index.docker.io/library/ubuntu:latest
Accept-Encoding: gzip
...
2022/05/18 10:10:48 --> GET https://index.docker.io/v2/library/ubuntu/blobs/sha256:d2e4e1f511320dfb2d0baff2468fcf0526998b73fe10c8890b4684bb7ef8290f
2022/05/18 10:10:48 GET /v2/library/ubuntu/blobs/sha256:d2e4e1f511320dfb2d0baff2468fcf0526998b73fe10c8890b4684bb7ef8290f HTTP/1.1
Host: index.docker.io
User-Agent: crane/(devel) go-containerregistry/(devel)
Authorization: <redacted>
X-Go-Container-Registry-For-Ref: index.docker.io/library/ubuntu@sha256:aa6c2c047467afc828e77e306041b7fa4a65734fe3449a54aa9c280822b0d87d
Accept-Encoding: gzip
...
```

NB: In the example above, a manifest GET for a child manifest in an index includes the ref of the parent index, and a blob GET for a child manifest includes the child manifest's ref-by-digest, not the parent index's. We can change this if we want to, or even include both if we really felt like it.

### DIFF
--- a/pkg/v1/remote/descriptor.go
+++ b/pkg/v1/remote/descriptor.go
@@ -243,6 +243,9 @@ func (f *fetcher) fetchManifest(ref name.Reference, acceptable []types.MediaType
 	if err != nil {
 		return nil, nil, err
 	}
+	if f.Ref != nil {
+		req.Header.Set("X-Go-Container-Registry-For-Ref", f.Ref.Name())
+	}
 	accept := []string{}
 	for _, mt := range acceptable {
 		accept = append(accept, string(mt))
@@ -307,6 +310,9 @@ func (f *fetcher) headManifest(ref name.Reference, acceptable []types.MediaType)
 	if err != nil {
 		return nil, err
 	}
+	if f.Ref != nil {
+		req.Header.Set("X-Go-Container-Registry-For-Ref", f.Ref.Name())
+	}
 	accept := []string{}
 	for _, mt := range acceptable {
 		accept = append(accept, string(mt))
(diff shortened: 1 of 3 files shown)


jonjohnsonjr:
Let's suggest this in an OCI meeting and maybe merge something for next release?

PR #1385, opened 2022-06-14
Feature/add new sub cmd tagged

Wilkhu90:
* Adding a new sub command gct to print all tagged images recursively

### DIFF
--- a/cmd/gcrane/cmd/gct.go
+++ b/cmd/gcrane/cmd/gct.go
@@ -1,0 +1,78 @@
+// Copyright 2018 Google LLC All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


cmd/gcrane/cmd/gct.go
```diff
@@ -0,0 +1,78 @@
...
+	"github.com/google/go-containerregistry/pkg/name"
+	"github.com/google/go-containerregistry/pkg/v1/google"
+	"github.com/spf13/cobra"
+)
+
+// NewCmdGct creates a new cobra.Command for the gct subcommand.
+func NewCmdGct() *cobra.Command {
+	recursive := false
+	cmd := &cobra.Command{
+		Use:   "gct",
```
imjasonh, line 31: Mostly a drive-by comment, but I don't think `gct` is a very descriptive name for this command. It may also be worth explaining in the help text how this differs from `gcrane ls`.

PR #1386, opened 2022-06-14
remote: Collect ETag in response, optimistically send that back in future updates

imjasonh:
When pulling a manifest, if the response includes an `ETag` header, hold on to it, and send that back in an `If-Match` request header when putting that manifest back.

This should prevent some race conditions during read-modify-write cycles, if the registry respects the `If-Match` header, since the resource's etag will have changed when the other client updated it.

TODO: Add test.

- https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/ETag#avoiding_mid-air_collisions
- https://github.com/opencontainers/distribution-spec/pull/251/

### DIFF
--- a/pkg/v1/mutate/image.go
+++ b/pkg/v1/mutate/image.go
@@ -39,6 +39,13 @@ type image struct {
 	digestMap       map[v1.Hash]v1.Layer
 }
 
+func (i *image) ETag() string {
+	if e, ok := i.base.(partial.WithETag); ok {
+		return e.ETag()
+	}
+	return ""
+}
+
 var _ v1.Image = (*image)(nil)
 
 func (i *image) MediaType() (types.MediaType, error) {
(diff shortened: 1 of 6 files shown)


jonjohnsonjr:
Maybe it's time to have formal unwrapping?

I was thinking something more along the lines of https://github.com/google/go-containerregistry/pull/1387 for the `partial` stuff, but it might be better to have an actual unwrapping API (borrowed from https://go.dev/blog/go1.13-errors) so we can check "does this (or something it wraps) implement `ETag`?" instead of just returning an empty string.

oofnish:
Hello, I know this PR is currently closed, but I'm interested in picking up and running with the ETag support here. Is it just tests that are missing for this, or would the "Formal unwrapping" approach be a requirement for a MVP here?

PR #1387, opened 2022-06-15
Allow config files to be mounted

jonjohnsonjr:
Because we wanted to reuse the layer uploading code, we wrap the config
bytes up as a Layer. Unfortunately, the partial.ConfigLayer
implementation doesn't preserve the fact that for mountable images, we
can perform a cross-repo mount.

This change updates partial.ConfigLayer to look for a ConfigLayer
implementation on the given image to see if it can simply dispatch to
that, falling back to calculating it from the config bytes otherwise.


imjasonh (approved):
Nice.

Issue #1321, opened 2022-03-22
FR: Support cross-origin blob mounting


rafibarash:
**This FR proposes adding a `origin` URL parameter, for the reason outlined below.**

`crane` & `gcrane` support efficient blob copies via [cross-repo blob mounting](https://docs.docker.com/registry/spec/api/#cross-repository-blob-mount) by adding the ['mount' and 'from'](https://github.com/google/go-containerregistry/blob/c6366444aa46d4d05854960485646f5d7a2bcd09/pkg/v1/remote/write.go#L273) url parameters on certain upload requests.

A CRBM http request via crane looks like so:

```
POST https://us-docker.pkg.dev/v2/<dest_proj>/<dest_repo>/<dest_pkg>/blobs/uploads/?from=<src_proj>/<src_repo>/<src_pkg>&mount=<src_sha256_digest>
```

For storage services that own multiple registries (e.g. Google owning `gcr.io`, `us-docker.pkg.dev`, `us-west1-docker.pkg.dev`, etc.), **encoding the source registry hostname in a url parameter allows for efficient copies from a different origin.**

This would look exactly the same as CRBM, but with an added `origin` parameter:

```
POST https://us-west1-docker.pkg.dev/v2/.../blobs/uploads?from=...&mount=...&origin=gcr.io
```

This change would be made in the [remote](https://github.com/google/go-containerregistry/blob/c6366444aa46d4d05854960485646f5d7a2bcd09/pkg/v1/remote/write.go#L273) pkg and propagate to `crane` & `gcrane` & any other clients using `remote`.

Also, I can't take credit for this proposal - @jonjohnsonjr originally proposed this.

imjasonh:
Is this supported by registries today?

This behavior isn't described in the spec: https://github.com/opencontainers/distribution-spec/blob/main/spec.md#mounting-a-blob-from-another-repository

rafibarash:
Hey Jason,

Cross-**repo** blob mounting is currently supported by registries, but this proposal is talking about Cross-**origin** blob mounting, which involves mounting a blob from another origin (hostname).

In the screenshot of the link you posted, I highlighted the part of the spec that says CRBM should be on the same registry.

![6cgmiSAoCpWiJWu](https://user-images.githubusercontent.com/27692412/159990182-acae67a7-856d-4bb0-a138-06d554a4da8e.png)

[...]

PR #1388, opened 2022-06-15
Implement cross-registry blob mounting experiment

jonjohnsonjr:
This is currently only implemented by Google but we'd like to upstream
this to the distribution spec. There's a small possibility that this
will break registries that don't implement the spec properly.

We signal the domain of the origin registry with the "origin"
querystring parameter.

Fixes https://github.com/google/go-containerregistry/issues/1321

Likely breaks with harbor because of https://github.com/google/go-containerregistry/pull/219 but they [SHOULD](https://github.com/opencontainers/distribution-spec/blob/main/spec.md#mounting-a-blob-from-another-repository) be returning a 202 here.

### DIFF
--- a/pkg/v1/remote/check.go
+++ b/pkg/v1/remote/check.go
@@ -39,7 +39,7 @@ func CheckPushPermission(ref name.Reference, kc authn.Keychain, t http.RoundTrip
 		client:  &http.Client{Transport: tr},
 		context: context.Background(),
 	}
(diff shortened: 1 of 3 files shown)


jonjohnsonjr:
Picking up https://github.com/google/go-containerregistry/pull/1322

imjasonh (approved):

PR #1389, opened 2022-06-17
bump deps using ./hack/bump-deps.sh

imjasonh:


konrad945:
Hi, @imjasonh @jonjohnsonjr what is the status of this review? It's open for some time and it fixes some CVEs (docker) that I would like to fix in my project.

imjasonh:
The sigs.k8s.io/json dep that gets bumped here replaced `interface{}` with `any`, introduced in Go 1.18.

Diff:
https://github.com/kubernetes-sigs/json/compare/227cbc7cc124...ff3dbbeab8d9

We can either:
- not bump the dep; future `./hack/bump-deps.sh`es will bump it again
- `replace` our dep back to the one that uses `interface{}`; users who depend on this won't pick up our `replace`, and users who `go install github.com/google/go-containerregistry/cmd/crane` won't get the `replace`, and will fail if they're building with Go <1.18
- assert that we can only be built with Go 1.18+, update CI accordingly, wait for 1.19 (currently in RC2) so that we can say we still support the last two Go releases.

jonjohnsonjr:
I'm ok with requiring 1.18.

PR #1396, opened 2022-06-24
Do not check /v2 endpoint on registry when RoundTripper is provided

joaopapereira:
This PR contains the following changes:
1. Bump staticcheck to the latest version and fix problems detected.
2. When a RoundTripper of the type `transport.Wrapper` is provided to the function `NewWithContext`, the function will assume that the called has already created the RoundTripper and made sure that the authentication happened and that the registry supports V2 of distribution spec.


jonjohnsonjr:
Thanks! Do you mind splitting the staticcheck stuff out into a separate PR just so it's easier to revert the roundtripper change if we need to?

jonjohnsonjr:
Ah the two separate commits is fine!

jonjohnsonjr (approved):

PR #1397, opened 2022-06-26
Returning the specific reason for the failure of ParseReference

UBarney:
I think it's better to return the specific reason for the failure of `ParseReference`, so that it is easier for users to debug. For example, when executing the following code, return the reason for the failure of `NewTag` (repo name cannot contain capital letters), so that the the user knows what went wrong and how to fix it (not everyone knows that repo names can't have capital letters...)
```go
	err := crane.Push(nil, "badRepo:1.0")
	if err != nil {
		panic(err.Error())
	}

```


UBarney:
> Contributor License Agreement

I have signed CLA

UBarney:
@jonjohnsonjr failed test has been fixed

PR #1401, opened 2022-06-29
Don't annotate refs by default, switch to OCI key

jonjohnsonjr:
A bunch of breaking changes that might be contentious, but I think this
is probably the right behavior.

1. Stick the full image ref in the image (not just e.g. ubuntu) so it's
   clear that it's the image and not just a tag.
2. Only do this when --annotate-ref is passed, so that these things are
   location-independent by default.
3. Switch from a crane annotation key to the standard OCI key.

### DIFF
--- a/cmd/crane/cmd/pull.go
+++ b/cmd/crane/cmd/pull.go
@@ -21,6 +21,7 @@ import (
 	"github.com/google/go-containerregistry/pkg/name"
 	v1 "github.com/google/go-containerregistry/pkg/v1"
 	"github.com/google/go-containerregistry/pkg/v1/cache"
+	"github.com/google/go-containerregistry/pkg/v1/empty"
 	"github.com/google/go-containerregistry/pkg/v1/layout"
 	"github.com/google/go-containerregistry/pkg/v1/remote"
 	"github.com/spf13/cobra"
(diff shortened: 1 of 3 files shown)


cmd/crane/cmd/pull.go
```diff
@@ -81,20 +85,42 @@ func NewCmdPull(options *[]crane.Option) *cobra.Command {
...
 				}
+				for ref, img := range imageMap {
+					opts := []layout.Option{}
+					if annotateRef {
+						parsed, err := name.ParseReference(ref, o.Name...)
+						if err != nil {
+							return err
+						}
+						opts = append(opts, layout.WithAnnotations(map[string]string{
+							"org.opencontainers.image.ref.name": parsed.Name(),
```
imjasonh, line 104: > No semantic restriction is given for the "org.opencontainers.image.ref.name" annotation of descriptors.

but also

> Implementor's Note: A common use case of descriptors with a "org.opencontainers.image.ref.name" annotation is representing a "tag" for a container image. For example, an image may have a tag for different versions or builds of the software. In the wild you often see "tags" like "v1.0.0-vendor.0", "2.0.0-debug", etc. Those tags will often be represented in an image-layout repository with matching "org.opencontainers.image.ref.name" annotations like "v1.0.0-vendor.0", "2.0.0-debug", etc.

Despite the spec asserting otherwise, do tools that interact with layouts tend to interpret these as tags, and not full image refs?

If so this would have been a problem before too, just curious if we're moving even further from accepted behavior.

jonjohnsonjr: It's straightforward to distinguish a tag from a full ref (`strings.Contains(v, "/")`), so I feel like it _should_ be okay to do this, but it could possibly break some stuff (which is why I don't want to do it by default anymore).

The grammar defined in https://github.com/opencontainers/image-spec/blob/main/annotations.md allows for a full ref, so I don't feel like we'd be doing anything particularly egregious, but @sudo-bmitch might disagree.

imjasonh (approved):

PR #1402, opened 2022-06-29
Wrap progress updates in a mutex

jonjohnsonjr:
Atomically incrementing the number of bytes written isn't sufficient if
we're sending the updates out of order.

I ran the progress tests 50 times after this change and they passed.

### DIFF
--- a/pkg/v1/remote/progress.go
+++ b/pkg/v1/remote/progress.go
@@ -1,0 +1,69 @@
+// Copyright 2022 Google LLC All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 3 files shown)


pkg/v1/remote/progress.go
```diff
@@ -0,0 +1,69 @@
...
+	sync.Mutex
+	updates    chan<- v1.Update
+	lastUpdate *v1.Update
+}
+
+func (p *progress) total(delta int64) {
+	atomic.AddInt64(&p.lastUpdate.Total, delta)
+}
+
+func (p *progress) complete(delta int64) {
```
imjasonh, line 35: How about `send(progress int64)`?

pkg/v1/remote/progress.go
```diff
@@ -0,0 +1,69 @@
...
+	v1 "github.com/google/go-containerregistry/pkg/v1"
+)
+
+type progress struct {
+	sync.Mutex
+	updates    chan<- v1.Update
+	lastUpdate *v1.Update
+}
+
+func (p *progress) total(delta int64) {
```
imjasonh, line 31: How about `incrTotal(delta int64)`?

imjasonh (approved):

PR #1403, opened 2022-07-01
Implement crane edit

jonjohnsonjr:
This works at least on linux with neovim. If stdin appears to be
interactive, crane edit will open an editor with the contents of the
target manifest/config/file. Otherwise, crane edit will read the desired
contents from stdin and just overwrite them for the target.

### DIFF
--- a/cmd/crane/cmd/root.go
+++ b/cmd/crane/cmd/root.go
@@ -20,6 +20,7 @@ import (
 	"path/filepath"
 
 	"github.com/docker/cli/cli/config"
+	"github.com/google/go-containerregistry/internal/cmd"
 	"github.com/google/go-containerregistry/pkg/crane"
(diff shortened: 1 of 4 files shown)


cmd/crane/cmd/edit.go
```diff
@@ -0,0 +1,417 @@
...
+	}
+
+	if dst == "" {
+		dst = src
+		ref, err := name.ParseReference(src, o.Name...)
+		if err != nil {
+			return nil, err
+		}
+		if _, ok := ref.(name.Digest); ok {
+			dst = fmt.Sprintf("%s@%s", ref.Context(), digest)
```
imjasonh, line 202: nit: `ref.Context().Digest(digest)` all over the place

imjasonh (approved):

imjasonh (approved):

Issue #1404, opened 2022-07-04
crane: unable to copy between to registries (registry.gitlab.com and docker.io)


ProbstDJakob:
### Describe the bug

We copy images from gitlab.com to docker.io and until version 0.9.0 this worked fine. With version 0.10.0 this stopped working. The process has not been changed (written as pipeline job) and all authentication tokens still work. The error we receive is the following:
```
2022/07/04 08:31:37 Copying from registry.gitlab.com/bbbatscale/bbbatscale/bbbatscale:4.6.0 to docker.io/bbbatscale/bbbatscale:4.6.0
Error: failed to copy image: POST https://index.docker.io/v2/bbbatscale/bbbatscale/blobs/uploads/?from=bbbatscale%2Fbbbatscale%2Fbbbatscale&mount=sha256%3Ab85a868b505ffd0342a37e6a3b1c49f7c71878afe569a807e6238ef08252fcb7&origin=REDACTED: UNAUTHORIZED: authentication required; [map[Action:pull Class: Name:bbbatscale/bbbatscale Type:repository] map[Action:push Class: Name:bbbatscale/bbbatscale Type:repository] map[Action:pull Class: Name:bbbatscale/bbbatscale/bbbatscale Type:repository]]
```
When executing the same command with crane v0.9.0 this works (same environment).

### To Reproduce

I am not quite sure if this is really the problem, but if so it could be reproduced by copying a _non-existing_ image from a public registry to docker.io.

### Expected behavior

When copying from one registry to another this should not result in an unauthorized error if the required permissions have been granted.

### Additional context

My uneducated guess is that as the URL `https://index.docker.io/[...]/uploads/?from=bbbatscale%2Fbbbatscale%2Fbbbatscale&[...]` states, it tries to use the image `bbbatscale/bbbatscale/bbbatscale` which most likely will be resolved to `docker.io/bbbatscale/bbbatscale/bbbatscale` and this image does not exist, which therefore might cause the unauthorized error.

- Output of `crane version`: `0.10.0`
- Registry used (e.g., GCR, ECR, Quay): registry.gitlab.com and docker.io

imjasonh:
@jonjohnsonjr sounds like another case of #1321 and a spec-non-compliant destination registry.

hobti01:
We're having the same issue since crane 0.10.0 and this really seems like a bug or at least a regression.

* `crane copy docker.io/bash:4 registry.gitlab.com/redacted/bash:4`
[...]

PR #1406, opened 2022-07-12
Fall back to no mount if registry misbehaves

jonjohnsonjr:
Sometimes we see authentication errors if the from repo doesn't exist,
so fall back to just initiating an upload without a mount attempt.

Fixes https://github.com/google/go-containerregistry/issues/1404


imjasonh (approved):

imjasonh:
Should we cut a v0.10.1 for this?

PR #1407, opened 2022-07-17
feat: allow pushing an image by digest

ktarplee:
When there are multiple images in the OCI directory the user can select the one you want by digest specifying it in the image reference (destination string).

## Testing
Pull down some sample images.  The first is a manifest.list and the second third is a plain manifest.
```shell
crane pull --format oci quay.io/csiaddons/volumereplication-operator:v0.3.0 saved.oci
crane pull --format oci quay.io/ceph/ceph-amd64:v17.2.1-20220623 saved.oci
podman run -d -p 127.0.0.1:5000:5000 --name registry docker.io/library/registry:2
go run ./cmd/crane push --index --insecure saved.oci localhost:5000/volumereplication-operator@sha256:cba18d9c2cd17f52fc4f1b8d19a3f6af18d29b7aded5ed0f21388da87fd6f543
go run ./cmd/crane push --index --insecure saved.oci localhost:5000/ceph-amd64@sha256:b4e95ad6b1a4fb02b3d5c3fc8cbdb91431d03fdf5c6d904f8a5cf1efb4053f27
podman rm -f registry
```

The current alternative for the user is to modify `saved.oci/index.json` to contain a single manifest descriptor of the image they want to push.  This can be done with `jq` but is rather tedious.

### DIFF
--- a/cmd/crane/cmd/push.go
+++ b/cmd/crane/cmd/push.go
@@ -50,6 +50,42 @@ func NewCmdPush(options *[]crane.Option) *cobra.Command {
 			if err != nil {
 				return err
 			}
+
+			// If the destination contains a digest then we should only push the manifest with that digest
+			if digestRef, ok := ref.(name.Digest); ok {
+				// convert the string to a v1.Hash
+				hashRef, err := v1.NewHash(digestRef.DigestStr())
+				if err != nil {
+					return err
+				}
+
+				switch t := img.(type) {
+				case v1.Image:
+					// ensure that the digest matches
+					hh, err := t.Digest()
+					if err != nil {
+						return err
+					}
+					if hashRef != hh {
+						return fmt.Errorf("requested manifest not present in index, missing %s", hh)
+					}
+				case v1.ImageIndex:
+					// Get a specific image
+					// Try ImageIndex first
+					img, err = t.ImageIndex(hashRef)
+					if err == nil {
+						break
+					}
+					// Fallback to plain Image
+					img, err = t.Image(hashRef)
+					if err != nil {
+						return err
+					}
+					// TODO there is no good way to handle the errors here since they are untyped and they contain information we do not know (the media type of the match).
+					//  I wish the findDescriptor() function was exposed or there was a more tolerant call to "Image" that did not care about Manifest vs Manifest List.
+				}
+			}
+
 			var h v1.Hash
 			switch t := img.(type) {
 			case v1.Image:


cmd/crane/cmd/push.go
```diff
@@ -50,6 +50,42 @@ func NewCmdPush(options *[]crane.Option) *cobra.Command {
...
+					}
+				case v1.ImageIndex:
+					// Get a specific image
+					// Try ImageIndex first
+					img, err = t.ImageIndex(hashRef)
+					if err == nil {
+						break
+					}
+					// Fallback to plain Image
+					img, err = t.Image(hashRef)
```
ktarplee, line 80: This is the ugly part.  It seems to me there is no good way to figure out which image type hashRef points to without redoing work (e.g., `findDescriptors()`) in `Image()` and `ImageIndex()`.

ktarplee:
I have implemented a different approach that allows one to do a `crane push path/to/something.oci reg.com/foo --match-name=index.docker.io/libary/busybox:latest` to select the image from the oci directory they would like to push.  It is implemented https://github.com/ktarplee/go-containerregistry/tree/name

That branch also adds a new command called `crane name` that fully qualifies a name (and optionally locks it to a digest in the registry).  For example it converts `busybox` to `index.docker.io/library/busybox:latest` which is handy in dealing with the OCI index annotations (index.json).

ktarplee:
I just imported go-containerregistry in my own project and implemented the features I needed.  It turns out to not be necessary to extend crane in this way to get this features I need so that is good news.

jonjohnsonjr:
Going through a backlog of email from when I was out, sorry to have missed this.

> Use k8s style selectors but use the manifest annotations instead. For example, add --selector or short form -l to the command to select the desired value of an annotation in the manifest. For example, crane push --index saved.oci reg.example.com/dest -l org.opencontainers.image.ref.name=quay.io/ceph/ceph-amd64:v17.2.1-20220623

Yes! I have had this idea myself. I think there's value here, especially when dealing with creative [ab]uses of image layouts.

> That branch also adds a new command called crane name that fully qualifies a name

This is interesting. I've thought about having a flag for e.g. `crane ls` and `crane digest` that emits a fully qualified ref to make it easier to chain commands, but exposing the name package just for normalizing refs would probably be useful too...

> I just imported go-containerregistry in my own project and implemented the features I needed. It turns out to not be necessary to extend crane in this way to get this features I need so that is good news.

I'd be curious to know what you're doing, but I suspect it's not public? I am certainly interested in any modifications to `crane` that would make it more flexible, but also you're following the expected path of graduating from crane to ggcr when it isn't flexible enough 👍

ktarplee:
@jonjohnsonjr Thank for replying on this MR.  I added the "name" command in this branch along with a few other improvements that are not on this MR.  https://github.com/ktarplee/go-containerregistry/tree/name   It might be useful to understand our use case so you can understand why I wanted to make these changes.  Doing a `crane pull` of many images to the same OCI directory and it works great.  We then take that directory to an air-gapped environment and we wanted to use `crane push` to push them.  We did not want to push all the images in the directory to the same repo or even the same registry.   So we needed the ability to use "matchers" by annotation or digest to select a specific image or image index and push that to a repo.  Then we can loop.  The `crane name` command comes into play when we script this since we need to know how crane will expand the name as it puts it in the annotations in the OCI's index.json file.

PR #1408, opened 2022-07-18
Redact sensitive information in redirected URLs

imjasonh:
This adds internal methods to redact potentially sensitive information
from URLs in error messages, especially when those URLs are the result
of server-side redirects.

We already redact potentially sensitive information like this as part of
transport.CheckError, but this isn't called when the error is the result
of an http.Client.Do (e.g., tcp dial error).

The specific use case where this can happen is a registry like GCR which
redirects blob requests to GCS with a sensitive access_token in the
query parameter. If the request to GCS fails due to tcp error, the error
message will include the sensitive access token.

This method of redaction relies on the original error being a
*url.Error, and redaction is accomplished by simply updating the error's
URL with a redacted equivalent.

### DIFF
--- a/internal/redact/redact.go
+++ b/internal/redact/redact.go
@@ -41,10 +41,8 @@ func FromContext(ctx context.Context) (bool, string) {
 // If the error is a *url.Error, this returns a *url.Error with the URL redacted.
 // Any other error type, or nil, is returned unchanged.
 func Error(err error) error {
-	if err == nil {
-		return nil
-	}
 	// If the error is a url.Error, we can redact the URL.
+	// Otherwise (including if err is nil), we can't redact.
 	var uerr *url.Error
 	if ok := errors.As(err, &uerr); !ok {
 		return err
--- a/pkg/v1/remote/descriptor.go
+++ b/pkg/v1/remote/descriptor.go
@@ -399,7 +399,7 @@ func (f *fetcher) headBlob(h v1.Hash) (*http.Response, error) {
 
 	resp, err := f.Client.Do(req.WithContext(f.context))
 	if err != nil {
-		return nil, err
+		return nil, redact.Error(err)
 	}
 
 	if err := transport.CheckError(resp, http.StatusOK); err != nil {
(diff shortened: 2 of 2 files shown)


internal/redact/redact.go
```diff
@@ -33,3 +36,77 @@ func FromContext(ctx context.Context) (bool, string) {
 	reason, ok := ctx.Value(redactKey).(string)
 	return ok, reason
 }
+
+// Error redacts potentially sensitive query parameter values in the URL from the error's message.
+func Error(err error) error {
```
jonjohnsonjr, line 41: -1 on this because we throw away error types, can you do this in here instead?

https://github.com/google/go-containerregistry/blob/d187a716b559771a3768caa9c4dba796cb4650b3/pkg/v1/remote/transport/error.go#L67-L87

jonjohnsonjr: Or redact this: https://pkg.go.dev/net/url#Error

jonjohnsonjr (changes requested):
LGTM just update the commit message, PR description, and godoc to reflect the new reality.

internal/redact/redact.go
```diff
@@ -33,3 +35,57 @@ func FromContext(ctx context.Context) (bool, string) {
 	reason, ok := ctx.Value(redactKey).(string)
 	return ok, reason
 }
+
+// Error redacts potentially sensitive query parameter values in the URL from the error's message.
+//
+// If the error is a *url.Error, this returns a *url.Error with the URL redacted.
+// Any other error type, or nil, is returned unchanged.
+func Error(err error) error {
+	if err == nil {
```
jonjohnsonjr, line 44: `errors.As` does a nil check if you want to be lazier.

pkg/v1/remote/descriptor.go
```diff
@@ -367,7 +368,7 @@ func (f *fetcher) fetchBlob(ctx context.Context, size int64, h v1.Hash) (io.Read
 
 	resp, err := f.Client.Do(req.WithContext(ctx))
 	if err != nil {
-		return nil, err
+		return nil, redact.Error(err)
```
jonjohnsonjr, line 371: Maybe we want to do this everywhere we do `f.Client.Do` (or create a `f.Do` that does it)? Or at least in `headBlob` and `blobExists` as well, given we expect this to happen for redirected things, which are usually blobs.

internal/redact/redact.go
imjasonh: I always want to be lazier 🏖️

jonjohnsonjr (approved):

PR #1413, opened 2022-07-20
Support the platform specific authentication of krane in "auth get" command

ingwonsong:
This PR enables to work "auth get" command with the platform specific keychains given in `krane` tool.

### DIFF
--- a/cmd/crane/cmd/auth.go
+++ b/cmd/crane/cmd/auth.go
@@ -32,7 +32,7 @@ import (
 )
 
 // NewCmdAuth creates a new cobra.Command for the auth subcommand.
-func NewCmdAuth(options *[]crane.Option, argv ...string) *cobra.Command {
+func NewCmdAuth(options []crane.Option, argv ...string) *cobra.Command {
 	cmd := &cobra.Command{
 		Use:   "auth",
 		Short: "Log in or access credentials",
@@ -63,7 +63,7 @@ func toCreds(config *authn.AuthConfig) credentials {
 }
 
 // NewCmdAuthGet creates a new `crane auth get` command.
-func NewCmdAuthGet(options *[]crane.Option, argv ...string) *cobra.Command {
+func NewCmdAuthGet(options []crane.Option, argv ...string) *cobra.Command {
 	if len(argv) == 0 {
 		argv = []string{os.Args[0]}
 	}
@@ -97,12 +97,7 @@ func NewCmdAuthGet(options *[]crane.Option, argv ...string) *cobra.Command {
 			if err != nil {
 				return err
 			}
-			keychain := authn.DefaultKeychain
-			if options != nil {
-				opts := crane.GetOptions(*options...)
-				keychain = opts.Keychain
-			}
-			authorizer, err := keychain.Resolve(reg)
+			authorizer, err := crane.GetOptions(options...).Keychain.Resolve(reg)
 			if err != nil {
 				return err
 			}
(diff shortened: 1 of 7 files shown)


imjasonh:
This is really interesting! So the idea is that with this change, `krane auth get gcr.io` will print out the auth token, possibly as derived from a compiled-in cred helper like ecr-login etc.

I think we'll want a simple end-to-end test of this behavior, for [GHCR](https://github.com/google/go-containerregistry/blob/main/.github/workflows/ghcr-auth.yaml) and [ECR](https://github.com/google/go-containerregistry/blob/main/.github/workflows/ecr-auth.yaml) at least. 

For ECR, that means calling `krane auth get ${{ env.AWS_ACCOUNT }}.dkr.ecr.${{ env.AWS_REGION }}.amazonaws.com` and checking that it doesn't fail, and that it prints something -- though we should be careful not to print that out to logs.

Could you add that to this PR?

ingwonsong:
@imjasonh Sure. I have just added the tests. 
In addition, I also added small extension.

Previously, `crane` and `krane` get a registry address through `stdio` only, like:
```
$ echo "example.registry.com" | krane auth get
```

But, I added to another way to specify the registry address as an argument as follows:
```
$ krane auth get example.registry.com
```

cmd/crane/cmd/auth.go
```diff
@@ -62,30 +63,46 @@ func toCreds(config *authn.AuthConfig) credentials {
...
 			}
-			reg, err := name.NewRegistry(strings.TrimSpace(string(b)))
+
+			reg, err := name.NewRegistry(registryAddr)
 			if err != nil {
 				return err
 			}
-			authorizer, err := authn.DefaultKeychain.Resolve(reg)
+			keychain := authn.DefaultKeychain
+			if options != nil {
```
jonjohnsonjr, line 101: I think you can drop this and just always pull out `crane.GetOptions(*options...).Keychain` given that crane options use authn.DefaultKeychain as the... default keychain.

I'd just make sure nothing is passing nil options into here.

cmd/gcrane/main.go
```diff
@@ -42,7 +42,7 @@ func main() {
 	root := cmd.New(use, short, []crane.Option{crane.WithAuthFromKeychain(gcrane.Keychain)})
 
 	// Add or override commands.
-	gcraneCmds := []*cobra.Command{gcmd.NewCmdList(), gcmd.NewCmdGc(), gcmd.NewCmdCopy(), cmd.NewCmdAuth("gcrane", "auth")}
+	gcraneCmds := []*cobra.Command{gcmd.NewCmdList(), gcmd.NewCmdGc(), gcmd.NewCmdCopy(), cmd.NewCmdAuth(nil, "gcrane", "auth")}
```
jonjohnsonjr, line 45: I'd pull the options from line 42 out and reuse them here so that `gcrane auth get` uses `google.Keychain`.

jonjohnsonjr (changes requested):

cmd/crane/cmd/auth.go
ingwonsong: Good idea. Done.

cmd/gcrane/main.go
ingwonsong: Done.

Now, `gcrane auth get` also works!

jonjohnsonjr (approved):

jonjohnsonjr:
Thanks for this change, this is awesome.

If I had the time, I would set up workload identity so that we could test `gcrane auth get` and `krane auth get` against GCP, but I'm fine without it 😅 

Looks like you just need to regenerate the docs for the presubmit.

PR #1419, opened 2022-07-27
Correct usage of authn.NewKeychainFromHelper in docs

ethan-lowman-dd:
`authn.NewFromHelper` is not a function, and should be `authn.NewKeychainFromHelper`.


imjasonh (approved):
Thanks! 👍

PR #1420, opened 2022-07-27
Ignore docker config if it's a directory

lcarva:
In a kubernetes environment, a volume from a Secret can be optionally
mounted. If the Secret does not exist, the mounted path is always a
directory.

This change modifies auth.DefaultKeychain to ignore the case where the
docker config file is unexpectedly a directory.

Signed-off-by: Luiz Carvalho <lucarval@redhat.com>

### DIFF
--- a/pkg/authn/keychain_test.go
+++ b/pkg/authn/keychain_test.go
@@ -21,6 +21,7 @@ import (
 	"io/ioutil"
 	"log"
 	"os"
+	"path"
 	"path/filepath"
 	"reflect"
 	"testing"
@@ -63,7 +64,7 @@ func setupConfigDir(t *testing.T) string {
 	fresh++
 	p := filepath.Join(tmpdir, fmt.Sprintf("%d", fresh))
 	t.Logf("DOCKER_CONFIG=%s", p)
-	os.Setenv("DOCKER_CONFIG", p)
+	t.Setenv("DOCKER_CONFIG", p)
 	if err := os.Mkdir(p, 0777); err != nil {
 		t.Fatalf("mkdir %q: %v", p, err)
 	}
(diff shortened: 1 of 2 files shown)


imjasonh (approved):
Thanks!

pkg/authn/keychain_test.go
```diff
@@ -107,6 +108,7 @@ func TestPodmanConfig(t *testing.T) {
 	fresh++
 	p := filepath.Join(tmpdir, fmt.Sprintf("%d", fresh))
 	os.Setenv("XDG_RUNTIME_DIR", p)
+	defer os.Unsetenv("XDG_RUNTIME_DIR")
```
imjasonh, line 111: I think we can use [`t.Setenv`](https://pkg.go.dev/testing#T.Setenv) here instead of `os.Setenv` and defer-unsetting.

pkg/authn/keychain_test.go
```diff
@@ -363,3 +365,27 @@ func TestNewKeychainFromHelper(t *testing.T) {
...
+	tmpdir := setupConfigDir(t)
+	// Create "config.json" as a directory, not a file to simulate optional
+	// secrets in Kubernetes.
+	err := os.Mkdir(path.Join(tmpdir, "config.json"), 0777)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	// Clear out files that may have been created by previous tests.
+	err = os.RemoveAll(path.Join(os.Getenv("HOME"), ".docker"))
```
imjasonh, line 379: Is this going to delete my ~/.docker directory if I run this test? 🤔

imjasonh (changes requested):

pkg/authn/keychain_test.go
lcarva: It shouldn't because `TestMain` sets `HOME` to a temporary location. I agree that it does look odd...

Let me see if I can refactor how `HOME` is set in the first place.

imjasonh: That's okay, let's keep it how it is unless we want to fix it in a separate PR. Thanks for clarifying how it works.

pkg/authn/keychain_test.go
lcarva: Nice! Updated.

pkg/authn/keychain_test.go
lcarva: Added a defer call to remove the `.docker/config.json` right after it is created in the corresponding test.

imjasonh (approved):

imjasonh (approved):

PR #1421, opened 2022-07-30
Add fuzzer and run it in CI

AdamKorcz:
This adds a fuzzer and runs it in the CI with [ClusterfuzzLite](https://google.github.io/clusterfuzzlite/).

I contribute fuzzers to several projects on OSS-Fuzz and have seen crashes in GCR by fuzzing projects that use its APIs. It would be great to have a setting where contributors can add fuzzers that will run and the GCR maintainers get notified about found issues. Integrating into OSS-Fuzz may also be an option, but for now ClusterfuzzLite should be enough to get started.

With this PR, ClusterfuzzLite will run fuzzers in the CI when PRs are made, and it can be extended beyond this PR with code coverage and batch fuzzing: https://google.github.io/clusterfuzzlite/running-clusterfuzzlite/github-actions/

### DIFF
--- a/.github/workflows/cflite.yaml
+++ b/.github/workflows/cflite.yaml
@@ -1,0 +1,26 @@
+name: ClusterFuzzLite PR fuzzing
+on:
+  workflow_dispatch:
+  pull_request:
+    paths:
+      - '**'
+permissions: read-all
+jobs:
+  PR:
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        sanitizer: [address]
+    steps:
+    - name: Build Fuzzers (${{ matrix.sanitizer }})
+      id: build
+      uses: google/clusterfuzzlite/actions/build_fuzzers@v1
+      with:
+        sanitizer: ${{ matrix.sanitizer }}
+        language: go
+    - name: Run Fuzzers (${{ matrix.sanitizer }})
+      id: run
+      uses: google/clusterfuzzlite/actions/run_fuzzers@v1
+      with:
+        github-token: ${{ secrets.GITHUB_TOKEN }}
--- a/.github/workflows/cflite.yaml
+++ b/.github/workflows/cflite.yaml
@@ -10,16 +10,14 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       fail-fast: false
-      matrix:
-        sanitizer: [address]
     steps:
-    - name: Build Fuzzers (${{ matrix.sanitizer }})
+    - name: Build Fuzzers
       id: build
       uses: google/clusterfuzzlite/actions/build_fuzzers@v1
       with:
-        sanitizer: ${{ matrix.sanitizer }}
+        sanitizer: 'address'
         language: go
-    - name: Run Fuzzers (${{ matrix.sanitizer }})
+    - name: Run Fuzzers
       id: run
       uses: google/clusterfuzzlite/actions/run_fuzzers@v1
       with:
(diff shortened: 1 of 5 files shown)


imjasonh:
Awesome!! Thanks for this 🔥🔥🔥

My only real question is whether there's a reason to prefer Go's built-in fuzzing, which just requires a fuzz test method and `go test -fuzz`. The Dockerfile seems unnecessary unless there's a benefit to running via cluster fuzz.

But in any case, thanks again for adding this, we'll definitely want this in some form.

.github/workflows/cflite.yaml
```diff
@@ -0,0 +1,26 @@
...
+    strategy:
+      fail-fast: false
+      matrix:
+        sanitizer: [address]
+    steps:
+    - name: Build Fuzzers (${{ matrix.sanitizer }})
+      id: build
+      uses: google/clusterfuzzlite/actions/build_fuzzers@v1
+      with:
+        sanitizer: ${{ matrix.sanitizer }}
```
imjasonh, line 20: Can we remove the matrix here if there's only one element/

.github/workflows/cflite.yaml
```diff
@@ -0,0 +1,26 @@
...
+        language: go
+    - name: Run Fuzzers (${{ matrix.sanitizer }})
+      id: run
+      uses: google/clusterfuzzlite/actions/run_fuzzers@v1
+      with:
+        github-token: ${{ secrets.GITHUB_TOKEN }}
```
imjasonh, line 26: Why does this need GITHUB_TOKEN?

If we keep this, can we limit its permissions to the needed subset? Hopefully this doesn't need more than `contents: read`.

AdamKorcz:
> Awesome!! Thanks for this 🔥🔥🔥
> 
> My only real question is whether there's a reason to prefer Go's built-in fuzzing, which just requires a fuzz test method and `go test -fuzz`. The Dockerfile seems unnecessary unless there's a benefit to running via cluster fuzz.
> 
> But in any case, thanks again for adding this, we'll definitely want this in some form.

Running the fuzzers in the CI with `go test -fuzz` has a few shortcomings for example:
1. The fuzzers are not building up corpus over time. They will start from scratch at every run. ClusterfuzzLite can build up corpus and use that every time.
2. Running the fuzzers with go test -fuzz will run all fuzzers at every PR for a given time. When more fuzzers are added, this adds extensive time to the CI. Considering 10 fuzzers that run 3 minutes each - which is little time for a fuzzer - it quickly adds up. Given the number of APIs that have a level of complexity, I believe GCR would quickly reach 10 fuzzers. ClusterfuzzLite aims to run only the fuzzers that cover code that is affected by a PR.
3. ClusterfuzzLite also enables batch fuzzing which can be useful for continuous testing, but also as a mechanism to keep the corpus up to date automatically.

In my opinion, ClusterfuzzLite doesn't add complexity but reduces it.

> The Dockerfile seems unnecessary unless there's a benefit to running via cluster fuzz

This is correct, the Dockerfile is for ClusterfuzzLite specifically.

.github/workflows/cflite.yaml
AdamKorcz: I am not sure _why_ ClusterfuzzLite needs the Github token, just that it is a requirement. 

[This discussion](https://github.com/google/clusterfuzzlite/issues/86) should cover the permissions required.

imjasonh: This claims that `GITHUB_TOKEN` is `required:false`: https://github.com/google/oss-fuzz/blob/129f8f2639015d11ee2efa23c7ea83c6818b1f55/infra/cifuzz/actions/run_fuzzers/action.yml#L38

Does the workflow pass without it? Or is it possible I'm looking at the wrong workflow?

imjasonh:
Closing in favor of https://github.com/google/oss-fuzz/pull/8627 which seems simpler.

Thanks Adam!

PR #1425, opened 2022-08-08
WIP: add References() methods to remote.Image and remote.ImageIndex

imjasonh:
This starts to sketch out support for the OCI ref types proposal in https://github.com/opencontainers/wg-reference-types/blob/main/docs/proposals/PROPOSAL_E.md, soon to be proposed for discussion and 🤞acceptance🤞 by OCI spec maintainers.

This adds a `References()` method to `pkg/v1/remote.Image` and `pkg/v1/remote.ImageIndex` to query for descriptors that refer to this image or index.

From here we can build a `Walk` method that recursively walks while it can find a `References()` method, e.g., to copy trees of things between registries, maintaining their structure.

The next step here would probably be to add optional support in `pkg/registry` for tests, so we can test upgrade/downgrade scenarios and general (Go) API functionality.

Future changes can also add support for the [artifact manifest type](https://github.com/opencontainers/wg-reference-types/blob/main/docs/proposals/PROPOSAL_E.md#artifact-spec), but that's a bit more involved and not yet supported by any registry, so I think we should wait to add that, and/or play with it in `internal` before committing to anything.

This is a draft to get feedback.

cc @jdolitsky

### DIFF
--- a/pkg/v1/manifest.go
+++ b/pkg/v1/manifest.go
@@ -28,6 +28,11 @@ type Manifest struct {
 	Config        Descriptor        `json:"config"`
 	Layers        []Descriptor      `json:"layers"`
 	Annotations   map[string]string `json:"annotations,omitempty"`
+
+	// This field is *experimental* and surrounding behavior is likely to change.
+	//
+	// See https://github.com/opencontainers/wg-reference-types/blob/main/docs/proposals/PROPOSAL_E.md#image-spec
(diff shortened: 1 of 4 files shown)


jdolitsky:
Had started on some of this a bit back, this changeset might be helpful: https://github.com/google/go-containerregistry/compare/main...oci-playground:main

LGTM so far 👍

PR #1437, opened 2022-08-17
fix: consider base image media type when appending layers

### LINKED ISSUE #1427: crane: mutate and append adds layer with docker mediaType regardless of what base image uses
### Describe the bug

`mutate` or `append` adds new layers as `DockerLayer` instead of `OCILayer` when the base image only has OCI mediaTypes.
 This behavior leads to mixed mediaTypes.
 
### To Reproduce

```
crane mutate thesayyn/debian:oci --append <some>.tar. <your-registry>:oci-mixed
``` 

### Expected behavior

The newly added layer should have the mediaType `application/vnd.oci.image.layer.v1.tar+gzip` instead of `application/vnd.docker.image.rootfs.diff.tar.gzip`. Perhaps this could be controlled by a global flag? Maybe `--preferred-media-types=oci|docker`?

### Additional context

- Output of `crane version`:  0.11.0
- Registry used (e.g., GCR, ECR, Quay): Docker

thesayyn:
As discussed with @jonjohnsonjr.  We still need to bikeshed name for the new `--oci` flag. 

Closes #1427


imjasonh:
How about `crane append --oci-empty`?

At some point we might have a `crane append --oci-layout-path=path/foo` to pull and append a layer then write it to a local OCI layout. We should make sure the names don't conflict and get too confusing.

thesayyn:
> At some point we might have a `crane append --oci-layout-path=path/foo` to pull and append a layer then write it to a local OCI layout. We should make sure the names don't conflict and get too confusing.

I agree.

> How about `crane append --oci-empty`?

This is a bit unclear I think. `--oci-empty-base`?

imjasonh:
`--oci-empty-base` sounds good. `crane` should fail if `--oci-empty-base` is set while `-b` is set, to be clear it only works when appending to an empty image.

imjasonh (approved):

thesayyn:
> `--oci-empty-base` sounds good. `crane` should fail if `--oci-empty-base` is set while `-b` is set, to be clear it only works when appending to an empty image.

I am not sure why the unit test is failing. do you have any pointers? using go1.19 locally.

thesayyn:
fixed the unit test. should be ready now

pkg/crane/append.go
```diff
@@ -47,9 +48,25 @@ func Append(base v1.Image, paths ...string) (v1.Image, error) {
 		return nil, fmt.Errorf("getting base image: %w", err)
 	}
 
+	if base == nil {
```
imjasonh, line 51: `isWindows` checks for `img==nil` already, we could remove that and put the `== nil` check before `isWindows`.

thesayyn: makes sense.

thesayyn: fixed it

imjasonh (approved):

Issue #1431, opened 2022-08-15
ggcr: Multiple k8s pull secrets for same registry doesn't work


dprotaso:
### Describe the bug

See: https://github.com/knative/serving/issues/13126

### To Reproduce

See linked issue:

dprotaso:
cc @jwcesign

imjasonh:
https://github.com/google/go-containerregistry/pull/1280 was supposed to handle the case where auth is correctly configured for `registry.com/foo` and `registry.com/bar` and possibly also just `registry.com` -- if your request is for `/foo`, you should get creds for foo, i.e., as specific as possible. If your request is for `/baz` you get creds for `registry.com`. This assumes that auth is configured correctly in all cases, and we should just be as specific as possible in selecting among those valid creds.

If there's a bug in that logic though, we should fix it -- the kaniko issue that inspired that change (https://github.com/GoogleContainerTools/kaniko/issues/687) was reopened because the behavior persisted.

This doesn't account for the case where there are two creds configured both for the same registry(+repo), but one is invalid and the other works. It sounds like kubelet will collect all matching auths and try them in a loop, and ggcr (currently) doesn't -- it will choose one at ~random and use it, even if it fails and the other would have worked.

It will be hard, but not impossible, for pkg/v1/remote to handle this fallback scenario, and try creds for `/foo` then fallback to creds for `/` if it gets a 403 response. That should match kubelet's behavior linked from the Knative issue above.

jwcesign:
I did test as follows:
first create secret:
```shell
root@cesign [12:38:20 AM] [+47.0°C] [~/git/group-service-acceleration] [main *]
-> # kubectl create secret docker-registry jwcesign \
--docker-server=https://index.docker.io/v1/ \
--docker-username=jwcesign \
--docker-password=xxx \
--docker-email=jwcesign@163.com

root@cesign [12:40:14 AM] [+45.0°C] [~/git/group-service-acceleration] [main *]
-> # kubectl create secret docker-registry cesign \
--docker-server=https://index.docker.io/v1/ \
--docker-username=cesign \
--docker-password=xxx \
--docker-email=jwcesign@gmail.com
```

then create service:
```shell
root@cesign [12:38:59 AM] [+47.0°C] [~/git/group-service-acceleration/config] [main *]
-> # cat group-service-c.yaml
apiVersion: serving.knative.dev/v1
kind: Service
metadata:
  name: group-service-c
spec:
  template:
    metadata:
[...]

Issue #1329, opened 2022-03-30
crane: should support multiple credentials to the same registry


acornett21:
### Describe the bug
When trying to pull an image from a registry ie `quay.io` with an auth file that contains multiple entries for the same registry, crane gives an `UNATHORIZED` error.

### To Reproduce
1. Have access to a registry with different accounts for different namespaces/images
2. Create an auth file with multiple entries for the same registry
```
{
  "auths": {
    "quay.io/telcoci/simple-demo-operator-bundle": {
      "auth": "XXXXXX"
    },
    "quay.io/telcoci": {
      "auth": "YYYYYY"
    },
    "quay.io": {
      "auth": "ZZZZZZ"
    }
  }
}
```

3. Run `crane pull  quay.io/telcoci/simple-demo-operator-catalog:0.0.3` (as an example)
4. See an error of `Error: GET https://quay.io/v2/telcoci/simple-demo-operator-catalog/manifests/0.0.3: UNAUTHORIZED: access to the requested resource is not authorized; map[]`

### Expected behavior
I'd expect `crane` to be able to use the proper `auth` value like podman, skopeo, buildah, etc.

### Additional context
This also does not work when using the crane api's. We are using `crane` as a library for a go based project found [here](https://github.com/redhat-openshift-ecosystem/openshift-preflight), so should be implemented in the `cli` as well as the `api`.

This also might relate to: 

-  #723 

but I think `crane` cli cmds use the files in the `authn` pkg

- Output of `crane version`: `0.8.0`
- Registry used (e.g., GCR, ECR, Quay): `quay` but I assume this exists for all registries.

imjasonh:
I believe https://github.com/google/go-containerregistry/pull/1280 should fix this.

You can install crane from head (`go install github.com/google/go-containerregistry/cmd/crane@main`) and try it out.

If so it might be worth cutting a release so folks don't have to depend on main to get this functionality.

acornett21:
@imjasonh Thanks for the reply. This still seems to be an issue, see the response in the issue [here](https://github.com/redhat-openshift-ecosystem/openshift-preflight/issues/495#issuecomment-1083736429)

jotadrilo:
@jonjohnsonjr I submitted a different bug fix for this issue

PR #1439, opened 2022-08-18
WIP: support multiple credentials

imjasonh:
This is an attempt to allow our authn and transport packages to cooperate and allow for keychains (incl `DefaultKeychain`) to provide multiple creds, which will be tried in turn by the transport, until one succeeds.

This is not optimal so far, since each time through `RoundTrip` will try all the provided credentials from the start until one works, when it should "remember" which one worked for this registry+repo and reuse that if possible next time against that registry+repo.

This also doesn't support bearer auth yet, only basic auth, since bearer auth requires refreshing tokens before checking if they can be used. That should be doable though, it's just work.

This changes `DefaultKeychain` to provide creds for both `"registry.com"` and `"registry.com/repo"`, and for backward compatibility adds an extension interface `MultiAuthenticator` which is an `Authenticator` capable of returning multiple auths.

We may also want to extend this to `MultiKeychain` so we can allow a multikeychain to try creds from the Docker config and then fallback to a cred helper if that first request fails -- today, we'll just fail if that first matching cred is invalid.

cc @jonjohnsonjr for 👀 

https://github.com/google/go-containerregistry/issues/1329
https://github.com/google/go-containerregistry/issues/1431

@jwcesign

### DIFF
--- a/pkg/authn/multikeychain.go
+++ b/pkg/authn/multikeychain.go
@@ -28,14 +28,50 @@ func NewMultiKeychain(kcs ...Keychain) Keychain {
 
 // Resolve implements Keychain.
 func (mk *multiKeychain) Resolve(target Resource) (Authenticator, error) {
+	var auths []Authenticator
 	for _, kc := range mk.keychains {
 		auth, err := kc.Resolve(target)
 		if err != nil {
 			return nil, err
 		}
 		if auth != Anonymous {
-			return auth, nil
+			auths = append(auths, auth)
 		}
 	}
-	return Anonymous, nil
+	if len(auths) == 0 {
+		return Anonymous, nil
+	}
+	return multiAuthenticator{auths: auths}, nil
+}
+
+type multiAuthenticator struct {
+	auths []Authenticator
+}
+
+func (ma multiAuthenticator) Authorization() (*AuthConfig, error) {
+	auths, err := ma.Authorizations()
+	if err != nil {
+		return nil, err
+	}
+	return &auths[0], nil
+}
+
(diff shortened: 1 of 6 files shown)


jwcesign:
For the case I did before, `ResolveMany` need to be implemented here:
https://github.com/google/go-containerregistry/blob/87b3a7966622dbfdb8cbb9c4b5068884540d387e/pkg/authn/kubernetes/keychain.go#L180

And `ResolveMany` should be call from here, same level with Resolve:
https://github.com/google/go-containerregistry/blob/87b3a7966622dbfdb8cbb9c4b5068884540d387e/pkg/authn/multikeychain.go#L32

pkg/authn/multikeychain.go
```diff
@@ -28,14 +28,50 @@ func NewMultiKeychain(kcs ...Keychain) Keychain {
 
 // Resolve implements Keychain.
 func (mk *multiKeychain) Resolve(target Resource) (Authenticator, error) {
+	var auths []Authenticator
 	for _, kc := range mk.keychains {
 		auth, err := kc.Resolve(target)
```
jwcesign, line 33: I think here should call `ResolveMany`, Resolve only return the first secret matched, so if resolve only return one matched, use `Authenticator`

jwcesign:
cc @imjasonh

cniry:
Hi, it would be nice

imjasonh:
Closing in favor of https://github.com/google/go-containerregistry/pull/1446 or whatever comes next.

PR #1443, opened 2022-08-24
feat: Add krane to release archive

MisterMX:
This adds krane to the release artifacts of this repo so it is not necessary to run `go install` to use it.

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -15,8 +15,10 @@ builds:
   flags:
   - -trimpath
   ldflags:
-  - "-s -w -X github.com/google/go-containerregistry/cmd/crane/cmd.Version={{.Version}}"
-  - "-s -w -X github.com/google/go-containerregistry/pkg/v1/remote/transport.Version={{.Version}}"
+    - -s
+    - -w
+    - -X github.com/google/go-containerregistry/cmd/crane/cmd.Version={{.Version}}
+    - -X github.com/google/go-containerregistry/pkg/v1/remote/transport.Version={{.Version}}
   goarch:
     - amd64
     - arm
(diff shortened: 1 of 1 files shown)


.goreleaser.yml
```diff
@@ -55,6 +55,31 @@ builds:
...
+  env:
+  - CGO_ENABLED=0
+  main: ./main.go
+  dir: ./cmd/krane
+  binary: krane
+  flags:
+  - -trimpath
+  ldflags:
+  - "-s -w -X github.com/google/go-containerregistry/cmd/crane/cmd.Version={{.Version}}"
+  - "-s -w -X github.com/google/go-containerregistry/pkg/v1/remote/transport.Version={{.Version}}"
```
06kellyjac, line 68:
```suggestion
  - -s
  - -w
  - -X github.com/google/go-containerregistry/cmd/crane/cmd.Version={{.Version}}
  - -X github.com/google/go-containerregistry/pkg/v1/remote/transport.Version={{.Version}}
```

not sure it needs 2 sets of `-s -w`
could either use a list like suggested or squash all into 1 item

MisterMX: I just built it with your suggestions and it seems to work. I think we can go on with this.

.goreleaser.yml
```diff
@@ -55,6 +55,33 @@ builds:
...
+  main: ./main.go
+  dir: ./cmd/krane
+  binary: krane
+  flags:
+  - -trimpath
+  ldflags:
+    - -s
+    - -w
+    - -X github.com/google/go-containerregistry/cmd/crane/cmd.Version={{.Version}}
+    - -X github.com/google/go-containerregistry/pkg/v1/remote/transport.Version={{.Version}}
```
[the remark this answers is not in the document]
MisterMX: You are right. I updated the other artifacts as well.

imjasonh (approved):

PR #1444, opened 2022-08-29
deps: update goreleaser-action for bug

asraa:
Signed-off-by: Asra Ali <asraa@google.com>

This updates goreleaser-action to fix this bug, ensures that v3.1.0 is chosen: https://github.com/goreleaser/goreleaser-action/issues/368

Otherwise, the action does not populate the artifacts output, which would lead to provenance generation failure.

@laurentsimon

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -17,7 +17,7 @@ jobs:
         with:
           go-version: 1.17
           check-latest: true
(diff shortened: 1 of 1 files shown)


asraa:
This may not be needed depending on version resolution, so feel free to close. just FYI the action will fail unless 3.1.0+ is used

imjasonh (approved):

PR #1446, opened 2022-08-31
Support multiple auth for same registry/repo

jwcesign:
Signed-off-by: jiang wei <jwcesign@gmail.com>

This is an attempt to allow our authn and transport packages to cooperate and allow for keychains (incl DefaultKeychain) to provide multiple creds, which will be tried in turn by the transport, until one succeeds.

This is not optimal so far, since each time through RoundTrip will try all the provided credentials from the start until one works, when it should "remember" which one worked for this registry+repo and reuse that if possible next time against that registry+repo.

This also doesn't support bearer auth yet, only basic auth, since bearer auth requires refreshing tokens before checking if they can be used. That should be doable though, it's just work.

This changes DefaultKeychain to provide creds for both "registry.com" and "registry.com/repo", and for backward compatibility adds an extension interface MultiAuthenticator which is an Authenticator capable of returning multiple auths.

We may also want to extend this to MultiKeychain so we can allow a multikeychain to try creds from the Docker config and then fallback to a cred helper if that first request fails -- today, we'll just fail if that first matching cred is invalid.

cc @jonjohnsonjr, @imjasonh, @dprotaso  for 👀

https://github.com/google/go-containerregistry/issues/1329
https://github.com/google/go-containerregistry/issues/1431
https://github.com/knative/serving/issues/13126

This PR is based on the ideas of https://github.com/google/go-containerregistry/pull/1439

### DIFF
--- a/pkg/authn/keychain.go
+++ b/pkg/authn/keychain.go
@@ -45,6 +45,8 @@ type Keychain interface {
 }
 
 // KeychainMany is an interface for resolving an image reference to multiple credentials.
+// NOTE: This name might change before (or even after) we cut a release. Will be removed as soon
+// as we find a better name for it.
 type KeychainMany interface {
 	ResolveMany(Resource) ([]Authenticator, error)
 }
@@ -66,6 +68,7 @@ const (
 	DefaultAuthKey = "https://" + name.DefaultRegistry + "/v1/"
 )
 
+// Resolve implements Keychain.
 func (dk *defaultKeychain) Resolve(target Resource) (Authenticator, error) {
 	auths, err := dk.ResolveMany(target)
 	if err != nil {
@@ -75,7 +78,7 @@ func (dk *defaultKeychain) Resolve(target Resource) (Authenticator, error) {
 	return auths[0], nil
 }
 
-// ResolveMany implements MultiKeychain.
+// ResolveMany implements KeychainMany.
 func (dk *defaultKeychain) ResolveMany(target Resource) ([]Authenticator, error) {
 	dk.mu.Lock()
 	defer dk.mu.Unlock()
--- a/pkg/authn/keychain.go
+++ b/pkg/authn/keychain.go
@@ -44,6 +44,7 @@ type Keychain interface {
 	Resolve(Resource) (Authenticator, error)
 }
 
+// KeychainMany is an interface for resolving an image reference to multiple credentials.
 type KeychainMany interface {
 	ResolveMany(Resource) ([]Authenticator, error)
 }
--- a/pkg/authn/keychain.go
+++ b/pkg/authn/keychain.go
@@ -44,6 +44,10 @@ type Keychain interface {
 	Resolve(Resource) (Authenticator, error)
 }
 
+type KeychainMany interface {
+	ResolveMany(Resource) ([]Authenticator, error)
+}
+
 // defaultKeychain implements Keychain with the semantics of the standard Docker
 // credential keychain.
 type defaultKeychain struct {
@@ -61,8 +65,17 @@ const (
 	DefaultAuthKey = "https://" + name.DefaultRegistry + "/v1/"
 )
 
-// Resolve implements Keychain.
 func (dk *defaultKeychain) Resolve(target Resource) (Authenticator, error) {
+	auths, err := dk.ResolveMany(target)
+	if err != nil {
+		return nil, err
+	}
+
+	return auths[0], nil
+}
+
+// ResolveMany implements MultiKeychain.
+func (dk *defaultKeychain) ResolveMany(target Resource) ([]Authenticator, error) {
 	dk.mu.Lock()
 	defer dk.mu.Unlock()
 
@@ -98,7 +111,7 @@ func (dk *defaultKeychain) Resolve(target Resource) (Authenticator, error) {
 	} else {
 		f, err := os.Open(filepath.Join(os.Getenv("XDG_RUNTIME_DIR"), "containers/auth.json"))
 		if err != nil {
-			return Anonymous, nil
+			return []Authenticator{Anonymous}, nil
 		}
 		defer f.Close()
 		cf, err = config.LoadFromReader(f)
@@ -110,7 +123,9 @@ func (dk *defaultKeychain) Resolve(target Resource) (Authenticator, error) {
 	// See:
 	// https://github.com/google/ko/issues/90
 	// https://github.com/moby/moby/blob/fc01c2b481097a6057bec3cd1ab2d7b4488c50c4/registry/config.go#L397-L404
-	var cfg, empty types.AuthConfig
+	var empty types.AuthConfig
+
+	var cfgs []types.AuthConfig
 	for _, key := range []string{
 		target.String(),
 		target.RegistryStr(),
@@ -119,25 +134,30 @@ func (dk *defaultKeychain) Resolve(target Resource) (Authenticator, error) {
 			key = DefaultAuthKey
 		}
 
-		cfg, err = cf.GetAuthConfig(key)
+		cfg, err := cf.GetAuthConfig(key)
 		if err != nil {
 			return nil, err
 		}
 		if cfg != empty {
-			break
+			cfgs = append(cfgs, cfg)
 		}
 	}
-	if cfg == empty {
-		return Anonymous, nil
+
+	if len(cfgs) == 0 {
+		return []Authenticator{Anonymous}, nil
 	}
 
-	return FromConfig(AuthConfig{
-		Username:      cfg.Username,
-		Password:      cfg.Password,
-		Auth:          cfg.Auth,
-		IdentityToken: cfg.IdentityToken,
-		RegistryToken: cfg.RegistryToken,
-	}), nil
+	auths := make([]Authenticator, 0, len(cfgs))
+	for _, cfg := range cfgs {
+		auths = append(auths, FromConfig(AuthConfig{
+			Username:      cfg.Username,
+			Password:      cfg.Password,
+			Auth:          cfg.Auth,
+			IdentityToken: cfg.IdentityToken,
+			RegistryToken: cfg.RegistryToken,
+		}))
+	}
+	return auths, nil
 }
 
 // fileExists returns true if the given path exists and is not a directory.
--- a/pkg/authn/kubernetes/keychain.go
+++ b/pkg/authn/kubernetes/keychain.go
@@ -177,6 +177,7 @@ type keyring struct {
 	creds map[string][]authn.AuthConfig
 }
 
+// Resolve implements Keychain.
 func (keyring *keyring) Resolve(target authn.Resource) (authn.Authenticator, error) {
 	auths, err := keyring.ResolveMany(target)
 	if err != nil {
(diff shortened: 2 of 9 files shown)


pkg/authn/keychain.go
```diff
@@ -44,6 +44,10 @@ type Keychain interface {
 	Resolve(Resource) (Authenticator, error)
 }
 
+type KeychainMany interface {
```
imjasonh, line 47: `MultiKeychain` is probably a better name, but then we have some conflict with the method `authn.NewMultiKeychain`, which does a different thing. 😕

pkg/authn/kubernetes/keychain.go
```diff
@@ -270,12 +279,14 @@ func urlsMatch(globURL *url.URL, targetURL *url.URL) (bool, error) {
...
 
-func toAuthenticator(configs []authn.AuthConfig) (authn.Authenticator, error) {
-	cfg := configs[0]
-
-	if cfg.Auth != "" {
-		cfg.Auth = ""
+func toAuthenticators(configs []authn.AuthConfig) []authn.Authenticator {
+	auths := make([]authn.Authenticator, 0, len(configs))
+	for _, cfg := range configs {
+		if cfg.Auth != "" {
```
imjasonh, line 285: Should this also do the comparison against `empty` that we do in `keychain.go`? It may even be worth having an `isEmpty` helper (unexported) on `AuthConfig` if we find ourselves doing the comparison a lot.

pkg/v1/remote/transport/basic.go
```diff
@@ -57,6 +71,19 @@ func (bt *basicTransport) RoundTrip(in *http.Request) (*http.Response, error) {
 				in.Header.Set("Authorization", hdr)
 			}
 		}
+		resp, err := bt.inner.RoundTrip(in)
+		if err != nil {
+			return nil, err
+		}
+
+		if resp.StatusCode == http.StatusUnauthorized || resp.StatusCode == http.StatusForbidden {
+			if last {
```
imjasonh, line 80: Could we define `last` nearer to where it's used here?

pkg/v1/remote/transport/basic.go
```diff
@@ -57,6 +71,19 @@ func (bt *basicTransport) RoundTrip(in *http.Request) (*http.Response, error) {
...
+				return resp, nil
+			}
+			continue
```
imjasonh, line 83: We probably want to debug-log each failure we get, in case it's helpful. 

https://pkg.go.dev/github.com/google/go-containerregistry/pkg/logs

pkg/v1/remote/transport/bearer.go
```diff
@@ -67,24 +69,48 @@ func stringSet(ss []string) map[string]struct{} {
...
 			in.Header.Set("Authorization", hdr)
 		}
 		return bt.inner.RoundTrip(in)
 	}
 
-	res, err := sendRequest()
-	if err != nil {
-		return nil, err
+	var res *http.Response
+	var err error
```
imjasonh, line 86: I think we can define this inside the loop, and change `return res, err` -> `return res, nil` on line 109. We'll either get an `err` and `return nil, err`, or we'll get a good `res` and `return res, nil`, and restructuring in this way makes this clearer at a glance.

imjasonh (commented):
This looks really promising so far!

I'd definitely like @jonjohnsonjr to take a look when he gets a chance, he usually catches things I don't, and this is a fiddly enough area I'd like to get another pair of eyes on it. 👀

pkg/authn/keychain.go
jwcesign: Haven't got better idea

pkg/authn/kubernetes/keychain.go
jwcesign: the checking is different, in `pkg/authn/keychain.go` is `if cfg != empty` but here only check `cfg.Auth`, maybe here has some reasons?

pkg/v1/remote/transport/basic.go
jwcesign: fixed

pkg/v1/remote/transport/bearer.go
jwcesign: fixed

pkg/v1/remote/transport/basic.go
jwcesign: BTW, should we rember which `bearerTokenList` work and use a `bearerTransport ` struct member to record it?

pkg/authn/kubernetes/keychain.go
```diff
@@ -177,7 +177,18 @@ type keyring struct {
...
 func (keyring *keyring) Resolve(target authn.Resource) (authn.Authenticator, error) {
+	auths, err := keyring.ResolveMany(target)
+	if err != nil {
+		return nil, err
+	}
+
+	return auths[0], nil
+}
+
+// ResolveMany implements MultiKeychain.
```
imjasonh, line 190:
```suggestion
// ResolveMany implements KeychainMany.
```

pkg/authn/multikeychain_test.go
```diff
@@ -70,16 +70,33 @@ func TestMultiKeychain(t *testing.T) {
...
 		want: Anonymous,
+	}, {
+		name: "match multiple keychains, take the first",
+		reg:  regOne,
+		kc: NewMultiKeychain(
+			fixedKeychain{regOne: one},
+			fixedKeychain{regOne: three, regTwo: two},
+			fixedKeychain{regOne: two},
+		),
+		want: one, // TODO: test that this actually has one,three,two internall and can return them all if asked.
```
imjasonh, line 81: What more would it take to do this TODO?

pkg/v1/remote/transport/bearer_test.go
```diff
@@ -192,12 +192,12 @@ func TestBearerTransportTokenRefresh(t *testing.T) {
 	if res.StatusCode != http.StatusOK {
 		t.Errorf("client.Get final StatusCode got %v, want: %v", res.StatusCode, http.StatusOK)
 	}
-	if got, want := transport.bearer.RegistryToken, refreshedToken; got != want {
+	if got, want := transport.bearer[0].RegistryToken, refreshedToken; got != want {
```
imjasonh, line 195: Could we have a test for a multi-keychain that has some good and some bad (and some expired and needing refreshing) tokens, like we have in base_test.go?

jwcesign: fixed

pkg/authn/multikeychain_test.go
jwcesign: deleted, i just copy from what u right, I think it's enough to add ut for 
> Could we have a test for a multi-keychain that has some good and some bad (and some expired and needing refreshing) tokens, like we have in base_test.go?

pkg/authn/kubernetes/keychain.go
jwcesign: fixed

pkg/authn/keychain.go
```diff
@@ -63,6 +68,16 @@ const (
...
 func (dk *defaultKeychain) Resolve(target Resource) (Authenticator, error) {
+	auths, err := dk.ResolveMany(target)
+	if err != nil {
+		return nil, err
+	}
+
+	return auths[0], nil
+}
+
+// ResolveMany implements MultiKeychain.
```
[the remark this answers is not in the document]
jwcesign: fixed

pkg/authn/keychain.go
```diff
@@ -44,6 +44,11 @@ type Keychain interface {
 	Resolve(Resource) (Authenticator, error)
 }
 
+// KeychainMany is an interface for resolving an image reference to multiple credentials.
+type KeychainMany interface {
```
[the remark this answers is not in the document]
jwcesign: done

pkg/v1/remote/transport/bearer_test.go
```diff
@@ -559,3 +560,108 @@ func TestInsufficientScope(t *testing.T) {
 		t.Error("didn't refresh insufficient scope")
 	}
 }
+
+func TestMultiAuthBrearRequest(t *testing.T) {
```
imjasonh, line 564:
```suggestion
func TestMultiAuthBearerRequest(t *testing.T) {
```

imjasonh (commented):
Thanks! One last typo, otherwise LGTM. 👍

pkg/v1/remote/transport/bearer_test.go
jwcesign: fixed

imjasonh (approved):

pkg/v1/remote/transport/bearer.go
```diff
@@ -67,24 +67,54 @@ func stringSet(ss []string) map[string]struct{} {
...
+		if refreshed {
+			res, err = sendRequest(bt.bearer[index].RegistryToken)
+			if err != nil {
+				return nil, err
+			}
+		}
+
+		if res.StatusCode == http.StatusOK {
+			return res, nil
+		}
```
jonjohnsonjr, line 110: I don't think this is quite right. We want to exit the loop early if the request was successful, but "successful" isn't necessarily a 200, right? We only want to keep going through the loop when the response still is a 401 with challenges.

jonjohnsonjr (changes requested):

pkg/v1/remote/transport/bearer.go
jwcesign: Fixed

jwcesign: cc @jonjohnsonjr

jwcesign:
Is this any more comments? @jonjohnsonjr

jwcesign:
cc @jonjohnsonjr Any new updates about this? like a new approach？

jonjohnsonjr:
I've been mulling over how to proceed, sorry for taking so long.

In general, I like the functionality this enables, but I don't want to commit to these (public) interfaces quite yet. The main issue I have is that this eagerly fetches credentials. I'd like to do this lazily so that we don't pay the full cost of fetching creds with each request. Making this lazy requires having some callback mechanism (rather than just returning a list) so that our transport can call it repeatedly to get the next credential to try.

I know that's a big change to this PR, so it's kind of annoying to make (again sorry), but I think it will simplify the code a lot.

tl;dr, we currently traverse the keychain tree until we find a matching credential, then use that forever. Your PR does a full tree traversal and collects all matching credentials, then tries to use them all. What I want is a traversal that returns the first matching credential, but maintains some state about the traversal so we can ask for the _next_ matching credential (rather than doing the full traversal up front).

That probably will look something like returning this?

```go
package authn

type Iterator interface {
  Next() (Authenticator, error)
}
```

Jotting down some more thoughts as we think about this:

1. Sometimes you need to refresh because your credentials are expired.
1. Sometimes you need to refresh because you asked for insufficient (or incorrect) scopes to perform the operation.
1. Sometimes you don't have sufficient permissions to perform the operation at all.

The third case here is what we're trying to solve by allowing multiple credentials and iterating over them. I'm not 100% sure how to differentiate between this case and the others...

jwcesign:
cc @jonjohnsonjr any new process about this?

jwcesign:
cc @jonjohnsonjr

PR #1451, opened 2022-09-22
Fix incorrect casing for version argument

dips-ogl:
If the instructions are copied directly and you use Latest as version, the script will fail with this error message:
`ERROR: 'Latest' not found amongst apt-cache madison results`

Fixed the casing.

### DIFF
--- a/cmd/crane/README.md
+++ b/cmd/crane/README.md
@@ -13,7 +13,7 @@ A collection of useful things you can do with `crane` is [here](recipes.md).
 Download [latest release](https://github.com/google/go-containerregistry/releases/latest):
 
 ```
-$ VERSION=TODO   # Latest, or other
+$ VERSION=TODO   # latest, or other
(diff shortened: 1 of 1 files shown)


cmd/crane/README.md
```diff
@@ -13,7 +13,7 @@ A collection of useful things you can do with `crane` is [here](recipes.md).
 Download [latest release](https://github.com/google/go-containerregistry/releases/latest):
 
 ```
-$ VERSION=TODO   # Latest, or other
+$ VERSION=TODO   # latest, or other
```
imjasonh, line 16: This is actually correct. It's intended to be a sentence, essentially "Use the latest release, or another one". No relation to the `:latest` tagging convention or Go's `@latest` convention.

If `VERSION=latest` (or `VERSION=Latest` for that matter), the `curl` command below will fail since neither of those are release names.

PR #1454, opened 2022-09-28
feat: debug-base oci image with apko

### LINKED ISSUE #1356: question: Include jq in debug container?
While I love minimal containers, ones that leave the package manager off make it difficult to work inside the container.

The GitLab example (thank you for that!) is an example where the job run inside the container - and with the package manager not present, it is challenging to add jq.

Since all the output of crane is json, would it be possible to install jq on the debug tag - or at least the os package manager so one can pull on their desired additional utils?

Thanks for any thoughts!

developer-guy:

Fixes #1356 

Successful action results in the forked repo 👇

_https://github.com/developer-guy/go-containerregistry/actions/runs/3846827929/jobs/6552572226_

/cc @imjasonh @cpanato @puerco

### DIFF
--- a/.github/workflows/release-debug-base.yaml
+++ b/.github/workflows/release-debug-base.yaml
@@ -1,0 +1,28 @@
+name: Create Debug Base Image Release
+
+on:
+  workflow_dispatch:
+  push:
+    branches: [ 'main']
+jobs:
+  build:
+    name: Release Debug Base OCI image
+    runs-on: ubuntu-latest
+
+    # https://docs.github.com/en/actions/reference/authentication-in-a-workflow
+    permissions:
+      id-token: write
+      packages: write
+      contents: read
+
+    steps:
+      - uses: actions/checkout@v3
+      - uses: imjasonh/setup-crane@v0.1
+      - id: apko
+        uses: chainguard-images/actions/apko-snapshot@main
+        with:
+          config: apko.yaml
+          image_refs: apko.images
+          base-tag: ghcr.io/${{ github.repository }}/debug-base
+          target-tag: latest
+          
--- a/.ko/debug/.ko.yaml
+++ b/.ko/debug/.ko.yaml
@@ -1,1 +1,1 @@
-defaultBaseImage: gcr.io/distroless/base:debug
+defaultBaseImage: ghcr.io/google/go-containerregistry/debug-base:latest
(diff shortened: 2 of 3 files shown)


.github/workflows/release-debug-base.yaml
```diff
@@ -0,0 +1,38 @@
...
+
+      - id: apko
+        uses: chainguard-images/actions/apko-snapshot@main
+        with:
+          config: apko.yaml
+          image_refs: apko.images
+          base-tag: ghcr.io/${{ github.repository }}/debug-base
+          target-tag: latest
+
+      - name: Emit Image Refs output
```
imjasonh, line 31: I think we could just skip this step. If I need to see the digests of the base image I can just use crane.

.github/workflows/release-debug-base.yaml
```diff
@@ -0,0 +1,38 @@
+name: Create Debug Base Image Release
+
+on:
+  workflow_dispatch:
+  push:
+    paths:
```
imjasonh, line 6: So I think there's a slight race here. If I push a change that touches apko.yaml and some Go code, the workflow to build the base image and the workflow to build the image on top of it will start separately at the same time. The base image workflow hopefully finishes first, but there's no guarantee. The result would be an inconsistently based crane debug image, which might be confusing and hard to track down.

Can this apko step just happen directly in the workflow to rebuild the crane debug image? It doesn't take long, and it should produce the same digest if packages haven't changed, so I don't mind doing it even if apko.yaml hasn't changed.

Actually that's another reason -- if the `busybox` package has changed in alpine but the apko.yaml hasn't, I'd still like to rebuild the image, so let's run this step as often as we can to try to pick up those package updates.

developer-guy: totally agree with you, I made some changes according to your feedbacks, thanks for bringing them to my attention.

.github/workflows/release-debug-base.yaml
```diff
@@ -0,0 +1,43 @@
+name: Create Debug Base Image Release
+
+on:
+  workflow_dispatch:
+  push:
+    branches: [ 'main', 'feature/**']
```
imjasonh, line 6:
```suggestion
    branches: ['main']
```

We don't really use feature branches, and if we did we might not want the debug base image rebuilt on every push to them, in case the feature changed the debug base image config in some way.

developer-guy: yeah, I forgot to remove it, I just set it for testing purposes for my fork, removed, thank you.

developer-guy:
kindly ping @imjasonh

Issue #1434, opened 2022-08-16
FR: Support OCI reference types


developer-guy:
Once I run `crane copy,` it'll (at the time of writing this) copy only the image itself, but the image may have other materials attached to it such as SBOMs, attestations, signatures, etc. `cosign copy` command supports that, but in crane, we do not have support for this, so, it'd be better copying these materials (if any) with the image itself optionally.

cc: @dentrax @imjasonh

imjasonh:
To put this in crane we'll need to rely on the new APIs and conventions proposed by the OCI reference types working group:

- proposed image-spec changes: https://github.com/opencontainers/image-spec/pull/934
- proposed distribution-spec changes: https://github.com/opencontainers/distribution-spec/pull/335

We expect these spec changes to be merged in soon 🤞, then for these changes to start trickling out to registry implementations over the next few years. The fallback behavior described in those changes relies on tag naming conventions, for compatibility with registries that haven't adopted the new specs yet.

@jdolitsky has been hacking on changes to go-containerregistry and crane to add support for these APIs here: https://github.com/google/go-containerregistry/compare/main...oci-playground:main

When these APIs and conventions are adopted and the changes above are merged into ggcr, the idea is you'll be able to `crane cp -R <src> <dst>` (`-R` for `--recursive`) and not only copy the source image, but all artifacts that reference it, SBOMs, signatures, attestations, cat pictures, whatever.

developer-guy:
It looks like these PRs are merged, so shall we start working on this issue?

imjasonh:
> It looks like these PRs are merged, so shall we start working on this issue?

See https://github.com/google/go-containerregistry/pull/1455

PR #1455, opened 2022-09-29
WIP: add support for new Referrers API

imjasonh:
Supercedes https://github.com/google/go-containerregistry/pull/1425 since I just wasn't feeling that PR, and I think enough changed in the acceptance of the spec changes that I wanted to start over.

API-wise I think this is pretty good, both in terms of `remote.Referrers(name.Digest)` and `img.Referrers()` and the CLI surface `crane attach <ref>` and `crane referrers <ref> -f foo.txt -t sbom`.

Everything should support fetching referrers for a manifest-by-digest that doesn't exist, and updating the manifest(-that-might-not-exist) when a manifest specifies a `subject`, going through the fallback tag flow and updating the index manifest, while deduping and sorting.

As with #1425 it needs tests, which will need pkg/registry support. A `Walk` method would be nice. Support for `ArtifactManifest` is commented out but mostly would be easy to add.

cc @jdolitsky

Progress toward https://github.com/google/go-containerregistry/issues/1434


jonjohnsonjr:
https://github.com/google/go-containerregistry/pull/1546

PR #1458, opened 2022-10-03
check if tar is compressed on push

pratikkaneria:
Fix: Push compressed (tar.gz) image to docker registry using crane

### DIFF
--- a/pkg/v1/tarball/image.go
+++ b/pkg/v1/tarball/image.go
@@ -223,7 +223,16 @@ func extractFileFromTar(opener Opener, filePath string) (io.ReadCloser, error) {
 			f.Close()
 		}
 	}()
-
+	iscompressed, err := gzip.Is(f)
+	if err != nil {
+		return nil, err
+	}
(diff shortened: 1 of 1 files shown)


imjasonh (commented):
This looks good so far! Thanks for the contribution.

Can you add a test for this? Either in a go test or an e2e test that uses crane to append a tar.gz and verifies that it works and produces a valid image.

pkg/v1/tarball/image.go
```diff
@@ -223,7 +223,20 @@ func extractFileFromTar(opener Opener, filePath string) (io.ReadCloser, error) {
 			f.Close()
 		}
 	}()
-
+	iscompressed, err := gzip.Is(f)
```
jonjohnsonjr, line 226: This will add an extra read of the file -- can you use `gzip.Peek` for this instead?

pratikkaneria: I will look into it, sorry for the delay as i was away from work

PR #1459, opened 2022-10-04
feat: add crane auth logout command

pbusko:
Signed-off-by: Pavel Busko <pavel.busko@sap.com>


cmd/crane/cmd/auth.go
```diff
@@ -203,3 +222,22 @@ func login(opts loginOptions) error {
 	log.Printf("logged in via %s", cf.Filename)
 	return nil
 }
+
+func logout(server string) error {
+	cf, err := config.Load(os.Getenv("DOCKER_CONFIG"))
```
imjasonh, line 227: This looks good! I think we might also want to look for podman-configured auth, which is configured in a different path, which `authn.DefaultKeychain` will also check for if `$DOCKER_CONFIG/config.json` isn't found.

PR #1462, opened 2022-10-13
Make ErrBadName checkable via errors.Is()

cjellick:
The function IsErrBadName says it is deprecated and directs the user to
use errors.Is(), but that will never return true because this error is
custom based on the tag value.

This fixes that problem by implementing an Is() function on ErrBadName
so that errors.Is() can properly identify the error as an ErrBadName.
Usage can now be: errors.Is(err, &ErrBadName{})

This exact problem was actually discussed when the deprecated warning was introduced: https://github.com/google/go-containerregistry/pull/1164/files#r742970357

### DIFF
--- a/pkg/name/errors_test.go
+++ b/pkg/name/errors_test.go
@@ -31,4 +31,8 @@ func TestBadName(t *testing.T) {
 	if err.Error() != "could not parse reference: @@" {
 		t.Errorf("Unexpected string: %v", err)
 	}
+	//var cerr *ErrBadName
+	if !errors.Is(err, &ErrBadName{}) {
+		t.Errorf("Not an ErrBadName using errors.Is: %v", err)
+	}
 }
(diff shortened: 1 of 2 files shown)


cjellick:
The linter didn't like the type assertion I used so I had to change impl to errors.As

pkg/name/errors_test.go
```diff
@@ -31,4 +31,8 @@ func TestBadName(t *testing.T) {
 	if err.Error() != "could not parse reference: @@" {
 		t.Errorf("Unexpected string: %v", err)
 	}
+	//var cerr *ErrBadName
```
imjasonh, line 34:
```suggestion
```

imjasonh (commented):
Nice! Thanks for this, just one comment.

imjasonh (approved):

Issue #1089, opened 2021-07-28
Container Registry: allow Path in URI as per RFC 3986


scodeman:
`error building image: registries must be valid RFC 3986 URI authorities` issue reported at https://github.com/GoogleContainerTools/kaniko/issues/1618  seem to be related to the fact that go-containerregistry project always assume a registry is made of Host part only in the URI, but there could be a Path section to the registry URI and it should not be prevented to be specified as a registry source 

https://github.com/google/go-containerregistry/blob/596751a1747114c7f502f8a99a55891865aa2d03/pkg/name/registry.go#L100

it seems the condition `url.Host != name` is not valid for checking RFC 3986 compliance.
Is there another reason for it ?
Could it be remove so to fix issues related to it

For example in my case I use a subrepo as a `docker.io` registry mirror and have to use a registry with Path, ex: `artifactory.mycompany.io/docker-io-mirror`

imjasonh:
I think the problem is that kaniko only supports mirror registries specified at the root registry, and not root+path. They call `name.NewRegistry` [here](https://github.com/GoogleContainerTools/kaniko/blob/1d9bc175c2070680c0d2d5b1d7531ffe53d713f7/pkg/image/remote/remote.go#L60-L64) when they could call [`name.NewRepository`](https://pkg.go.dev/github.com/google/go-containerregistry/pkg/name#NewRepository), which supports parsing not only the registry component (Host only, e.g., `my.co`) but also the repository component including path (e.g., `my.co/cache`).

Briefly browsing the kaniko code I think this should be relatively easy to fix on their end.

imjasonh:
Let's find out whether it's actually that easy to fix 😄 https://github.com/GoogleContainerTools/kaniko/pull/1707

(Closing this issue so we can continue discussion on that PR and issue)

thw0rted:
Is this actually fixed?  `echo $PASS | crane auth login -u $USER registry.example.com/reponame --password-stdin` still produces

> Error: registries must be valid RFC 3986 URI authorities: registry.example.com/reponame

while a `docker login` with the same arguments works fine.

PR #1465, opened 2022-10-18
Allow paths in registries

### LINKED ISSUE #1464: ggcr: Allow path in registry URIs
### Describe the bug

This was already raised in #1089, but it seems that working with registries that contain a path part before the repositories' names isn't well supported. This is because it's assumed (and enforced) that the registry's name is only the `host[:port]` part of the URI but some registries, for example JFrog Artifactory, provide the Docker V2 endpoint at a certain path.

The suggested solution in #1089 doesn't seem to work when dealing with the `remote` package.

### To Reproduce

```go
reference, _ := name.ParseReference("example.jfrog.io/artifactory/api/docker/my-repository/my-image:latest")
_, err := remote.Image(reference)
fmt.Println(err.Error())
```

Would give:
```
GET https://example.jfrog.io/v2/artifactory/api/docker/my-repository/my-image/manifests/latest: : Not Found
```

But if we move the `/v2` to the correct place it'll work (commands and output shortened for brevity):
```sh
$ curl https://example.jfrog.io/artifactory/api/docker/my-repository/v2/_catalog
{
  "repositories" : [ "my-image" ]
}
$ curl -v https://example.jfrog.io/artifactory/api/docker/my-repository/v2/my-image/manifests/latest
...
< HTTP/1.1 200 OK
...
```

### Expected behavior

First, the ability to construct a `Registry` with a path part should be provided. Second, that path should be honored when constructing the resources' URLs.

### Additional context

- Version: All up to the latest (`v0.11.0` and current `main`).
- Registry used: JFrog Artifactory, but it's not unlikely that others exist.

nivbend:
This fixes #1464.

I've opted to allow manually constructing the reference from a `Registry` by adding the `Repository` method. This of course assumes the user of the library knows about this case and how to handle it.

Another way of going about it would be to split the entire path and iterate the placement of `/v2` until we hit a `200 OK`, but I don't think a default behavior of "path knocking" on registries would be desired. We can consider adding this as a utility function perhaps.

Notice that because `Registry` is embedded `Repository`, this also added a `Repository` to that struct.

So we allow two ways of constructing a reference: One is through `registry.Repository("name").Tag("latest")`, and the other is by replacing the `Repository` as suggested in #1089.

The second part here is adding `SplitURL` to construct resources' URLs properly (in `fetcher.url`) and matching against the correct hostname in `matchesHost`.

I haven't tested all the cases here, only the one I was interested in, so it's quite possible that more changes are required for other types of authentication or different areas of the libraries.

### DIFF
--- a/pkg/v1/remote/descriptor_test.go
+++ b/pkg/v1/remote/descriptor_test.go
@@ -26,6 +26,7 @@ import (
 	"testing"
 
 	"github.com/google/go-cmp/cmp"
+
 	v1 "github.com/google/go-containerregistry/pkg/v1"
 	"github.com/google/go-containerregistry/pkg/v1/types"
 )
@@ -70,7 +71,7 @@ func TestGetSchema1(t *testing.T) {
 
 	want := `unsupported MediaType: "application/vnd.docker.distribution.manifest.v1+prettyjws", see https://github.com/google/go-containerregistry/issues/377`
 	// Should fail based on media type.
-	if _, err := desc.Image(); err != nil {
+	if _, err = desc.Image(); err != nil {
 		if errors.Is(err, &ErrSchema1{}) {
 			t.Errorf("Image() = %v, expected remote.ErrSchema1", err)
 		}
@@ -82,7 +83,7 @@ func TestGetSchema1(t *testing.T) {
 	}
 
 	// Should fail based on media type.
-	if _, err := desc.ImageIndex(); err != nil {
+	if _, err = desc.ImageIndex(); err != nil {
 		var s1err ErrSchema1
 		if errors.Is(err, &s1err) {
 			t.Errorf("ImageImage() = %v, expected remote.ErrSchema1", err)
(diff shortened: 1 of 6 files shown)


pkg/v1/remote/descriptor_test.go
```diff
@@ -70,7 +71,7 @@ func TestGetSchema1(t *testing.T) {
 
 	want := `unsupported MediaType: "application/vnd.docker.distribution.manifest.v1+prettyjws", see https://github.com/google/go-containerregistry/issues/377`
 	// Should fail based on media type.
-	if _, err := desc.Image(); err != nil {
+	if _, err = desc.Image(); err != nil {
```
imjasonh, line 74: nit: These `:=` changes are not necessary.

imjasonh (commented):
> Another way of going about it would be to split the entire path and iterate the placement of /v2 until we hit a 200 OK, but I don't think a default behavior of "path knocking" on registries would be desired. We can consider adding this as a utility function perhaps.

Yeah I agree, I don't think we want to do that.

If we proceed this route I'd like to make sure our doc comments are really thorough, since it will start to get confusing that `name.NewRegistry` can take path elements, and have further `.Repository` elements added on top.

nivbend:
Seems like we can work around this on the caller code (see discussion on #1464), closing. Thanks for your time anyhow and sorry to bother.

PR #1469, opened 2022-10-20
split pkg/v1/daemon out into its own module

imjasonh:
Draft mode since I want to think about it a little more and see how it works for clients that upgrade into the newly split module.


jonjohnsonjr:
Want to rebase?

jonjohnsonjr (approved):

imjasonh:
This doesn't remove `docker/docker` as a dep, it just moves it to an indirect one.

This is because we still depend on docker/cli for config parsing in [`cmd/crane/cmd/auth.go`](https://github.com/google/go-containerregistry/blob/07c767c707c69fa25de1147a588b5560a260a62e/cmd/crane/cmd/auth.go#L26) (and `root.go`) and [`pkg/authn/keychain.go`](https://github.com/google/go-containerregistry/blob/07c767c707c69fa25de1147a588b5560a260a62e/pkg/authn/keychain.go#L23). If we replaced that config parsing with our own (possibly just a fork) we might be able to fully jettison `docker/cli` as a dep.

We also depend on `docker/distribution` mainly for authchallenge parsing in `pkg/remote/transport`, but that's a separate module with only a couple small deps (just opencontainers/image-spec and go-digest), so that's not so bad to keep.

PR #1473, opened 2022-10-23
crane: add catalog argument use annotation

brandtkeller:
Following [Cobra Recommended Syntax](https://github.com/spf13/cobra/blob/main/command.go#L50) for "Use", I added the required argument annotation to the command.

While the registry argument may be assumed, I believe this improves the user experience by explicitly defining required arguments. Runtime without a registry `Error: accepts 1 arg(s), received 0` and use of the `--help` flag does not define what the expected argument should be.

### DIFF
--- a/cmd/crane/cmd/catalog.go
+++ b/cmd/crane/cmd/catalog.go
@@ -24,7 +24,7 @@ import (
 // NewCmdCatalog creates a new cobra.Command for the repos subcommand.
 func NewCmdCatalog(options *[]crane.Option) *cobra.Command {
 	return &cobra.Command{
-		Use:   "catalog",
+		Use:   "catalog {REGISTRY}",
 		Short: "List the repos in a registry",
 		Args:  cobra.ExactArgs(1),
 		RunE: func(_ *cobra.Command, args []string) error {
--- a/cmd/crane/cmd/catalog.go
+++ b/cmd/crane/cmd/catalog.go
@@ -16,17 +16,31 @@ package cmd
 
 import (
 	"fmt"
+	"os"
+	"strings"
 
 	"github.com/google/go-containerregistry/pkg/crane"
 	"github.com/spf13/cobra"
 )
 
 // NewCmdCatalog creates a new cobra.Command for the repos subcommand.
-func NewCmdCatalog(options *[]crane.Option) *cobra.Command {
+func NewCmdCatalog(options *[]crane.Option, argv ...string) *cobra.Command {
+
+	if len(argv) == 0 {
+		argv = []string{os.Args[0]}
+	}
+
+	baseCmd := strings.Join(argv, " ")
+	eg := fmt.Sprintf(`  # list the repos for reg.example.com
+  $ echo "reg.example.com" | %s catalog
+  # or
+  $ %s catalog reg.example.com`, baseCmd, baseCmd)
+
 	return &cobra.Command{
-		Use:   "catalog {REGISTRY}",
-		Short: "List the repos in a registry",
-		Args:  cobra.ExactArgs(1),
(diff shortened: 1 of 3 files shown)


brandtkeller:
also considered using the same language/syntax as [auth get](https://github.com/google/go-containerregistry/blob/main/cmd/crane/cmd/auth.go#L80).

cmd/crane/cmd/catalog.go
```diff
@@ -24,7 +24,7 @@ import (
 // NewCmdCatalog creates a new cobra.Command for the repos subcommand.
 func NewCmdCatalog(options *[]crane.Option) *cobra.Command {
 	return &cobra.Command{
-		Use:   "catalog",
+		Use:   "catalog {REGISTRY}",
```
imjasonh, line 27:
```suggestion
		Use:   "catalog [REGISTRY]",
```

Yeah let's use the same form we use elsewhere.

It might also help to add a usage section with some examples, so folks know what kinds of values to expect `[REGISTRY]` to mean.

brandtkeller: Updated to square brackets & included a standard example using some existing patterns.
```
crane catalog --help
List the repos in a registry

Usage:
  crane catalog [REGISTRY] [flags]

Examples:
  # list the repos for reg.example.com
  $ echo "reg.example.com" | crane catalog
  # or
  $ crane catalog reg.example.com

Flags:
  -h, --help   help for catalog

Global Flags:
      --allow-nondistributable-artifacts   Allow pushing non-distributable (foreign) layers
      --insecure                           Allow image references to be fetched without TLS
      --platform platform                  Specifies the platform in the form os/arch[/variant][:osversion] (e.g. linux/amd64). (default all)
  -v, --verbose                            Enable debug logs
```

cmd/crane/cmd/catalog.go
```diff
@@ -16,17 +16,31 @@ package cmd
...
-func NewCmdCatalog(options *[]crane.Option) *cobra.Command {
+func NewCmdCatalog(options *[]crane.Option, argv ...string) *cobra.Command {
+
+	if len(argv) == 0 {
+		argv = []string{os.Args[0]}
+	}
+
+	baseCmd := strings.Join(argv, " ")
+	eg := fmt.Sprintf(`  # list the repos for reg.example.com
+  $ echo "reg.example.com" | %s catalog
```
imjasonh, line 35: This doesn't work today, the repo name has to be a positional arg.

cmd/crane/cmd/catalog.go
```diff
@@ -16,17 +16,31 @@ package cmd
...
+  # or
+  $ %s catalog reg.example.com`, baseCmd, baseCmd)
```
imjasonh, line 37:
```suggestion
  $ %s catalog reg.example.com`, baseCmd)
```

imjasonh (approved):

imjasonh (approved):

imjasonh:
```
-  $ /tmp/go-build1373274557/b001/exe/main catalog reg.example.com
+  $ crane catalog reg.example.com
```

lolwut

How does this work for [`crane auth login`](https://github.com/google/go-containerregistry/blob/main/cmd/crane/doc/crane_auth_login.md) but not here 🤔

jonjohnsonjr:
Apologies in advance: https://github.com/google/go-containerregistry/blob/426de7d42a8710cc65106984022f46311710e1bc/cmd/crane/cmd/root.go

imjasonh (approved):

brandtkeller:
Are there any additional steps that I need to make prior to this being clear for merge?

imjasonh (approved):

PR #1474, opened 2022-10-23
Make checkRegistry and checkRepository functions public

hectorj2f:
Signed-off-by: Hector Fernandez <hector@chainguard.dev>

This PR aims to make these check functions public, so these validations can be easily reused from other packages/projects.

cc @mattmoor

### DIFF
--- a/pkg/name/registry.go
+++ b/pkg/name/registry.go
@@ -94,7 +94,8 @@ func (r Registry) Scheme() string {
 	return "https"
 }
 
-func checkRegistry(name string) error {
+// CheckRegistry validates a registry name is compliant to the RFC 3986.
+func CheckRegistry(name string) error {
 	// Per RFC 3986, registries (authorities) are required to be prefixed with "//"
 	// url.Host == hostname[:port] == authority
 	if url, err := url.Parse("//" + name); err != nil || url.Host != name {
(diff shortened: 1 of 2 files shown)


imjasonh:
Would it be enough to call [`NewRegistry`](https://pkg.go.dev/github.com/google/go-containerregistry/pkg/name#NewRegistry) or [`NewRepository`](https://pkg.go.dev/github.com/google/go-containerregistry/pkg/name#NewRepository], and ignore the returned value? These call `check*` internally.

I'd like to have fewer exported methods in here if we can.

hectorj2f:
Sure, I can adjust my function calls. I just found a bit unnecessary to understand why I need to set different levels of validations in these functions. Also I'd need to check whether the input value is a registry or registry/repo to use whether NewRepository or NewRegistry. That is why I proposed to expose the validation themselves.

```
parts := strings.SplitN(source.OCI, ociRepoDelimiter, 2)
		if len(parts) == 2 && (strings.ContainsRune(parts[0], '.') || strings.ContainsRune(parts[0], ':')) {
			_, err := NewRepository(source.OCI, registryfuncs.StrictValidation)
			if err != nil {
				errs = errs.Also(apis.ErrInvalidValue(source.OCI, "oci", err.Error()))
			}
		} else {
			_, err := NewRegistry(source.OCI, registryfuncs.StrictValidation)
			if err != nil {
				errs = errs.Also(apis.ErrInvalidValue(source.OCI, "oci", err.Error()))
			}
		}
```

PR #1481, opened 2022-10-28
remote: fetchBlob: ensure we close response body

johnstcn:
If the remote server returns an incorrrect value for Content-Length we will end up not closing the response body and potentially leaking. This change ensures that we do so, and adds a corresponding unit test.


jonjohnsonjr (approved):
Thanks!

johnstcn:
Hold off on this for now, sorry!

PR #1487, opened 2022-11-07
Add support for zstd compression

LFrobeen:
This PR adds support for zstandard compression to the tarball layer implementation. It adds the application/vnd.oci.image.layer.v1.tar+zstd media type, a LayerCompression type and some extra logic in the tarball package. Moreover, this PR modifies [pkg/v1/partial/compressed.go](https://github.com/northflank/go-containerregistry/pull/2/files#diff-6632d2ac8e3d36f5972180430ed71640533372e8e1e3da9b377e8a5e7641c5af) so that it is able to unpack zstd compressed layers.

When creating a tarball layer, one can specify the compression algorithm using WithCompression and the compression level with WithCompressionLevel. By default it will still use gzip compression.
The changes should be fully backwards compatible.

### DIFF
--- a/internal/zstd/zstd.go
+++ b/internal/zstd/zstd.go
@@ -1,46 +1,28 @@
-// Copyright 2020 Google LLC All Rights Reserved.
-//
-// Licensed under the Apache License, Version 2.0 (the "License");
-// you may not use this file except in compliance with the License.
-// You may obtain a copy of the License at
-//
-//    http://www.apache.org/licenses/LICENSE-2.0
-//
-// Unless required by applicable law or agreed to in writing, software
-// distributed under the License is distributed on an "AS IS" BASIS,
-// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
-// See the License for the specific language governing permissions and
-// limitations under the License.
-
-// Package gzip provides helper functions for interacting with gzipped streams.
-package gzip
+package zstd
 
 import (
 	"bufio"
 	"bytes"
-	"compress/gzip"
-	"io"
-
 	"github.com/google/go-containerregistry/internal/and"
+	"github.com/klauspost/compress/zstd"
+	"io"
 )
 
-var MagicHeader = []byte{'\x1f', '\x8b'}
+var MagicHeader = []byte{'\x28', '\xb5', '\x2f', '\xfd'}
 
 // ReadCloser reads uncompressed input data from the io.ReadCloser and
 // returns an io.ReadCloser from which compressed data may be read.
-// This uses gzip.BestSpeed for the compression level.
+// This uses zstd level 1 for the compression.
 func ReadCloser(r io.ReadCloser) io.ReadCloser {
-	return ReadCloserLevel(r, gzip.BestSpeed)
+	return ReadCloserLevel(r, 1)
 }
 
 // ReadCloserLevel reads uncompressed input data from the io.ReadCloser and
 // returns an io.ReadCloser from which compressed data may be read.
-// Refer to compress/gzip for the level:
-// https://golang.org/pkg/compress/gzip/#pkg-constants
 func ReadCloserLevel(r io.ReadCloser, level int) io.ReadCloser {
 	pr, pw := io.Pipe()
 
-	// For highly compressible layers, gzip.Writer will output a very small
+	// For highly compressible layers, zstd.Writer will output a very small
 	// number of bytes per Write(). This is normally fine, but when pushing
 	// to a registry, we want to ensure that we're taking full advantage of
 	// the available bandwidth instead of sending tons of tiny writes over
@@ -52,7 +34,7 @@ func ReadCloserLevel(r io.ReadCloser, level int) io.ReadCloser {
 	go func() error {
 		// TODO(go1.14): Just defer {pw,gw,r}.Close like you'd expect.
 		// Context: https://golang.org/issue/24283
-		gw, err := gzip.NewWriterLevel(bw, level)
+		gw, err := zstd.NewWriter(bw, zstd.WithEncoderLevel(zstd.EncoderLevelFromZstd(level)))
 		if err != nil {
 			return pw.CloseWithError(err)
 		}
@@ -63,7 +45,7 @@ func ReadCloserLevel(r io.ReadCloser, level int) io.ReadCloser {
 			return pw.CloseWithError(err)
 		}
 
-		// Close gzip writer to Flush it and write gzip trailers.
+		// Close zstd writer to Flush it and write zstd trailers.
 		if err := gw.Close(); err != nil {
 			return pw.CloseWithError(err)
 		}
@@ -86,7 +68,7 @@ func ReadCloserLevel(r io.ReadCloser, level int) io.ReadCloser {
 // UnzipReadCloser reads compressed input data from the io.ReadCloser and
 // returns an io.ReadCloser from which uncompessed data may be read.
 func UnzipReadCloser(r io.ReadCloser) (io.ReadCloser, error) {
-	gr, err := gzip.NewReader(r)
+	gr, err := zstd.NewReader(r)
 	if err != nil {
 		return nil, err
 	}
@@ -105,7 +87,7 @@ func UnzipReadCloser(r io.ReadCloser) (io.ReadCloser, error) {
 
 // Is detects whether the input stream is compressed.
 func Is(r io.Reader) (bool, error) {
-	magicHeader := make([]byte, 2)
+	magicHeader := make([]byte, 4)
 	n, err := r.Read(magicHeader)
 	if n == 0 && err == io.EOF {
 		return false, nil
--- a/internal/zstd/zstd_test.go
+++ b/internal/zstd/zstd_test.go
@@ -1,0 +1,101 @@
+// Copyright 2020 Google LLC All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package zstd
+
+import (
+	"bytes"
+	"fmt"
+	"io"
+	"testing"
+)
+
+func TestReader(t *testing.T) {
+	want := "This is the input string."
+	buf := bytes.NewBufferString(want)
+	zipped := ReadCloser(io.NopCloser(buf))
+	unzipped, err := UnzipReadCloser(zipped)
+	if err != nil {
+		t.Error("UnzipReadCloser() =", err)
+	}
+
+	b, err := io.ReadAll(unzipped)
+	if err != nil {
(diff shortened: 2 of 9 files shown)


pkg/v1/types/types.go
```diff
@@ -24,7 +24,9 @@ const (
 	OCIManifestSchema1             MediaType = "application/vnd.oci.image.manifest.v1+json"
 	OCIConfigJSON                  MediaType = "application/vnd.oci.image.config.v1+json"
 	OCILayer                       MediaType = "application/vnd.oci.image.layer.v1.tar+gzip"
+	OCILayerZStd                   MediaType = "application/vnd.oci.image.layer.v1.tar+zstd"
 	OCIRestrictedLayer             MediaType = "application/vnd.oci.image.layer.nondistributable.v1.tar+gzip"
+	OCIRestrictedLayerZStd         MediaType = "application/vnd.oci.image.layer.nondistributable.v1.tar+zstd"
```
imjasonh, line 29: I'd prefer not to add this for now until we have a use case for it.

We might be able to get rid of non-distributable layers entirely: https://github.com/opencontainers/image-spec/pull/965

imjasonh (commented):
Wow, thanks for this! I've just taken a quick look so far but it looks great to me.

imjasonh:
It looks like there are some conflicts preventing CI from running. Can you rebase and we can see what the tests have to say?

pkg/v1/types/types.go
LFrobeen: Ok, that makes sense. I'll remove it.

LFrobeen:
Thanks for the feedback @imjasonh. I've removed the `OCIRestrictedLayerZStd` type and rebased the changes onto the main branch. 
So far I ran the unit tests locally and they appear to be successful. However, I should definitely add more tests for the new functionality. I was thinking about adding some more test cases to `tarbal/image_test.go` and `tarbal/layer_test.go`. Do you think that would be enough already, or are there other places where I should add more tests?

imjasonh:
> Thanks for the feedback @imjasonh. I've removed the `OCIRestrictedLayerZStd` type and rebased the changes onto the main branch. So far I ran the unit tests locally and they appear to be successful. However, I should definitely add more tests for the new functionality. I was thinking about adding some more test cases to `tarbal/image_test.go` and `tarbal/layer_test.go`. Do you think that would be enough already, or are there other places where I should add more tests?

That seems sufficient honestly. If this is something we expect to expose via a flag in `crane` it might be nice to have an e2e test for it, but I don't think we need to tackle that in this PR.

cc @jonjohnsonjr in case he has thoughts

pkg/v1/tarball/layer.go
```diff
@@ -90,11 +102,26 @@ func (l *layer) MediaType() (types.MediaType, error) {
...
+// WithCompression is a functional option for overriding the default
+// compression algorithm used for compressing uncompressed tarballs.
+// Also updates the mediaType to "application/vnd.oci.image.layer.v1.tar+zstd"
+// if zstd compression is selected.
+func WithCompression(compression LayerCompression) LayerOption {
+	return func(l *layer) {
+		if compression == ZStd {
+			l.mediaType = types.OCILayerZStd
+		} else {
+			l.mediaType = types.DockerLayer
```
jonjohnsonjr, line 114: I would expect `WithCompression(None)` to set `l.mediaType` to `DockerUncompressedLayer` and produce uncompressed layers. WDYT?

I am also OK with punting on this for now, returning an error or documenting that None is invalid.

internal/zstd/zstd.go
```diff
@@ -0,0 +1,114 @@
...
+	// the available bandwidth instead of sending tons of tiny writes over
+	// the wire.
+	// 64K ought to be small enough for anybody.
+	bw := bufio.NewWriterSize(pw, 2<<16)
+
+	// Returns err so we can pw.CloseWithError(err)
+	go func() error {
+		// TODO(go1.14): Just defer {pw,gw,r}.Close like you'd expect.
+		// Context: https://golang.org/issue/24283
+		gw, err := zstd.NewWriter(bw, zstd.WithEncoderLevel(zstd.EncoderLevelFromZstd(level)))
```
jonjohnsonjr, line 52: Nit: can these `gw` be `zw`?

internal/zstd/zstd.go
```diff
@@ -0,0 +1,114 @@
...
+		defer r.Close()
+
+		return nil
+	}()
+
+	return pr
+}
+
+// UnzipReadCloser reads compressed input data from the io.ReadCloser and
+// returns an io.ReadCloser from which uncompessed data may be read.
```
jonjohnsonjr, line 84:
```suggestion
// returns an io.ReadCloser from which uncompressed data may be read.
```

jonjohnsonjr: (And in the gzip version, old typo...)

internal/gzip/zip.go
```diff
@@ -113,34 +113,5 @@ func Is(r io.Reader) (bool, error) {
...
-	io.Reader
-	Peek(n int) ([]byte, error)
-}
-
-// Peek detects whether the input stream is gzip compressed.
-//
-// If r implements Peek, we will use that directly, otherwise a small number
-// of bytes are buffered to Peek at the gzip header, and the returned
-// PeekReader can be used as a replacement for the consumed input io.Reader.
-func Peek(r io.Reader) (bool, PeekReader, error) {
```
jonjohnsonjr, line 130: Can you add a stub for `gzip.Peek` and `zstd.Peek`?

internal/zstd/zstd_test.go
```diff
@@ -0,0 +1,101 @@
...
+			t.Error("UnzipReadCloser: expected errRead, got", err)
+		}
+	}
+
+	zr := ReadCloser(io.NopCloser(fr))
+	if _, err := zr.Read(nil); err != errRead {
+		t.Error("ReadCloser: expected errRead, got", err)
+	}
+
+	//zr = ReadCloserLevel(io.NopCloser(strings.NewReader("zip me")), -10)
```
jonjohnsonjr, line 97: Need to clean this up?

pkg/v1/tarball/layer.go
```diff
@@ -19,25 +19,37 @@ import (
...
 	v1 "github.com/google/go-containerregistry/pkg/v1"
 	"github.com/google/go-containerregistry/pkg/v1/types"
 )
 
+// LayerCompression is an enumeration of the supported compression algorithms for tarball layers
+type LayerCompression comp.Compression
+
+// The collection of known MediaType values.
+const (
+	None LayerCompression = "none"
```
jonjohnsonjr, line 40: I don't love these symbols being `tarball.None`, etc...

I am tempted to just expose a `compression` package so that we don't have to duplicate these everywhere... @imjasonh wdyt?

internal/zstd/zstd.go
LFrobeen: Fixed.

internal/zstd/zstd_test.go
LFrobeen: Fixed.

pkg/v1/tarball/layer.go
LFrobeen: Yes I agree that `tarball.None` looks weird. I would also be in favor of creating a non-internal `compression` package.

pkg/v1/tarball/layer.go
LFrobeen: That makes sense, but I'm not quite sure how easy this would be to implement. My understanding is that the decision whether layers are produced with or without compression is made by consumers of the Layer interface when they access the data via `.Compressed()` or `.Uncompressed()`. Hence I'd have to modify all the code sites that use `.Compressed()` to make `WithCompression(None)` work, right?

I was thinking about alternatively just returning uncompressed data from the `.Compressed()` method when `WithCompression(None)` is set. But I felt that would conflict with semantics of the `.Compressed()` method which may be a bit confusing. Then again, this PR assumes the consumers of the `.Compressed()` method are agnostic to the compression format anyway, so one could argue that no compression can still be regarded "some kind of very inefficient compression format".

If you agree with this idea, I would implement it that way. Otherwise, I think the best option might be to drop `WithCompression(None)` altogether for now.

internal/gzip/zip.go
LFrobeen: Sure, I'll just restore the original functionality, yes?

pkg/v1/tarball/layer.go
jonjohnsonjr: > But I felt that would conflict with semantics of the .Compressed() method which may be a bit confusing.

Yeah that's fair. Having both the `Compressed` and `Uncompressed` methods is a design mistake. It would be better if we deprecated `Uncompressed` and renamed `Compressed`, but it's too late for that. I am fine with dropping this for now, but I think we want to return an error if we don't want to support it.

pkg/v1/tarball/layer.go
```diff
@@ -90,11 +102,26 @@ func (l *layer) MediaType() (types.MediaType, error) {
...
 type LayerOption func(*layer)
 
+// WithCompression is a functional option for overriding the default
+// compression algorithm used for compressing uncompressed tarballs.
+// Also updates the mediaType to "application/vnd.oci.image.layer.v1.tar+zstd"
+// if zstd compression is selected.
+func WithCompression(compression LayerCompression) LayerOption {
+	return func(l *layer) {
+		if compression == ZStd {
+			l.mediaType = types.OCILayerZStd
```
jonjohnsonjr, line 112: This conflicts a bit with `layer.WithMediaType`. I might just drop the `else` here? We could also warn or error or document this fact.

pkg/v1/tarball/layer.go
LFrobeen: I created a public `pkg/compression` package that contains the `Compression` enum.

pkg/v1/tarball/layer.go
LFrobeen: I modified the `WithCompression` method to ignore calls the `compression.None` argument and use `gzip` instead. Additionally it will print this warning:
```
Compression type 'none' is not supported for tarball layers; using gzip compression.
```

pkg/v1/tarball/layer.go
LFrobeen: I removed the mediaType override for both zstd and gzip in order to be consistent. If `WithCompression(compression.ZStd)` is used, then the user should also set `WithMediaType(`types.OCILayerZStd)`. I've mentioned that in the updated documentation for `WithCompression`. When the layer is created via `LayerFromOpener`, the method will check whether the layer's `mediaType` matches its `compression` and print a warning if not. What do you think about this approach?

jonjohnsonjr:
There are a handful of submodules here (sorry) -- you might need to `go mod tidy` some others:

```
$ find | grep go\.mod | grep -v \.git
./cmd/krane/go.mod
./pkg/authn/k8schain/go.mod
./pkg/authn/kubernetes/go.mod
./go.mod
```

LFrobeen:
Oh yes, I totally overlooked those. I ran `go mod tidy` in all the submodules now. Indeed `github.com/klauspost/compress v1.15.11 // indirect` was missing from `pkg/authn/k8schain/go.mod`.

imjasonh:
Not annoying at all!

LFrobeen:
Awesome, thanks for triggering the tests! It looks like all checks have passed this time.

jonjohnsonjr (approved):

jonjohnsonjr:
Thanks for sticking with this -- I'd like to play with it a bit before cutting a release, so I might break the API in small ways, but it looks pretty close to what we want.

Issue #671, opened 2020-02-05
Add Gitlab + crane example


marklagendijk:
In #605 a debug image was added to allow for usage similar to kaniko, so that crane can be easily used from Gitlab.
However, there seems to be no documentation on how to use it with Gitlab.

marklagendijk:
Non-verified example:
``` yaml
docker-release-tag:
  stage: docker-release-tag
  only:
    refs:
      - /^release\/.+/
  image:
    name: gcr.io/go-containerregistry/crane:debug
    entrypoint: [""]
  script:
    - mkdir /root/.docker
    - echo "{\"auths\":{\"$CI_REGISTRY\":{\"username\":\"$CI_REGISTRY_USER\",\"password\":\"$CI_REGISTRY_PASSWORD\"}}}" > /root/.docker/config.json
    - /ko-app/crane tag $CI_REGISTRY_IMAGE:$CI_COMMIT_SHORT_SHA CI_COMMIT_TAG
```

jonjohnsonjr:
I don't use GitLab so I can't verify this example, but I'd be open to a PR with something like this.

We don't have a README under cmd/crane but it might make sense to start one (and also link to the [generated docs](https://github.com/google/go-containerregistry/blob/master/cmd/crane/doc/crane.md).)

marklagendijk:
I tested the example today and updated it to a working version.

marklagendijk:
I updated the example again. I made it a minimal example because that still fully shows how to use crane with GitLab.

jonjohnsonjr:
>     - /ko-app/crane tag $CI_REGISTRY_IMAGE:$CI_COMMIT_SHORT_SHA latest

I just cut a release of `ko` to pick up https://github.com/google/ko/pull/127 and re-built the latest/debug images. `crane` should be properly on the PATH now, so I think you can drop the `/ko-app/`.

marklagendijk:
Awesome! Thanks!

PR #1497, opened 2022-12-01
docs: pull latest instead of debug

AndrewCharlesHay:
I don't think it is standard to advise people to pull the debug image. I would imagine that would be like giving people the beta as the standard

### DIFF
--- a/cmd/crane/README.md
+++ b/cmd/crane/README.md
@@ -93,7 +93,7 @@ docker-tag-latest:
     refs:
       - main
   image:
-    name: gcr.io/go-containerregistry/crane:debug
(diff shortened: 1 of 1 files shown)


jonjohnsonjr:
To satisfy my own curiosity, I tried to figure out why it was this way. This example was proposed in https://github.com/google/go-containerregistry/issues/671 before we had a `crane auth login` command, so they were manually creating a config file, which needed busybox. The login command removed the need for this, but I neglected to move it back to the non-debug tag.

jonjohnsonjr (approved):
Thanks!

AndrewCharlesHay:
Thanks @jonjohnsonjr!

PR #1502, opened 2022-12-06
Use the default retry predicate in transport

jonjohnsonjr:
We have two levels of retries here.

The first is in the retryTransport, which currently just tests IsTemporary. This has a short backoff and was intended to handle blips in the network.

The second level was intended for higher level issues, e.g. if an upload PUT failed, we have to retry more than just a single HTTP request, because we might have to restart the upload with a new POST and PATCH.

This change propagates the default retryable predicate (from the scond level) down to the retryTransport. This also adds a list of HTTP codes that are retryable, so that the retryTransport can also handle server-side errors that should be retried.

Note that this doesn't respect remote.WithRetry{Backoff,Predicate} because callers can just override the entire transport via remote.WithTransport. This change tries to do the right thing by default without changing behavior for folks who seem to already know what they're doing.

Originally, I started down the path of wrapping various remote methods in a retry.Retry function (similar to writes) and decided against it due to the size of the change and how difficult it would be to maintain.

In order to reduce retry ampliciation, this adds a little context wrapper to signal to retryTransport not to retry (because the retry is happening at a higher level).

This also reduces the number of steps in the default backoff for retryTransport because more kinds of things are being retried, and I want to reduce the blast radius if something is erroneously retried.

### DIFF
--- a/internal/retry/retry.go
+++ b/internal/retry/retry.go
@@ -76,3 +76,19 @@ func Retry(f func() error, p Predicate, backoff wait.Backoff) (err error) {
 	wait.ExponentialBackoff(backoff, condition)
 	return
 }
+
+type contextKey string
+
+var key = contextKey("never")
+
+// Never returns a context that signals something should not be retried.
+// This is a hack and can be used to communicate across package boundaries
+// to avoid retry amplification.
+func Never(ctx context.Context) context.Context {
+	return context.WithValue(ctx, key, true)
+}
+
+// Ever returns true if the context was wrapped by Never.
+func Ever(ctx context.Context) bool {
+	return ctx.Value(key) == nil
+}
--- a/pkg/v1/remote/check.go
+++ b/pkg/v1/remote/check.go
@@ -35,11 +35,10 @@ func CheckPushPermission(ref name.Reference, kc authn.Keychain, t http.RoundTrip
 	// authorize a push. Figure out how to return early here when we can,
 	// to avoid a roundtrip for spec-compliant registries.
 	w := writer{
-		repo:    ref.Context(),
-		client:  &http.Client{Transport: tr},
-		context: context.Background(),
+		repo:   ref.Context(),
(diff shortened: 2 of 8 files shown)


imjasonh (approved):

sushain97:
@jonjohnsonjr 👋 We recently upgraded to this version and are frequently seeing `file already closed` errors. It appears that the transport here reuses the existing body for retrying the request. Since `net/http` closes the body when it's done with a request, that appears unsafe. Have you seen this issue?

parvathikp100:
> @jonjohnsonjr 👋 We recently upgraded to this version and are frequently seeing `file already closed` errors. It appears that the transport here reuses the existing body for retrying the request. Since `net/http` closes the body when it's done with a request, that appears unsafe. Have you seen this issue?

+1 to this, @jonjohnsonjr we noticed a surge in `PATCH: read {path}: file already closed error` failures post version upgrade to v0.15.2. 

The PATCH request is made here - https://github.com/google/go-containerregistry/blob/808e3548ee866e772d1332488cd8b94eb5df9b28/pkg/v1/remote/write.go#L269. 

The introduction of retry.Never(ctx) seems to be the major change in functions like streamBlob(), commitManifest() and initiateUpload() which have seen errors since the upgrade.

Can you please help verify if the changes introduced in this PR (like reduction of the backoff steps) haven't reduced the overall number of retries hence causing this error?

Also is it safe to re-use the same request in roundtripper here - 
```
func (t *retryTransport) RoundTrip(in *http.Request) (out *http.Response, err error) {
	roundtrip := func() error {
		out, err = t.inner.RoundTrip(in)
		if !retry.Ever(in.Context()) {
			return nil
		}
		if out != nil {
			for _, code := range t.codes {
				if out.StatusCode == code {
					return CheckError(out)
				}
			}
		}
		return err
	}
	retry.Retry(roundtrip, t.predicate, t.backoff)
	return
}
```

PR #1508, opened 2022-12-11
Update Arch Linux install instructions

kpcyrd:
I've uploaded an official `crane` package that contains binaries for `crane`, `gcrane` and `krane`.

Cheers!


imjasonh (approved):
Thanks!

PR #1517, opened 2022-12-23
Fix DNS lookup failure caused by too-short dial timeout

brandond:
If the user's primary nameserver is unavailable, and they are using the default resolver timeout of 5 seconds, the dial will fail before it can attempt to use any backup nameservers.

This change raises the dial timeout to a value that should allow failover through multiple nameservers. With a value of 16 seconds, the 4th nameserver has 1 second to respond before the dial times out.

Signed-off-by: Brad Davidson <brad.davidson@rancher.com>

### DIFF
--- a/pkg/v1/remote/options.go
+++ b/pkg/v1/remote/options.go
@@ -106,7 +106,10 @@ var DefaultTransport http.RoundTripper = &http.Transport{
 		// By default we wrap the transport in retries, so reduce the
 		// default dial timeout to 5s to avoid 5x 30s of connection
 		// timeouts when doing the "ping" on certain http registries.
-		Timeout:   5 * time.Second,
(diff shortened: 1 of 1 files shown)


brandond:
x-ref: https://github.com/rancher/wharfie/pull/11

imjasonh (commented):
@jonjohnsonjr

jonjohnsonjr:
Attempted to put together an ~alternative to this: https://github.com/google/go-containerregistry/pull/1521

Reverts this back to the default dialer timeout and fixes the issue with https://github.com/google/go-containerregistry/pull/1165 in a different manner.

brandond:
That's great; a parallel ping would be preferred to just increasing the timeout again.

jonjohnsonjr:
Closing in favor of https://github.com/google/go-containerregistry/pull/1521, let me know if you still see issues.

PR #1521, opened 2022-12-27
Race http fallback ping

jonjohnsonjr:
This implements a "happy eyeballs" (RFC 6555) style of race in order to speed up http fallback during registry pings.

This heavily borrows code from [net/dial.go](https://cs.opensource.google/go/go/+/master:src/net/dial.go;l=447;drc=38cfb3be9d486833456276777155980d1ec0823e) which implements this same kind of race for dual stack DNS.

Rather than waiting for the initial "https" ping to time out completely before attempting to use "http", we will instead start a second fallback ping after 300ms and return whichever response comes back first.

Partially reverts https://github.com/google/go-containerregistry/pull/1165 given that it negatively impacts dns fallback (https://github.com/google/go-containerregistry/pull/1517).

### DIFF
--- a/pkg/v1/remote/options.go
+++ b/pkg/v1/remote/options.go
@@ -103,10 +103,7 @@ const (
 var DefaultTransport http.RoundTripper = &http.Transport{
 	Proxy: http.ProxyFromEnvironment,
 	DialContext: (&net.Dialer{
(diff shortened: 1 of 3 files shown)


imjasonh (approved):
TIL "happy eyeballs"

PR #1529, opened 2023-01-10
add white background

rj-vinodh:
it is difficult to view a svg file with transparent background while use dark theme. please find the attached image

image


imjasonh (commented):
The new file just looks white to me, is there some rendering issue on GitHub's part?

Maybe the new white background is actually the top-most layer, and obscures the diagram? 🤷

PR #1532, opened 2023-01-12
fix: Clarify crane install steps

okozachenko1203:

PR #1533, opened 2023-01-12
clarify crane download readme

dtanner:
The download instructions include the $VERSION in the path twice, which isn't a valid link. 
After getting it to work for a specific release, I figured people might also want instructions to download the latest release, which uses a different URL path, so added that example as well.

### DIFF
--- a/cmd/crane/README.md
+++ b/cmd/crane/README.md
@@ -19,7 +19,7 @@ $ curl -sL "https://github.com/google/go-containerregistry/releases/latest/downl
 
 Download a specific version:
 ```
-$ VERSION=TODO   # Version number without leading v 
+$ VERSION=TODO   # Version number without leading v
 $ OS=Linux       # or Darwin, Windows
 $ ARCH=x86_64    # or arm64, x86_64, armv6, i386, s390x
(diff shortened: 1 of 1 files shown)


imjasonh:
Thanks for this contribution! Another recent PR, https://github.com/google/go-containerregistry/pull/1532, also proposed this change.

In any case, either you or that contributor will have to sign the Google CLA before we can merge this change.

dtanner:
ah thanks missed the redundancy.  FWIW the CLA is signed.

cmd/crane/README.md
```diff
@@ -11,12 +11,18 @@ A collection of useful things you can do with `crane` is [here](recipes.md).
...
+```
+$ OS=Linux       # or Darwin, Windows
+$ ARCH=x86_64    # or arm64, x86_64, armv6, i386, s390x
+$ curl -sL "https://github.com/google/go-containerregistry/releases/latest/download/go-containerregistry_${OS}_${ARCH}.tar.gz" > go-containerregistry.tar.gz
+```
 
+Download a specific version:
 ```
-$ VERSION=TODO   # Latest, or other
+$ VERSION=TODO   # Version number without leading v
```
imjasonh, line 22:
```suggestion
$ VERSION=TODO   # Version number without leading v
```

jonjohnsonjr (approved):

PR #1538, opened 2023-01-24
hash: use generic instantiation

howardjohn:
This enables doing something like
https://github.com/opencontainers/go-digest/issues/71#issuecomment-1330069987 to override the hashing method (example: replace with sha256simd).

When an importer does not overriden, the library should behave exactly the same.

See https://github.com/google/go-containerregistry/issues/1330

### DIFF
--- a/pkg/registry/registry_test.go
+++ b/pkg/registry/registry_test.go
@@ -15,8 +15,7 @@
 package registry_test
 
 import (
-	"crypto/sha256"
-	"encoding/hex"
+	"bytes"
 	"fmt"
 	"io/ioutil"
 	"log"
@@ -27,6 +26,7 @@ import (
 	"testing"
 
 	"github.com/google/go-containerregistry/pkg/registry"
+	v1 "github.com/google/go-containerregistry/pkg/v1"
 )
 
 const (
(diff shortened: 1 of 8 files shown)


jonjohnsonjr:
Oh that's wonderful. A few more places:

```
registry/manifest.go
113:		rd := sha256.Sum256(m.blob)
140:		rd := sha256.Sum256(m.blob)
156:		rd := sha256.Sum256(b.Bytes())

registry/registry_test.go
50:	h := sha256.Sum256([]byte(s))
```

```
legacy/tarball/write.go
66:	rawDigest := sha256.Sum256([]byte(s))
```

howardjohn:
> legacy/tarball/write.go

Thanks, missed those ones! Updated a bit. Note the ignored errors match what the old function did internally so I think its safe

jonjohnsonjr (approved):

pkg/registry/registry_test.go
```diff
@@ -47,8 +47,8 @@ const (
 )
 
 func sha256String(s string) string {
-	h := sha256.Sum256([]byte(s))
-	return hex.EncodeToString(h[:])
+	h, _, _ := v1.SHA256(bytes.NewReader([]byte(s)))
```
imjasonh, line 50: strings.NewReader could be better here

imjasonh:
We might want a lint check that `crypto/sha256` doesn't sneak back in in a future change, and to steer folks toward `crypto.SHA256`. I know I'm going to forget about this, old habits die hard.

Issue #1539, opened 2023-01-24
SLSA provenance verification broken for 0.13.0


imjasonh:
https://github.com/google/go-containerregistry/releases/tag/v0.13.0 was cut earlier, and the `verification` phase of the release failed: https://github.com/google/go-containerregistry/actions/runs/3998908027/jobs/6862265563

```
Download assets
  gh -R "$GITHUB_REPOSITORY" release download "$GITHUB_REF_NAME" -p "*.tar.gz"
  gh -R "$GITHUB_REPOSITORY" release download "$GITHUB_REF_NAME" -p "attestation.intoto.jsonl"
no assets match the file pattern
```

And indeed, there isn't any `attestation.intoto.jsonl` in the release, there's only [`multiple.intoto.jsonl`](https://github.com/google/go-containerregistry/releases/download/v0.13.0/multiple.intoto.jsonl).

The provenance generation and verification steps of the 0.12.0 and 0.12.1 and 0.11.0 releases also had breakages:
- https://github.com/google/go-containerregistry/pull/1468
- https://github.com/google/go-containerregistry/pull/1489
- https://github.com/google/go-containerregistry/pull/1444

Unless we can be sure that provenance is generated and can be verified correctly, I'm sort of inclined to remove the attestation until they can be made more reliable. It's worse to me to have misleading or broken or unverifiable provenance than to have no provenance at all.

@asraa @laurentsimon

laurentsimon:
/cc @ianlewis 
The`attestation.intoto.jsonl` was renamed to `multiple.intoto.jsonl` and the semver of the builder did not reflect that change.
Let me send a PR to update the file name in the verification part. Sorry about this. Nothing is broken, the attestation still exists and verification will work.

PR #1540, opened 2023-01-24
Update attestation name

laurentsimon:
Fix https://github.com/google/go-containerregistry/issues/1539

NOTE: we will release the slsa verifier Action Installer in a few weeks, so we'll be able to remove the script that manually downloads the slsa-verifier binary.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -72,7 +72,7 @@ jobs:
         run: |
           set -euo pipefail
           gh -R "$GITHUB_REPOSITORY" release download "$GITHUB_REF_NAME" -p "*.tar.gz"
-          gh -R "$GITHUB_REPOSITORY" release download "$GITHUB_REF_NAME" -p "attestation.intoto.jsonl"
+          gh -R "$GITHUB_REPOSITORY" release download "$GITHUB_REF_NAME" -p "multiple.intoto.jsonl"
       - name: Verify assets
         env:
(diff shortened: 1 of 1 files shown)


imjasonh (approved):

.github/workflows/release.yml
```diff
@@ -72,7 +72,7 @@ jobs:
         run: |
           set -euo pipefail
           gh -R "$GITHUB_REPOSITORY" release download "$GITHUB_REF_NAME" -p "*.tar.gz"
-          gh -R "$GITHUB_REPOSITORY" release download "$GITHUB_REF_NAME" -p "attestation.intoto.jsonl"
+          gh -R "$GITHUB_REPOSITORY" release download "$GITHUB_REF_NAME" -p "multiple.intoto.jsonl"
```
asraa, line 75: if you want to preserve the old name i believe there's an input `provenance-name` as well
https://github.com/slsa-framework/slsa-github-generator/tree/77395259ff7f705e2578f4552b79d90c06f49554/internal/builders/generic#workflow-inputs

ianlewis: An alternative to that would be to use the `provenance-name` output which will always be the name of the provenance artifact regardless of the input.
https://github.com/slsa-framework/slsa-github-generator/blob/main/internal/builders/generic/README.md#workflow-outputs

imjasonh: I don't have a strong preference for the filename either way, so long as verification works. Thanks for making this change!

PR #1541, opened 2023-01-24
Support artifactType, for images whose config.mediaType is not a config

imjasonh:
If an index contains an image whose config.mediaType is not the normal config type (i.e., it's an artifact, packaged as an image for registry compatibility), the index's descriptor for the image will include `artifactType`.

This is work toward supporting the OCI 1.1 referrers fallback tag scheme, where referrers are managed by the client and pushed as indexes to the registry, tagged with `:sha256-abcdef...`.

cc @jdolitsky


jdolitsky (approved):
👍

jonjohnsonjr (approved):
Overall LGTM. There are probably some things we missed, but let's let this bake for a bit before we cut a release.

PR #1543, opened 2023-01-26
Support for OCI 1.1+ referrers via fallback tag

jdolitsky:
Builds on top of #1541 

Crane subcommands "attach" and "refs" dropped in this PR, but can be found here: https://github.com/google/go-containerregistry/compare/main...jdolitsky:attach-pt-3?expand=1

cc @imjasonh

### DIFF
--- a/pkg/v1/mutate/mutate.go
+++ b/pkg/v1/mutate/mutate.go
@@ -115,9 +115,33 @@ func Config(base v1.Image, cfg v1.Config) (v1.Image, error) {
 	return ConfigFile(base, cf)
 }
 
-// Annotatable represents a manifest that can carry annotations.
-type Annotatable interface {
-	partial.WithRawManifest
+// Subject mutates the subject on an image or index manifest.
+//
+// The input is expected to be a v1.Image or v1.ImageIndex, and
+// returns the same type. You can type-assert the result like so:
+//
+//	img := Subject(empty.Image, subj).(v1.Image)
+//
+// Or for an index:
+//
+//	idx := Subject(empty.Index, subj).(v1.ImageIndex)
+//
+// If the input is not an Image or ImageIndex, the result will
+// attempt to lazily annotate the raw manifest.
+func Subject(f partial.WithRawManifest, subject v1.Descriptor) partial.WithRawManifest {
+	if img, ok := f.(v1.Image); ok {
+		return &image{
+			base:    img,
+			subject: &subject,
+		}
+	}
+	if idx, ok := f.(v1.ImageIndex); ok {
+		return &index{
+			base:    idx,
+			subject: &subject,
+		}
+	}
+	return arbitraryRawManifest{a: f, subject: &subject}
 }
 
 // Annotations mutates the annotations on an annotatable image or index manifest.
@@ -137,7 +161,7 @@ type Annotatable interface {
 //
 // If the input Annotatable is not an Image or ImageIndex, the result will
 // attempt to lazily annotate the raw manifest.
-func Annotations(f Annotatable, anns map[string]string) Annotatable {
+func Annotations(f partial.WithRawManifest, anns map[string]string) partial.WithRawManifest {
 	if img, ok := f.(v1.Image); ok {
 		return &image{
 			base:        img,
@@ -150,12 +174,13 @@ func Annotations(f Annotatable, anns map[string]string) Annotatable {
 			annotations: anns,
 		}
 	}
-	return arbitraryRawManifest{f, anns}
+	return arbitraryRawManifest{a: f, anns: anns}
 }
 
 type arbitraryRawManifest struct {
-	a    Annotatable
-	anns map[string]string
+	a       partial.WithRawManifest
+	anns    map[string]string
+	subject *v1.Descriptor
 }
 
 func (a arbitraryRawManifest) RawManifest() ([]byte, error) {
@@ -177,6 +202,9 @@ func (a arbitraryRawManifest) RawManifest() ([]byte, error) {
 		}
 	} else {
 		m["annotations"] = a.anns
+	}
+	if a.subject != nil {
+		m["subject"] = a.subject
 	}
 	return json.Marshal(m)
 }
--- a/pkg/v1/remote/descriptor.go
+++ b/pkg/v1/remote/descriptor.go
@@ -17,6 +17,8 @@ package remote
 import (
 	"bytes"
 	"context"
+	"encoding/json"
+	"errors"
 	"fmt"
 	"io"
 	"net/http"
@@ -59,7 +61,7 @@ type Descriptor struct {
 	v1.Descriptor
 	Manifest []byte
 
-	// So we can share this implementation with Image..
+	// So we can share this implementation with Image.
 	platform v1.Platform
 }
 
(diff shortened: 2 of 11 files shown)


pkg/v1/remote/descriptor.go
```diff
@@ -237,6 +239,49 @@ func (f *fetcher) url(resource, identifier string) url.URL {
...
+	}
+	// TODO: What am I supposed to do with a 400 here?
+	var im v1.IndexManifest
+	if err := json.Unmarshal(b, &im); err != nil {
+		return nil, nil, err
+	}
+
+	// If filter applied, filter out by artifactType and add annotation
+	// See https://github.com/opencontainers/distribution-spec/blob/main/spec.md#listing-referrers
+	if filter != nil {
```
jdolitsky, line 266: was not sure if wanting to actually filter the contents of the fallback tag for the client, but would like to at least start using something like `WithFilter()`. Can also be dropped for later

pkg/v1/remote/referrers.go
```diff
@@ -0,0 +1,43 @@
...
+	return f.fetchReferrers(o.context, o.filter, d)
+}
+
+// TODO: Implement this once real referrers API endpoint is used
+type ReferrersNextPage struct{}
+
+// TODO: Implement this once real referrers API endpoint is used
+func (nextPage *ReferrersNextPage) Referrers() (*v1.IndexManifest, *ReferrersNextPage, error) {
+	panic("not yet implemented")
+}
```
jdolitsky, line 43: open to other suggestions - picturing the case with pagination, where the `ReferrersNextPage` has stored the contents of the `Link` header, then called like this for example:

```go
var index *v1.IndexManifest
var next *remote.ReferrersNextPage
var err error
for next == nil {
	if index == nil { 
		index, next, err = Referrers(d)      // First time through the loop
	} else { 
		index, next, err = next.Referrers(d) // If we got here, we are on the next page
	}
	if err != nil {
		return err
	}
	processIndex(index)
}
```

jdolitsky: after discussion w/ @imjasonh, the paging can be hidden from the client (similar to tag listing)

If needed later, can add an option to limit it etc

pkg/v1/remote/descriptor.go
```diff
@@ -237,6 +239,49 @@ func (f *fetcher) url(resource, identifier string) url.URL {
 	}
 }
 
+// https://github.com/opencontainers/distribution-spec/blob/main/spec.md#referrers-tag-schema
+func fallbackTag(d name.Digest) name.Tag {
+	return d.Context().Tag(fmt.Sprintf("%s-%s", d.Digest().Algorithm, d.Digest().Hex))
```
imjasonh, line 244: I think I wrote this, but maybe we should just do:

```
return d.Context().Tag(strings.Replace(d.DigestStr(), ":", "-", 1))
```

This also lets us drop the new `d.Digest()` method in `pkg/name`.

pkg/v1/remote/descriptor.go
```diff
@@ -237,6 +239,49 @@ func (f *fetcher) url(resource, identifier string) url.URL {
...
+}
+
+func (f *fetcher) fetchReferrers(ctx context.Context, filter map[string]string, d name.Digest) (*v1.IndexManifest, error) {
+	// The registry doesn't support the Referrers API endpoint, so we'll use the fallback tag scheme.
```
imjasonh, line 248:
```suggestion
	// Assume the registry doesn't support the Referrers API endpoint, so we'll use the fallback tag scheme.
```

pkg/v1/remote/descriptor.go
```diff
@@ -237,6 +239,49 @@ func (f *fetcher) url(resource, identifier string) url.URL {
...
+	b, _, err := f.fetchManifest(fallbackTag(d), []types.MediaType{types.OCIImageIndex})
+	if err != nil {
+		return nil, err
+	}
+	var terr *transport.Error
+	if ok := errors.As(err, &terr); ok && terr.StatusCode == http.StatusNotFound {
+		// Not found just means there are no attachments yet. Start with an empty manifest.
+		return nil, nil
```
imjasonh, line 256: We should return an empty `IndexManifest` here, with at least the mediaType set.

pkg/v1/remote/descriptor.go
```diff
@@ -237,6 +239,49 @@ func (f *fetcher) url(resource, identifier string) url.URL {
...
+	}
+	// TODO: What am I supposed to do with a 400 here?
```
imjasonh, line 258: I think we should return that error, e.g., `transport.CheckError(200, 404)` and then handle 404 like you do above.

pkg/v1/remote/descriptor.go
imjasonh: I like the filtering, let's keep it.

It might make sense as a separate method we can unit test separately, and maybe eventually export separately in a future change.

We also might be able to lean on https://pkg.go.dev/github.com/google/go-containerregistry@v0.13.0/pkg/v1/match

pkg/v1/remote/image.go
```diff
@@ -139,6 +139,12 @@ func (r *remoteImage) Descriptor() (*v1.Descriptor, error) {
 	return r.descriptor, err
 }
 
+func (r *remoteImage) Referrers(filter map[string]string) (*v1.IndexManifest, error) {
+	return r.fetcher.fetchReferrers(r.context, filter, r.Ref.Context().Digest(r.descriptor.Digest.String()))
+}
+
+// TODO: img.Subject()?
```
imjasonh, line 146: Let's do this in a separate PR, like we did with `img.ArtifactType`.

pkg/v1/remote/referrers_test.go
```diff
@@ -0,0 +1,185 @@
...
+	if err != nil {
+		t.Fatal(err)
+	}
+	if d := cmp.Diff([]v1.Descriptor{leafDesc}, index.Manifests); d != "" {
+		t.Fatalf("referrers diff after second push (-want,+got): %s", d)
+	}
+
+	// Try applying filters and verify number of manifests and and annotations
+	goodFilter := map[string]string{types.OCIFilterArtifactType: "application/testing123"}
+	index, err = remote.Referrers(rootRefDigest, remote.WithFilter(goodFilter))
```
imjasonh, line 147: Can we change the method signature of `WithFilter` to something like `WithFilter(key, value)` and allow it to be passed potentially multiple times to build up the same map?

```
remote.Referrers(d,
    remote.WithFilter("type", "sbom"),
    remote.WithFilter("flavor", "salty"))
```

pkg/v1/types/types.go
```diff
@@ -44,6 +44,10 @@ const (
 	DockerVendorPrefix = "vnd.docker"
 )
 
+// OCI 1.1+ related
+const OCIFilterArtifactType = "artifactType"
+const OCIAnnotationFiltersApplied = "org.opencontainers.referrers.filtersApplied"
```
imjasonh, line 49:
```suggestion
// OCIFilterArtifactType is the filter parameter value that indicates returned referrers should be filtered by artifact type.
const OCIFilterArtifactType = "artifactType"

// OCIAnnotationFiltersApplied is the annotation that, when present on an index returned when querying referrers, indicates a filter was applied to the results.
const OCIAnnotationFiltersApplied = "org.opencontainers.referrers.filtersApplied"
```

imjasonh (commented):
This looks good!

pkg/name/digest.go
```diff
@@ -49,6 +50,12 @@ func (d Digest) DigestStr() string {
 	return d.digest
 }
 
+// Digest returns the digest component of the Digest, parsed as a v1.Hash.
+func (d Digest) Digest() v1.Hash {
```
jonjohnsonjr, line 54: If we keep this, call it `Hash`.

pkg/v1/mutate/mutate.go
```diff
@@ -115,9 +115,33 @@ func Config(base v1.Image, cfg v1.Config) (v1.Image, error) {
...
+//
+//	img := Subject(empty.Image, subj).(v1.Image)
+//
+// Or for an index:
+//
+//	idx := Subject(empty.Index, subj).(v1.ImageIndex)
+//
+// If the input is not an Image or ImageIndex, the result will
+// attempt to lazily annotate the raw manifest.
+func Subject(f partial.WithRawManifest, subject v1.Descriptor) partial.WithRawManifest {
```
jonjohnsonjr, line 131: My understanding was that an index couldn't have a `subject` (for reasons I don't agree with).

I'm fine with this as-is because I disagree with the wg, but want to make sure this was intentional.

pkg/v1/types/types.go
```diff
@@ -44,6 +44,10 @@ const (
 	DockerVendorPrefix = "vnd.docker"
 )
 
+// OCI 1.1+ related
```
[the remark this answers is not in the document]
jdolitsky: The annotation was in the spec, but not the querystring param. Dropped the former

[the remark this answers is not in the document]
jdolitsky: both dropped

pkg/v1/remote/referrers_test.go
jdolitsky: yee

imjasonh (approved):
@jonjohnsonjr anything else?

jonjohnsonjr (approved):

PR #1544, opened 2023-01-26
Enforce proper sha256 usage

howardjohn:
followup to https://github.com/google/go-containerregistry/pull/1538#issuecomment-1403631330


imjasonh (approved):
Thanks!

PR #1545, opened 2023-01-26
Allow crane edit to generate non-image artifacts

jonjohnsonjr:
Instead of using mutate, just do this ourselves to allow for arbitrary config file contents rather than mutating the OCI config struct.

### DIFF
--- a/internal/cmd/edit.go
+++ b/internal/cmd/edit.go
@@ -17,6 +17,7 @@ package cmd
 import (
 	"archive/tar"
 	"bytes"
+	"encoding/json"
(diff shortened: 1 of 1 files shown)


imjasonh (commented):
I'm not super convinced we need to do more work to enable crane to produce invalid configs. Can this just live in a fork and not be supported behavior in crane?

jonjohnsonjr:
I have re-framed the change in terms of more generic usefulness.

imjasonh (approved):
Now this I can get behind! 😍

PR #1546, opened 2023-01-27
Support for OCI 1.1+ referrers via API

jdolitsky:
Follow-up to #1543, use the referrers API endpoint if the registry supports it, otherwise fallback to digest tag

### DIFF
--- a/pkg/registry/registry.go
+++ b/pkg/registry/registry.go
@@ -109,7 +109,7 @@ func Logger(l *log.Logger) Option {
 	}
 }
 
-// Logger overrides the logger used to record requests to the registry.
+// WithReferrersSupport enables the referrers API endpoint (OCI 1.1+)
 func WithReferrersSupport(enabled bool) Option {
 	return func(r *registry) {
 		r.referrersEnabled = enabled
--- a/pkg/registry/registry.go
+++ b/pkg/registry/registry.go
@@ -30,9 +30,10 @@ import (
 )
 
 type registry struct {
-	log       *log.Logger
-	blobs     blobs
-	manifests manifests
+	log              *log.Logger
+	blobs            blobs
+	manifests        manifests
+	referrersEnabled bool
 }
 
 // https://docs.docker.com/registry/spec/api/#api-version-check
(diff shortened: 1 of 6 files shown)


pkg/v1/remote/descriptor.go
```diff
@@ -479,3 +492,20 @@ func (f *fetcher) blobExists(h v1.Hash) (bool, error) {
...
+		return
+	}
+	if v, ok := filter["artifactType"]; ok {
+		tmp := []v1.Descriptor{}
+		for _, desc := range im.Manifests {
+			if desc.ArtifactType == v {
+				tmp = append(tmp, desc)
+			}
+		}
+		im.Manifests = tmp
```
imjasonh, line 509: This could also just return `tmp` and line 270 and 290 could be `return applyFilter(...), nil`

imjasonh (commented):
Can we add support for referrers to `pkg/registry`, and use that to test this path?

pkg/registry/registry.go
```diff
@@ -104,3 +108,10 @@ func Logger(l *log.Logger) Option {
 		r.blobs.log = l
 	}
 }
+
+// WithReferrersSupport enables the referrers API endpoint (OCI 1.1+)
+func WithReferrersSupport(enabled bool) Option {
+	return func(r *registry) {
+		r.referrersEnabled = enabled
+	}
+}
```
jdolitsky, line 117: this is disabled by default, but we could go the other direction too

pkg/v1/remote/write.go
```diff
@@ -580,19 +580,40 @@ func unpackTaggable(t Taggable) ([]byte, *v1.Descriptor, error) {
...
+	}
+	defer resp.Body.Close()
+
+	if err := transport.CheckError(resp, http.StatusOK, http.StatusNotFound, http.StatusBadRequest); err != nil {
+		return err
+	}
+	if resp.StatusCode == http.StatusOK {
+		// The registry supports Referrers API. The registry is responsible for updating the referrers list.
+		return nil
+	}
```
imjasonh, line 605: Sounds like this will all go away once https://github.com/opencontainers/distribution-spec/pull/379 is in, do we want to just start supporting that in here?

imjasonh (commented):
Very nice! I think we could merge like this, just a couple questions.

pkg/v1/remote/write.go
jdolitsky: i think id rather see it merged first, then come back to change here

imjasonh (approved):

imjasonh:
I'll let @jonjohnsonjr clickey the mergey

jonjohnsonjr (approved):

PR #1550, opened 2023-01-31
fix(mutate): also set timestamps only present in some formats

### LINKED ISSUE #1523: ggcr: `mutate.Time` should set `AccessTime` and `ChangeTime`
### Describe the bug

Layers stored in the PAX format can contain file access times and creation times in addition to modification times. However, the `layerTime` function [only sets `ModTime`](https://github.com/google/go-containerregistry/blob/9db616f1dab14326675b8f0b8b360a3066fe3348/pkg/v1/mutate/mutate.go#L439).

I believe this to be the root cause of [issue 2005](https://github.com/GoogleContainerTools/kaniko/issues/2005) in Kaniko.

### Expected behavior

Images mutated with `mutate.Time` should set all timestamps in the image, including `AccessTime` and `ChangeTime`:

> ```go
> // Time sets all timestamps in an image to the given timestamp.
> func Time(img v1.Image, t time.Time) (v1.Image, error) {
> ```

BronzeDeer:
fixes #1523 

Currently, layerTime only mutates the modification timestamp which is present in all tar formats. When working with layers GNU or PAX header formats, access and change timestamps are also generated/recorded. This breaks functionality like reproducible builds which rely on having control over all variable fields (esp. timestamps)  in the layers

This commit adds a special case to the layerTime(Layer, Time) function which will also set the additional timestamps if the header is of GNU or PAX format

Signed-off-by: Joel Pepper <pepper@bronze-deer.de>

### DIFF
--- a/pkg/v1/mutate/mutate.go
+++ b/pkg/v1/mutate/mutate.go
@@ -437,6 +437,13 @@ func layerTime(layer v1.Layer, t time.Time) (v1.Layer, error) {
 		}
 
 		header.ModTime = t
+
+		//PAX and GNU Format support additional timestamps in the header
+		if header.Format == tar.FormatPAX || header.Format == tar.FormatGNU {
+			header.AccessTime = t
+			header.ChangeTime = t
+		}
+
 		if err := tarWriter.WriteHeader(header); err != nil {
 			return nil, fmt.Errorf("writing tar header: %w", err)
 		}


pkg/v1/mutate/mutate.go
```diff
@@ -437,6 +437,13 @@ func layerTime(layer v1.Layer, t time.Time) (v1.Layer, error) {
 		}
 
 		header.ModTime = t
+
+		//PAX and GNU Format support additional timestamps in the header
+		if header.Format == tar.FormatPAX || header.Format == tar.FormatGNU {
```
imjasonh, line 442: Is it worth always setting these times, even if it's not a PAX tar?

imjasonh (commented):
Thanks for this change!

pkg/v1/mutate/mutate.go
BronzeDeer: Since they only exist in those Formats, I don't know how the archive/tar package will react when trying to write a header for another while those fields are set. I would expect it to handle that gracefully, but since I'm not sure of that I programmed defensively

imjasonh (approved):

BronzeDeer:
@imjasonh can we get this merged into trunk? Then I can go ahead and create a PR for kaniko, which has had broken reproducible builds for over a year

PR #1551, opened 2023-02-01
Allow setting Content-Type in crane edit manifest

jonjohnsonjr:
This uses the --media-type flag.

### DIFF
--- a/internal/cmd/edit.go
+++ b/internal/cmd/edit.go
@@ -84,18 +84,18 @@ func NewCmdEditConfig(options *[]crane.Option) *cobra.Command {
 
 // NewCmdManifest creates a new cobra.Command for the manifest subcommand.
 func NewCmdEditManifest(options *[]crane.Option) *cobra.Command {
-	var dst string
+	var (
+		dst string
+		mt  string
+	)
 	cmd := &cobra.Command{
 		Use:   "manifest",
 		Short: "Edit an image's manifest.",
-		Example: `  # Edit ubuntu's config file
-  crane edit config ubuntu
-
-  # Overwrite ubuntu's config file with '{}'
-  echo '{}' | crane edit config ubuntu`,
+		Example: `  # Edit ubuntu's manifest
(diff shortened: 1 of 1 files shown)


tianon:
Can this new flag default to the `mediaType` in my edited document? :pleading_face:

(I want to be able to convert docs from Docker media types to OCI media types :bow:)

tianon:
I suppose I should report back that I successfully tested this PR for converting an image from Docker media types to OCI media types and all went well. :smile:

imjasonh (approved):

imjasonh:
> I suppose I should report back that I successfully tested this PR for converting an image from Docker media types to OCI media types and all went well. :smile:

Would a command just to OCIfy a docker image be helpful? That's something we've found a need for in the past and then punted on.

tianon:
Hmmm, as a flag to `crane cp` maybe?  I don't think I'd personally use it, though, since it's just as easy to munge the 2-3 media types myself and I don't actually do this kind of munging very often (and hopefully even less over time!)

I ran into this while testing something _else_ the first time, and it was only the second time that I was trying to OCI-ify an existing image that I decided to really bug Jon about it in earnest since it seemed like an easy win for the overall ergonomics of `crane edit` matching my expectations of how it was about to work. :joy:

tianon:
Mostly just to make Jon cringe, I'd be happy to write a `jq` one-liner for the docs so you can suggest users do `crane manifest ... | jq '...' | crane edit manifest ...` if that's helpful :smile:

jonjohnsonjr:
> Would a command just to OCIfy a docker image be helpful? That's something we've found a need for in the past and then punted on.

I'm 99% sure there's a PR to do this somewhere.

tianon:
> > Would a command just to OCIfy a docker image be helpful? That's something we've found a need for in the past and then punted on.
> 
> I'm 99% sure there's a PR to do this somewhere.

Looks like that's #1293 (also relevant is #1245)

PR #1559, opened 2023-02-07
skip tls verification if default transport is used with insecure option

### LINKED ISSUE #1553: ggcr: library's crane.Pull behavior with crane.Insecure option does not match the CLI's --insecure flag behavior
### Describe the bug

The `crane.Pull` behavior appears to be inconsistent with the CLI when attempting to reach either a registry using HTTP, or a registry that uses HTTPS but with a self-signed certificate.

The `crane` CLI accepts the `--insecure` flag which seems to support either case transparently. To accomplish this with the library, you'd have to pass both the crane.Insecure[1] option and crane.WithTransport[2] to get the same behavior as observed in the crane CLI's code.

[1] https://github.com/google/go-containerregistry/blob/main/cmd/crane/cmd/root.go#L59
[2] https://github.com/google/go-containerregistry/blob/main/cmd/crane/cmd/root.go#L75-L78

### To Reproduce

I just used caddy to throw up a quick reverse-proxy with "internal" certs, and modified my host file to point the `--from` domain to my own machine. Then targeted existing images from the origin registry.

```
caddy reverse-proxy --to https://quay.io --internal-certs --from https://registry.example.com:8443
```

A code snippet like the following works as I expect.

```Go
        // copied from crane's CLI
	rt := remote.DefaultTransport.(*http.Transport).Clone()
	rt.TLSClientConfig = &tls.Config{
		InsecureSkipVerify: true, //nolint: gosec
	}

	img, err := crane.Pull("registry.example.com:8443/fedora/fedora:38", crane.Insecure, crane.WithTransport(rt))
	if err != nil {
		panic(err)
	}
	fmt.Println(img.Layers())
```

If you pull out the Transport definition, it throws an error that I believe indicated that the `client sent an HTTP request to an HTTPS server`. 

### Expected behavior

I expected the Insecure flag to crane to behave like the CLI, in that the self-signed certificate is ignored when pulling the image, and the image downloads as expected. To that end, I would imagine that passing crane.Insecure would also set the crane.WithTransport option similar to how the CLI does it, or that the Pull logic would handle the transport definition internally using the insecure option's value as an input (if no transport option was passed in).

With the same reverse-proxy setup described above, you should be able to `crane pull --insecure registry.example.com:8443/fedora/fedora:38 fedora38.tgz` and have it succeed.

### Additional context

I noticed that there's a Scheme handler https://github.com/google/go-containerregistry/blob/main/pkg/name/registry.go#L75 - Just looking at that, I'm actually unsure exactly what scheme is being used in cases where both crane.Insecure and crane.WithTransport are passed. I'm not sure if I'm missing something, or this particular block that derives the scheme is out of scope, but it seemed odd that a case where both options were passed seems to behave correctly. I would have thought for sure that we'd still be getting `http` given what I'm seeing there.

- Version of the module: v0.13.0
- Registry used (e.g., GCR, ECR, Quay): n/a

komish:
Fixes #1553 

This PR addresses an inconsistency between the Crane library and the Crane CLI (which works correctly) with regards to interactions with image registries behind untrusted (read: self-signed) certificates.

This implementation internally tracks the Insecure/WithTransport options being set to allow `makeOptions` to decide whether or not a slightly modified default transport needs to be added to the remote options. If the right combination is not set, this does not add any default transport, relying on whatever default was to be used (same behavior as before).

This PR _does not_ change the Crane CLI because the `http.RoundTripper` that it builds out ultimately takes into account docker config values that a library caller should invoke manually. In doing so, a caller would be defining their own transport, and the logic here would not take effect. https://github.com/google/go-containerregistry/blob/main/cmd/crane/cmd/root.go#L88

Signed-off-by: Jose R. Gonzalez <jose@flutes.dev>


pkg/crane/options.go
```diff
@@ -47,26 +51,43 @@ func makeOptions(opts ...Option) Options {
...
 func WithTransport(t http.RoundTripper) Option {
 	return func(o *Options) {
 		o.Remote = append(o.Remote, remote.WithTransport(t))
+		o.transportSet = true
 	}
 }
 
 // Insecure is an Option that allows image references to be fetched without TLS.
+// This will also allow for untrusted (e.g. self-signed) certificates in cases where
+// the default transport is used (i.e. when WithTransport is not defined).
```
jonjohnsonjr, line 87:
```suggestion
// the default transport is used (i.e. when WithTransport is not used).
```

jonjohnsonjr (commented):
This LGTM overall. Do you mind adding a test to cover this? I'd probably be lazy and add a private `transport` to `Options` when you call `Insecure` and `WithTransport`, then just check that `InsecureSkipVerify` is set for that.

pkg/crane/options.go
komish: This suggestion should be incorporated in the latest force-push.

komish:
@jonjohnsonjr Sure thing. I went ahead and added tests to make sure we're setting the internal boolean values (that trigger the logic we want), and then a test to make sure we're adding the transport. As you suggested, I added a private transport value to facilitate the test, and rewired the logic to set that value. Made the test much simpler.

pkg/crane/options.go
```diff
@@ -47,26 +52,44 @@ func makeOptions(opts ...Option) Options {
...
 type Option func(*Options)
 
 // WithTransport is a functional option for overriding the default transport
-// for remote operations.
+// for remote operations. Setting a transport will override the Insecure option's
+// configuration allowing for image registries to use untrusted certificates.
 func WithTransport(t http.RoundTripper) Option {
 	return func(o *Options) {
 		o.Remote = append(o.Remote, remote.WithTransport(t))
+		o.transportSet = true
```
jonjohnsonjr, line 83: What do you think about `o.transport = t` instead of the bool, then have `makeOptions` check for `opt.transport == nil` rather than having both the bool and the transport?

komish: I'm thinking we'd still want to track the insecure bool, but otherwise that could work and removes a bit of a redundancy. I'll put it together and send it here momentarily (along with the Boilerplate check fix).

komish: Implemented in https://github.com/google/go-containerregistry/commit/5112fe351dcc371dfe427daf1870978a4c08053e

jonjohnsonjr (approved):

jonjohnsonjr:
Thanks!

Issue #1557, opened 2023-02-07
question: how can I copy image between registries for multiple OS/arch ?


developer-guy:
I noticed that the platform flag in copy command only accepts a single value, and if you specify multiple platform flags, the last one will win. So, it'd be good to support taking multiple values for the platform flag to be able to specify the exact platforms that people care about.

```shell
# will copy an image built for linux/amd64
$ crane copy --platform linux/amd64

# will copy an image built for linux/arm64
$ crane copy --platform linux/arm64

# will copy an image built for linux/amd64, it ignores linux/arm64, but it'd be copy this image for each platform.
$ crane copy --platform linux/arm64 --platform linux/amd64
```

jonjohnsonjr:
If you copy the whole multi-arch image, you get everything, but I assume you only want to copy things you care about?

I can interpret this three ways:
1. Copy linux/arm64 and linux/amd64 images separately, do not push a multi-platform image.
2. Synthesize a new multi-platform image that only contains linux/arm64 and linux/amd64.
3. Copy only the linux/arm64 and linux/amd64 child images, then attempt to push the multi-platform image as-is, hoping that the registry supports [sparse indexes](https://github.com/opencontainers/distribution-spec/pull/310).

The first one would be straightforward to implement.

The second one I would hesitate to add to `copy` but maybe we can have a flag for the behavior. This was requested in https://github.com/google/go-containerregistry/issues/1143 already.

The third one would make sense, but probably wouldn't work with almost all existing registries.

Any of those seem like what you want?

jonjohnsonjr:
FWIW you could hack around this a bit with `crane edit manifest`, but it's certainly not elegant.

```
$ crane manifest ubuntu | jq '.manifests = ([.manifests | .[] | select (.platform.os == "linux" and (.platform.architecture == "amd64" or .platform.architecture == "arm64"))])'
{
  "manifests": [
    {
      "digest": "sha256:c985bc3f77946b8e92c9a3648c6f31751a7dd972e06604785e47303f4ad47c4c",
      "mediaType": "application/vnd.oci.image.manifest.v1+json",
      "platform": {
        "architecture": "amd64",
        "os": "linux"
      },
      "size": 424
    },
    {
      "digest": "sha256:61bd0b97000996232eb07b8d0e9375d14197f78aa850c2506417ef995a7199a7",
      "mediaType": "application/vnd.oci.image.manifest.v1+json",
      "platform": {
        "architecture": "arm64",
[...]

Issue #1144, opened 2021-10-10
RFC: `crane amend`


jonjohnsonjr:
See also:
* https://github.com/google/go-containerregistry/pull/891
* https://github.com/google/go-containerregistry/issues/1137
* https://github.com/google/go-containerregistry/issues/1143

There is a pretty obvious need for a command to create/manipulate manifest lists, analogous to `crane append`. Originally I was hoping to extend `crane append` to do this, but I think having separate commands for images vs indexes could simplify the UX quite a bit. Also, I just really like the name `crane amend`.

I would love input from the community around use cases and ideas around the API surface. It's possible that we might need more than one command to cover everything, but the use cases I have in mind seem very related. 

#### 1. Constructing a new manifest list

This would mirror `crane append` with no base. I am unsure exactly how to express something like overriding the platform of an image. We discussed this quite a bit in https://github.com/google/go-containerregistry/pull/891 so that might be a reasonable path forward.

#### 2. Appending to an existing manifest list

This seems like i could just be a variant of the first use case where you supply an existing index ​to amend.

#### 3. Merging existing manifest lists

This  feels different enough that it might make sense as a separate command or possibly the composition of two commands?

@imjasonh has implemented this as https://github.com/imjasonh/combine

#### 4. Reducing an existing manifest list to a subset of platforms

Not sure exactly how we would want to express this -- a flag of which platforms/positions to keep or which platforms/positions to remove? For most folks, it's probably most natural to say "I want only these platforms' because it's more declarative and context-independent, but I'm not sure if we want something more expressive/general than this.

### Open Questions

1. Should it be possible to create an index that references arbitrary mediaTypes, e.g. a layer

thesayyn:
I believe many of these ideas eventually boil down to merging layouts on disk and avoiding tricks like argument positions.
I get the feeling that the maintainers don't want to introduce yet another and are obliged to maintain that forever.

Thinking out loud.

### 4. i'd see this as being supported in `pull` either with 

#### newly introduced options for platform selection
- `--platform-only` to get exact set of platforms 
- a softer version would be a combination of  `--platform-exclude`  and `--platform-include`. possibly fetching images that not listed via these two.

but then you left wondering how about other properties such as `annotations` `os.features`  `os.version` (recently added). this feature won't be complete ever

#### a brand new option  `--select`
this option can express everything from `platforms` to `features` with power of [CEL](https://github.com/google/cel-go)
[...]

Issue #1143, opened 2021-10-08
crane: how to copy a sub-set of a image platforms?


rgl:
How can we copy the images of a manifest list that match a given platform list?

Using `crane --all` we can copy all the images from a manifest list.

Using `crane --platform` we can can copy a single image that matches the given singular platform, but how about a sub-set? Something like `crane --platform linux/amd64,linux/arm64`?

This would be useful for copying a sub-set of the [`debian:bullseye-slim`](https://hub.docker.com/_/debian) images (e.g. just the `linux/amd64` and `linux/arm64` platform images).

rgl:
> Do you want to copy both images separately to the destination? Or do you want to take the existing manifest list, filter out all the platforms you don't care about, and push a _new_ manifest list with just those two platforms?

The second one :-)

jonjohnsonjr:
X-ref: https://github.com/containers/skopeo/pull/1511

gpsmit:
I would love to have this feature as well.
Today I pushed 5GB worth of windows images when im only interested in linux arm & and images

PR #1561, opened 2023-02-08
Implement crane index subcommand

jonjohnsonjr:
crane index filter allows platform-based filtering of manifests in an index. This is primarily useful for folks who want to use only some entries in a base image without having to copy irrelevant platforms.

crane index append allows appending manifests to an existing index or creating a new index from scratch.

Fixes https://github.com/google/go-containerregistry/issues/1144
Fixes https://github.com/google/go-containerregistry/issues/1143
Fixes https://github.com/google/go-containerregistry/issues/1557

See also:
* https://github.com/google/go-containerregistry/pull/891
* https://github.com/google/go-containerregistry/issues/1144#issuecomment-1044845681

### DIFF
--- a/cmd/crane/cmd/util.go
+++ b/cmd/crane/cmd/util.go
@@ -15,7 +15,6 @@
 package cmd
 
 import (
-	"fmt"
 	"strings"
 
 	v1 "github.com/google/go-containerregistry/pkg/v1"
@@ -75,14 +74,7 @@ func platformToString(p *v1.Platform) string {
 	if p == nil {
 		return "all"
 	}
-	platform := ""
-	if p.OS != "" && p.Architecture != "" {
-		platform = p.OS + "/" + p.Architecture
-	}
-	if p.Variant != "" {
-		platform += "/" + p.Variant
-	}
-	return platform
+	return p.String()
 }
 
 func parsePlatform(platform string) (*v1.Platform, error) {
@@ -90,26 +82,5 @@ func parsePlatform(platform string) (*v1.Platform, error) {
 		return nil, nil
 	}
 
-	p := &v1.Platform{}
-
-	parts := strings.SplitN(platform, ":", 2)
-	if len(parts) == 2 {
-		p.OSVersion = parts[1]
-	}
-
-	parts = strings.Split(parts[0], "/")
-
-	if len(parts) > 3 {
-		return nil, fmt.Errorf("failed to parse platform '%s': too many slashes", platform)
-	}
-
-	p.OS = parts[0]
-	if len(parts) > 1 {
-		p.Architecture = parts[1]
-	}
-	if len(parts) > 2 {
-		p.Variant = parts[2]
-	}
-
-	return p, nil
+	return v1.ParsePlatform(platform)
 }
--- a/cmd/crane/cmd/util.go
+++ b/cmd/crane/cmd/util.go
@@ -21,6 +21,35 @@ import (
 	v1 "github.com/google/go-containerregistry/pkg/v1"
 )
 
+type platformsValue struct {
+	platforms []v1.Platform
+}
+
+func (ps *platformsValue) Set(platform string) error {
+	if ps.platforms == nil {
+		ps.platforms = []v1.Platform{}
+	}
+	p, err := parsePlatform(platform)
+	if err != nil {
+		return err
+	}
+	pv := platformValue{p}
+	ps.platforms = append(ps.platforms, *pv.platform)
+	return nil
+}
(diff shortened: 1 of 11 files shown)


tianon:
I gave Jon badly worded feedback that made him split the one command he'd created into two separate commands :sob:

"filter" for something that mutates seems like an odd word choice - his original choice was `crane amend`, and I thought that was potentially a large source of user confusion when it'll appear _right next_ to `crane append` (and "append" vs "amend" is not great - they're very related words but the two commands will do very different things and even operate on entirely different sets of objects)

IMO something like `crane index amend` would be totally reasonable, but it feels a little weird next to `crane manifest`, `crane config`, and `crane blob` that are all single-object-type getters _and_ it feels like overkill to `crane index` just for a _single_ subcommand so I don't know what to recommend :disappointed:

I'm sorry I made this harder Jon :sob: :heart: :bow:

imjasonh:
Don't beat yourself up @tianon 😄, I recommended `index filter` and `index append` because they feel like two separate operations to me. One removes manifests from an index, one adds them.

I think I like having `crane index` as a place to do index-specific operations. A lot of `crane` implicitly deals with images (`pull`, `append`, `export`), but I don't think it's worth having a `crane image` sub-tree. Agnostic stuff like `manifest`, `digest`, `blob` also make sense at the root and should stay there.

Since `index append` is different than `(image) append`, I think that's a good place to start filling out an index-specific sub-command.

cmd/crane/cmd/util.go
```diff
@@ -21,6 +21,35 @@ import (
...
 
+type platformsValue struct {
+	platforms []v1.Platform
+}
+
+func (ps *platformsValue) Set(platform string) error {
+	if ps.platforms == nil {
+		ps.platforms = []v1.Platform{}
+	}
+	p, err := parsePlatform(platform)
```
imjasonh, line 32: Should we use `v1.ParsePlatform` instead?

cmd/crane/cmd/util.go
```diff
@@ -21,6 +21,35 @@ import (
 	v1 "github.com/google/go-containerregistry/pkg/v1"
 )
 
+type platformsValue struct {
```
imjasonh, line 24: I could see things like `ko` using this too, is there a good way to share this?

pkg/v1/platform.go
```diff
@@ -85,6 +85,47 @@ func (p Platform) Equals(o Platform) bool {
...
+//
+// For OSFeatures and Features, Satisfies will return true if this Platform's
+// fields contain a superset of the values in the spec's fields (order ignored).
+func (p Platform) Satisfies(spec Platform) bool {
+	return satisfies(spec.OS, p.OS) &&
+		satisfies(spec.Architecture, p.Architecture) &&
+		satisfies(spec.Variant, p.Variant) &&
+		satisfies(spec.OSVersion, p.OSVersion) &&
+		satisfiesList(spec.OSFeatures, p.OSFeatures) &&
+		satisfiesList(spec.Features, p.Features)
```
imjasonh, line 103: Can we just leave features and osfeatures in the dustbin of history? Are they ever practically used?

cmd/crane/cmd/util.go
jonjohnsonjr: Nice.

pkg/v1/platform.go
jonjohnsonjr: I wanna say no just because OSFeatures is in the spec. We could probably drop Features, but it's no work to keep it given that we have to implement OSFeatures anyway.

cmd/crane/cmd/util.go
jonjohnsonjr: We should, but I'd like to play with refactoring `ko` separately before trying to guess the API we want.

pkg/v1/platform.go
tianon: It sounds like Jason's volunteering to open a spicy image-spec discussion :eyes: :hot_pepper:

(`features` is in the spec too: https://github.com/opencontainers/image-spec/blob/24027e1093941c9785f19475c8c1a4a8bef23c29/image-index.md?plain=1#L48-L86)

jonjohnsonjr:
@thesayyn FYI, would be interested in your feedback.

I'm not sure if I'm ready to go full-on CEL integration :P

imjasonh (approved):

thesayyn:
sorry, I was off last week due to the earthquake. this looks as good as it could get without having a DSL or using CEL.

PR #1563, opened 2023-02-09
tarball: pass imageToTags

thepwagner:
In `tarball.MultiWrite()`,  the `map[name.Reference]v1.Image` is inverted to a `map[v1.Image][]string` by 3 independent helpers. This PR modifies to invert the map once, then pass the result between helpers.

I'd made this change as part of a failed attempt at `tarball.WriteIndex()` - I realize now why that was a bad idea, but this snippet seemed worth saving.


imjasonh (approved):

imjasonh:
Thanks!

PR #1566, opened 2023-02-16
crane: add --flatten for index append

jonjohnsonjr:
While it's possible that folks might want to append an index to an index, it's more likely that they would want to merge them. This flag gives them an option to do either.

### DIFF
--- a/cmd/crane/cmd/index.go
+++ b/cmd/crane/cmd/index.go
@@ -260,7 +260,7 @@ The platform for appended manifests is inferred from the config file or omitted 
 	cmd.Flags().StringVarP(&newTag, "tag", "t", "", "Tag to apply to resulting image")
 	cmd.Flags().StringSliceVarP(&newManifests, "manifest", "m", []string{}, "References to manifests to append to the base index")
 	cmd.Flags().BoolVar(&dockerEmptyBase, "docker-empty-base", false, "If true, empty base index will have Docker media types instead of OCI")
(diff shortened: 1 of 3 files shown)


tianon:
I've tested this successfully with a list of images that contains both indexes and a non-index manifest and it flattened correctly :eyes:

cmd/crane/cmd/index.go
```diff
@@ -229,6 +260,7 @@ The platform for appended manifests is inferred from the config file or omitted
 	cmd.Flags().StringVarP(&newTag, "tag", "t", "", "Tag to apply to resulting image")
 	cmd.Flags().StringSliceVarP(&newManifests, "manifest", "m", []string{}, "References to manifests to append to the base index")
 	cmd.Flags().BoolVar(&dockerEmptyBase, "docker-empty-base", false, "If true, empty base index will have Docker media types instead of OCI")
+	cmd.Flags().BoolVar(&flatten, "flatten", false, "If true, appending an index will append each of its children rather than the index itself")
```
imjasonh, line 263:
```suggestion
	cmd.Flags().BoolVar(&flatten, "flatten", true, "If true, appending an index will append each of its children rather than the index itself")
```

because 

> it's more likely that they would want to merge them

Or maybe this means we want a `crane index concat` or something?

jonjohnsonjr: Done, but I can't believe you would willfully break all our existing users.

imjasonh (approved):

PR #1576, opened 2023-02-22
copy: Add auth-pairs keychain to support multi-repo auth

jotadrilo:
Signed-off-by: Joseda Rios <jdrios@vmware.com>

This PR adds a new `Keychain` to handle multiple repository authentication. See https://github.com/google/go-containerregistry/issues/1329.

The underlying implementation caches `Authenticator`s after a target is attempted to be resolved to avoid unnecessary I/O. It also fallbacks to the `DefaultKeychain` if no auth pairs match the target.

This is useful when a process requires using different auth for the same registry (host). For example, today is not possible to copy an image between two different GCR projects unless your service account has Reader/Writer roles in both projects, which is unlikely.

## Current Behavior

`crane` fails accessing the destination repository.

```shell
❯ crane auth login -u _json_key -p "$(cat jotadrilo-svc-account.json)" gcr.io
2023/02/22 20:13:46 logged in via /Users/jotadrilo/.docker/config.json

❯ crane \
     copy \
     gcr.io/jotadrilo/demo/busybox:latest \
     gcr.io/jotadrilo2/demo/busybox:test

2023/02/22 20:10:31 Copying from gcr.io/jotadrilo/demo/busybox:latest to gcr.io/jotadrilo2/demo/busybox:test
Error: failed to copy index: GET https://gcr.io/v2/token?scope=repository%3Ajotadrilo2%2Fdemo%2Fbusybox%3Apush%2Cpull&service=gcr.io: DENIED: Token exchange failed for project 'jotadrilo2'. Caller does not have permission 'storage.buckets.get'. To configure permissions, follow instructions at: https://cloud.google.com/container-registry/docs/access-control
exit status 1
```

## New Behavior

`crane` copies the image successfully.

```shell
❯ crane auth login -u _json_key -p "$(cat jotadrilo-svc-account.json)" gcr.io
2023/02/22 20:13:46 logged in via /Users/jotadrilo/.docker/config.json

❯ DOCKER_CONFIG="${PWD}/jotadrilo2" crane auth login -u _json_key -p "$(cat jotadrilo2-svc-account.json)" gcr.io
2023/02/22 20:13:46 logged in via /Users/jotadrilo/personal/go-containerregistry/cmd/crane/jotadrilo2/config.json

❯ crane \
      copy \
      gcr.io/jotadrilo/demo/busybox:latest \
      gcr.io/jotadrilo2/demo/busybox:test \
      --auth-pair gcr.io/jotadrilo2/demo/busybox:jotadrilo2

2023/02/22 20:20:24 Copying from gcr.io/jotadrilo/demo/busybox:latest to gcr.io/jotadrilo2/demo/busybox:test
2023/02/22 21:00:18 existing blob: sha256:f78e6840ded1aafb6c9f265f52c2fc7c0a990813ccf96702df84a7dcdbe48bea
2023/02/22 21:00:20 pushed blob: sha256:c00904b679bb7edaf852620402b7eb164ceea14d9338240addfed833061c36ca
2023/02/22 21:00:22 gcr.io/jotadrilo2/demo/busybox@sha256:7b74ce83a3a90d10e9b4c7473dd3dbec515a409b0cedfb0de342832bb582c26e: digest: sha256:7b74ce83a3a90d10e9b4c7473dd3dbec515a409b0cedfb0de342832bb582c26e size: 502
2023/02/22 21:00:23 existing blob: sha256:205dae5015e78dd8c4d302e3db4eb31576fac715b46d099fe09680ba28093a7a
2023/02/22 21:00:26 pushed blob: sha256:04bfb6f8e19b909595598e4a8f068f337df3a299b637daaf937fb80e004dfae2
2023/02/22 21:00:27 gcr.io/jotadrilo2/demo/busybox@sha256:417bf965b4178fd94d3cbaf8e16b3c3a500e2b2cca000e8fe9a24e8403bbfbe9: digest: sha256:417bf965b4178fd94d3cbaf8e16b3c3a500e2b2cca000e8fe9a24e8403bbfbe9 size: 502
2023/02/22 20:20:27 gcr.io/jotadrilo2/demo/busybox:test: digest: sha256:7d80e0e207005f658e9b6ddffa577f2d1ddc1f7c9f7da3f0c2bec99f01bfdbd9 size: 683
```

### DIFF
--- a/cmd/crane/cmd/copy.go
+++ b/cmd/crane/cmd/copy.go
@@ -15,20 +15,30 @@
 package cmd
 
 import (
+	"github.com/google/go-containerregistry/pkg/authn"
 	"github.com/google/go-containerregistry/pkg/crane"
 	"github.com/spf13/cobra"
 )
 
 // NewCmdCopy creates a new cobra.Command for the copy subcommand.
 func NewCmdCopy(options *[]crane.Option) *cobra.Command {
-	return &cobra.Command{
+	authPairs := &authPairsValue{}
+
+	cmd := &cobra.Command{
 		Use:     "copy SRC DST",
 		Aliases: []string{"cp"},
 		Short:   "Efficiently copy a remote image from src to dst while retaining the digest value",
 		Args:    cobra.ExactArgs(2),
 		RunE: func(_ *cobra.Command, args []string) error {
 			src, dst := args[0], args[1]
+
+			*options = append(*options, crane.WithAuthFromKeychain(authn.NewAuthPairsKeychain(authn.WithAuthPairs(authPairs.authPairs))))
+
 			return crane.Copy(src, dst, *options...)
 		},
 	}
+
+	cmd.Flags().Var(authPairs, "auth-pair", "Authentication pairs in the form REMOTE:DOCKER_CONFIG. Example: acme.io/path/to/image:/tmp/acme-io-docker-config-dir")
+
+	return cmd
 }
--- a/cmd/crane/cmd/copy.go
+++ b/cmd/crane/cmd/copy.go
@@ -32,7 +32,7 @@ func NewCmdCopy(options *[]crane.Option) *cobra.Command {
 		RunE: func(_ *cobra.Command, args []string) error {
 			src, dst := args[0], args[1]
 
-			*options = append(*options, crane.WithAuthFromKeychain(authn.NewAuthPairsKeychain(authn.WithAuthPairs(authPairs.authPairs))))
+			*options = append(*options, crane.WithAuthFromKeychain(authn.NewAuthPairsKeychain(authPairs.authPairs)))
 
 			return crane.Copy(src, dst, *options...)
 		},
--- a/cmd/crane/cmd/util.go
+++ b/cmd/crane/cmd/util.go
@@ -15,8 +15,10 @@
 package cmd
 
 import (
+	"fmt"
 	"strings"
 
+	"github.com/google/go-containerregistry/pkg/authn"
 	v1 "github.com/google/go-containerregistry/pkg/v1"
 )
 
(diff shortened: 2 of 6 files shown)


jotadrilo:
@imjasonh I saw there is a PR to fix the same issue.

Feel free to close this one if you believe this implementation is not aligned with your design principles.

jotadrilo:
Hello @jonjohnsonjr, did you have a chance to check this?

jonjohnsonjr:
Sorry for not responding to this for so long. I was having a hard time understanding the issue.

If https://github.com/google/go-containerregistry/issues/1329 were fixed, we wouldn't need this right? You could just have a config file that respected subdirectories?

IMO, I'd be more inclined to fork the config parser stuff from docker. We used to have our own implementation, but I wanted to use their upstream to avoid hardcoding a bunch of magical values. It would probably be better to just hardcode those magical values than to add to this complexity.

jotadrilo:
> If #1329 were fixed, we wouldn't need this right? You could just have a config file that respected subdirectories?

Indeed, this MR aims to fix #1329. However, the current format on the config file does not respect "subdirectories": it is a map of hostnames to authn objects.

> IMO, I'd be more inclined to fork the config parser stuff from docker. We used to have our own implementation, but I wanted to use their upstream to avoid hardcoding a bunch of magical values. It would probably be better to just hardcode those magical values than to add to this complexity.

I don't fully follow you, but I am probably missing details on what was implemented in the past.

What do you mean by magical values specifically?

It would be nice if upstream adds some flexibility to the docker config file schema, so it is not just a `map[host]:authn` JSON.

I didn't want to add a new, custom, schema in crane, just to cover this uncovered case in the docker config file schema. Hence, this implementation only adds the possibility of mapping a repository to a specific docker config that you have previously crafted (either with `crane login` or with `docker login`) using specific credentials. See the "New Behavior" section in the PR description.

PR #1582, opened 2023-03-02
harbor-registry use basic auth

testwill:
issue #1493: go-containerregistry catalog harbor-registry fail


imjasonh:
I'm a little confused about this change. The last update in #1493 sounds like this is a bug in Harbor. Is this a temporary workaround for that issue? Is there a Harbor issue we should link so we can remove the workaround when it's no longer needed?

testwill:
https://github.com/goharbor/harbor/issues/17834
This is a temporary solution, they haven't dealt with it for a long time, and then they said that we need to replace it with basic auth

PR #1586, opened 2023-03-06
crane: add serve subcommand

### LINKED ISSUE #1579: FR: implement crane serve
Currently, go-containerregistry has its own registry implementation which is mostly used for testing. It would be useful to expose this functionality for others to use via `crane serve` command.

imjasonh:
Fixes #1579 

Previously, #1211, which was shelved until we had an on-disk implementation, which never materialized. If/when that comes along we can support a persistent `crane serve`.

### DIFF
--- a/cmd/crane/cmd/serve.go
+++ b/cmd/crane/cmd/serve.go
@@ -1,0 +1,69 @@
+// Copyright 2023 Google LLC All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package cmd
+
+import (
+	"errors"
+	"fmt"
+	"log"
+	"net/http"
+	"os"
+	"time"
+
+	"github.com/spf13/cobra"
+
+	"github.com/google/go-containerregistry/pkg/registry"
+)
+
+func NewCmdServe() *cobra.Command {
+	port := os.Getenv("PORT")
+	if port == "" {
+		port = "8080"
+	}
+
+	return &cobra.Command{
+		Use:   "serve",
+		Short: "Serve an in-memory registry implementation",
+		Long: `This sub-command serves an in-memory registry implementation on port :8080 (or $PORT)
+
+The command blocks while the server accepts pushes and pulls.
+
+Contents are only stored in memory, and when the process exits, pushed data is lost.`,
+		Args: cobra.NoArgs,
+		RunE: func(cmd *cobra.Command, _ []string) error {
+			ctx := cmd.Context()
+			s := &http.Server{
+				Addr:              fmt.Sprintf(":%s", port),
+				ReadHeaderTimeout: 5 * time.Second, // prevent slowloris, quiet linter
+				Handler:           registry.New(),
+			}
+			log.Printf("serving on port %s", port)
+
+			errCh := make(chan error)
+			go func() { errCh <- s.ListenAndServe() }()
+
+			<-ctx.Done()
+			log.Println("shutting down...")
+			if err := s.Shutdown(ctx); err != nil {
+				return err
+			}
+
+			if err := <-errCh; !errors.Is(err, http.ErrServerClosed) {
+				return err
+			}
+			return nil
+		},
+	}
+}
--- a/cmd/crane/cmd/root.go
+++ b/cmd/crane/cmd/root.go
@@ -94,7 +94,7 @@ func New(use, short string, options []crane.Option) *cobra.Command {
 		},
 	}
 
-	commands := []*cobra.Command{
+	root.AddCommand(
 		NewCmdAppend(&options),
 		NewCmdAuth(options, "crane", "auth"),
 		NewCmdBlob(&options),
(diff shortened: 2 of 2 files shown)


cmd/crane/cmd/serve.go
```diff
@@ -0,0 +1,69 @@
...
+	"time"
+
+	"github.com/spf13/cobra"
+
+	"github.com/google/go-containerregistry/pkg/registry"
+)
+
+func NewCmdServe() *cobra.Command {
+	port := os.Getenv("PORT")
+	if port == "" {
```
thesayyn, line 32: could we support `kernel selected port (0)`?

thesayyn: that'd be really helpful for spawning multiple instances of registry. my use case is bazel actions

imjasonh: Sure, that seems reasonable. Is there a way you'd prefer to do that? Some quick googling found [`freeport`](https://pkg.go.dev/github.com/phayes/freeport#GetFreePort) which looks promising, but I've never used it.

thesayyn: my implementation of registry looks like this; https://github.com/bazel-contrib/rules_oci/blob/main/e2e/custom_registry/registry/registry.go using `net.Listen`

thesayyn: I prefer this as it guarantees that you'll be allocated a port by the kernel.

cmd/crane/cmd/serve.go
```diff
@@ -0,0 +1,75 @@
...
+The command blocks while the server accepts pushes and pulls.
+
+Contents are only stored in memory, and when the process exits, pushed data is lost.`,
+		Args: cobra.NoArgs,
+		RunE: func(cmd *cobra.Command, _ []string) error {
+			ctx := cmd.Context()
+
+			port := os.Getenv("PORT")
+			if port == "" {
+				porti, err := freeport.GetFreePort()
```
thesayyn, line 46: I have a concern here. since freeport closes the port after it found it, it may have been already allocated to some other process by the time registry starts listening on this port. 

there's a tiny chance that this could happen but could happen when there's pressure on the number of free ports or kernel settings are different. (eg kernel marks the closed port as available immediately.)

thesayyn: from a blog post i read a while back; 

The most important problem here is that the port isn’t guaranteed to still be available when the second program tries to use it. In practice under normal circumstances it almost always will be available for two reasons. The first is that in the common case that you’re using a TCP port your TCP stack will put the port into the so-called TIME_WAIT state. When this happens the kernel will reserve the port (actually, the interface/port tuple) for some amount of time before allowing another program to use it. On Linux this defaults to two minutes, so if you can get the second program to bind in under two minutes you’re usually in the clear. The second reason is that in practice the kernel on most systems---and at the very least on Linux---will try to allocate ports in this way pseudo-sequentially. What this means is that if you bind on port 0 and actually bind to port 55441, then the next process that tries to bind to port 0 will probably get port 55442 (if it’s available), and the next process will get port 55443, and so on. Eventually this will wrap around which means that two processes that bind to port 0 could eventually get allocated the same port. In practice unless the number of free ports is extremely small usually it will take a long time to wrap around and therefore this isn’t usually an issue.

imjasonh: Oh. Uh. I guess?

I guess my question is, what are you trying to do with `crane serve`? If you're talking about starting and stopping so many local registries that this race is a concern, are you sure invoking `crane serve` is the best way to do this?

You could relatively easily write a lil Go program that spins up a thousand registries and makes sure none of them have conflicting ports, if that's what you're going for. Or, without having `crane` use `freeport` directly, have something that requetss 1000 free ports and runs `PORT=<port> crane serve` if you want. `crane` probably won't gain the ability to choose a free non-conflicting port across `crane` invocations that doesn't have the race you're talking about.

I didn't realize the intention here was to be able to spin up a bunch of non-conflicting local registries all in-memory -- that's probably pretty quickly going to end up in a bad place, since those copies won't be able to share any of your relatively limited in-memory storage.

alexeagle: I think the concern is just because Bazel is a scheduler, and if you're using `crane serve` inside an action (build step) then Bazel may schedule several concurrent actions. The concurrent job count is usually a heuristic based on the number of CPUs, so I guess typically 4-72 is the upper-bound of how many copies of crane we might try to run.

imjasonh: Makes sense. I guess in that case asking bazel to invoke one command that starts and manages N registries is out of the question.

Is https://github.com/bazel-contrib/rules_oci/blob/main/e2e/custom_registry/registry/registry.go a working solution for you? Is the ask then to just move that into crane?

thesayyn: > Is the ask then to just move that into crane?

just the part where net.Listen is preferred over freeport. net.Listen will guarantee that we are still allocating the port after having found it therefore no risk of race-condition due to port pressure or kernel configuration. 

> You could relatively easily write a lil Go program that spins up a thousand registries and makes sure none of them have conflicting ports

this still wouldn't work as this issue is independent of the scheduler.

imjasonh: Now using `net.Listener` as in your example. Thanks for the suggestion!

thesayyn (approved):
awesome

thesayyn:
just a thought; maybe `crane registry serve` is better than `crane serve`? I like the explicitness of the commands.

imjasonh:
> just a thought; maybe `crane registry serve` is better than `crane serve`? I like the explicitness of the commands.

`crane registry serve` sgtm

jonjohnsonjr (approved):

thesayyn:
it looks like it's been a while since the latest release hehe. It may be a new release time already. 🫡

imjasonh:
Yeah I can cut a release next week.

thesayyn:
> Yeah I can cut a release next week.

awesome

imjasonh:
> > Yeah I can cut a release next week.
> 
> awesome

Jon did it: https://github.com/google/go-containerregistry/releases/tag/v0.14.0

thesayyn:
You guys are awesome

Issue #1587, opened 2023-03-07
question: crane auth logout


Danielku15:
We plan to use `crane auth login` in our CI pipeline (filled via secrets) and want to remove after further actions those credentials again. 
I don't see a `crane auth logout` in the docs. How can we remove these credentials again?

imjasonh:
This has been requested before (recently, https://github.com/google/go-containerregistry/pull/1459) and it's something I think would be useful. If you're interested in sending a PR I'd be happy to review it. If not, I can throw something together when I have some time.

PR #1589, opened 2023-03-09
add crane auth logout

imjasonh:
Fixes https://github.com/google/go-containerregistry/issues/1587

NB: This doesn't clear creds from Podman-configured auth, which `authn.DefaultKeychain` uses if a Docker config file isn't found. This might be confusing, but `crane auth login` doesn't use that either, so 🤷.

It might make more sense to always `login` using `DOCKER_CONFIG` and `logout` of Docker- or Podman-configured auth, feedback welcome.


imjasonh:
Some things I've learned about the `Erase` method, and by extension, `docker logout`:

- it doesn't unset cred helpers for the server
- it doesn't complain if creds already aren't defined for the server

I think I'd like to fix both of those for `crane auth logout`, but I'd be willing to be convinced that following `docker`'s behavior exactly is better.

jonjohnsonjr (approved):

PR #1592, opened 2023-03-09
bump deps using ./hack/bump-deps.sh

imjasonh:
Also bump Go versions used to build and test to 1.19 and 1.20, the only two supported Go versions


imjasonh:
Removed `Build (1.18)` and `Test (1.18)` as required presubmits so this can merge. I'll add the 1.20 equivalents as required after this merges.

jonjohnsonjr (approved):

Issue #1590, opened 2023-03-09
question: Benefit of Crane Flatten


AndrewCharlesHay:
What does `Flattening an image's layers into a single layer` do exactly?
Does it reduce the size of the image?
What are the benefits and drawbacks?
Is it similar to [Docker Squash](https://github.com/goldmann/docker-squash)?

imjasonh:
Here's an example. If an image had three layers:

```
layer 1 (bottom): file at /tmp/foo containing the text "foo"
layer 2 (middle): file at /tmp/bar containing the text "bar"
layer 3 (top): file at /tmp/foo containing the text "blah"
```

Flattening this image would produce an image with one layer:

```
layer: file at /tmp/foo containing the text "blah", file at /tmp/bar containing the text "bar"
```

In this example it reduced the size of the image -- the version of the file with contents `"foo"` is gone from the image because it was shadowed by the top-layer file with the same name.

This can be a drawback in some cases, due to caching. If there's a base layer that's shared by many images, you don't need to pull the contents of all three files to pull the example image -- the bottom two layers might already be present, so you'd only pull `"blah"`. Common base images like `ubuntu`, `debian`, `alpine`, etc. are very often shared by many images, and their layers are heavily cached and don't usually have to be pulled to pull a new image based on those images.

`crane flatten` sounds similar to Docker Squash. I've never used it though, so I don't know exactly what similarities or differences there might be. At a glance it looks like Docker Squash uses the Docker Daemon to manage images, which normally requires privilege to run. `crane flatten` can pull, flatten and push an image without involving Docker, and without privilege. There may be other differences.

AndrewCharlesHay:
@imjasonh Awesome thank you! I personally feel like this information would be useful in the documentation for the command

PR #1594, opened 2023-03-13
docs: add explination

AndrewCharlesHay:
Copied @imjasonh's response from https://github.com/google/go-containerregistry/issues/1590 to documentation


imjasonh:
This file is generated from go code. If we want to keep this we should either add it to go code docs, or (my preference) maybe add a new doc that's not generated, that's linked to from the generated docs.

Issue #1600, opened 2023-03-15
v0.14.0 release failed provenance validation


imjasonh:
The release workflow failed during validation: https://github.com/google/go-containerregistry/actions/runs/4430435438/jobs/7772326986#step:4:18

```
Verifying go-containerregistry_Windows_arm64.tar.gz
/home/runner/work/_temp/98950ea0-ea5c-4090-bec7-965fdc0ecfc2.sh: line 6: ./slsa-verifier-linux-amd64: No such file or directory
```

See also: https://github.com/ko-build/ko/issues/978

cc @ianlewis @laurentsimon

PR #1601, opened 2023-03-15
Fix SLSA verification

laurentsimon:
closes https://github.com/google/go-containerregistry/issues/1600

Changes in this PR:
- Updated the command line, since the Action uses v2.x installation
- Updated the binary call. The Action installs it in the `PATH` and the binary name is `slsa-verifier`


imjasonh (approved):
Thanks!

Issue #1595, opened 2023-03-14
Add support for filepath-based `authn.Keychain`


hasheddan:
Currently, it is possible to configure the location of the auth config file in the [default keychain](https://github.com/google/go-containerregistry/blob/1cfe1fc25f233b40aa5d3b0edd572ed5c3f854c9/pkg/authn/keychain.go#L81), but only if there is not already a config file at the default path. Furthermore, the name of the file (`config.json`) is hard-coded and cannot be overridden. It would be helpful to support passing an arbitrary path to credentials to be used, perhaps as part of a multi-keychain. This would be straightforward to implement in consume in the same manner as other alternative keychain implementations, such as `github` and `k8schain`.

imjasonh:
This seems valuable. There are at least a couple ways we could do it, and I'm curious which one fits your needs best:

1. document an env var that `authn.DefaultKeychain` will look for _before_ considering `${DOCKER_CONFIG}` (e.g., `GGCR_DOCKER_CONFIG`) which specifies the full path of the file. For example, `GGCR_DOCKER_CONFIG=test/cfg.json crane ls ...`.
2. implement `authn.NewConfigKeychain(filename string)` that returns the otherwise normal `DefaultKeychain` implementation, but pointed at the full filename instead of `${DOCKER_CONFIG}/config.json`, for example:

```
remote.Image(ref,
    remote.WithAuthFromKeychain(authn.NewConfigKeychain("test/cfg.json"))
```

@jonjohnsonjr might have even better ideas.

hasheddan:
@imjasonh I think both could be implemented as independent functionality, but I was thinking more in the vein of (2) as it allows the natural keychain precedence to be used with potentially multiple config files.

PR #1603, opened 2023-03-16
authn: Add NewConfigKeychain to load a config from explicit path

imjasonh:
Fixes https://github.com/google/go-containerregistry/issues/1595

cc @hasheddan

### DIFF
--- a/pkg/authn/keychain.go
+++ b/pkg/authn/keychain.go
@@ -50,6 +50,9 @@ type Keychain interface {
 type defaultKeychain struct {
 	mu sync.Mutex
 
+	once sync.Once
+	cfg  types.AuthConfig
+
 	configFilePath string
 }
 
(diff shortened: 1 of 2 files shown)


pkg/authn/keychain.go
```diff
@@ -97,6 +117,34 @@ func (dk *defaultKeychain) Resolve(target Resource) (Authenticator, error) {
...
+	}
+	return cf, nil
+}
+
+func (dk *defaultKeychain) Resolve(target Resource) (Authenticator, error) {
+	dk.mu.Lock()
+	defer dk.mu.Unlock()
+
+	var cf *configfile.ConfigFile
+	if dk.configFilePath == "" {
```
jonjohnsonjr, line 137: Should we sync.Once this so we don't lock the _entire_ `Resolve`? Why was the mutex added originally?

jonjohnsonjr:
Want to rebase this given the refresher stuff?

jonjohnsonjr (approved):

PR #1604, opened 2023-03-18
Support `Warning` header aggregation and reporting

imjasonh:
https://github.com/opencontainers/distribution-spec/pull/393 seems to be on a good path, so I figured I'd put in a better implementation of warning collection/reporting in ggcr.

- adds `remote.WithWarnCollector` and `crane.WithWarnCollector` which take a `remote.WarnCollector` interface
- pipes through warning collection to all the places we make HTTP calls
- adds an unexported implementation of `WarnCollector` in `cmd/crane` that dedupes warnings and reports them to `logs.Warn` with coloring (unless [`NO_COLOR`](https://no-color.org/) is set)
- adds `registry.WithWarning` to `pkg/registry` that probabilistically reports the given warning (`WithWarning(0.1, "hi")` warns `"hi`" 10% of the time)

This also updates `cmd/registry` to add demo warnings some % of the time; I'll revert this before merging, this is just for demo purposes. Speaking of demo:

```
$ go run ./cmd/registry &
$ go run ./cmd/crane cp alpine localhost:1338/alpine
...
2023/03/18 16:40:16 [WARNING]: 60% of the time, it works every time.
2023/03/18 16:40:16 [WARNING]: This registry is cool.
```

Logic for parsing the warning message out of the header is pretty janky, and could use some tests.

Clients that use go-containerregistry won't suddenly start reporting warnings without changes. They'll need to inject a `WarnCollector` like `crane` does -- if we think `crane`'s approach is generally useful we can make it available in `pkg/crane` itself.


cmd/crane/cmd/root.go
```diff
@@ -146,3 +152,27 @@ func (ht *headerTransport) RoundTrip(in *http.Request) (*http.Response, error) {
...
+}
+
+func (d *warnCollector) Report() {
+	warns := make([]string, 0, len(d.warns))
+	for k := range d.warns {
+		warns = append(warns, k)
+	}
+	sort.Strings(warns)
+	prefix := "\033[1;33m[WARNING]\033[0m:"
+	if v, set := os.LookupEnv("NO_COLOR"); set && v != "0" {
```
liggitt, line 171: there's also various detection approaches to try to see if the writer is a terminal that we think can display color output - xref https://github.com/kubernetes/kubernetes/blob/fe91bc257b505eb6057eb50b9c550a7c63e9fb91/staging/src/k8s.io/kubectl/pkg/util/term/term.go - it might make sense to only enable color output automatically when we think we're in an interactive terminal

jonjohnsonjr:
I think there are a couple ways we could go with this that will be slightly better than adding `WithWarnCollector` to `crane` and `remote`.

We already have [`logs.Warn`](https://pkg.go.dev/github.com/google/go-containerregistry@v0.14.0/pkg/logs#Warn) (used [here](https://github.com/google/go-containerregistry/pull/1604/files#diff-04d49325bf28b04678340481ee8de048890fc2fa5701460121459ada4b0d3081R175) in this PR currently). We could just have warnings be written to `logs.Warn` where we set `logs.Warn` to do warncollectory things.

There's also [`remote.WithProgress`](https://pkg.go.dev/github.com/google/go-containerregistry@v0.14.0/pkg/v1/remote#WithProgress) that we currently use to emit very specific kinds of updates. We could do something similar (or even use the same option) to send clients more kinds of events, e.g. these Warnings.

I really dislike the `logs` package -- it would probably be better if we were just emitting events and having consumers do things with them...

There's also the [`slog`](https://pkg.go.dev/golang.org/x/exp/slog) stuff that has its own `Warn` level, which maybe we could replace the `logs` package with 🤔 

How do you feel about this _not_ being a public API, and instead we have crane pass in a custom transport that does all these warncollectory things until we have a happier public interface?

cmd/crane/cmd/root.go
```diff
@@ -146,3 +152,27 @@ func (ht *headerTransport) RoundTrip(in *http.Request) (*http.Response, error) {
 	}
 	return ht.inner.RoundTrip(in)
 }
+
+type warnCollector struct {
+	warns map[string]struct{}
+}
+
+func (d *warnCollector) Add(warning string) {
+	d.warns[warning] = struct{}{}
```
liggitt, line 161: one thing to watch out for is accumulation without cleanup in a long-running process... if the lifetime of the collector is expected to be ~short, this is probably fine. if the lifetime could be long-running, this could grow without bound

imjasonh: In `crane`'s context I think we can expect this to be pretty short-lived, most invocations are at most "copy this image to this image" and there's not a lot of opportunity for it to do anything else.

But I agree this is a good thing to consider for long-running things that may collect logs of warnings.

imjasonh:
> How do you feel about this not being a public API, and instead we have crane pass in a custom transport that does all these warncollectory things until we have a happier public interface?

I'm fine with that. I'll get something along these lines up shortly.

cmd/crane/cmd/root.go
imjasonh: Thanks, this was helpful. I omitted the `github.com/moby/term.IsTerminal` call for now mainly to avoid the dependency. We might add it later, it looks pretty small but it seems to bring in `github.com/Azure/go-ansiterm` for some reason too.

cmd/crane/cmd/root.go
```diff
@@ -146,3 +159,76 @@ func (ht *headerTransport) RoundTrip(in *http.Request) (*http.Response, error) {
...
+	for k := range wt.warns {
+		warns = append(warns, k)
+	}
+	sort.Strings(warns)
+	prefix := "\033[1;33m[WARNING]\033[0m:"
+	if nocolor() {
+		prefix = "[WARNING]:"
+	}
+	for _, w := range warns {
+		fmt.Fprintln(os.Stderr, prefix, w)
```
jonjohnsonjr, line 211:
```suggestion
		// TODO: Consider using logs.Warn here if we move this out of crane.
		fmt.Fprintln(os.Stderr, prefix, w)
```

jonjohnsonjr (approved):

Issue #1515, opened 2022-12-19
crane: feature request mutate --workdir


thesayyn:
### Describe the bug

Currently, the `mutate` command has no option to set the working dir of the config. 

### Expected behavior

`mutate` should have a  `--workdir` option as it is commonly used.

### Additional context

- Output of `crane version`
- Registry used (e.g., GCR, ECR, Quay)

thesayyn:
this is still relevant

PR #1615, opened 2023-03-26
Adding `mutate` `--workdir` option to set the working directory

eglookz:
Will add the feature #1515


imjasonh:
You'll also need to run `./hack/update-codegen.sh` to regenerate the Markdown docs to include the new flag.

Otherwise looks good to me! Thanks for adding this.

eglookz:
@imjasonh I apologize, I made a regrettable mistake in the penultimate commit.

PR #1624, opened 2023-04-05
Refresh authn.DefaultKeychain creds every 5 min

jonjohnsonjr:
Sometimes Authenticators stick around for a while and eventually expire. This changes modifies Authenticators returned by DefaultKeychain to re-resolve after 5 minutes to ensure the underlying creds are fresh.

The hard-coded 5 minutes should prevent us from spamming the underlying Keychain (it can be expensive) while also preventing the creds from expiring.

### DIFF
--- a/pkg/authn/keychain.go
+++ b/pkg/authn/keychain.go
@@ -18,6 +18,7 @@ import (
 	"os"
 	"path/filepath"
 	"sync"
+	"time"
 
 	"github.com/docker/cli/cli/config"
 	"github.com/docker/cli/cli/config/configfile"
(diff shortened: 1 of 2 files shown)


pkg/authn/keychain.go
```diff
@@ -178,3 +179,62 @@ func (w wrapper) Resolve(r Resource) (Authenticator, error) {
 	}
 	return FromConfig(AuthConfig{Username: u, Password: p}), nil
 }
+
+type refreshingKeychain struct {
```
imjasonh, line 183: wdyt about exporting this as `authn.RefreshingKeychain(inner authn.Keychain, d time.Duration)`

imjasonh (approved):
Love it.

pkg/authn/keychain.go
jonjohnsonjr: Sure, done.

PR #1628, opened 2023-04-06
fix(options): pass retryBackoff and retryPredicate options to transport

aslafy-z:


pkg/v1/remote/options.go
```diff
@@ -89,6 +89,7 @@ var retryableStatusCodes = []int{
 	http.StatusBadGateway,
 	http.StatusServiceUnavailable,
 	http.StatusGatewayTimeout,
+	http.StatusTooManyRequests,
```
imjasonh, line 92: I'd prefer to make this change as a separate PR. Retrying in the event of rate limits is a complex subject, and automatically retrying here can make problems worse instead of better.

imjasonh (commented):
@jonjohnsonjr aside from the 429 change, do you see a reason we shouldn't retry with the custom semantics when we're also injecting a custom transport?

pkg/v1/remote/options.go
aslafy-z: Ok, I will move this change to a separate PR.

jonjohnsonjr:
I had these be intentionally separate because if you want to control the transport retry behavior, you should be using `remote.WithTransport`.

aslafy-z:
@jonjohnsonjr if so, I don't get what are all these options for..

pkg/v1/remote/options.go
aslafy-z: Done with #1636 and #1635

block-danm:
@jonjohnsonjr Could you expand on
>  if you want to control the transport retry behavior, you should be using remote.WithTransport.

please?

There are three settings that control retry behavior - status codes, predicate, and backoff - and currently all three are set in different ways:
* the status codes that trigger a retry attempt are set from `o.retryStatusCodes`, so they account for any settings provided with `config.WithRetryStatusCodes()`;
* the retry predicate is _always_ set to the `defaultRetryPredicate` defined in `options.go`;
* the backoff isn't explicitly set in `makeOptions()`, so it falls back to [the default](https://github.com/google/go-containerregistry/blob/dbcd01c402b2f05bcf6fb988014c5f37e9b13559/pkg/v1/remote/transport/retry.go#L25) defined in `retry.go`.

This inconsistency is quite unexpected and confusing, and this PR would immediately resolve that confusion.


I am also very curious about @aslafy-z 's [question](https://github.com/google/go-containerregistry/pull/1628#issuecomment-1500164538) - if `config.WithRetryBackoff()` does not affect the retry backoff used by the actual transport, what _is_ it for?

aslafy-z:
@imjasonh @jonjohnsonjr Can you please reconsider this PR? Thank you

aslafy-z:
This PR is still desired.

aslafy-z:
I just rebased and fixed conflicts to this PR. Now that https://github.com/google/go-containerregistry/pull/1635 is merged, can we merge that one? Thank you

hayk99:
is this going to be merged soon?

aslafy-z:
I would love to see this merged..

aslafy-z:
This is not stale.

markusthoemmes:
Ack on this PR. @jonjohnsonjr or @imjasonh it feels like the current implementation is a bug in that the respective retry mechanisms aren't consistently applied. Or am I missing something?

aslafy-z:
@Subserial Any chance a release gets cut anytime soon?

PR #1633, opened 2023-04-06
Implement remote.Pusher

jonjohnsonjr:
The Pusher handles multiplexing across multiple repositories and deduplicating in-flight uploads. This removes the need to do an awkward dance around re-using auth handshakes by passing custom transports.

This also fixes limitations in remote.MultiWrite around streaming layers and multiple repositories.

### DIFF
--- a/pkg/v1/remote/multi_write_test.go
+++ b/pkg/v1/remote/multi_write_test.go
@@ -21,21 +21,52 @@ import (
 	"net/http"
 	"net/http/httptest"
 	"net/url"
+	"os"
 	"strings"
 	"testing"
 
 	"github.com/google/go-containerregistry/pkg/authn"
+	"github.com/google/go-containerregistry/pkg/logs"
 	"github.com/google/go-containerregistry/pkg/name"
 	"github.com/google/go-containerregistry/pkg/registry"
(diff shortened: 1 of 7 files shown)


pkg/v1/remote/multi_write_test.go
```diff
@@ -21,21 +21,52 @@ import (
...
 	"github.com/google/go-containerregistry/pkg/v1/random"
 	"github.com/google/go-containerregistry/pkg/v1/remote/transport"
+	"github.com/google/go-containerregistry/pkg/v1/stream"
 	"github.com/google/go-containerregistry/pkg/v1/types"
 	"github.com/google/go-containerregistry/pkg/v1/validate"
 )
 
+func streamable(t *testing.T) v1.Layer {
+	t.Helper()
+	rl, err := random.Layer(1024, types.OCIUncompressedLayer)
```
imjasonh, line 44: Isn't this a long way to say `NopCloser(LimitReader(Rand, n))`?

imjasonh (approved):

pkg/v1/remote/multi_write_test.go
jonjohnsonjr: I wanted it to be a valid filesystem for some reason.

PR #1635, opened 2023-04-07
feat(remote): make retryStatusCodes configurable

aslafy-z:
Add `retryStatusCodes` option and `remote.WithRetryStatusCodes` to customize retryable status codes list.

PR #1636, opened 2023-04-07
feat(remote): retry on TooManyRequests

aslafy-z:
Add HTTP 429 `TooManyRequests` to the list of retryable status codes.

### DIFF
--- a/pkg/v1/remote/options.go
+++ b/pkg/v1/remote/options.go
@@ -89,6 +89,7 @@ var retryableStatusCodes = []int{
 	http.StatusBadGateway,
 	http.StatusServiceUnavailable,
 	http.StatusGatewayTimeout,
+	http.StatusTooManyRequests,
 	499,
 }
 


imjasonh:
As previously commented at https://github.com/google/go-containerregistry/pull/1628#discussion_r1160081257

> Retrying in the event of rate limits is a complex subject, and automatically retrying here can make problems worse instead of better.

If we go ahead with https://github.com/google/go-containerregistry/pull/1635 (which mostly seems fine) and you pass in 429 as a retryable code, that seems fine to me, but I wouldn't want to make this the new behavior for every dependency of this code.

But to Jon's comment in https://github.com/google/go-containerregistry/pull/1628#issuecomment-1499536861 I wonder if you might be better off just passing in `WithTransport` and having complete control over the retry behavior yourself, without having to add a new option.

If you've tried that and it was insufficient or painful in some way, that would be helpful for motivating this change.

aslafy-z:
@imjasonh I agree, this change can break things in some setups. I'd prefer to reuse the pre-configured Transport you `go-containerregistry` provides with all the retry behaviors. Let's close this PR and go ahead with #1635.

PR #1638, opened 2023-04-07
Cleanup: Switch the debug image to `cgr.dev/chainguard/busybox`

mattmoor:
:broom: This doesn't have `glibc` (like `base`) and just has the `busybox` necessities.

/kind cleanup

### DIFF
--- a/.ko/debug/.ko.yaml
+++ b/.ko/debug/.ko.yaml
@@ -1,1 +1,1 @@
-defaultBaseImage: gcr.io/distroless/base:debug
+defaultBaseImage: cgr.dev/chainguard/busybox


jonjohnsonjr (approved):

antonvdl:
Using chainguard busybox is causing issues in GitLab CI/CD with Gitlab runners.
There is a non-working /bin/bash symlink to /bin/busybox which results in an error; gitlab-runner checks if /bin/bash is there and uses this because of the symlink is present.

As soon as /bin/bash is not present the shell detection in gitlab-runner will look further and uses the correct busybox shell.

PR #1644, opened 2023-04-11
Add remote.Puller

jonjohnsonjr:
This PR adds a Puller implementation and uses it in `crane ls` and `crane catalog` to stream results for each page.

### DIFF
--- a/pkg/v1/remote/catalog.go
+++ b/pkg/v1/remote/catalog.go
@@ -25,8 +25,9 @@ import (
 	"github.com/google/go-containerregistry/pkg/v1/remote/transport"
 )
 
-type catalog struct {
+type Catalogs struct {
 	Repos []string `json:"repositories"`
+	Next  string   `json:"next,omitempty"`
 }
 
 // CatalogPage calls /_catalog, returning the list of repositories on the registry.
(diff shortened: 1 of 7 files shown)


pkg/v1/remote/catalog.go
```diff
@@ -25,8 +25,9 @@ import (
 	"github.com/google/go-containerregistry/pkg/v1/remote/transport"
 )
 
-type catalog struct {
+type Catalogs struct {
```
imjasonh, line 28: Why plural?

pkg/v1/remote/descriptor.go
```diff
@@ -107,7 +107,11 @@ func Head(ref name.Reference, options ...Option) (*v1.Descriptor, error) {
 		return nil, err
 	}
 
-	return f.headManifest(o.context, ref, allManifestMediaTypes)
+	return f.head(o.context, ref)
+}
+
+func (f *fetcher) head(ctx context.Context, ref name.Reference) (*v1.Descriptor, error) {
+	return f.headManifest(ctx, ref, allManifestMediaTypes)
 }
```
imjasonh, line 115: This method doesn't add much, can we just have callers pass `allManifestMediaTypes` themselves?

imjasonh (approved):

pkg/v1/remote/catalog.go
jonjohnsonjr: `remote.Catalog` is already defined 😅

pkg/v1/remote/descriptor.go
jonjohnsonjr: Done.

Issue #1643, opened 2023-04-11
gcrane: Cannot list more than ~1000 entries in an artifact registry


nickstenning:
### Describe the bug

Running `gcrane ls` on an Artifact Registry target with more than 1000 images in it only returns the first ~1000 entries, and there seems to be no way to get the remainder:

```console
$ gcrane ls us-central1-docker.pkg.dev/my-project-name/my-repository |  wc -l
1005
```

It's clear from the output that `gcrane` is just emitting the first ~1000 entries alphabetically. There are many more than 1000 top-level tags in this repository:

<img width="703" alt="gcrane-output" src="https://user-images.githubusercontent.com/3602/231118140-a3a7a421-6b79-442c-843e-d4ac36f07e37.png">

### To Reproduce

Create an Artifact Registry, and tag >1000 images into it. Then run `gcrane ls ...` on the base of the repository.

### Expected behavior

I'd like `gcrane ls` to either a) list _all_ of the first-level tags in the repository, or to expose some kind of pagination/max-size option that allows me to get access to results beyond the first 1000.

### Additional context

```console
$ gcrane version
v0.14.0
```

Registry used: Google Artifact Registry

The main reason this is a problem for me right now is that I'd like to use the output of `gcrane ls` to start multiple parallel `gcrane cp -r` processes (via `parallel` or `xargs -P`) but I can't do that unless I know that the output correctly enumerates every entry in the Artifact Registry.

Issue #984, opened 2021-04-19
GCrane fails to list all tags available on Quay.io registries


davidalger:
Run the following two commands and compare the output. The former returns all 147 tags, the latter only 50 randomly ordered tags.

* `docker run --rm gcr.io/go-containerregistry/crane ls quay.io/prometheus/prometheus`
* `docker run --rm gcr.io/go-containerregistry/gcrane ls quay.io/prometheus/prometheus`

The verbose output reveals crane is passing `n=1000` to the Quay API and GCrane is not. Both appear to possibly be making paginated calls, although apparently not successfully.

jonjohnsonjr:
I think I'm OK with this behavior. `gcrane` is specifically for google container registries, whereas `crane` is meant to be a generic client.

There are a couple ways to fix this, but I'm not sure if it's worth fixing...

jonjohnsonjr:
> we're using gcrane to copy images from Quay to GCR so that Quay outages won't affect the node-exporter daemonset we have running. So we have a daily Gitlab job which basically runs gcrane ls on a set of images, and then copies all tags to us.gcr.io/myimagemirror/quay.io/image/name allowing us to then update chart values to use our registry and/or use a mutating webhook controller for enforcement.

I see. In theory we could make `gcrane cp -r` work for this use case, but I'm happy to just fix `gcrane ls` to do pagination.

> I haven't tested it, but I have a sneaking suspicion that the same issue may exist in crane for registries which have more than 1k tags.

The `crane` implementation does pagination as you might expect. The `gcrane` implementation doesn't, because GCR doesn't do pagination, but it just so happens to return the tags you would expect anyway.

ImJasonH:
> `gcrane` implements a superset of crane commands and it's ls provides enriched output (even when working with non-GCR repositories, it's useful when mirroring images). An ls shouldn't simply be redacted just because I listed a non-GCR registry.

Could we solve this by having `crane` provide the features you need from `gcrane`?

We could add `crane ls --full` or something, that outputs the full names of images, instead of just tags.

I think in general we should discourage (maybe even disallow?) using `gcrane` against anything but GCR.

davidalger:
> We could add crane ls --full or something, that outputs the full names of images, instead of just tags.

[...]

PR #1645, opened 2023-04-11
gcrane: Use page size of 10,000 for googley things

jonjohnsonjr:
We dropped down to 1000 for the sake of ECR, but that doesn't make much sense, really.

Per https://cloud.google.com/artifact-registry/quotas#limits this will return up to 10k items.

Fixes https://github.com/google/go-containerregistry/issues/1643


imjasonh (approved):

imjasonh:
Found https://github.com/google/go-containerregistry/issues/984

PR #1651, opened 2023-04-17
Introduce GGCR_CREDS_CACHE

jonjohnsonjr:
If set, ggcr will store and load ping responses and token responses to
avoid re-authenticating on every invocation. Currently, this writes out
files that it will ignore after 10 minutes (based on file modtime) to
avoid using an expired token.

We can do more clever things in the future like respecting ExpiresIn
from the token response.

On windows this won't work because renameio doesn't work on windows.

### DIFF
--- a/pkg/v1/remote/transport/bearer.go
+++ b/pkg/v1/remote/transport/bearer.go
@@ -29,8 +29,120 @@ import (
 	"github.com/google/go-containerregistry/internal/redact"
 	"github.com/google/go-containerregistry/pkg/authn"
 	"github.com/google/go-containerregistry/pkg/logs"
-	"github.com/google/go-containerregistry/pkg/name"
 )
+
+func newBearer(ctx context.Context, pr *PingResponse, target resource, auth authn.Authenticator, t http.RoundTripper, scopes []string) (http.RoundTripper, error) {
+	key := fmt.Sprintf("token/%s", url.QueryEscape(target.String()))
+	if credCache != nil {
+		b, err := credCache.Get(key)
+		if err != nil || b == nil {
+			logs.Debug.Printf("Transport.credCache.Get(%q) = %v", key, err)
+		} else {
+			tr := TokenResponse{}
+			if err := json.Unmarshal(b, &tr); err != nil {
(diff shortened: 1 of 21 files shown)


pkg/v1/remote/transport/bearer.go
```diff
@@ -29,9 +29,121 @@ import (
...
+// Some registries don't have "token" in the response. See #54.
+type TokenResponse struct {
+	Token        string `json:"token"`
+	AccessToken  string `json:"access_token,omitempty"`
+	RefreshToken string `json:"refresh_token"`
+	ExpiresIn    int    `json:"expires_in"`
+}
+
+// NewBearer returns a bearer transport and the registry TokenResponse for reuse.
+func NewBearer(ctx context.Context, pr *PingResponse, reg resource, auth authn.Authenticator, t http.RoundTripper, scopes []string) (http.RoundTripper, *TokenResponse, error) {
```
imjasonh, line 73: Do `NewBearer` and `OldBearer` need to be exported?

pkg/v1/remote/transport/ping.go
```diff
@@ -39,22 +40,38 @@ const (
 // 300ms is the default fallback period for go's DNS dialer but we could make this configurable.
 var fallbackDelay = 300 * time.Millisecond
 
-type pingResp struct {
-	challenge challenge
+type PingResponse struct {
```
imjasonh, line 43: I think we could get away with keeping these fields unexported, and have a separate unexported type with equivalent exported fields that we actually use to marshal. This jumps through some hoops, but it means our public API doesn't change any.

imjasonh:
We should also document this before merging, so that folks are aware what gets cached, why, and with some benchmarks about how much faster it can make things.

For paranoid types, is it worthwhile to have separate env vars to enable ping-response-caching (which is not sensitive and shouldn't expire) and token-response-caching (which are sensitive but expire quickly)?

And, while I'm thinking of it, how likely is it that ping responses will change in a way that caching makes more painful? Maybe we should expire those every hour or something (configurable?), just in case.

PR #1652, opened 2023-04-17
Change return type of remote.Referrers

jonjohnsonjr:
Actually using this was cumbersome because it just returns a struct. This is a breaking change, but I don't think it's a huge deal because not too many people should be using this yet (and we're < 1.0.0).


jonjohnsonjr:
cc @jdolitsky

imjasonh:
What methods did you find yourself needing from the interface that you didn't get from the struct?

jonjohnsonjr:
> What methods did you find yourself needing from the interface that you didn't get from the struct?

E.g. if I want to save referrers to an OCI layout, I can just do that with a `v1.ImageIndex`. Currently, I have to iterate over each entry and have a separate `remote.Get` call for each one.

imjasonh:
Makes sense, thanks for that example.

imjasonh (approved):

jdolitsky (approved):
gotta make sure to update in cosign! I can do

PR #1671, opened 2023-04-24
add `registry.Repo("foo")`

imjasonh:
```
reg, _ := name.NewRegistry("foo.io")
repo := reg.Repo("bar").Tag("v3") // -> foo.io/bar:v3
```

### DIFF
--- a/pkg/name/registry.go
+++ b/pkg/name/registry.go
@@ -53,7 +53,6 @@ func (r Registry) String() string {
 
 // Repo returns a Repository in the Registry with the given name.
 func (r Registry) Repo(repo ...string) Repository {
-	// TODO: Panic if repo is empty?
 	return Repository{Registry: r, repository: path.Join(repo...)}
 }
 
--- a/pkg/name/registry.go
+++ b/pkg/name/registry.go
@@ -50,6 +50,11 @@ func (r Registry) String() string {
 	return r.Name()
 }
 
+// Repo returns a Repository in the Registry with the given name.
+func (r Registry) Repo(repo string) Repository {
+	return Repository{Registry: r, repository: repo}
+}
+
 // Scope returns the scope required to access the registry.
 func (r Registry) Scope(string) string {
 	// The only resource under 'registry' is 'catalog'. http://goo.gl/N9cN9Z
--- a/pkg/name/registry.go
+++ b/pkg/name/registry.go
@@ -17,6 +17,7 @@ package name
 import (
 	"net"
 	"net/url"
+	"path"
 	"regexp"
 	"strings"
 )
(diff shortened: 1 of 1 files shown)


pkg/name/registry.go
```diff
@@ -50,6 +50,11 @@ func (r Registry) String() string {
 	return r.Name()
 }
 
+// Repo returns a Repository in the Registry with the given name.
+func (r Registry) Repo(repo string) Repository {
```
jonjohnsonjr, line 54: This idea might be bad, but hear me out: consider making this variadic.

I'm in a weird spot with `registry.k8s.io` where I was to call `google.List` on `registry.k8s.io`, but I can't because it doesn't parse. They redirect to AR but with an included path, but effectively calling list on `registry.k8s.io` is ~equivalent to `/v2/_catalog` because of GCR's custom response.

If we make this variadic, then `Reference.Repo()` could implement the same interface. So could `Repository.Repo(...)`. Essentially we get a way to either get the repo for a given thing _or_ get a child repo for a given thing... wdyt?

The variadic is kinda weird for ref, but if (for some reason) you wanted to get a child repo from a ref, instead of doing `ref.Context().Repo("child") or `ref.Repo().Repo("child")`, you could just do `ref.Repo("child")`.

For `len(repos) > 1`, I think `path.Join`ing them is reasonable.

imjasonh: Is there a difference between `Repo("a", "b", "c")` and `Repo(path.Join("a", "b", "c"))`? It's more verbose but clearer IMO.

I'm not sure it's clear what `NewTag("reg.io/repo:tag").Repo("child")` does -- does it give you `reg.io/repo/child:tag`, or strip the tag? Same for digest. `NewTag("reg.io/repo:tag").Context().Repo("child")` is more verbose but clearer, because I'm explicitly stepping up into the repo, and leaving the tag behind.

I'm fine with `Repository` having `Repo(child string)`. I don't have an immediate need for it personally, but I could see it. Making that variadic seems like it's maybe trying to be too clever. If we want to make it variadic we can do that later, so long as it's not part of an interface like `Reference`.

pkg/name/registry.go
```diff
@@ -50,6 +51,12 @@ func (r Registry) String() string {
 	return r.Name()
 }
 
+// Repo returns a Repository in the Registry with the given name.
+func (r Registry) Repo(repo ...string) Repository {
+	// TODO: Panic if repo is empty?
```
imjasonh, line 56: This is also kind of a wart. What should we do if someone does `NewRegistry("foo.io").Repo()`? Now they have a `name.Repository` with an empty repository, which they can pass to `remote.List` and send a request to `https://foo.io/v2/tags/list` where it will fail. Same for `remote.WriteLayer` and `https://foo.io/v2/blobs/uploads`.

Should we just panic here in that case?

jonjohnsonjr: > where it will fail

It depends on the registry. Some registries will not fail here, which is ~most of the point of adding this.

imjasonh: Okay cool. I won't panic, but I will look at it funny if I ever see it. :)

jonjohnsonjr, line 56:
```suggestion
```

jonjohnsonjr (approved):

PR #1672, opened 2023-04-24
Add remote.Reuse for Pusher/Puller

jonjohnsonjr:
This is a big change, but it helps callers out a lot.

The Pusher/Puller interfaces allow us to deduplicate a bunch of work
(largely, ping an auth), but they only work if you actually use them.

It's a huge pain to migrate callers from remote.{Image,Index,...}
interfaces to start using Pusher/Puller because the remote functions can
be called from anywhere, which means plumbing pushers and pullers around
the entire callgraph, which is super painful.

Happily, though, most callers already plumb remote.Options around the
callgraph so that they can set their options in one place and have them
be consistent throughout their application.

This change takes advantage of that fact by introducing remote.Reuse,
which takes either a Puller or a Pusher, and calls equivalent methods on
said pusher/puller whenever you pass remote.Reuse into a remote
function.

The end result is that we can get almost all of the performance wins of
using Puller/Pusher directly with a 3 line change to most applications
rather than refactoring everything.


pkg/v1/remote/puller.go
```diff
@@ -61,32 +70,76 @@ func (r *reader) init(ctx context.Context) error {
...
+
+func (p *Puller) get(ctx context.Context, ref name.Reference, acceptable []types.MediaType, platform v1.Platform) (*Descriptor, error) {
+	f, err := p.fetcher(ctx, ref.Context())
+	if err != nil {
+		return nil, err
+	}
+	return f.get(ctx, ref, acceptable, platform)
+}
+
+// Layer is like remote.Layer, but avoids re-authenticated when possible.
```
imjasonh, line 120:
```suggestion
// Layer is like remote.Layer, but avoids re-authenticating when possible.
```

imjasonh (approved):
This is awesome. 🔥

Issue #1674, opened 2023-04-25
question: Feature request to add rand.Source or Seed to the pkg/v1/random functions


ktarplee:
As a user of the library, I would like to be able to generate random test images but do so in a deterministic way.   I would like to have options from the random.Index(), random.Image(), and random.Layer() functions that allow me to set the `rand.Source` or the seed to use.

During unit testing I find the random images useful however they are always different (even when I am re-running a test) so it is a larger cognitive load to track what is happening in my test.  I cannot add a conditional breakpoint on the sha256 in one of these images.  I do not care what the digest is or what the contents of the images are, I just want them to stay constant unless I change the seed.  Calling `rand.Seed()` is deprecated now so that is not a good option.

The only place a random number generator is used is in Layer() and actually two different random number generators are called.  `crypto/rand` and `math/rand`.

PR #1678, opened 2023-04-27
fix: remove current time from random.Image history

ktarplee:
The history contained the `time.Now()` which I missed in my prior PR #1675 .  Using the current time makes the random images non-reproducible which defeats the purpose of setting the random source. I removed the CreatedBy time since it is optional according to the OCI spec.  We could instead use the random source to generate a random time but the time is not used for anything important so I do not see the value in doing that.

I added tests that revealed the issue then fixed the issue.

Related to #1674

### DIFF
--- a/pkg/v1/random/image.go
+++ b/pkg/v1/random/image.go
@@ -22,7 +22,6 @@ import (
 	"fmt"
 	"io"
 	"math/rand"
-	"time"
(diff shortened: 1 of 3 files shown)


jonjohnsonjr (approved):
Thanks again!

PR #1686, opened 2023-05-04
WIP: internal/test helpers

imjasonh:
Opening this to solicit feedback.

It's internal now, after we settle on an API we all like I'll take a pass at applying it to our own tests, and eventually make this an external API so folks can use it in their own tests (myself included 😆 )

### DIFF
--- a/internal/test/must.go
+++ b/internal/test/must.go
@@ -4,7 +4,7 @@
 // you may not use this file except in compliance with the License.
 // You may obtain a copy of the License at
 //
-//	http://www.apache.org/licenses/LICENSE-2.0
+//    http://www.apache.org/licenses/LICENSE-2.0
 //
 // Unless required by applicable law or agreed to in writing, software
 // distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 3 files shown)


internal/test/must.go
```diff
@@ -0,0 +1,120 @@
...
+func T(t *testing.T) Muster {
+	return testMuster{t}
+}
+
+type digester interface {
+	Digest() (v1.Hash, error)
+}
+
+type Muster interface {
+	RandomImage(int64, int64) v1.Image
```
jonjohnsonjr, line 49: I'd rather not take a dep on `random` if we can avoid it.

Can this be `Image(v1.Image, error) v1.Image` called as `must.Image(random.Image(...))`?

imjasonh: Great idea. Done!

internal/test/must.go
```diff
@@ -0,0 +1,117 @@
...
+// test.T(t).Digest(img)
+//
+// or
+//
+// must := must.T(t)
+// must.Digest(img)
+
+// func TestFoo(t *testing.T) {
+//   must := test.T(t)
+//   img := must.RandomImage(1024, 5)
```
jonjohnsonjr, line 33:
```suggestion
//   img := must.Image(random.Image(1024, 5))
```

imjasonh: Dropped this comment section anyway, in favor of the test, which will actually stay up to date.

PR #1701, opened 2023-05-16
Do not reuse pushers for pullers

jonjohnsonjr:
This causes issues with broken registry implementations that 403 if you have requested push scopes even for pull operations.

See https://github.com/sigstore/cosign/issues/2973


dmitris (approved):
LGTM, thanks.  We can close my https://github.com/sigstore/cosign/pull/2972 and instead use this fix, with an upgrade of the new version of github.com/google/go-containerregistry (probably v0.15.2)

imjasonh (approved):

PR #1705, opened 2023-05-19
fixed a goroutine leak

ktarplee:
If the rc is not closed (which happens when the files with the same digest already exists, due to the same image being added to the OCI layout directory twice) then 
```go
if s, err := os.Stat(file); err == nil && !s.IsDir() && (s.Size() == size || size == -1) {
	return nil
}
```
will not close the io.ReadCloser leaking a goroutine [here](https://github.com/ktarplee/go-containerregistry/blob/009eea0d1c371a8bc3ba6898a2e88d6fab35eecf/internal/gzip/zip.go#L53)


ktarplee:
Closes #1706

imjasonh (approved):
Nice!

Issue #1470, opened 2022-10-21
`crane auth token $registry`


imjasonh:
@jonjohnsonjr had a good idea

`crane auth token` is a command that goes through the ping and token endpoints to exchange configured creds for a short-lived registry token, and prints it to stdout.

This could be useful for testing with curl, e.g:

```
curl -H "Authorization: Bearer $(crane auth token registry.biz)" https://registry.biz/v2/repo/manifests/tag | jq
```

imjasonh:
We basically have this already, in `crane auth get $registry`, it just prints out the whole JSON token response, instead of just the token.

Maybe this doesn't need to be a whole new command, it just needs something like a [recipe](https://github.com/google/go-containerregistry/blob/main/cmd/crane/recipes.md) to document it:

```
curl -H "Authorization: Bearer $(crane auth get registry.biz | jq .Secret)" https://registry.biz/v2/repo/manifests/tag | jq
```

PR #1709, opened 2023-05-19
Add crane auth token

jonjohnsonjr:
This makes it easier to just use curl.

Fixes https://github.com/google/go-containerregistry/issues/1470


tianon:
This is really neat :smile:

imjasonh (approved):

PR #1711, opened 2023-05-22
Limit size of manifest

AdamKorcz:
This adds a limit to the size of manifest that can get pulled.

Since the manifest is untrusted, this imposes a security risk to go-containerregistry users; If the size of the manifest is excessive compared to the ressources available on the machine running go-containerregistry, then a large manifest could cause Denial of Service by exhausting the memory of the machine.

### DIFF
--- a/pkg/v1/remote/fetcher.go
+++ b/pkg/v1/remote/fetcher.go
@@ -32,6 +32,12 @@ import (
 	"github.com/google/go-containerregistry/pkg/v1/types"
 )
 
+const (
+	kib           = 1024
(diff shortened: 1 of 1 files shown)


jonjohnsonjr (approved):
We should probably make this configurable, but 100mb should be way higher than anything would reasonably hit, so I think it's fine to merge as is.

jonjohnsonjr (approved):
We should probably make this configurable, but 100mb should be way higher than anything would reasonably hit, so I think it's fine to merge as is.

PR #1721, opened 2023-06-05
Make ErrSchema1 checkable via errors.Is()

Laitr0n:
The `TestGetSchema1` said it should fail based on media type while gave the unsupported MediaType. But it not return the true.

This fixes that problem by implementing an `Is()` function on `ErrSchema1`
so that `errors.Is()` can properly identify the error as an ErrSchema1.
Usage can now be: `errors.Is(err, &ErrSchema1{schema : "foo"})`

### DIFF
--- a/pkg/v1/remote/descriptor.go
+++ b/pkg/v1/remote/descriptor.go
@@ -16,6 +16,7 @@ package remote
 
 import (
 	"context"
+	"errors"
 	"fmt"
 
 	"github.com/google/go-containerregistry/pkg/logs"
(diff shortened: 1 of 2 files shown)


imjasonh (approved):

imjasonh:
cc @jonjohnsonjr

jonjohnsonjr (changes requested):
I think embedding the mediaType that we saw is both unnecessary and makes this really complicated.

Can you update this to just be:

```go
var ErrSchema1 = errors.New("unsupported schema1 mediaType, see https://github.com/google/go-containerregistry/issues/377")
```

And fix any tests that it breaks?

jonjohnsonjr (approved):

pkg/v1/remote/descriptor.go
```diff
@@ -33,20 +34,11 @@ var allManifestMediaTypes = append(append([]types.MediaType{
 // ErrSchema1 indicates that we received a schema1 manifest from the registry.
 // This library doesn't have plans to support this legacy image format:
 // https://github.com/google/go-containerregistry/issues/377
-type ErrSchema1 struct {
-	schema string
-}
+var ErrSchema1 = errors.New("unsupported schema1 mediaType, see https://github.com/google/go-containerregistry/issues/377")
```
jonjohnsonjr, line 37:
```suggestion
var ErrSchema1 = errors.New("see https://github.com/google/go-containerregistry/issues/377")
```

pkg/v1/remote/descriptor.go
```diff
@@ -33,20 +34,11 @@ var allManifestMediaTypes = append(append([]types.MediaType{
...
 func newErrSchema1(schema types.MediaType) error {
-	return &ErrSchema1{
-		schema: string(schema),
-	}
-}
-
-// Error implements error.
-func (e *ErrSchema1) Error() string {
-	return fmt.Sprintf("unsupported MediaType: %q, see https://github.com/google/go-containerregistry/issues/377", e.schema)
+	return fmt.Errorf("mediaType: %s, %w", schema, ErrSchema1)
```
jonjohnsonjr, line 41:
```suggestion
	return fmt.Errorf("unsupported MediaType: %q, %w", schema, ErrSchema1)
```

jonjohnsonjr (changes requested):
Oh that's clever. Added a couple suggestions so that we can keep the exact same error string but have the `%w` wrapping make `errors.Is` work.

I have a feeling someone, somewhere is checking for the string value of this error specifically because `errors.Is` doesn't work, so I want to avoid breaking that.

jonjohnsonjr (approved):
thanks!

PR #1728, opened 2023-06-13
crane: Respect cmd.OutOrStdout

kyleconroy:
When calling Crane programmatically, it's difficult to capture the output as it's written directly to stdout. This change uses Cobra's built-in support for changing stdout to an io.Writer.


jonjohnsonjr (approved):
Thanks!

Issue #629, opened 2019-12-10
Oauth grant_type=password is inaccessible


jonjohnsonjr:
Before https://github.com/google/go-containerregistry/pull/628 we would fall back to the oauth flow for username/password if the token flow failed. This is not great behavior, since it almost always will fail, so we removed that fallback. As a result, there's actually no way to use username/password with the oauth flow.

Docker has a similar codepath that is reachable if the [`ForceOAuth`](https://github.com/docker/distribution/blob/84f47e7bb3bb3521d7d20615d42078df306fb01d/registry/client/auth/session.go#L196) option is set in their `TokenHandlerOptions`, but as far as I can tell there's no existing code that actually exercises that path in the docker client.

It's unclear to me if any registries support this or any clients need it, and there's no easy way to tell if we should take that path from the registry response.

splucs:
@jonjohnsonjr friendly ping on the PR for this: https://github.com/google/go-containerregistry/pull/1740

PR #1740, opened 2023-06-21
Enable using oauth flow with username and password

splucs:
## Problem
In issue https://github.com/google/go-containerregistry/issues/626, it was reported that when executing the refresh bearer token codepath will fallbacks, only the latest error is returned. This caused the "endpoint not implemented" errors (e.g. 404 and 405) to end up obfuscating the real permissioning errors (e.g. 403). This was originally fixed in PR https://github.com/google/go-containerregistry/pull/628 by only enabling the oauth flow if an identity token is provided. This did, however, make the codepath to using the oauth flow with username and password unreachable. This is tracked in issue https://github.com/google/go-containerregistry/issues/629.

The change https://github.com/google/go-containerregistry/pull/628 does, however, introduce a condition where a fallback to another auth flow should only happen if the previous auth flow returns status code 404, denoting that it is not supported. We can use this to solve the original problem without gating on a specific type of credentials.

This PR enables fallback between the oauth and basic flows for both username/password and identity tokens. The next auth flow is only attempted if the previous one returns 404 or 405. The first non-404/405 response is returned.

## Issue
fixes https://github.com/google/go-containerregistry/issues/629

## Testing
Tested by adding unit tests for the fallback with basic credentials.


splucs:
@jonjohnsonjr it seems you were the one that has the highest activity and context on this area of the codebase. Could you review this or point me to someone who can review this? Thanks.

pkg/v1/remote/transport/bearer.go
```diff
@@ -230,20 +229,29 @@ func (bt *bearerTransport) Refresh(ctx context.Context, auth *authn.AuthConfig)
 		content []byte
 		err     error
 	)
+
+	first, second := bt.refreshBasic, bt.refreshOauth
```
jonjohnsonjr, line 233: Let me know if I understand the intention of this change correctly:

Previously, we would only enter `refreshOauth` if `IdentityToken` was set, which means we would ignore `Username` and `Password` (hence, the TODO I left).

Now, (if `IdentityToken` is unset) we will try `refreshBasic` first, falling back to `refreshOauth` if we encounter a 404 or 405, so we will end up going down the Username/Password path after `refreshBasic` fails.

If `IdentityToken` _is_ set, we reverse that order and first try `refreshOauth`, falling back to `refreshBasic`, which matches the current behavior.

Can you provide any evidence that this is the right thing to do on 404/405? (I believe you, I would just love to have a comment in here somewhere as justification, even if it's just something simple like "foo.example.com returns a 404/5 in this case").

splucs: This logic somewhat copies what `containerd` does: https://github.com/containerd/containerd/blob/v1.7.2/remotes/docker/authorizer.go#L297-L321
`containerd` is a commonly used container runtime in kubernetes environments and it's well battle tested, so I believe it to be a good source of truth in the absence of an RFC that specifies when to use oauth vs basic.

The only difference is the preference for the basic flow when an username/password is provided and for the oauth flow when an identity token is provided. `containerd` always tries oauth first. I made it this way because it seems this repo had a prior art that did this: https://github.com/google/go-containerregistry/pull/566

I personally think we should copy what `containerd` does (i.e. always try oauth first) and I'm happy to change, but I'm also happy to hear why did we choose to select the flow preference based on the type of credentials provided.

jonjohnsonjr: > I personally think we should copy what containerd does (i.e. always try oauth first) and I'm happy to change, but I'm also happy to hear why did we choose to select the flow based on the type of credentials provided.

I'd like to keep basic first, mostly because it's more widely supported and will reduce the number of roundtrips for registries that don't support the oauth flow. This is at the expense of registries that _only_ support the oauth flow, but that doesn't affect me, personally, so I'd selfishly bias towards basic first 😅

splucs: That sounds ok to me. The registry I'm working with should behave nicely by returning a 405 if the method is a GET on the token endpoint.

I believe the implementation suffices as is, or would you like to use basic first if we see an identity token as well?
Also, are there any other changes you'd like to see in the PR?

splucs:
@jonjohnsonjr is there anything else needed before approval?

splucs:
@mattmoor @jonjohnsonjr friendly ping as this is still impacting us. Is there any more concerns/comments with this approach?

splucs:
@jonjohnsonjr friendly ping on the PR for this: https://github.com/google/go-containerregistry/pull/1740

splucs:
@jonjohnsonjr @mattmoor @imjasonh friendly ping

splucs:
@jonjohnsonjr @mattmoor @imjasonh ping

vahan-sahakyan-op (approved):
🚢

PR #1742, opened 2023-06-22
mutate: limit size of layer before copying

AdamKorcz:
A large tarReader can crash go-containerregistry with an out of memory panic. This limits the size of `header.Size`.

### DIFF
--- a/pkg/v1/mutate/mutate.go
+++ b/pkg/v1/mutate/mutate.go
@@ -35,6 +35,7 @@ import (
 )
 
 const whiteoutPrefix = ".wh."
+const maxFileSize = 1024 * 1024 * 20 // 20mb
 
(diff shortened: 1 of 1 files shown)


jonjohnsonjr:
It shouldn't unless you're buffering the output.

AdamKorcz:
> It shouldn't unless you're buffering the output.

That would be a valid use case, right?

jonjohnsonjr:
> That would be a valid use case, right?

Absolutely, but we shouldn't limit the file size for _all_ consumers to avoid OOM when you're buffering.

An alternative to this PR would be using a [`LimitedReader`](https://pkg.go.dev/io#LimitedReader) on the consuming side if you want to limit how much you read from the flattened filesystem.

PR #1745, opened 2023-06-24
crane: support mutating index annotations

jonjohnsonjr:


cmd/crane/cmd/mutate.go
```diff
@@ -48,23 +48,54 @@ func NewCmdMutate(options *[]crane.Option) *cobra.Command {
...
-					return err
-				}
-				if desc.MediaType.IsIndex() {
-					return errors.New("mutating annotations on an index is not yet supported")
+			desc, err := crane.Get(ref, *options...)
+			if err != nil {
+				return fmt.Errorf("pulling %s: %w", ref, err)
+			}
+			if desc.MediaType.IsIndex() {
+				if len(annotations) != 0 {
```
imjasonh, line 56: `if IsIndex && len(annotations) != 0`

pkg/crane/push.go
```diff
@@ -43,6 +43,16 @@ func LoadTag(path, tag string, opt ...Option) (v1.Image, error) {
 	return tarball.ImageFromPath(path, &t)
 }
 
+// Put writes the remote.Taggable to a registry as dst.
+func Put(put remote.Taggable, dst string, opt ...Option) error {
```
imjasonh, line 47: Is `dst` expected to be a repo, tag or digest? Can we make that clearer from the comment and/or validation and error messages?

PR #1746, opened 2023-06-25
Drop localhost to support crane registry serve in a container

mattmoor:
This allows you to run `docker run -e PORT=1234 --name foo.local cgr.dev/chainguard/crane registry serve` and then `--link foo.local` to push to `foo.local:1234`.

Right now it only listens on the loopback interface, so unless it's run in a K8s-style pod with something that exposes and proxies things (e.g. like Knative / Cloud Run), it is inaccessible.

### DIFF
--- a/cmd/crane/cmd/serve.go
+++ b/cmd/crane/cmd/serve.go
@@ -54,7 +54,7 @@ Contents are only stored in memory, and when the process exits, pushed data is l
 			if port == "" {
 				port = "0"
 			}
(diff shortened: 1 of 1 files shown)


jonjohnsonjr (approved):

imjasonh (approved):

ewhauser:
@mattmoor - this causes network error warnings on MacOS: 

Screenshot 2023-11-19 at 1 52 47 PM

Bazel's `rules_oci` makes use of crane so you end up with a warning for every image you are building which is untenable.

PR #1748, opened 2023-07-03
Adjust referrers fallback behavior to be optional

tianon:
I was talking with @jonjohnsonjr about using `crane cp` in a use case where I never want it to accidentally surprise me with the fallback tags, so this implements a very basic means of making that behavior optional.

I figured having a PR would give us a stronger straw-man to discuss the finer details around, especially since I don't feel like I know this library/codebase well enough to make actual informed recommendations for how I'd like to see this implemented. :sweat_smile:

(As always, happy to rebase, amend, adjust, rewrite, reword, close, let someone else take over, etc as desired!)

### DIFF
--- a/pkg/v1/remote/referrers.go
+++ b/pkg/v1/remote/referrers.go
@@ -20,6 +20,7 @@ import (
 	"errors"
 	"io"
 	"net/http"
+	"os"
 	"strings"
 
 	"github.com/google/go-containerregistry/pkg/name"
@@ -71,7 +72,7 @@ func (f *fetcher) fetchReferrers(ctx context.Context, filter map[string]string, 
 		if err != nil {
 			return nil, err
 		}
-	} else {
+	} else if os.Getenv("GGCR_REF_FALLBACK") != "" {
 		// The registry doesn't support the Referrers API endpoint, so we'll use the fallback tag scheme.
 		b, _, err = f.fetchManifest(ctx, fallbackTag(d), []types.MediaType{types.OCIImageIndex})
 		var terr *transport.Error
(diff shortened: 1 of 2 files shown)


tianon:
Delicious tests pointing out that this is a breaking change as-is because it's opt-in instead of opt-out, as suggested by Jon -- I'm happy to adjust tests or behavior as desired, but y'all maintainers probably need to decide which direction you'd like to go and give me some suggestion (or send me away). :bow: :joy:

tianon:
"for Jon"

tianon:
Begone, ye stale bot

pkg/v1/remote/referrers.go
```diff
@@ -81,6 +82,9 @@ func (f *fetcher) fetchReferrers(ctx context.Context, filter map[string]string,
 		} else if err != nil {
 			return nil, err
 		}
+	} else {
+		// TODO just returning "empty" here doesn't feel quite right -- maybe transport.CheckError again but this time letting http.StatusBadRequest return an error?
+		return empty.Index, nil
```
tianon, line 87: Coming back to this almost a year later and feeling like a nice `fmt.Errorf` is probably more appropriate here.  Thoughts?  Maybe the variable needs to be opt-out instead of opt-in, also?  (`GGCR_REF_NO_FALLBACK` or `GGCR_REF_NEVER_FALLBACK` ?)

tianon:
https://github.com/google/go-containerregistry/commit/38b69ff497f02cb384eba8fd292f1e732684560b if you wanna see where I finally got cheeky and have applied my own patch to my personal builds :smile:

tianon:
uno reverse card

No bot, _you're_ stale!!!

tianon:
It seems like I'm due for another stale bot ping, so I'm going to pre-empt it with an appropriate link to https://nostalebots.xyz/

PR #1798, opened 2023-09-28
feat: Add zstd compression to crane append and mutate

ReillyBrogan:
Support for zstd-compression was added in #1487. Add handling for this to crane append and mutate which should allow other projects that use Crane to begin creating zstd images (notably [rules_oci](https://github.com/bazel-contrib/rules_oci/issues/333) which is used by distroless amongst other projects).

### DIFF
--- a/pkg/compression/compression.go
+++ b/pkg/compression/compression.go
@@ -15,6 +15,8 @@
 // Package compression abstracts over gzip and zstd.
 package compression
 
+import "errors"
+
 // Compression is an enumeration of the supported compression algorithms
 type Compression string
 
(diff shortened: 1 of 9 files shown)


pkg/crane/append.go
```diff
@@ -90,7 +107,11 @@ func getLayer(path string, layerType types.MediaType) (v1.Layer, error) {
 		return stream.NewLayer(f, stream.WithMediaType(layerType)), nil
 	}
 
-	return tarball.LayerFromFile(path, tarball.WithMediaType(layerType))
+	if layerType == types.OCILayerZStd {
+		return tarball.LayerFromFile(path, tarball.WithMediaType(layerType), tarball.WithCompression(compression.ZStd))
```
slonopotamus, line 111: I don't think this works properly. Imagine you gave it a `.tar.gz` file and asked for zstd compression. No re-compression will happen. You will have zstd in manifest, but contents of the blob will be gzip.

Note that there is a bug even before this PR: if one calls `crane append` with a zstd file, it will blindly append it with gzip in manifest.

pkg/compression/compression.go
```diff
@@ -24,3 +26,25 @@ const (
...
+		return errors.New(`must be one of "none", "gzip, or "zstd"`)
+	}
+}
+
+// Used in Cobra help text
+func (e *Compression) Type() string {
+	return "Compression"
+}
+
+var ErrZStdNonOci = errors.New("ZSTD compression can only be used with an OCI base image")
```
slonopotamus, line 50: This statement is not true. There exists `vnd.docker.image.rootfs.diff.tar.zstd` type. See https://github.com/moby/buildkit/pull/2344

slonopotamus:
See #1827 for alternative implementation.

slonopotamus:
<dummy comment to keep stalebot away>

PR #1800, opened 2023-09-28
Add stream.NewCompressedLayer for handling compressed input

### LINKED ISSUE #1799: ggcr: Support streaming already-compressed layers
I have a use case where I have an io.ReadCloser representing an already-compressed layer I wish to upload to a registry. As far as I can tell, the only mechanism is `stream.NewLayer` which expects an _uncompressed_ stream (often from a tarball). The layer will need to compute Digest/DiffID/Size during the single read pass like we do in the uncompressed version. While I could uncompress the input stream with gzip.Reader and provide that as the NewLayer input argument, I do not want the compression parameters to change and want the input's checksum preserved.

Specific case of https://github.com/google/go-containerregistry/issues/1011

// More FR than bug, but I can not remove the label

bmoylan:
Closes #1799

### DIFF
--- a/pkg/v1/stream/layer.go
+++ b/pkg/v1/stream/layer.go
@@ -42,7 +42,9 @@ var (
 
 // Layer is a streaming implementation of v1.Layer.
 type Layer struct {
-	blob        io.ReadCloser
+	// exactly one reader should be non-nil
+	uncompressed, compressed io.ReadCloser
+
 	consumed    bool
 	compression int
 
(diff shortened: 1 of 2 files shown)


bmoylan:
Hi @jonjohnsonjr could I request a review here? Happy to update the APIs based on any feedback, but right now I am forced to fall back to the distribution/distribution client to orchestrate all the blob operations manually. More context in #1799. Thanks in advance!

PR #1811, opened 2023-10-11
feat: implement gc command

thesayyn:
This is needed for a case where we use oci-layout as the registry storage directory and there are unreferenced blobs. We want to remove them during pull into the storage directory. I marked this as hidden to get some mileage in rules_oci.


cmd/crane/cmd/pull.go
```diff
@@ -30,8 +30,8 @@ import (
 // NewCmdPull creates a new cobra.Command for the pull subcommand.
 func NewCmdPull(options *[]crane.Option) *cobra.Command {
 	var (
-		cachePath, format string
-		annotateRef       bool
+		cachePath, format  string
+		annotateRef, prune bool
```
imjasonh, line 34: I don't think it's appropriate to have a command for pulling be responsible for GCing. Maybe a separate hidden command that prunes an OCI layout directory would be appropriate, which you could run after pulling into that OCI layout directory.

It might make sense for layout.GC to move to an internal package, or at least have a comment saying it's experimental and shouldn't be relied upon outside of this path.

It'd also be great to see tests for it, since it's fairly complex and deletes data. There are some interesting corner cases in GCing that it would be great to have covered by tests.

thesayyn: hmm, i'd be okay with having this feature as a separate command. i believe in that case, we should name this `crane gc` and specifically say that it only supports local oci layouts

thesayyn: about corner cases, are there any specific cases you are aware of? i will add tests for basic cases but would be great to hear from you as well.

thesayyn: Done @imjasonh

thesayyn: Added some basic tests.

thesayyn:
Test failures seems unrelated. probably expired token.

cmd/crane/cmd/pull.go
thesayyn: ping @imjasonh

thesayyn: ping @jonjohnsonjr

imjasonh (approved):

PR #1823, opened 2023-10-24
feat: gcrane - Increase GCRBackoff Steps to 5 and decrease Factor to 5.0

### LINKED ISSUE #424: Gcrane: unrecognized HTTP status: 429 Too Many Requests

klemmari1:
[ExponentialBackoff](https://github.com/google/go-containerregistry/blob/main/internal/retry/wait/kubernetes_apimachinery_wait.go#L117-L119) breaks the loop when `backoff.Steps == 1`. This might cause the `gcrane cp` command to fail with 429 errors when we have only 2 steps to `GCRBackoff` because the wait period is only ~1 minute.

For some reason I saw 429 errors from GCR even after a wait of 15 minutes. Do a couple of improvements to remediate this issue:
- Increase `Steps` to 5 to get more retries and a longer wait time on the last step.
- Decrease `Factor` to 5.0 to get more retries in a more tight schedule. Last wait time will be the `Cap` which is 1 hour.
- Set `b.Steps = 1` when `b.Duration > b.Cap`.

Fix the same issue as [this PR](https://github.com/google/go-containerregistry/pull/512).

Fixes #424

### DIFF
--- a/internal/retry/wait/kubernetes_apimachinery_wait.go
+++ b/internal/retry/wait/kubernetes_apimachinery_wait.go
@@ -89,7 +89,7 @@ func (b *Backoff) Step() time.Duration {
 		b.Duration = time.Duration(float64(b.Duration) * b.Factor)
 		if b.Cap > 0 && b.Duration > b.Cap {
 			b.Duration = b.Cap
-			b.Steps = 0
+			b.Steps = 1
 		}
 	}
 
--- a/pkg/gcrane/copy.go
+++ b/pkg/gcrane/copy.go
@@ -57,7 +57,7 @@ func GCRBackoff() retry.Backoff {
 		Duration: 6 * time.Second,
 		Factor:   5.0,
 		Jitter:   0.1,
-		Steps:    6,
+		Steps:    5,
 		Cap:      1 * time.Hour,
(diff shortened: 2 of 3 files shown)


internal/retry/wait/kubernetes_apimachinery_wait.go
```diff
@@ -89,7 +89,7 @@ func (b *Backoff) Step() time.Duration {
 		b.Duration = time.Duration(float64(b.Duration) * b.Factor)
 		if b.Cap > 0 && b.Duration > b.Cap {
 			b.Duration = b.Cap
-			b.Steps = 0
+			b.Steps = 1
```
klemmari1, line 92: Set the Steps to 1 so that we run a final time with `Duration` = `Cap` which was not run until now.

PR #1824, opened 2023-10-25
Add `mutate --volumes` option to set the volumes

gzm0:


gzm0:
Just checking in here. Is there anything I can do to help move this forward?

jamesreprise:
+1, getting this merged would be very useful!

mpdn:
+1, volume support is what is blocking us from adopting `rules_oci` at the moment, and support in that is predicated on `crane` getting support for it. It would be really useful!

@jonjohnsonjr I noticed your "never offended by friendly pings" and you recently approving a PR - is this something you could maybe review? 😄

PR #1834, opened 2023-11-01
Ignore malformed secrets when creating keychain

### LINKED ISSUE #1833: ggcr: Ignore malformed secrets
The [NewFromPullSecrets](https://github.com/google/go-containerregistry/blob/dbcd01c402b2f05bcf6fb988014c5f37e9b13559/pkg/authn/kubernetes/keychain.go#L161) function iterates through a slice of secrets when building a keychain. The iteration [may fail](https://github.com/google/go-containerregistry/blob/dbcd01c402b2f05bcf6fb988014c5f37e9b13559/pkg/authn/kubernetes/keychain.go#L171-L190) if one of the secrets contain malformed data.

When using the [New](https://github.com/google/go-containerregistry/blob/dbcd01c402b2f05bcf6fb988014c5f37e9b13559/pkg/authn/kubernetes/keychain.go#L72C3-L72C3) function, the slice of secrets is inferred from different sources. If one of those secrets is malformed the keychain building process fails.

I propose changing the behavior of `NewFromPullSecrets` to simply log a warning when it comes across a malformed secret.

lcarva:
This changes the behavior of the [NewFromPullSecrets](https://github.com/google/go-containerregistry/blob/dbcd01c402b2f05bcf6fb988014c5f37e9b13559/pkg/authn/kubernetes/keychain.go#L161) function to ignore secrets that contain malformed data. This allows processing of the other secrets which may be sufficient for authentication.

Resolves #1833.


lcarva:
@imjasonh , @jonjohnsonjr wdyt?

lcarva:
It would be great to have a review here :pray:

pkg/authn/kubernetes/keychain.go
```diff
@@ -186,7 +187,8 @@ func NewFromPullSecrets(ctx context.Context, secrets []corev1.Secret) (authn.Key
 			}
 			parsed, err := url.Parse(value)
 			if err != nil {
-				return nil, fmt.Errorf("Entry %q in dockercfg invalid (%w)", value, err)
+				logs.Warn.Printf("entry %q in dockercfg secret %s/%s invalid (%s); ignoring", secret.Namespace, secret.Name, value, err)
+				continue
```
chitrangpatel, line 191: Instead of continuing. 
Can we throw an error if none of the secrets work? May be thats a failure mode?

lcarva: Done!

chitrangpatel (approved):
Thanks for these changes!

enarha:
Looking for that fix to be merged. Thank you.

PR #1848, opened 2023-11-21
feat: allow port and disk path to be overriden

thesayyn:
Cross-ref: https://github.com/bazel-contrib/rules_oci/issues/418

### DIFF
--- a/cmd/crane/cmd/serve.go
+++ b/cmd/crane/cmd/serve.go
@@ -37,11 +37,11 @@ func newCmdRegistry() *cobra.Command {
 }
 
 func newCmdServe() *cobra.Command {
-	var disk bool
+	var address, disk string
 	cmd := &cobra.Command{
 		Use:   "serve",
 		Short: "Serve an in-memory registry implementation",
-		Long: `This sub-command serves an in-memory registry implementation on an automatically chosen port (or $PORT)
+		Long: `This sub-command serves an in-memory registry implementation on an automatically chosen port (:0), $PORT or --address
 
 The command blocks while the server accepts pushes and pulls.
 
@@ -54,7 +54,12 @@ Contents are only stored in memory, and when the process exits, pushed data is l
 			if port == "" {
 				port = "0"
 			}
-			listener, err := net.Listen("tcp", ":"+port)
+			listenOn := ":" + port
+			if address != "" {
+				listenOn = address
+			}
+
+			listener, err := net.Listen("tcp", listenOn)
 			if err != nil {
 				log.Fatalln(err)
 			}
@@ -62,12 +67,14 @@ Contents are only stored in memory, and when the process exits, pushed data is l
 			port = fmt.Sprintf("%d", porti)
 
 			bh := registry.NewInMemoryBlobHandler()
-			if disk {
-				tmp := os.TempDir()
-				log.Printf("storing blobs in %s", tmp)
-				bh = registry.NewDiskBlobHandler(tmp)
+			if cmd.Flags().Changed("blobs-to-disk") {
+				path := os.TempDir()
+				if disk != "" && disk != " " {
+					path = disk
+				}
+				log.Printf("storing blobs in %s", path)
+				bh = registry.NewDiskBlobHandler(path)
 			}
-
 			s := &http.Server{
 				ReadHeaderTimeout: 5 * time.Second, // prevent slowloris, quiet linter
 				Handler:           registry.New(registry.WithBlobHandler(bh)),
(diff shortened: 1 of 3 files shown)


cmd/crane/cmd/serve.go
```diff
@@ -37,11 +37,11 @@ func newCmdRegistry() *cobra.Command {
 }
 
 func newCmdServe() *cobra.Command {
-	var disk bool
+	var address, disk string
 	cmd := &cobra.Command{
 		Use:   "serve",
 		Short: "Serve an in-memory registry implementation",
-		Long: `This sub-command serves an in-memory registry implementation on an automatically chosen port (or $PORT)
+		Long: `This sub-command serves an in-memory registry implementation on an automatically chosen port (:0), $PORT or --address
```
imjasonh, line 44: Does address assume it includes the `:` then? Should it always be `localhost:NNNN` ?

cmd/crane/cmd/serve.go
```diff
@@ -54,20 +54,27 @@ Contents are only stored in memory, and when the process exits, pushed data is l
...
 			porti := listener.Addr().(*net.TCPAddr).Port
 			port = fmt.Sprintf("%d", porti)
 
 			bh := registry.NewInMemoryBlobHandler()
-			if disk {
-				tmp := os.TempDir()
-				log.Printf("storing blobs in %s", tmp)
-				bh = registry.NewDiskBlobHandler(tmp)
+			if cmd.Flags().Changed("blobs-to-disk") {
+				path := os.TempDir()
```
imjasonh, line 71: Should this be [`MkdirTemp`](https://pkg.go.dev/os#MkdirTemp) ?

cmd/crane/cmd/serve.go
```diff
@@ -89,7 +96,11 @@ Contents are only stored in memory, and when the process exits, pushed data is l
 			return nil
 		},
 	}
-	cmd.Flags().BoolVar(&disk, "blobs-to-disk", false, "Store blobs on disk")
+	cmd.Flags().StringVarP(&disk, "blobs-to-disk", "", "", "Store blobs on disk")
+	// allow --blobs-to-disk to work without a value
+	cmd.Flags().Lookup("blobs-to-disk").NoOptDefVal = " "
```
imjasonh, line 101:
```suggestion
	cmd.Flags().Lookup("blobs-to-disk").NoOptDefVal = "[[[UNSET]]]"
```

(optional)

The thinking was that if someone might do `--disk=' '` (especially if they're doing something like `--disk=$DISK` and get unexpected behavior, and this makes it a bit harder to stumble into.

cmd/crane/cmd/serve.go
thesayyn: hmm maybe, maybe I didn't change this at all but happy to change now.

cmd/crane/cmd/serve.go
thesayyn: oh it can't be localhost always, that was the behavior in https://github.com/google/go-containerregistry/pull/1746 but that didn't work in a container. 

address is basically can be one of these with no restrictions

`:0` (kernel chosen) 
`localhost:0` (kernel chosen port on loopback interface)
`0.0.0.0:$port` 

sky is the limit here.

thesayyn: i just tried to preserve the old behavior here while allowing users to override the address string.

cmd/crane/cmd/serve.go
thesayyn: hmm makes sense didn't think about that. fixing now.

thesayyn: now that I think about it more, I can just deprecate --blobs-to-disk by printing a warning but still let it override the address.

cmd/crane/cmd/serve.go
thesayyn: `crane registry serve --disk=$(mktemp -d)` works nicely.

```
2023/11/29 12:47:34 storing blobs in /var/folders/c3/qcpmmp4s0_1cy0b3bxjwcsqh0000gn/T/tmp.B4TRj2kT
2023/11/29 12:47:34 serving on port 63643
```

cmd/crane/cmd/serve.go
thesayyn: yeah, that's better. i changed it to mkdirtemp :)

cmd/crane/cmd/serve.go
thesayyn: this is not necessary anymore.

cmd/crane/cmd/serve.go
thesayyn: and if you do `crane registry serve --disk=$(mktemp -d) --blobs-to-disk`

```
Flag --blobs-to-disk has been deprecated, and will stop working in a future release. use --disk=$(mktemp -d) instead.
Error: --disk and --blobs-to-disk can't be used together
```

imjasonh (approved):

thesayyn:
ready to go

PR #1866, opened 2024-01-19
validate.Layer: handle validation for layer content that is not compressed

### LINKED ISSUE #1826: crane: `validate` fails with invalid header message on image with uncompressed layers mediaType
### Describe the bug

`crane validate` fails on an image that appears valid.

### To Reproduce

I've put up a sample image on dockerhub.
```
crane validate --remote jabrown85/crane-validate-failure 
```

### Expected behavior

I expect `crane validate` to pass without error.

### Additional context

`crane` output is an invalid header error.

```
$ crane validate --remote jabrown85/crane-validate-failure
FAIL: jabrown85/crane-validate-failure: validating layers: gzip: invalid header
Error: validating layers: gzip: invalid header
```

The image runs fine (busybox)

```
$ docker run -it jabrown85/crane-validate-failure
/ # exit 0
```

I believe the issue is the manifest looks like this. The base layer is gzipped but other layers are not.
```
$ crane manifest jabrown85/crane-validate-failure | jq
{
  "schemaVersion": 2,
  "mediaType": "application/vnd.docker.distribution.manifest.v2+json",
  "config": {
    "mediaType": "application/vnd.docker.container.image.v1+json",
    "digest": "sha256:65989925e192fceef77b2f6a18901cbb0fdc52783e4890e05726bb6c183a480a",
    "size": 625
  },
  "layers": [
    {
      "mediaType": "application/vnd.docker.image.rootfs.diff.tar.gzip",
      "digest": "sha256:90e01955edcd85dac7985b72a8374545eac617ccdddcc992b732e43cd42534af",
      "size": 727978
    },
    {
      "mediaType": "application/vnd.docker.image.rootfs.diff.tar",
      "digest": "sha256:aa94aec5af6b55423b52074e5223d6bee9896a45558fc2307aceaccac371e476",
      "size": 2048
    },
    {
      "mediaType": "application/vnd.docker.image.rootfs.diff.tar",
      "digest": "sha256:7f194e7fb955ec2e1a0ba1a010faf25a202973ca686226a7482a5847e45a1f1b",
      "size": 2048
    }
  ]
}
```

Another important detail is I'm using Docker Desktop for Mac _with_ the beta containerd storage system. I believe the storage system change is leading to this mixed situation in imgutil, but mixing seems valid.

- Output of `crane version`
```
$ crane version
0.16.1
```
hub.docker.com

nmiyake:
Fixes #1826

### DIFF
--- a/pkg/v1/validate/layer.go
+++ b/pkg/v1/validate/layer.go
@@ -21,6 +21,8 @@ import (
 	"encoding/hex"
 	"errors"
 	"fmt"
+	"github.com/google/go-containerregistry/internal/compression"
+	comp "github.com/google/go-containerregistry/pkg/compression"
 	"io"
 	"strings"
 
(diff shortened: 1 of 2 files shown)


nmiyake:
@bmoylan for reference

pkg/v1/validate/layer.go
```diff
@@ -21,6 +21,8 @@ import (
 	"encoding/hex"
 	"errors"
 	"fmt"
+	"github.com/google/go-containerregistry/internal/compression"
+	comp "github.com/google/go-containerregistry/pkg/compression"
```
bmoylan, line 25: sort imports

bmoylan (commented):
👍 Thanks @nmiyake this should fix the test failures I have in #1865

bmoylan (approved):

PR #1869, opened 2024-01-23
feat: mutate on oci-layout

thesayyn:
#1433 but NOT COMPLETE!

Make most of the crane cli able to mutate images on oci-layout.

The idea is to introduce a local version for some functions in `v1/remote` that does the identical operation on an oci-layout 

cc @developer-guy as you were interested in this.

### DIFF
--- a/cmd/crane/cmd/root.go
+++ b/cmd/crane/cmd/root.go
@@ -47,6 +47,7 @@ func New(use, short string, options []crane.Option) *cobra.Command {
 	insecure := false
 	ndlayers := false
 	platform := &platformValue{}
+	local := ""
 
 	wt := &warnTransport{}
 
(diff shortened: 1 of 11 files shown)


cmd/crane/cmd/root.go
```diff
@@ -137,7 +142,8 @@ func New(use, short string, options []crane.Option) *cobra.Command {
 	root.PersistentFlags().BoolVar(&insecure, "insecure", false, "Allow image references to be fetched without TLS")
 	root.PersistentFlags().BoolVar(&ndlayers, "allow-nondistributable-artifacts", false, "Allow pushing non-distributable (foreign) layers")
 	root.PersistentFlags().Var(platform, "platform", "Specifies the platform in the form os/arch[/variant][:osversion] (e.g. linux/amd64).")
-
+	root.PersistentFlags().StringVar(&local, "local", "", "Use a local oci-layout as remote registry")
```
imjasonh, line 145: This feels like a big change, since it affects all commands. Would it be worth having some e2e tests that set up a local layout and `crane pull --local=<that>`, `crane mutate`, etc.?

PR #1871, opened 2024-01-25
feat: implement local puller and pusher

thesayyn:
Take two on https://github.com/google/go-containerregistry/issues/1433

Most of the crane cli surface will be able to work with images on oci-layout.


pkg/v1/local/layer.go
```diff
@@ -0,0 +1,44 @@
+package local
```
imjasonh, line 1: Is there a reason this shouldn't just go in the layout package? 🤔

pkg/.DS_Store
imjasonh, line 1: Should remove this, and probably add .DS_Store to .gitignore

imjasonh:
Can you fix the build errors so we can see if any other tests are affected?

```
Error: pkg/v1/local/puller.go:63:21: undefined: errors.ErrUnsupported
```

pkg/v1/local/layer.go
thesayyn: Not really, could go into layout package.

thesayyn:
> Can you fix the build errors so we can see if any other tests are affected?
> 
> ```
> Error: pkg/v1/local/puller.go:63:21: undefined: errors.ErrUnsupported
> ```

Done

cmd/crane/cmd/export.go
```diff
@@ -70,20 +70,14 @@ func NewCmdExport(options *[]crane.Option) *cobra.Command {
...
-					img, err = desc.Schema1()
-					if err != nil {
-						return fmt.Errorf("pulling schema 1 image %s: %w", src, err)
-					}
-				} else {
-					img, err = desc.Image()
-					if err != nil {
-						return fmt.Errorf("pulling Image %s: %w", src, err)
-					}
+				var ok bool
```
thesayyn, line 77: This is wrong. inspect media type to see if it's schema1

pkg/v1/layout/pusher.go
```diff
@@ -0,0 +1,183 @@
...
+	}
+	b, desc, err := unpackTaggable(t)
+	if err != nil {
+		return err
+	}
+	if err := lp.path.WriteBlob(desc.Digest, io.NopCloser(bytes.NewBuffer(b))); err != nil {
+		return err
+	}
+
+	if img, ok := mf.(v1.Image); ok {
```
thesayyn, line 131: TODO: image index

pkg/v1/remote/image.go
```diff
@@ -62,12 +63,40 @@ var _ partial.CompressedImageCore = (*remoteImage)(nil)
...
-
-	return desc.Image()
+	desc, err := Artifact(ref, options...)
+	if err != nil {
+		return nil, err
+	}
+	if img, ok := desc.(v1.Image); ok {
+		return img, nil
+	}
+	if idx, ok := desc.(v1.ImageIndex); ok && opts.platform.String() != "" {
```
thesayyn, line 77: Better way to do this?

pkg/v1/layout/puller.go
```diff
@@ -0,0 +1,132 @@
...
+	"github.com/google/go-containerregistry/pkg/v1/partial"
+	"github.com/google/go-containerregistry/pkg/v1/remote"
+	specsv1 "github.com/opencontainers/image-spec/specs-go/v1"
+)
+
+type puller struct {
+	path Path
+}
+
+func NewPuller(path Path) remote.Puller {
```
jonjohnsonjr, line 33: I think we want to define a new interface in `v1` that both `remote.Puller` and `layout.Source` would implement.

In my mind I'd want:

```
package v1

type Source interface{
 // minimal subset of methods needed to read things
}

type Sink interface{
 // minimal subset of methods needed to write things

}
```

pkg/v1/remote/puller.go
```diff
@@ -24,27 +25,40 @@ import (
 	"github.com/google/go-containerregistry/pkg/v1/types"
 )
 
-type Puller struct {
+type Puller interface {
```
jonjohnsonjr, line 28: We want to keep this a concrete struct.

pkg/v1/layout/puller.go
aw185176: That would remove the dependency between `layout` + `remote` and avoid the need to convert `remote.Puller` into a struct below. Spitballing the set of methods:

```go
type Source interface {
  Layer(...)
  Head(...)
  Get(...)
  Artifact(...)
}

type Sink interface {
  Push(...)
  Upload(...) // Maybe not?
}
```

The verbs are a bit clunky for non-remote implementors, but I don't see a clean way around that.

pkg/v1/remote/puller.go
thesayyn: Agreed, will make the change

pkg/v1/layout/puller.go
thesayyn: Hmm, i can't put source and sink into v1 root folder as it creates cycle between `v1` -> `partial.Artifact` -> `v1`


Would putting this into its own directory be acceptable?

```
package github.com/google/go-containerregistry/cmd/crane
        imports github.com/google/go-containerregistry/cmd/crane/cmd
        imports github.com/google/go-containerregistry/internal/cmd
        imports github.com/google/go-containerregistry/internal/verify
        imports github.com/google/go-containerregistry/pkg/v1
        imports github.com/google/go-containerregistry/pkg/v1/partial
        imports github.com/google/go-containerregistry/pkg/v1: import cycle not allowed
```

thesayyn: okay i have created the sinksource (name TBD) directory.

pkg/v1/remote/puller.go
thesayyn: Done.

pkg/v1/layout/puller.go
thesayyn: This is now ready for another look

aw185176: @jonjohnsonjr what do you think about `v1.Artifact` to get around the import cycle?

thesayyn: that one also sounds fair, though now v1.Artifact would depend on bunch of partial interfaces.

thesayyn: anything i could do to push this over the edge @jonjohnsonjr? I'd really like to land this, as i have some people who's been waiting for it for a long time.

PR #1875, opened 2024-02-05
Ensure we can access layer by diffID on a mutated image

natalieparellano:
Previously, `calling LayerByDiffID()` immediately after adding a layer would error with the layer not found.

By ensuring that we call "compute" on the image before trying to access the diffID map, we ensure that the layer is found.


natalieparellano:
@codecov-commenter I disagree! Maintainers, any opposition to merging this one?

PR #1921, opened 2024-04-10
Fix windows race condition when writing image with duplicate layers

dgannon991:
This is a potential solution for #1920, any feedback would be greatly appreciated.

### DIFF
--- a/pkg/v1/layout/write.go
+++ b/pkg/v1/layout/write.go
@@ -22,6 +22,8 @@ import (
 	"io"
 	"os"
 	"path/filepath"
+	"runtime"
+	"sync"
 
 	"github.com/google/go-containerregistry/pkg/logs"
 	v1 "github.com/google/go-containerregistry/pkg/v1"
@@ -36,6 +38,8 @@ import (
 var layoutFile = `{
     "imageLayoutVersion": "1.0.0"
 }`
+
+var renameMutex sync.Mutex
 
 // AppendImage writes a v1.Image to the Path and updates
 // the index.json to reference it.
@@ -259,6 +263,11 @@ func (l Path) writeBlob(hash v1.Hash, size int64, rc io.ReadCloser, renamer func
 	}
 
 	renamePath := l.path("blobs", finalHash.Algorithm, finalHash.Hex)
+
+	if runtime.GOOS == "windows" {
+		renameMutex.Lock()
+		defer renameMutex.Unlock()
+	}
 	return os.Rename(w.Name(), renamePath)
 }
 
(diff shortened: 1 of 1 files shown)


slonopotamus:
I'd say a more robust solution is "try to rename, if it fails, but destination exists, just remove the source and pretend you're fine". This would also cover the case when multiple `crane` processes stomp on the same file.

slonopotamus:
I had a _very_ similar issue in [git-lfs](https://github.com/git-lfs/git-lfs) in the past. Here's the analysis: https://github.com/git-lfs/git-lfs/pull/3880#issue-512345192 and here's the final fix that solved the issue: https://github.com/git-lfs/git-lfs/commit/662a6248198594a59c5a652b1ab543d8948c302a

dgannon991:
I'll give that a go, thank you for the feedback.

dgannon991:
@slonopotamus I've written this up and given it a test, and I can confirm it works:
```go
	err = os.Rename(w.Name(), renamePath)
	var badInputErr *os.LinkError
	if err != nil && errors.As(err, &badInputErr) && badInputErr.Err == syscall.ERROR_ACCESS_DENIED {
		if _, err := os.Stat(renamePath); err == nil {
			// windows os.Rename() issue #1920
			// the destination already exists, just remove the target and continue
			return os.Remove(w.Name())
		}
	}
	return err
```

I do have to say though of the two solutions I think I'm leaning towards the first one, as it makes os.Rename behave more like linux (in that it's an atomic transaction.) Would you mind if I left this up for a second opinion? I'm more than happy to swap over to this approach, but I worry it would mask other issues outside of the scope of this ticket (as you mentioned, running two crane processes at the same time, it got me thinking that realistically that /should/ be an error!)

slonopotamus:
Sure, I'm just passing by, and simply gave my thoughts.

dgannon991:
@slonopotamus - thank you for taking the time, the thoughts were definitely appreciated, and it was great to have a confirm that it was an issue in other places!

dgannon991:
Hi Guys,
Is there anything more I need to do on this one? Just aware it will auto close at some point!
Cheers,
David

pkg/v1/layout/write.go
```diff
@@ -37,6 +39,8 @@ var layoutFile = `{
     "imageLayoutVersion": "1.0.0"
 }`
 
+var renameMutex sync.Mutex
```
imjasonh, line 42:
```suggestion
// renameMutex guards os.Rename calls in AppendImage on Windows only.
var renameMutex sync.Mutex
```

imjasonh (approved):

Issue #1832, opened 2023-10-31
question: Support OCI artifacts with artifactType


knqyf263:
I tried to generate a manifest with `config.mediaType: "application/vnd.oci.empty.v1+json"` and `artifactType: "application/vnd.example+type"` as described below.
https://github.com/opencontainers/image-spec/blob/93f6e65855a1e046b42afbad0ad129a3d44dc971/manifest.md#guidelines-for-artifact-usage

Especially, I want to generate the second example.

```
{
  "schemaVersion": 2,
  "mediaType": "application/vnd.oci.image.manifest.v1+json",
  "artifactType": "application/vnd.example+type",
  "config": {
    "mediaType": "application/vnd.oci.empty.v1+json",
    "digest": "sha256:44136fa355b3678a1146ad16f7e8649e94fb4fc21fe77e8310c060f61caaff8a",
    "size": 2
  },
  "layers": [
    {
      "mediaType": "application/vnd.example+type",
      "digest": "sha256:e258d248fda94c63753607f7c4494ee0fcbe92f1a76bfdac795c9d84101eb317",
      "size": 1234
    }
  ]
}
```

I used [empty.Image](https://github.com/google/go-containerregistry/blob/dbcd01c402b2f05bcf6fb988014c5f37e9b13559/pkg/v1/empty/image.go#L26) and the mutate package, like [mutate.ConfigMediaType](https://github.com/google/go-containerregistry/blob/dbcd01c402b2f05bcf6fb988014c5f37e9b13559/pkg/v1/mutate/mutate.go#L542).

```
content := []byte(`hello`)
artifactType := "application/vnd.example+type"
img, err := mutate.AppendLayers(empty.Image, static.NewLayer(content, artifactType))
if err != nil {
	panic(err)
}

img = mutate.ConfigMediaType(img, artifactType)
```

However, it generates the following manifest.

```
{
  "schemaVersion": 2,
  "mediaType": "application/vnd.docker.distribution.manifest.v2+json",
  "config": {
    "mediaType": "application/vnd.example+type",
    "size": 233,
    "digest": "sha256:185de42e4ed8eb4fc529c9a3ea02d808c1dfe4466737c0bcf30a5d7ad855088f"
  },
  ...
```

`mutate.ConfigMediaType(img, "application/vnd.oci.empty.v1+json")` also doesn't work.

[...]

PR #1931, opened 2024-04-23
feat: Add v1.Manifest 'artifactType' field from OCI spec

bmoylan:
OCI has released v1.1.0 of the Image Spec which includes the artifactType field in the manifest json schema: https://github.com/opencontainers/image-spec/blob/v1.1.0/manifest.md#image-manifest-property-descriptions

Relevant to #1832 closed as stale

### DIFF
--- a/pkg/registry/manifest.go
+++ b/pkg/registry/manifest.go
@@ -379,6 +379,8 @@ func (m *manifests) handleReferrers(resp http.ResponseWriter, req *http.Request)
 	target := elem[len(elem)-1]
 	repo := strings.Join(elem[1:len(elem)-2], "/")
 
+	queryArtifactType := req.URL.Query().Get("artifactType")
+
 	// Validate that incoming target is a valid digest
 	if _, err := v1.NewHash(target); err != nil {
 		return &regError{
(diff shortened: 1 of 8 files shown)


cdupuis:
@bmoylan, Thanks for doing the work on this. Any chance to get this into main any time soon? It would be super helpful as we start to see more OCI 1.1 adoption across the ecosystem.

bmoylan:
👋 Hey @jonjohnsonjr @imjasonh sorry for the pings but it would be great to get this reviewed when possible, thanks!

kipz:
Yeah, it'd be great to get this in.

kipz:
@bmoylan I've added a couple of extra bits for better `artifactType` here: https://github.com/bmoylan/go-containerregistry/pull/2

* Set artifactType in fallback descriptor (if present)
* Similarly for in-memory registry: set artifactType instead of config mediaType (if present)
* Send artifactType filter to registry if supplied in call to referrers API

nmiyake:
I noticed this issue as well and it would be great to get this PR merged!

Another issue with regards to diffs between the current implementation and the OCI 1.1 referrers specification is that the specification states that the referrer descriptors `MUST` have the annotations from the `Image` or `ImageIndex`, but currently this library doesn't do that.

I opened PR #1998 to fix the "annotations" issue. I believe that merging these 2 PRs should make `go-containerregistry` compliant in its behavior.

I didn't see an existing issue for this, so I filed #1997 to explicitly describe the issues.

takirala:
Any idea why this went stale? The PR solves a real problem and the approach seems intuitive. Hey @jonjohnsonjr or @imjasonh any chance you could take a look at this? Thanks!

malt3:
This is now implemented as part of #2269.

PR #1939, opened 2024-05-06
Support copying multiple platforms.

ameir:
`crane index filter` is a great way to copy multi-platform images without copying undesired platforms. Unfortunately, that command requires that the source image support multiple platforms (i.e., has a manifest list/index). In a scenario where you may be programmatically copying images, some source images may not host a manifest list/index, and additional checks must be performed to decide between `crane index filter` and `crane copy`.

This PR updates`crane copy` so that it behaves similar to `crane index filter`, but does not require a manifest list/index. In a single-platform scenario, it behaves much as it did before. In a multi-platform scenario, it filters the manifest list/index to the desired platforms, and copies only those layers.

Example command:
```
crane copy --platform linux/amd64 --platform linux/arm64 redis localhost:58273/redis
```

### DIFF
--- a/cmd/crane/cmd/copy.go
+++ b/cmd/crane/cmd/copy.go
@@ -23,6 +23,7 @@ import (
 
 // NewCmdCopy creates a new cobra.Command for the copy subcommand.
 func NewCmdCopy(options *[]crane.Option) *cobra.Command {
+	platforms := &platformsValue{}
 	allTags := false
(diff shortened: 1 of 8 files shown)


ameir:
lifecycle/frozen

ameir:
@imjasonh / @jonjohnsonjr do either of you have cycles to take a look at this PR? Thanks!

PR #1940, opened 2024-05-07
pkg/registry: Export errNotFound

Eun:
The registry logic checks for the unexported error `errNotFound` during blob handling.  
When implementing a custom `BlobHandler` outside of this package the developer is unable to use the unexported `errNotFound` variable, therefore the custom `BlobHandler` cannot signal the correct error to the registry package which results in a wrong execution flow.

This PR exposes the `ErrNotFound` error so it can be used from outside the registry package, this aligns with other packages in this module that use the same approach e.g. [cache](https://github.com/google/go-containerregistry/blob/main/pkg/v1/cache/cache.go#L49)

### DIFF
--- a/pkg/registry/blobs.go
+++ b/pkg/registry/blobs.go
@@ -51,14 +51,14 @@ func isBlob(req *http.Request) bool {
 // BlobHandler represents a minimal blob storage backend, capable of serving
 // blob contents.
 type BlobHandler interface {
-	// Get gets the blob contents, or errNotFound if the blob wasn't found.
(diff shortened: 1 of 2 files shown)


Eun:
Can anybody take a look at this?

tylerwilliams:
Just ran into this issue myself. It would be great if we could get this CL merged :)

tylerwilliams:
@jonjohnsonjr would you mind giving this a look?

Eun:
Any update on this one?

PR #1947, opened 2024-05-14
fix: redact.URL uses (*URL).Redacted to omit basic-auth password

bmoylan:
When a request URL contains basic auth information, that URL segment is ending up in the logs unredacted. Use the standard library's `Redacted()` function to fix this.


bmoylan:
Hi @jonjohnsonjr @imjasonh this small change should fix leaking tokens into logs, hoping you can take a quick look!

bmoylan:
not stale! @jonjohnsonjr @imjasonh could you take a look at this? Thanks!

jonjohnsonjr (approved):

PR #1966, opened 2024-06-27
Refined registry credential lookup

zregvart:
The Container Tools project defines a lookup for authentication from most-specific to least-specific[1]. This change implements that credential lookup.

[1] https://github.com/containers/image/blob/main/docs/containers-auth.json.5.md#format


zregvart:
@jonjohnsonjr, @imjasonh and @matthewrobertson I'm curious about your thoughts on adopting this change?

zregvart:
> @jonjohnsonjr, @imjasonh and @matthewrobertson I'm curious about your thoughts on adopting this change?

I'm gauging the interest here, do the maintainers wish to incorporate this change?

chmeliik:
@jonjohnsonjr @imjasonh @matthewrobertson would you be interested in aligning the credentials lookup with the containers-auth.json spec? This would be useful to avoid needing various [workarounds](https://github.com/konflux-ci/build-trusted-artifacts/blob/7a9a66fe70481511d9a45ff5dedc5dd291210aaf/select-oci-auth.sh#L1) for tools that use this library for authentication, such as [cosign](https://github.com/sigstore/cosign).

PR #1974, opened 2024-07-10
crane: Add timestamp to flatten layer

nivbend:
When flattening an image the blob's `created` field is the same as the original image's, while the flattened layer's `created` field is a zero timestamp. This makes it impossible to figure out when was the flattened image actually created and distinguish it from the original.

Set both fields to the current time.


nivbend:
Fixes #1996.

nivbend:
@imjasonh @thesayyn @jonjohnsonjr Would you have some time to review this minor fix? Much appreciated.

Stephanie0829:
Hi! I wanted to open a similar issue/PR. @jonjohnsonjr @Subserial @imjasonh Mind reviewing this so these changes can be merged (much appreciated)?

brandonmichigangithub:
I'm going to close this as https://github.com/google/go-containerregistry/pull/2117 does the exact same.

Issue #1980, opened 2024-07-18
crane: Windows images failed to pull after pushing to local hosted Docker Registry instance


laozc:
### Describe the bug

Some Windows image derived from windows-host-process-container-base-image were not able to pull by containerd
after they were pushed to local hosted Docker Registry instance using crane.

### To Reproduce

crane push windows-image.tar localhost:5000/imgetag:windows

containerd may report failure when pulling this image due to corrupted image content.

### Expected behavior

The push should not resulted in corrupted image blob.
containerd should be able to pull the image.

### Additional context

- Output of `crane version`: main
- Registry used (e.g., GCR, ECR, Quay)
Docker Registry

laozc:
Based on a further investigation on the issue,
we found that the base layer of windows-host-process-container-base-image is not gzip compressed

$ file blobs/sha256/f2784c8bc8525977aef9b6cd74e9a899274db7d82cd4b028e0ae049270927585 <= this is the base layer of windows-host-process-container-base-image
blobs/sha256/f2784c8bc8525977aef9b6cd74e9a899274db7d82cd4b028e0ae049270927585 POSIX tar archive (GNU)

$ file blobs/sha256/f1912f0046069c73564ef7d8cd9257698d5a9a2c4d4da8910dbb35e1865b3cb5 <= this is our layer to build upon windows-host-process-container-base-image
blobs/sha256/f1912f0046069c73564ef7d8cd9257698d5a9a2c4d4da8910dbb35e1865b3cb5: gzip compressed data, original size modulo 2^32 5632

This may break the image blob after the image gets uploaded to a remote registry.

PR #1981, opened 2024-07-18
Fixes detection if the image contains compressed layers

laozc:
This PR fixes the detection for a image with compressed layers.

Currently only the first base layer is probed to determine whether the image contains compressed layers.
However, the image may consist from a uncompressed base layer and compressed upper layers.
Typically for the image windows-host-process-container-base-image shipped from MSFT,
the base layer is a tar without compression.

Fixes: #1980

### DIFF
--- a/pkg/v1/tarball/image.go
+++ b/pkg/v1/tarball/image.go
@@ -96,7 +96,7 @@ func Image(opener Opener, tag *name.Tag) (v1.Image, error) {
 		return nil, err
 	}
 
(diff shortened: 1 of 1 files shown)


akhilerm:
@jonjohnsonjr Can you take a look at this patch and see if it can be taken in?

akhilerm:
ping @imjasonh Can you take a look at this change? Is this something that can be taken into crane?

PR #2009, opened 2024-09-14
fix: remote.Referrers: Set remoteIndex.ref to resolve panics in Index manifests

bmoylan:
When `remote.Referrers()` returns a `v1.Index`, fetching derivative objects (like referenced manifests) requires keeping the Reference to the original repository.

Without setting the `ref` field, we get nil pointer dereference panics in (*remoteIndex).childDescriptor() https://github.com/google/go-containerregistry/blob/005bb7197dad1a6bffcda9405b91990331b1f406/pkg/v1/remote/index.go#L206-L207

### DIFF
--- a/pkg/v1/remote/referrers.go
+++ b/pkg/v1/remote/referrers.go
@@ -89,6 +89,7 @@ func (f *fetcher) fetchReferrers(ctx context.Context, filter map[string]string, 
 	}
 	idx := &remoteIndex{
 		fetcher:   *f,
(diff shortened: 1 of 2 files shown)


bmoylan:
Hi @jonjohnsonjr @imjasonh hoping you can take a look at this panic fix. I hope I'm not missing any side effects with the `*remote.Index` now holding a `ref` that doesn't actually exist in the remote's manifest store (since the referrers index is created on-the-fly)

PR #2013, opened 2024-09-19
Use OCI manifest format for images from scratch

Silvanoc:
Images built from scratch where still using the outdated Docker v2 image format (recognizable in the mediaType `application/vnd.docker.distribution.manifest.v2+json`). This patch replaces that outdate format with the current one based in the OCI standard (marked by the mediaType `application/vnd.oci.image.manifest.v1+json`).

### DIFF
--- a/pkg/v1/empty/image.go
+++ b/pkg/v1/empty/image.go
@@ -29,7 +29,7 @@ type emptyImage struct{}
 
 // MediaType implements partial.UncompressedImageCore.
 func (i emptyImage) MediaType() (types.MediaType, error) {
-	return types.DockerManifestSchema2, nil
+	return types.OCIManifestSchema1, nil
 }
 
 // RawConfigFile implements partial.UncompressedImageCore.


jonjohnsonjr:
> The v1 file layout and manifests are no longer used in Moby and Docker, except in docker save and docker load.

The v1 file layout is not the same as docker schema 2, which is modern and ~equivalent to OCI v1.

Silvanoc:
> > The v1 file layout and manifests are no longer used in Moby and Docker, except in docker save and docker load.
> 
> The v1 file layout is not the same as docker schema 2, which is modern and ~equivalent to OCI v1.

You are right on that. Nevertheless, the issue remains, that Docker image v2 is not conformant with OCI v1 and therefore some tools that are OCI-only will have issues with them.

Now the question is if this projects wants to stay consistent with the IMHO Docker inconsistent behavior on images built from scratch compared with those available in the Docker Hub Library. Apparently the change in the Library is quite new (last 3 months or so), so perhaps BuildX will adapt its behavior building from scratch in the upcoming future.

Silvanoc:
> Now the question is if this projects wants to stay consistent with the IMHO Docker inconsistent behavior on images built from scratch compared with those available in the Docker Hub Library. Apparently the change in the Library is quite new (last 3 months or so), so perhaps BuildX will adapt its behavior building from scratch in the upcoming future.

I am mixing here the fact that BuildKit is being developed by the Moby project and Docker BuildX is surely simply taking over their default, whereas the Docker Hub is directly being provided by Docker Inc. So possibly Docker Inc. is also unhappy with the mixture.

Silvanoc:
It looks to me as if Docker schema v2 should be replaced by OCI v1, even in Moby: https://github.com/moby/buildkit/blob/5cd5a63de5d37512c24f20085607509b749df762/exporter/containerimage/writer.go#L490

PR #2023, opened 2024-10-17
Detect zstd in crane append

jonjohnsonjr:
We assumed everything was gzip, which is not always true!


brancz:
Can confirm that this patch works for us!

imjasonh (approved):

pkg/crane/append.go
```diff
@@ -90,6 +90,21 @@ func getLayer(path string, layerType types.MediaType) (v1.Layer, error) {
 		return stream.NewLayer(f, stream.WithMediaType(layerType)), nil
 	}
 
+	// This is dumb but the tarball package assumes things about mediaTypes that aren't true
+	// and doesn't have enough context to know what the right default is.
+	f, err = os.Open(path)
+	if err != nil {
+		return nil, err
+	}
+	defer f.Close()
```
hunshcn, line 99: Maybe we should close it immediately after PeekCompression?

slonopotamus:
1. Does this work properly if blob is feeded through stdin?
2. I'm not sure this is enough to properly upload a zstd image, https://github.com/google/go-containerregistry/pull/1827 involves more changes (and also fixes the issue you're handling here).

Issue #2045, opened 2025-01-21
question: what is the Go version support policy?


spencerschrock:
The newest release (v0.20.3) includes a change from #2025 which bumps the Go directive from `1.18` to `1.23.0`.  As far as I know, the reason wasn't due to a feature required in Go 1.23, but rather "update everything to latest so that things build in CI." This change propagates to dependent libraries, so we are also required to use Go 1.23.

Would you consider walking back the 1.23 requirement if you don't need the features? Or supporting the last two versions of Go (1.22 too  in this case).

There are other ways to use the latest version of Go in your CI, without requiring it for the project:

Instead of 
```yaml
    - uses: actions/setup-go@v5
      with:
        go-version-file: go.mod
```

There is also an `actions/setup-go` [stable alias](https://github.com/actions/setup-go/blob/f111f3307d8850f501ac008e886eec1fd1932a34/README.md#using-stableoldstable-aliases) for the latest Go release:
```yaml
    - uses: actions/setup-go@v5
      with:
        go-version: stable

```

@imjasonh would this work for you?

edit: I see one of your dependencies requires at least `1.22.7`, so the request would be to use that instead of `1.23.0` (although `pkg/authn/k8schain/go.mod` and `pkg/authn/kubernetes/go.mod` would require 1.23.0 for theirs)

PR #2047, opened 2025-01-24
chore(deps): relax go directive to permit 1.22.x

dnwe:
Setting the go directive to 1.23.0 requires all consumers of the module
to be building with 1.23.0 or newer release of Go 1.23 and to update
their own go.mod accordingly — this seems unnecessarily restrictive for
a library module, particularly as the code itself doesn't currently use
any modern language constructs and builds fine even with older Go
versions.

Instead set the go directive to 1.22.0 and use the toolchain directive
to recommend the latest 1.23.x when building locally.

Note: this also includes a run of `hack/bump-deps.sh` to ensure all deps
are compatible with go1.22.0 as well

Fixes https://github.com/google/go-containerregistry/issues/2045

### DIFF
--- a/hack/bump-deps.sh
+++ b/hack/bump-deps.sh
@@ -18,28 +18,35 @@ set -o errexit
 set -o nounset
 set -o pipefail
 
+# go directive in go.mod defines module compatibility
+export GOVERSION=1.22.0
+# gotoolchain directive in go.mod recommends compile toolchain version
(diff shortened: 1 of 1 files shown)


dnwe:
@imjasonh would you be happy to take a look at this PR?

haydentherapper:
@imjasonh @jonjohnsonjr Bumping this PR for review, this is needed for us to stay on Go 1.22 in sigstore.

PR #2052, opened 2025-01-31
Support ~/.config podman path in default keychain

bduffany:
According to [podman's auth docs](https://docs.podman.io/en/v5.1.0/markdown/podman-login.1.html#authfile-path):

> Default is `${XDG_RUNTIME_DIR}/containers/auth.json` on Linux, and `$HOME/.config/containers/auth.json` on Windows/macOS.

The current default keychain supports the `$XDG_RUNTIME_DIR` path but not the `$HOME/.config` path. This PR fixes that.

Note: podman's behavior is actually a little more subtle than what these docs would imply:
- The `$HOME/.config/containers` path is actually supported on all platforms, not just Windows/macOS. The `$HOME/.config` part of the path is computed as `$XDG_CONFIG_HOME` if defined, otherwise it defaults to `$HOME/.config`. The logic for that is [here](https://github.com/containers/podman/blob/c131c9d0382e8dedd460d8a9844a0cc2e65ecb71/vendor/github.com/containers/image/v5/pkg/docker/config/config.go#L157-L160).
- The `$XDG_RUNTIME_DIR` is indeed Linux-only, though, and takes precedence over the `~/.config` path. The logic for this precedence is [here](https://github.com/containers/podman/blob/c131c9d0382e8dedd460d8a9844a0cc2e65ecb71/vendor/github.com/containers/image/v5/pkg/docker/config/config.go#L147-L156), and the relevant code from `getPathToAuth` is [here](https://github.com/containers/podman/blob/c131c9d0382e8dedd460d8a9844a0cc2e65ecb71/vendor/github.com/containers/image/v5/pkg/docker/config/config.go#L573-L577).

### DIFF
--- a/pkg/authn/keychain.go
+++ b/pkg/authn/keychain.go
@@ -95,9 +95,13 @@ func (dk *defaultKeychain) ResolveContext(_ context.Context, target Resource) (A
 
 	// First, check $HOME/.docker/config.json
 	foundDockerConfig := false
+	configHome := os.Getenv("XDG_CONFIG_HOME")
 	home, err := homedir.Dir()
 	if err == nil {
 		foundDockerConfig = fileExists(filepath.Join(home, ".docker/config.json"))
+		if configHome == "" {
+			configHome = filepath.Join(home, ".config")
+		}
 	}
 	// If $HOME/.docker/config.json isn't found, check $DOCKER_CONFIG (if set)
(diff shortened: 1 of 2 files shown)


bduffany:
Friendly ping! It would simplify things quite a bit for us if this were upstreamed.

PR #1998, opened 2024-08-13
Set annotations on descriptors created by referrers API

nmiyake:
Implement the behavior described by the OCI 1.1 specification at https://github.com/opencontainers/distribution-spec/blob/main/spec.md#listing-referrers for annotations for descriptors in the ImageIndex returned by the referrers API.

Addresses #1997

Issue #1997, opened 2024-08-13
ggcr: descriptors in ImageIndex returned by referrers API do not fulfill requirements stated in OCI spec


nmiyake:
### Describe the bug
The [OCI 1.1 specification for the "referrers" API](https://github.com/opencontainers/distribution-spec/blob/main/spec.md#listing-referrers) states that, for the descriptors in the returned referrers index, `The descriptors MUST include an artifactType field that is set to the value of the artifactType in the image manifest or index, if present` and `The descriptors MUST include annotations from the image manifest or index.`.

However, the current `go-containerregistry` library does not fulfill either of these requirements.

### To Reproduce
For annotations:

1. Create a `v1.Image` with a non-empty `annotations` field and `subject` field and use `remote.Write` to write the image
2. Call `remote.Referrers` on the subject and examine the manifest entry
3. Current behavior is that the descriptor for the image from step 1 does not have an `annotations` field

For `artifactType`:

1. Create a `v1.Image` that specifies a value for the `artifactType` field (can be done by including this field in the output of the `RawManifest()` implementation of `v1.Image`) and `subject` field and use `remote.Write` to write the image
2. Call `remote.Referrers` on the subject and examine the manifest entry
3. Current behavior is that the `artifactType` field of the descriptor for the image from step 1 is set to the value of `Config.MediaType`

### Expected behavior
For annotations:
* The descriptor for image should have an `annotations` field that matches the `annotations` field of the image

For `artifactType`:
* The descriptor for the image should have an `artifactType` value that matches the `aritfactType` field of the image

### Additional context
Full text of the referrers specification:

> Upon success, the response MUST be a JSON body with an image index containing a list of descriptors. The Content-Type header MUST be set to application/vnd.oci.image.index.v1+json. Each descriptor is of an image manifest or index in the same \<name\> namespace with a subject field that specifies the value of \<digest\>. The descriptors MUST include an artifactType field that is set to the value of the artifactType in the image manifest or index, if present. If the artifactType is empty or missing in the image manifest, the value of artifactType MUST be set to the config descriptor mediaType value. If the artifactType is empty or missing in an index, the artifactType MUST be omitted. The descriptors MUST include annotations from the image manifest or index. If a query results in no matching referrers, an empty manifest list MUST be returned.

[...]

PR #2068, opened 2025-02-24
Set annotations on descriptors created by referrers API 

ilopezluna:
Follow up of #1998
This PR just verifies that we see those values in the descriptor as mentioned [here](https://github.com/google/go-containerregistry/pull/1998#discussion_r1842980363)

### DIFF
--- a/pkg/registry/manifest.go
+++ b/pkg/registry/manifest.go
@@ -426,6 +426,7 @@ func (m *manifests) handleReferrers(resp http.ResponseWriter, req *http.Request)
 			Config struct {
 				MediaType string `json:"mediaType"`
 			} `json:"config"`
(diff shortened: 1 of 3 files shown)


cmurphy:
@ilopezluna it looks like you haven't signed the Google CLA, you might get traction on this PR if you follow up on that step.

ilopezluna:
> @ilopezluna it looks like you haven't signed the Google CLA, you might get traction on this PR if you follow up on that step.

Thanks @cmurphy already signed it, do I have to something else?

cmurphy:
@jonjohnsonjr @imjasonh any chance you could have a look at this? This is a partial fix for https://github.com/google/go-containerregistry/issues/1997 which was automatically closed due to inactivity.

PR #2118, opened 2025-07-15
Migrate from docker/distribution to distribution/distribution

dereknola:
## Changes
- Move from docker/distribution to distribution/distribution. 
- Cleanup documentation references to point to new docs locations

**Note:** that the same version `v2.8.3` is used, but this now enables future upgrades to this dependency, as docker/distribution was sunset back in 2023. Future upgrades to v3.0.0+ now become possible with this PR.


dereknola:
Side stepped with https://github.com/google/go-containerregistry/commit/19a36cd8c44dd465a84f9d6ffea3009023f73660

PR #2121, opened 2025-07-28
isBlob: avoid panic on invalid path

malt3:
The array access of `elem[len(elem)-1]` can be invalid if `len(elem) == 0`.


imjasonh (approved):

imjasonh:
cc @jonjohnsonjr

PR #2144, opened 2025-10-19
Add resumable download support via range requests

ericcurtin:
So we can resume pulls

### DIFF
--- a/pkg/registry/blobs.go
+++ b/pkg/registry/blobs.go
@@ -302,13 +302,19 @@ func (b *blobs) handle(resp http.ResponseWriter, req *http.Request) *regError {
 		}
 
 		if rangeHeader != "" {
-			start, end := int64(0), int64(0)
+			start, end := int64(0), size-1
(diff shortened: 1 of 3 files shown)


ericcurtin:
This is important, it is getting common for people to put AI models in layers. They can be many GBs in size increasing the chances of interrupted downloads being problematic.

ericcurtin:
HI @brandonmichigangithub @Subserial

I'd appreciate it if you could take a look at this PR when you have a moment. As mentioned, resumable downloads are becoming increasingly important for large container layers, like those used for AI models, where download interruptions are more likely.

I've also noticed that the PR queue is quite long and the merge velocity seems low. I completely understand this is common for popular and busy projects. This got me thinking about the project's long-term bandwidth.

Has the team considered adding more co-maintainers from the community to help manage the backlog of issues and PRs? There are likely several people (myself included) who would be willing to step up, share the load, and help keep the project moving.

Thanks for your time and for maintaining this project!

PR #2152, opened 2025-11-13
feat(transport): add resumable transport for remote resources

derekhjray:
Add resumable transport support for pulling images from registry, which is useful for huge image pulling operation, and avoid downloading all image/layer data again when error hanpens

### DIFF
--- a/pkg/v1/remote/transport/resumable.go
+++ b/pkg/v1/remote/transport/resumable.go
@@ -9,15 +9,24 @@ import (
 	"strconv"
 	"strings"
 	"sync/atomic"
+	"time"
 
 	"github.com/google/go-containerregistry/pkg/logs"
 )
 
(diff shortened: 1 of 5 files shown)


pkg/v1/remote/transport/resumable.go
```diff
@@ -0,0 +1,242 @@
...
+	}
+
+	if err = rb.resume(err); err == nil {
+		if n == 0 {
+			// zero bytes read, try reading again with new response.Body
+			goto resume
+		}
+
+		// already read some bytes from previous response.Body, returns and waits for next Read operation
+	}
```
Subserial: Please avoid labels and express this as a loop.

derekhjray: > Please avoid labels and express this as a loop.

modified

PR #2157, opened 2025-11-13
Fix reviewdog arguments

Subserial:
`fail-on-error: true` was deprecated in favor of `fail-level: error`.


Subserial:
Reviewdog was a red herring for the dependabot issues.

PR #2165, opened 2025-11-21
feat(tarball): add tarball option to override default attributes

derekhjray:
Add option to override default layer attributes, in case of re-calculating digest or diff_id will use too much time for huge layer


Subserial (changes requested):
Looks good, just add tests and I'll approve this.

PR #2166, opened 2025-11-21
fix(ping): fix auto detect service scheme issue

derekhjray:
Add auto registry scheme(https/http) detect for Ping().

If registry auto redirect http to https( or https to http), and we try to use name.Insecure option, Ping() func has chances stopped by http.Client for too many redirects. And the fixture just replace original scheme with successful request's scheme


Subserial (changes requested):
Please add a test for this case.

derekhjray:
> Please add a test for this case.

tests added

PR #2223, opened 2026-02-27
Adds local file support to the `crane index` subcommand 

edwardthiele:
This PR adds support to the `crane index` subcommand to support local OCI layout directories and tarballs. It unifies the handling of local and remote references, allowing users to create, modify, and inspect indices across both local filesystems and remote registries.

## Key Features
* **Local OCI Layout Support:** `crane index append` can now target local directories. If the target directory does not exist, it is automatically initialized as a new OCI layout.
* **Mixed Reference Support:** Manifests from local tarballs, local OCI layouts, and remote registries can be combined in a single command. For example:
    ```bash
    crane index append -m ./image.tar -m gcr.io/distroless/base -t ./my-layout
    ```
* **Index Listing:** A new `crane index list` subcommand allows for inspecting the digests, media types, and platform information of any index (local or remote).
* **Platform Metadata Preservation:** The tool automatically extracts platform information from image configurations to ensure the resulting index manifest correctly records `os/arch` metadata.
* **Nesting & Flattening Control:** Supports the `--flatten` flag to control whether appended indices are expanded into their constituent manifests or added as nested indices.
* **Docker Compatibility:** Includes the `--docker-empty-base` flag for creating indices with Docker-specific media types when required.

## Verification
Comprehensive integration tests in `cmd/crane/index_test.sh` verify:
* Appending local tarballs/layouts to local indices.
* Appending remote manifests to local indices.
* Appending local manifests to remote indices.
* Nested index creation and platform metadata preservation.
* Docker media type compatibility.

> **Note:** All tests have been verified against both the local filesystem and a live Artifact Registry repository.

### DIFF
--- a/pkg/v1/remote/transport/ping.go
+++ b/pkg/v1/remote/transport/ping.go
@@ -15,6 +15,7 @@
 package transport
 
 import (
+	"cmp"
 	"context"
 	"errors"
 	"fmt"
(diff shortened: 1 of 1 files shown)


mattmoor:
cc @jonjohnsonjr

PR #2227, opened 2026-03-06
mutate: reject path traversal and symlink escape in Extract

KevinZhao:
## Summary

`mutate.Extract()` (used by `crane export` and `crane flatten`) passes tar entry names and symlink targets to the output stream without validating they stay within the extraction root. A malicious container image can exploit this to:

- **Path traversal**: Write files outside the extraction directory via `../../` prefixes (`filepath.Clean` preserves leading `../` sequences)
- **Symlink escape**: Create symlinks pointing to arbitrary host paths (`header.Linkname` is never validated)
- **Absolute path injection**: Target fixed host locations via absolute paths in entry names

This affects any consumer that extracts the output of `mutate.Extract()`, including `crane export <image> - | tar xf - -C /dir`.

## Changes

- Skip tar entries where `header.Name` has a `../` prefix or is an absolute path
- Skip symlink/hardlink entries where `header.Linkname` has a `../` prefix or is an absolute path
- Add table-driven tests covering all five attack vectors plus a positive case for safe relative symlinks

## References

- Docker CVE-2018-15664: symlink-exchange attack in `docker cp`
- Docker CVE-2019-14271: path traversal in `docker cp`
- Snyk Zip Slip: https://security.snyk.io/research/zip-slip-vulnerability


KevinZhao:
Found and fixed the issues from the first commit's CI:

1. **Test failures** (commit 2 already fixed): \`TestCraneFilesystem\` and \`TestEditFilesystem\` failed because absolute paths like \`/some/file\` were being skipped. Changed behavior to normalize (strip leading \`/\`) instead of skip, which preserves entries while preventing host-path injection.

2. **golangci-lint** (commit 3, just pushed): \`tarball.LayerFromReader\` is deprecated (SA1019). Replaced with \`tarball.LayerFromOpener\` in the test helper.

All checks pass locally: \`golangci-lint run\`, \`go test -race ./...\`, \`./hack/presubmit.sh\`.

PR #2263, opened 2026-04-13
fix: handle missing DiffIDs in tarball Manifest() to prevent panic

### LINKED ISSUE #2192: ggcr: panic on getting image layers with config `application/vnd.buildkit.cacheconfig.v0` from tar
### Describe the bug

When calling [`v1.Image.Layers()`](https://pkg.go.dev/github.com/google/go-containerregistry@v0.20.7/pkg/v1#Image.Layers) from an imported 'tar' image whose config is media type `application/vnd.buildkit.cacheconfig.v0`, the function will panic.

This is due to the layer code attempting to read layer information from `DiffIDs` array which does not exist in the cacheconfig type (linked [here](https://github.com/google/go-containerregistry/blob/main/pkg/v1/tarball/image.go#L368)). 

### To Reproduce

1. Export a build cache image to tar using [`tarball.WriteToFile()`](https://pkg.go.dev/github.com/google/go-containerregistry@v0.20.7/pkg/v1/tarball#WriteToFile)
2. Import the tar using the [`tarball.ImageFromPath()`](https://pkg.go.dev/github.com/google/go-containerregistry@v0.20.7/pkg/v1/tarball#ImageFromPath)
3. Call the [`Layers()`](https://pkg.go.dev/github.com/google/go-containerregistry@v0.20.7/pkg/v1#Image.Layers) function

Below is a script that will reproduce, if the first CLI argument is a reference to a build cache image

```go
package main

import (
	"fmt"
	"os"
	"path"

)

func main() {
	if len(os.Args) < 2 {
		fmt.Println("please provide the image reference as an arugment")
		os.Exit(1)
	}

	imageRef, err := name.ParseReference(os.Args[1])
	if err != nil {
		fmt.Println("unable to parse name reference", err)
		os.Exit(1)
	}

	remoteImg, err := remote.Image(imageRef, remote.WithAuthFromKeychain(authn.DefaultKeychain))
	if err != nil {
		fmt.Println("unable to get remote image", err)
		os.Exit(1)
	}
	
	tmpTar := path.Join(os.TempDir(), "test.tar")
	defer func() {
		if err = os.Remove(tmpTar); err != nil {
			fmt.Println("error removing tar", err)
		}
	}()
	fmt.Printf("writing tar to %s\n", tmpTar)

	if err = tarball.WriteToFile(tmpTar, imageRef, remoteImg); err != nil {
		fmt.Println("unable to get write image", err)
		os.Exit(1)
	}

	fmt.Println("importing image")
	img, err := tarball.ImageFromPath(tmpTar, nil)

	if err != nil {
		fmt.Println("unable to read exported image", err)
		os.Exit(1)
	}

	fmt.Println("getting layers")
	layers, err := img.Layers() // this will panic
	if err != nil {
		fmt.Println("unable to get layers", err)
		os.Exit(1)
	}

	fmt.Printf("found %d layers\n", len(layers))
}
```

### Expected behavior

`Layers()` should still return as expected, with `[]v1.Layer`

### Additional context

A Build cache image can be built & pushed to registry via the github action `docker/build-push-action` and setting `cache-to: type=registry,ref=$REF` where $REF is the image ref to push the cache to

Yanhu007:
Fixes #2192

## Problem

`(*compressedImage).Manifest()` accesses `cfg.RootFS.DiffIDs[i]` without bounds checking. When loading a tar image whose config is a non-standard type (e.g. `application/vnd.buildkit.cacheconfig.v0`), `DiffIDs` is empty, causing:

```
panic: runtime error: index out of range [0] with length 0
```

## Fix

Add a bounds check before accessing `DiffIDs[i]`. When the index exceeds the DiffIDs length (as with buildkit cache configs), fall through to read the layer directly from the tar archive to compute its digest and size, similar to the existing non-foreign-layer path.

### DIFF
--- a/pkg/v1/tarball/image.go
+++ b/pkg/v1/tarball/image.go
@@ -374,6 +374,25 @@ func (c *compressedImage) Manifest() (*v1.Manifest, error) {
 		if err != nil {
 			return nil, err
 		}
+		if i >= len(cfg.RootFS.DiffIDs) {
+			// Config formats like buildkit cacheconfig don't have DiffIDs.
+			// Fall through to read the layer from the tar directly.
+			l, err := extractFileFromTar(c.opener, p)
+			if err != nil {
+				return nil, err
+			}
+			defer l.Close()
+			sha, size, err := v1.SHA256(l)
+			if err != nil {
+				return nil, err
+			}
+			c.manifest.Layers = append(c.manifest.Layers, v1.Descriptor{
+				MediaType: types.DockerLayer,
+				Size:      size,
+				Digest:    sha,
+			})
+			continue
+		}
 		diffid := cfg.RootFS.DiffIDs[i]
 		if d, ok := c.imgDescriptor.LayerSources[diffid]; ok {
 			// If it's a foreign layer, just append the descriptor so we can avoid


Subserial:
Hey @Yanhu007, you'll need to sign the CLA before we can accept this PR.

PR #2266, opened 2026-04-14
remote: use DeleteScope (with "delete" action) for manifest deletion

### LINKED ISSUE #2184: ggcr: remote.Delete() fails or registries which require additional token scope/action
### Describe the bug

remote.Delete() fails for registries which require additional token scope/action over "pull,push"
specifically IBM Cloud Registry requires either "*" or "delete" action in the token

### To Reproduce
remote.Delete() results in error
```
DELETE https://icr.io/v2/testing-te-1-dev/scratchtest/manifests/sha256:8f9f519f9a82865e4255f057f0f840fa6cf75091346a546d33e7cf8310010721: DENIED: You are not authorized to access the specified resource. See https://cloud.ibm.com/docs/Registry?topic=Registry-troubleshoot-resource; [map[Action:delete Class: Name:testing-te-1-dev/scratchtest Type:repository]]
```

### Expected behavior

Successful delete of the manifest referenced. 

### Additional context
I notice that the v1/remote/transport/scope.go DeleteScope value is unreferenced / unimplemented.

Modifying PushScope in scope.go to "push,pull,*" gets delete working, probably want to actually implement use of a distinct DeleteScope though?

- Version of the module v0.20.7
- Registry used IBM Cloud Container Registry (ICR)

alliasgher:
## Description

Fixes #2184.

`DeleteScope` was defined as an alias for `PushScope` (`"push,pull"`) with a comment "For now DELETE is PUSH". Registries that require an explicit `delete` token action — such as IBM Cloud Container Registry — rejected delete requests because the token only asked for `push,pull`.

### Changes

1. **`transport/scope.go`**: Update `DeleteScope` to `"push,pull,delete"` to include the standard `delete` action.
2. **`write.go`**: Add `makeDeleteClient` which creates an `http.Client` whose transport uses `DeleteScope` instead of `PushScope`.
3. **`pusher.go`**: `Pusher.Delete` now calls `makeDeleteClient` instead of reusing the shared push writer. This ensures the token includes the `delete` action regardless of whether a push client was previously created.

### Compatibility

The extra `delete` action in the scope is harmless for registries that don't require it — they ignore unknown actions or grant the same token. No existing tests were broken.


Subserial:
Can you add tests to show that the delete scope is propagated? I don't see the value in my manual testing.

alliasgher:
Added `TestDeleteRequestsDeleteScope` in 159ae77. It stands up a fake registry + token endpoint, proxies the client via `DialContext`, and asserts both that `transport.DeleteScope` literally contains `"delete"` and that the scope parameter on the token request contains `"delete"`.

Subserial:
Looks good! Once the linter errors are resolved I'll approve and merge this change.

Subserial (approved):

mattmoor:
How many registries was this tested against?

Our registry rejects the new scope, and while we're happy to fix this, I am worried that this is going to break others 🤔

Issue #2238, opened 2026-03-24
crane flatten silently drops symlinks with absolute targets since v0.21.3


ebizboy:
## Description

Since v0.21.3, `crane flatten` silently drops all symlinks that have absolute targets (e.g., `/lib64` → `/usr/lib64`). This makes the resulting flattened image unusable for most Linux-based container images.

## Root Cause

PR #2227 ("mutate: reject path traversal and symlink escape in Extract") introduced logic that skips symlink/hardlink entries with absolute `Linkname`:

```go
if header.Typeflag == tar.TypeSymlink || header.Typeflag == tar.TypeLink {
    linkTarget := filepath.Clean(header.Linkname)
    if strings.HasPrefix(linkTarget, "..") || filepath.IsAbs(linkTarget) {
        continue  // entry is silently dropped
    }
}
```

While rejecting path traversal (`..`) makes sense, rejecting absolute symlink targets is too aggressive. Absolute symlinks like `/lib64` → `/usr/lib64` are standard in virtually all Linux container images (Debian, Ubuntu, Bitnami, etc.).

## Steps to Reproduce

```bash
# Using crane v0.21.3
crane flatten <any-debian-based-image>:tag -t <registry>/test:flattened

# Check the result
docker run --rm --platform linux/amd64 <registry>/test:flattened ls /lib64/ld-linux-x86-64.so.2
# Output: qemu-x86_64: Could not open '/lib64/ld-linux-x86-64.so.2': No such file or directory
```

## Expected Behavior

Symlinks with absolute targets should be preserved during flatten, as they were in v0.21.1 and earlier.

## Actual Behavior

All symlinks with absolute targets are silently dropped from the flattened output, breaking the dynamic linker and rendering the image non-functional.

## Environment

- crane version: v0.21.3
- Works correctly on: v0.21.1
- Tested with: Bitnami Airflow (debian-12 based) multi-arch image (amd64 + arm64)

## Suggested Fix

For symlink targets, normalize absolute paths instead of skipping them — similar to how `header.Name` absolute paths are already handled (stripped of leading `/`):

```go
if header.Typeflag == tar.TypeSymlink || header.Typeflag == tar.TypeLink {
    linkTarget := filepath.Clean(header.Linkname)
    if strings.HasPrefix(linkTarget, "..") {
        continue  // only reject path traversal
    }
    // Absolute symlink targets are valid and should be preserved
}
```

PR #2279, opened 2026-04-25
pkg/v1/mutate: preserve relative symlinks that stay within rootfs in Extract

### LINKED ISSUE #2268: crane: 2238 started dropping symlinks, but it seems agressive for relative ones too
### Describe the bug

#2238 reported a bug with absolute symlinks, but it seems dropping relative symlinks also are aggresive

It makes sense to drop symlinks that escape the rootfs of the container but not within the rootfs.

Common examples, glibc, and many C tools when doing `make install` creates relative symlinks.

```text
Relative symlinks found:
- image: tailscale
  symlinks:
    - rootfs/usr/local/lib/containers/tailscale/var/lock -> ../run/lock
    - rootfs/usr/local/lib/containers/tailscale/var/run -> ../run
- image: glibc
  symlinks:
    - rootfs/usr/local/bin/ld.so -> ../lib/ld-linux-x86-64.so.2
- image: lldpd
  symlinks:
    - rootfs/usr/local/lib/containers/lldpd/var/lock -> ../run/lock
    - rootfs/usr/local/lib/containers/lldpd/var/run -> ../run
- image: nfs-utils
  symlinks:
    - rootfs/usr/local/lib/containers/rpc-statd/var/lock -> ../run/lock
    - rootfs/usr/local/lib/containers/rpc-statd/var/run -> ../run
    - rootfs/usr/local/lib/containers/rpcbind/var/lock -> ../run/lock
    - rootfs/usr/local/lib/containers/rpcbind/var/run -> ../run
- image: nut-client
  symlinks:
    - rootfs/usr/local/lib/containers/nut-client/var/lock -> ../run/lock
    - rootfs/usr/local/lib/containers/nut-client/var/run -> ../run
```

I guess so it makes sense to drop symlinks starting with `../` that escapes the rootfs

anishesg:
The `extract()` function in `pkg/v1/mutate/mutate.go` had all symlink filtering removed in the #2250 revert. Without any filtering, images can contain symlinks with absolute targets (e.g. `/etc/passwd`) or relative targets that escape the rootfs (e.g. `../../../tmp/evil`), both of which are unsafe when layers are extracted to disk.

At the same time, #2268 reports that a prior version of the filtering was over-aggressive: it dropped all relative symlinks that contained `..` anywhere in their target, even safe ones like `../lib/ld-linux-x86-64.so.2` from `usr/local/bin/` — a pattern common in glibc and C toolchains.

This change adds targeted filtering back in `extract()`: absolute `Linkname` values are always dropped, and relative ones are resolved against the symlink's own directory using `filepath.Join`+`filepath.Clean`; only those whose resolved path starts with `..` (i.e. would leave the rootfs) are dropped. Safe relative symlinks that resolve inside the rootfs are preserved unchanged.

A new table-driven test `TestExtractSymlinkFiltering` covers the four cases: safe relative symlink that stays inside rootfs (preserved), another common glibc pattern (preserved), absolute target (dropped), and an escaping relative target (dropped).

Fixes #2268


pkg/v1/mutate/mutate.go
```diff
@@ -297,6 +297,22 @@ func extract(img v1.Image, w io.Writer) error {
...
+			// Reject symlinks and hardlinks whose targets escape the image
+			// rootfs. Absolute targets always point to host paths. Relative
+			// targets are resolved against the symlink's own directory: if the
+			// clean result starts with ".." the link would leave the rootfs.
+			// Relative symlinks that stay within the rootfs (common for glibc,
+			// C toolchains, etc.) are preserved unchanged.
+			if header.Typeflag == tar.TypeSymlink || header.Typeflag == tar.TypeLink {
+				if filepath.IsAbs(header.Linkname) {
+					continue
+				}
```
Subserial: This code is called for both `crane export` and `crane flatten`. #2238 shows us a use-case for when absolute links should be preserved. So, there is still open discussion on whether we should prune absolute links at all. I suggest for now skipping this check on absolute paths,

anishesg: Good point, removed the absolute-path check. Now only relative targets that resolve outside the rootfs are dropped; absolute targets pass through unchanged. Updated the test and comment to match, referencing #2238.

Subserial (approved):

Subserial:
Whoops, I tried manually merging on the UI and messed up the formatting. Could you re-merge for submit? Sorry about that.

PR #2286, opened 2026-05-01
fix(mutate): preserve config blob and layers for non-Docker OCI artifacts

blackwell-systems:
## Summary

`compute()` in `image.go` unconditionally parses the config as a Docker image config and derives layers from `RootFS.DiffIDs`. For non-Docker OCI artifacts (Helm charts, WASM modules, arbitrary blobs), the config has no `RootFS.DiffIDs`, causing:

1. **`Layers()` returns empty.** `remote.Write` uses `Layers()` to determine which blobs to upload. On a cross-repository push, layer blobs are never uploaded, and the registry rejects the manifest with `MANIFEST_BLOB_UNKNOWN`.

2. **The config blob digest is corrupted.** `RawConfigFile()` re-marshals the config through the Docker `ConfigFile` struct, producing a different digest than the original. The manifest references a blob that was never uploaded.

Repro: `crane copy registry-1.docker.io/bitnamicharts/nginx:13.2.14 <dest>` (Helm chart with `application/vnd.cncf.helm.config.v1+json` config).

## Fix

Detect non-image configs by checking the config media type against the two known image config types (`DockerConfigJSON`, `OCIConfigJSON`). When the config is neither:

- **`computeArtifact()`**: new method that preserves the original config blob (no re-marshal, no digest corruption) and builds the layer list from manifest descriptors and added layers.
- **`Layers()`**: falls back to base image layers + added layers instead of deriving from empty `DiffIDs`.
- **`RawConfigFile()`**: returns the original raw bytes from the base image instead of re-marshaling.

Standard Docker/OCI image workflows are unchanged (all existing tests pass).

## Testing

Added `TestNonImageArtifactPreservesConfig` which creates an artifact with `application/vnd.cncf.helm.config.v1+json` config type, mutates annotations, and verifies:
- `Layers()` does not error
- Config media type is preserved in the manifest
- Config blob digest matches the raw config content (no corruption)
- Annotations are present

Full `./pkg/...` test suite passes across all packages.

Fixes #2251


pkg/v1/mutate/mutate_test.go
```diff
@@ -619,6 +619,59 @@ func TestMutateMediaType(t *testing.T) {
...
+	}).(v1.Image)
+
+	// Layers() should not panic or return an error.
+	layers, err := img.Layers()
+	if err != nil {
+		t.Fatalf("Layers() failed on artifact: %v", err)
+	}
+	// Empty image has no layers; that's fine, it shouldn't error.
+	_ = layers
+
```
Subserial: Is this doing anything? It should be removed if it isn't part of the test.

Subserial:
Confirmed with the image in the original bug that this prevents corrupting the image on upload and download. I can approve this change after comments and linter errors are addressed.

pkg/v1/mutate/image.go
```diff
@@ -170,6 +199,69 @@ func (i *image) compute() error {
 	return nil
 }
 
+// computeArtifact handles compute() for non-Docker OCI artifacts whose config
+// is not an image config. It preserves the original config blob and derives
+// layers from the manifest descriptors rather than RootFS.DiffIDs.
+func (i *image) computeArtifact(manifest *v1.Manifest, diffIDMap map[v1.Hash]v1.Layer, digestMap map[v1.Hash]v1.Layer, _ *v1.Manifest) error {
```
Subserial: Why add the manifest as a parameter if it isn't used? Please remove this parameter from the signature if it isn't relevant.

pkg/v1/mutate/image.go
```diff
@@ -170,6 +199,69 @@ func (i *image) compute() error {
...
+	}
+	manifest.Subject = i.subject
+
+	i.manifest = manifest
+	i.diffIDMap = diffIDMap
+	i.digestMap = digestMap
+	i.computed = true
+	return nil
+}
+
```
Subserial, line 193: Compared to the function this section is called from, most of the logic is duplicated here. Can you inline what's different with non-image configs in the original function?

pkg/v1/mutate/image.go
```diff
@@ -189,6 +281,21 @@ func (i *image) Layers() ([]v1.Layer, error) {
...
+			return nil, err
+		}
+		for _, add := range i.adds {
+			if add.Layer != nil {
+				layers = append(layers, add.Layer)
+			}
+		}
+		return layers, nil
+	}
+
```
Subserial: Same comment here. This logic already exists and is duplicated. Please reduce the footprint of this change.

pkg/v1/mutate/image.go
```diff
@@ -52,6 +52,19 @@ func (i *image) MediaType() (types.MediaType, error) {
 	return i.base.MediaType()
 }
 
+// isImageConfig reports whether the given media type represents a container
+// image config that uses Docker-style RootFS.DiffIDs and History fields.
+// Non-image OCI artifacts (Helm charts, WASM modules, etc.) use custom config
+// media types that do not contain these fields.
```
Subserial: Please reduce this comment. The stated purpose is barely relevant to what the function actually does. Save the explanation for when the function is used.

Subserial (changes requested):
Sorry for noting this just now, but this code is very poor quality and duplicates several code paths that already exist. I've left comments, but from now on please review your generated changes to reduce this introduction of code debt.

blackwell-systems:
Thanks for the detailed feedback. I've refactored to inline the artifact logic into `compute()` as suggested. The duplicated paths are eliminated and net change is -70 lines. Branch is up to date with main.

pkg/v1/mutate/image.go
Subserial: To fulfill this, you can remove this block and change the above if statement to:

```
if err := i.compute(); errors.Is(err, stream.ErrNotComputed) || (i.manifest != nil && !isImageConfig(i.manifest.Config.MediaType) {
```

pkg/v1/mutate/image.go
```diff
@@ -70,33 +98,33 @@ func (i *image) compute() error {
...
 
-	m, err := i.base.Manifest()
-	if err != nil {
-		return err
+		configFile.RootFS.DiffIDs = diffIDs
+		configFile.History = history
 	}
-	manifest := m.DeepCopy()
+
+	// Append added layers to the manifest (shared by both paths).
```
Subserial: Please remove this comment.

Subserial (changes requested):
Some of the comments seem to refer to images vs artifacts, which should be updated to images vs non-image artifacts. 

Also, please do not remove arbitrary whitespace and comments from untouched code.

Subserial (approved):
I was being a bit picky on the wording. LGTM, thanks for responding to the feedback on this PR.

EronWright:
Thanks @Subserial and @blackwell-systems for the fix!

PR #2291, opened 2026-05-06
transport: block redirects from token server to private/link-local addresses (SSRF fix)

evilgensec:
## Summary

`refreshOauth` and `refreshBasic` in `bearer.go` construct an `http.Client` without a `CheckRedirect` handler. Go's default redirect policy follows up to 10 hops unconditionally, forwarding the request to any URL returned by the server — including private and link-local addresses.

A malicious or compromised token server (or a registry that controls the `realm` parameter in `WWW-Authenticate`) can therefore bypass the existing `validateRealmURL` check by first responding with a valid-looking realm URL and then issuing an HTTP `302` redirect to an internal service:

```
WWW-Authenticate: Bearer realm="https://legit-token-server.example.com/token",...
→ GET https://legit-token-server.example.com/token   ← passes validateRealmURL
← HTTP/1.1 302 Found
   Location: http://169.254.169.254/latest/meta-data/iam/security-credentials/ROLE
→ GET http://169.254.169.254/...                      ← never checked
← {"Token": "AQo...", ...}                            ← AWS session token
→ stored as bt.bearer.RegistryToken
→ sent as Authorization: Bearer <AWS_SESSION_TOKEN> to attacker's registry
```

This is mechanically distinct from the DNS-based SSRF explicitly noted as out-of-scope in the `validateRealmURL` comment — redirect-based SSRF is fully preventable at the HTTP client layer.

## Fix

Add `realmRedirectCheck`, which applies the same scheme and private-IP restrictions as `validateRealmURL` to every redirect hop. Wire it as `CheckRedirect` on both `http.Client` instances used for token fetches.

## Test

`TestTokenServerRedirectSSRF` spins up a loopback token server that redirects to a second loopback "internal service" and confirms that `bt.refresh()` returns an error rather than following the redirect.

## Relation to prior work

PR #2243 added `validateRealmURL` to block direct private-IP realm values. This PR closes the remaining redirect-bypass path that #2243 left open.

### DIFF
--- a/pkg/v1/remote/transport/bearer.go
+++ b/pkg/v1/remote/transport/bearer.go
@@ -105,6 +105,24 @@ func fromChallenge(reg name.Registry, auth authn.Authenticator, t http.RoundTrip
 	}, nil
 }
 
+// realmRedirectCheck returns an http.Client.CheckRedirect function that applies
+// the same private-address and scheme restrictions as validateRealmURL to every
+// redirect hop during a token-fetch request. Without this guard a malicious
+// token server can issue an HTTP redirect to an internal service (e.g. the
+// cloud instance-metadata endpoint at 169.254.169.254), effectively bypassing
+// the initial validateRealmURL check and causing SSRF.
+func realmRedirectCheck(insecure bool) func(*http.Request, []*http.Request) error {
+	return func(req *http.Request, via []*http.Request) error {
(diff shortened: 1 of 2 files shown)


evilgensec:
@googlebot I signed it.

evilgensec:
Closing to reopen with updated commit.

PR #2293, opened 2026-05-06
remote: validate foreign layer URLs to prevent SSRF (fixes #2259)

evilgensec:
## Summary

OCI and Docker manifests may include a \`urls\` field in layer descriptors
specifying alternative sources for foreign layers.  Without validation, a
malicious registry can set these URLs to private or link-local addresses
(e.g. \`http://169.254.169.254/latest/meta-data/iam/security-credentials/\`),
causing the client to exfiltrate cloud credentials when pulling an image.

### Attack scenario 1 — direct private IP in \`urls\`

\`\`\`json
{
  "mediaType": "application/vnd.docker.image.rootfs.foreign.diff.tar.gzip",
  "digest": "sha256:...",
  "size": 1024,
  "urls": ["http://169.254.169.254/latest/meta-data/iam/security-credentials/my-role"]
}
\`\`\`

The client calls \`Compressed()\`, which iterates \`d.URLs\` and fetches each
URL via \`fetchBlobURL\`.  On AWS/GCP/Azure the metadata service returns IAM
tokens.

### Attack scenario 2 — redirect-based bypass (added in this update)

Initial URL validation checks only the URL literal.  A public CDN under
attacker control can pass validation and then issue a redirect:

1. \`"urls": ["https://cdn.attacker.com/layer.tar.gz"]\` — passes the IP check.
2. \`cdn.attacker.com\` returns \`HTTP 302 → http://169.254.169.254/credentials\`.
3. Go's HTTP client follows the redirect transparently, leaking cloud creds.

### Fix

1. Add \`validateForeignURL\`: rejects non-HTTP(S) schemes and private/loopback
   IP literals.  HTTP is only allowed when the registry itself uses HTTP
   (insecure mode), matching the precedent in \`transport.validateRealmURL\`.
2. Add \`fetchForeignBlobURL\` on \`*fetcher\`: reuses the existing transport but
   sets a \`CheckRedirect\` hook that passes every redirect destination through
   \`validateForeignURL\`, closing the redirect-bypass path.
3. \`Compressed()\` routes foreign layer fetches through \`fetchForeignBlobURL\`
   (not the shared \`fetchBlobURL\`) so both the initial URL and each redirect
   hop are validated.

## Test plan

- \`TestValidateForeignURL\`: unit tests covering loopback, link-local
  (169.254.169.254), RFC-1918, unspecified, disallowed schemes, insecure HTTP.
- \`TestPullingForeignLayerSSRF\`: end-to-end test — manifest with a metadata
  URL is rejected before any request reaches 169.254.169.254.
- \`TestPullingForeignLayerSSRFViaRedirect\` (new): two \`httptest\` servers —
  attacker server redirects to loopback victim; confirms \`Compressed()\` returns
  a "private or link-local" error before any data is returned.

Fixes #2259.


pkg/v1/remote/image.go
```diff
@@ -29,6 +32,74 @@ import (
...
+	if err != nil {
+		return nil, err
+	}
+	if resp.StatusCode != http.StatusOK {
+		resp.Body.Close()
+		return nil, fmt.Errorf("GET %s: unexpected status %s", u.String(), resp.Status)
+	}
+	return verify.ReadCloser(resp.Body, size, h)
+}
+
```
Subserial: Please reduce the verbosity of function comments, limited to what the function does or prevents (SSRF), not necessarily describing how the vulnerability is exploited.

This and associated tests should be moved to pkg/v1/remote/fetcher.go, so that we can keep the method near its receiver definition.

pkg/v1/remote/image.go
```diff
@@ -185,21 +256,43 @@ func (rl *remoteImageLayer) Compressed() (io.ReadCloser, error) {
...
+			err error
+		)
+		if lu.foreign {
+			// Use fetchForeignBlobURL so that any HTTP redirect from the
+			// foreign server is also validated against private-IP ranges,
+			// preventing a redirect-based SSRF bypass.
+			rc, err = rl.ri.fetcher.fetchForeignBlobURL(ctx, lu.u, d.Size, rl.digest, insecure)
+		} else {
+			rc, err = rl.ri.fetcher.fetchBlobURL(ctx, lu.u, d.Size, rl.digest)
+		}
```
Subserial: This comment is verbose because it is intended as a directive to an assistant.

This implementation constructs an array of a bespoke type when wholly unnecessary. Instead, you should:
* Replace the "urls" array with a single value. Instead of iterating on the array, call "fetchBlobURL" on the single URL, maintaining the error in a "lastErr" variable. Maintain the iteration's original comment on this section.
* Collect the foreign layers in an array "foreignURLs" and perform the original iteration on this array, always calling "fetchForeignBlobURL" and collecting the last error in "lastErr".

Subserial (changes requested):
Looks good, but needs some code cleanup before merge.

Subserial (approved):

PR #2295, opened 2026-05-07
remote: block SSRF via private-IP Location headers in blob uploads

adilburaksen:
## Summary

`nextLocation()` in `write.go` extracts the `Location` header from
registry HTTP responses during blob upload and uses it verbatim as the
target for the next request in the upload sequence (POST → PATCH → PUT).

A malicious or compromised registry can return a `Location` header
pointing at a private or link-local IP address (e.g. `169.254.169.254`
for cloud instance metadata, `10.x.x.x/192.168.x.x` for internal
services), causing the client to send PATCH/PUT requests — including
the full container layer blob as the request body — to internal network
endpoints that are not directly reachable by the attacker.

## Attack Scenario

1. Attacker controls or compromises a registry at `registry.example.com`.
2. Victim runs `crane push registry.example.com/image:tag` (or any tool
   using go-containerregistry to push images).
3. Client sends `POST /v2/image/blobs/uploads/` to initiate a blob upload.
4. Attacker's registry responds:
   ```
   HTTP/1.1 202 Accepted
   Location: http://169.254.169.254/latest/meta-data/iam/security-credentials/role
   ```
5. `nextLocation()` returns the IMDS URL without any validation.
6. Client sends `PATCH http://169.254.169.254/...` with the layer blob
   as the body — making a request to the cloud metadata service from
   the client's network context.

## Fix

Add an IP-literal blocklist in `nextLocation()` that rejects cross-host
redirects targeting loopback, link-local, private (RFC 1918), and
unspecified addresses. Same-host redirects (different path on the same
registry) are always allowed so that test servers using `127.0.0.1`
and self-referential registries are unaffected.

This mirrors the `validateRealmURL()` fix from #2243 for the
`WWW-Authenticate` realm SSRF, extended to cover the blob upload
`Location` header code path.

## Test

Added `TestNextLocationSSRFProtection` which verifies that:
- Cross-host redirects to private/link-local IPs are rejected.
- Same-host redirects (including to loopback, for local test registries)
  are allowed.


adilburaksen:
Gentle ping — happy to address any review feedback or adjust the fix if needed.

Subserial (approved):

PR #2296, opened 2026-05-07
fix: limit HTTP response body reads to prevent OOM

evilgensec:
## Problem

Three paths read HTTP response bodies from a registry with no size limit, allowing a malicious or compromised registry to exhaust client memory:

**1. `pkg/v1/remote/referrers.go`**
```go
b, err = io.ReadAll(resp.Body)  // Referrers API — no size cap
```
The manifest fetch path caps at `manifestLimit` (100 MiB), but the Referrers fast path bypasses this entirely.

**2. `pkg/v1/remote/transport/error.go` — `CheckError` and `retryError`**
```go
b, err := io.ReadAll(resp.Body)  // fires on every non-2xx response
```
Triggered before any authentication succeeds (e.g., on the initial 401 challenge), so a malicious registry can exhaust memory without the client ever authenticating.

**3. `pkg/v1/remote/transport/bearer.go` — `refreshOauth` and `refreshBasic`**
```go
return io.ReadAll(resp.Body)  // token endpoint responses
```
A registry whose `WWW-Authenticate` `realm` points to a large-response server (or a compromised token endpoint) can cause OOM during every token refresh.

## Fix

- Referrers: apply the existing `manifestLimit` (100 MiB) constant.
- Error bodies and token responses: add `maxErrorBodySize` (64 KiB) — sufficient for any structured registry error JSON or token response.

All existing tests pass.

### DIFF
--- a/pkg/v1/remote/transport/bearer.go
+++ b/pkg/v1/remote/transport/bearer.go
@@ -397,7 +397,7 @@ func (bt *bearerTransport) refreshOauth(ctx context.Context) ([]byte, error) {
 		return nil, err
 	}
 
-	return io.ReadAll(resp.Body)
+	return io.ReadAll(io.LimitReader(resp.Body, maxErrorBodySize))
 }
 
 // https://docs.docker.com/registry/spec/auth/token/
@@ -441,5 +441,5 @@ func (bt *bearerTransport) refreshBasic(ctx context.Context) ([]byte, error) {
 		return nil, err
 	}
 
-	return io.ReadAll(resp.Body)
+	return io.ReadAll(io.LimitReader(resp.Body, maxErrorBodySize))
 }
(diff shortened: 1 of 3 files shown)


pkg/v1/remote/transport/bearer.go
```diff
@@ -397,7 +397,7 @@ func (bt *bearerTransport) refreshOauth(ctx context.Context) ([]byte, error) {
 		return nil, err
 	}
 
-	return io.ReadAll(resp.Body)
+	return io.ReadAll(io.LimitReader(resp.Body, maxErrorBodySize))
```
Subserial: For readability purposes, this should use a different constant, since this is not inherently handling an error.

evilgensec: Changes implemented as proposed.

Subserial (approved):

PR #2299, opened 2026-05-15
fix: preserve per-occurrence layer identity in mutate.Image.Layers()

### LINKED ISSUE #2034: ggcr:
### Describe the bug

It's possible to mutate an image with a duplicate layer that has the same diff ID as a previous layer (from a parent image), but the underlying blob differs due to a difference in compression (e.g. level).

### To Reproduce

```sh
docker volume create ggcr-test-data
docker run -d -p 1234:5000 --name ggcr-test --volume ggcr-test-data:/var/lib/registry registry:2

mkdir -p tmp/docker tmp/kaniko
echo $'FROM alpine:latest\nWORKDIR /home' >tmp/docker/Dockerfile
echo $'FROM localhost:1234/docker:latest\nRUN echo test' >tmp/kaniko/Dockerfile

docker build -t localhost:1234/docker:latest tmp/docker
docker push localhost:1234/docker:latest

docker run \
    --network=host \
    -v $(pwd)/tmp/kaniko:/workspace \
    gcr.io/kaniko-project/executor:latest \
    --dockerfile /workspace/Dockerfile \
    --destination "localhost:1234/kaniko:latest" \
    --context dir:///workspace/ \
    --force-build-metadata
```

Output (end):

```
INFO[0000] Pushing image to localhost:1234/kaniko:latest
error pushing image: failed to push to destination localhost:1234/kaniko:latest: PUT http://localhost:1234/v2/kaniko/manifests/latest: MANIFEST_BLOB_UNKNOWN: blob unknown to registry; sha256:4f4fb700ef54461cfa02571ae0db9a0dc1e0cdb5577484a6d75e68dc38e8acc1
```

Thus we see that `4f4fb700ef54461cfa02571ae0db9a0dc1e0cdb5577484a6d75e68dc38e8acc1` from the Docker build is missing and replaced with the one from Kaniko (both the empty layer, different compression result). This is a manifest<->layer inconsistency.

### Expected behavior

The mutation does one of the following:

- Looks through image layers for a matching diff ID and replaces the given layer with it
- Returns an error due to matching diff ID but otherwise mismatched digest/blob

Or perhaps something else should happen, I haven't researched the subject in-depth. I do think ggcr shouldn't allow invalid images to be created, though.

### Additional context

This can easily be worked around outside the library by using `LayerByDiffID(...)` before calling `mutate.Append` as done here: https://github.com/coder/kaniko/commit/131d6094d78162cb70441e3fd5a1b524a3d899b9. But I don't think it's a good idea to allow mutate to create invalid images.

- Version of the module: v0.20.2
- Registry used (e.g., GCR, ECR, Quay): registry:2

iahsanGill:
Fixes #2034.

When a base image and an appended layer share a diff ID but have different blob digests — same uncompressed content, different compression — `mutate.Image.Layers()` was returning the appended layer twice instead of `[base, appended]`. The manifest itself still listed both digests correctly, but downstream code that walks `Layers()` to upload blobs would push the appended layer twice and never upload the base's blob, so the final manifest PUT failed with `MANIFEST_BLOB_UNKNOWN`. The reporter hit this through kaniko; another commenter ran into the same thing via `crane append` with RPM-extracted tarballs.

The root cause is in `pkg/v1/mutate/image.go`. `Layers()` walked the rootfs diff IDs and resolved each one via `LayerByDiffID`, which has no way to disambiguate two layers that share a diff ID — the addendum overwrites the base entry in `diffIDMap` during `compute()`, so both diff-ID lookups return the same (appended) layer.

The manifest layer descriptors are already unambiguous: one entry per occurrence, each with its own digest. Walking those and resolving via `LayerByDigest` keeps the same length and order but preserves per-occurrence identity. `LayerByDigest` naturally handles the case because `digestMap` is keyed on (unique) digests, and base-image digests fall through to `i.base.LayerByDigest`.

Added `TestAppendLayers_DuplicateDiffID` covering it — uses a thin wrapper that reports the inner layer's DiffID with its own Digest/Size/Compressed, which is how the bug shows up in practice. Confirmed the test fails on `main` (returns the wrong layer at index 0) and passes with this change. Full `pkg/v1/mutate` suite passes with `-race`.

### DIFF
--- a/pkg/v1/mutate/image.go
+++ b/pkg/v1/mutate/image.go
@@ -189,13 +189,16 @@ func (i *image) Layers() ([]v1.Layer, error) {
 		return nil, err
 	}
 
-	diffIDs, err := partial.DiffIDs(i)
-	if err != nil {
-		return nil, err
-	}
-	ls := make([]v1.Layer, 0, len(diffIDs))
-	for _, h := range diffIDs {
-		l, err := i.LayerByDiffID(h)
+	// Walk manifest layer descriptors by digest rather than rootfs diff
+	// IDs. Two layers can legitimately share a diff ID — same uncompressed
+	// content, different compression — and produce distinct digests. The
+	// manifest preserves the per-occurrence digest; LayerByDiffID does not,
+	// which previously caused duplicate-diff-ID layers to collapse to a
+	// single entry in the returned slice and break blob upload for
+	// downstream pushers (see #2034).
+	ls := make([]v1.Layer, 0, len(i.manifest.Layers))
+	for _, desc := range i.manifest.Layers {
+		l, err := i.LayerByDigest(desc.Digest)
 		if err != nil {
 			return nil, err
 		}
--- a/pkg/v1/mutate/mutate_test.go
+++ b/pkg/v1/mutate/mutate_test.go
@@ -1050,3 +1050,137 @@ func (rc *tokenReleasingReadCloser) Close() error {
 	}
 	return nil
 }
+
+// duplicateDiffIDLayer wraps a v1.Layer so that the wrapper reports the
+// inner layer's DiffID but its own (distinct) Digest, Size, and Compressed
+// stream. This models the real-world case behind #2034: two layers built
+// from the same uncompressed tar but with different compression settings
+// share a diff ID and have distinct blob digests.
+type duplicateDiffIDLayer struct {
+	v1.Layer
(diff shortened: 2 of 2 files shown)


Subserial (commented):
LGTM, I can merge this once the language checks pass.

Subserial (approved):

PR #2302, opened 2026-05-18
transport: allow bearer realm at same host:port as registry

### LINKED ISSUE #2258: question: is bearer realm check over-zealous?
In #2243 a check was introduced on the bearer realm returned by remote registries. This rejects realm URls where the host is a localhost or private-range IP address.

We use go-containerregistry heavily in a project where the test suite uses an authenticated private registry it runs on 127.0.0.1 and #2243 prevents use of this with the error:

```
invalid realm in www-authenticate: realm host "127.0.0.1" is a private or link-local address
```

Would it be reasonable to:

* Be able to disable the realm check via an option?
* Always allow a localhost / 127.0.0.1 realm URL *if* the remote registry we are making requests against is running on 127.0.0.1?

It's not entirely clear to me, either, why #2243 appears to reject URLs with numeric local IP addresses, but not the `localhost` hostname?

CC @evilgensec for any thoughts?

iahsanGill:
## Summary

Fixes #2258.

The realm-URL validation introduced in #2243 rejects realms whose host resolves to a private, loopback, or link-local IP. This is the right default for the cross-host SSRF case (a malicious registry pointing the token endpoint at `169.254.169.254` or at a sister internal service), but it breaks legitimate setups where a user explicitly targets a private/loopback registry that returns a same-host token endpoint — e.g. test suites running an authenticated registry on `127.0.0.1`, or on-prem deployments that colocate registry and token server.

This PR adds the same-host:port exception agreed in the issue thread (cc @evilgensec, @Subserial): if the realm URL host:port equals the registry host:port the user is already talking to, the private/link-local check is skipped. The realm is by definition not crossing a trust boundary the user did not already trust by passing the reference.

The match is strict — host **and** port — so that the "different port on the same loopback host" case (e.g. a realm pointed at a local database or metadata-service port that happens to live alongside the registry) remains blocked.

This mirrors the same-host exception pattern already used by the blob upload Location-header check in [`pkg/v1/remote/write.go`](pkg/v1/remote/write.go).

## Changes

| Where | Change |
|---|---|
| `validateRealmURL` | Add `registryHost` parameter; return `nil` early when `u.Host == registryHost`. Scheme check still applies. |
| `realmRedirectCheck` | Threads `registryHost` through to `validateRealmURL` so redirects honor the same exception. |
| `fromChallenge` callsite | Passes `reg.RegistryStr()`. |
| `refreshOauth` / `refreshBasic` callsites | Pass `bt.registry.RegistryStr()`. |
| `TestValidateRealmURLUnspecified` | Updated for new signature (empty `registryHost` preserves prior behavior). |
| `TestValidateRealmURLSameHost` (new) | Covers: same host:port allowed (loopback, private, IPv6 loopback); same host **different** port still blocked; cross-host loopback/metadata still blocked; scheme check still applies. |

`pkg/v1/remote/...` tests all pass locally; `go vet ./pkg/v1/remote/transport/...` clean.

## Out of scope

- Adding `localhost` (DNS) to the private blocklist, as suggested by @evilgensec in the issue. That would be a separate change to the threat-model surface and worth its own PR.
- Honoring `Retry-After` on transport-level retries (separate follow-up to #2301).

## Test plan

- [x] `TestValidateRealmURLSameHost` — table-driven, 7 cases
- [x] `TestValidateRealmURLUnspecified` — still passes with new signature
- [x] `TestTokenServerRedirectSSRF` — still rejects cross-host redirects after the closure signature change
- [x] Full `go test ./pkg/v1/remote/...` green

### DIFF
--- a/pkg/v1/remote/transport/bearer.go
+++ b/pkg/v1/remote/transport/bearer.go
@@ -90,7 +90,7 @@ func fromChallenge(reg name.Registry, auth authn.Authenticator, t http.RoundTrip
 	// registry can supply a realm pointing at an internal service or cloud
 	// metadata endpoint (e.g. 169.254.169.254), causing SSRF when the client
 	// subsequently fetches a token.
-	if err := validateRealmURL(realm, pr.Insecure); err != nil {
+	if err := validateRealmURL(realm, reg.RegistryStr(), pr.Insecure); err != nil {
 		return nil, fmt.Errorf("invalid realm in www-authenticate: %w", err)
 	}
 	service := pr.Parameters["service"]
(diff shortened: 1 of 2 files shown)


Subserial (changes requested):
LGTM, but can you reduce the long comments? Just mentioning "always allow realm URLs matching the registry URL" is fine.

iahsanGill:
Sure, I will push the changs with short commetns...

Subserial (approved):

PR #2303, opened 2026-05-18
mutate: verify layer digests in Extract and Time

momenashrafff:
## What

`extractLayer` and `layerTime` in `pkg/v1/mutate/mutate.go` read a layer through an `archive/tar` reader. `archive/tar` returns `io.EOF` from `Next()` as soon as it reads the two-block end-of-archive marker, so the loop stops before the underlying layer stream reaches its own `io.EOF`.

`internal/verify`'s reader checks the layer's SHA-256 against the expected digest only when the wrapped stream returns `io.EOF` (`verifyReader.Read`). Because the tar loop stops early that point is never reached, and `Close()` does not verify either. As a result `mutate.Extract` and `mutate.Time`, and `crane export` / `crane edit` which build on `mutate.Extract`, accept a layer whose contents do not match the layer digest listed in the manifest.

`crane pull` and `pkg/v1/validate` are not affected: they read each layer to `io.EOF`, so the digest check fires.

## Fix

`pkg/v1/validate/layer.go` already drains the stream after its `tar.Reader` loop for this reason:

```go
// Discard any trailing padding that the tar.Reader doesn't consume.
if _, err := io.Copy(io.Discard, hashUncompressed); err != nil {
```

This change adds the same drain after the `tar.Reader` loop in `extractLayer` and `layerTime`, so the underlying verifying reader reaches `io.EOF` and the layer digest is checked.

## Testing

- A layer whose bytes do not match the manifest's layer digest: `crane export` built from this branch fails with `error verifying sha256 checksum` instead of succeeding. Built from current `main` it succeeds.
- Honest images extract and pull unchanged.
- `go test ./pkg/v1/mutate/...` passes.

>🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/pkg/v1/mutate/mutate.go
+++ b/pkg/v1/mutate/mutate.go
@@ -365,6 +365,15 @@ func extractLayer(tarWriter *tar.Writer, fileMap map[string]bool, layer v1.Layer
 			}
 		}
 	}
+
+	// Drain any bytes the tar.Reader did not consume (trailing data after the
+	// end-of-archive marker) so the underlying verifying reader reaches io.EOF
+	// and the layer's digest is verified. Without this, a layer whose contents
+	// do not match the manifest's layer digest is extracted without error.
+	// pkg/v1/validate/layer.go performs the same drain.
+	if _, err := io.Copy(io.Discard, layerReader); err != nil {
+		return fmt.Errorf("verifying layer: %w", err)
+	}
 	return nil
 }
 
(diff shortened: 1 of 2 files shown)


Subserial:
Please provide tests that model the scenario you are describing.

momenashrafff:
Hello @Subserial, thank you for your comment

I just added tests in pkg/v1/mutate/mutate_test.go (5f9072a).

TestExtractVerifiesLayerDigest and TestTimeVerifiesLayerDigest each build a layer whose Uncompressed() stream is wrapped by internal/verify against a digest that doesn't match the bytes it serves, which is the case described in the PR. Both assert the operation fails with a checksum error, and both have a matching-digest case to cover the normal path.

I reverted the fix locally to check the tests actually catch it: the mismatch cases pass through silently without the drain and fail with it. go test ./pkg/v1/mutate/... is green.

Subserial (approved):

PR #2306, opened 2026-05-19
Reject symlinks in OCI layout blobs

mosskappa:
## Summary

This change makes OCI layout blob reads reject symlinks and other non-regular files before returning blob contents.

`layout.Path.Blob` and `layout.Path.Bytes` now share the same open path, which:

- uses `Lstat` to reject symlinks before opening the blob path
- rejects non-regular files
- verifies the opened file still matches the original filesystem entry

This prevents local OCI layouts from resolving `blobs/<algorithm>/<hex>` entries through symlinks when used by layout readers and commands such as `crane push` or `crane index append`.

## Tests

Run in Docker on Linux so symlink behavior is exercised:

```bash
docker run --rm -v ${PWD}:/src -w /src golang:1.25 go test ./pkg/v1/layout ./cmd/crane/cmd -run 'TestLayoutBlobSymlink|TestLayoutAppendImageSymlink|TestCranePushLayoutSymlinkBlobNotUploaded' -count=1 -v
```

All targeted tests passed.


pkg/v1/layout/blob.go
```diff
@@ -23,15 +24,52 @@ import (
...
+	}
+	defer f.Close()
+	return io.ReadAll(f)
 }
 
 func (l Path) blobPath(h v1.Hash) string {
 	return l.path("blobs", h.Algorithm, h.Hex)
 }
+
+func (l Path) openBlob(h v1.Hash) (*os.File, error) {
```
Subserial, line 45: Can you replace the multiple other tests with just a unit test for this function, in a file named blob_test.go? The other tests for this short function seem far removed from this file.

Subserial (changes requested):
I'm not quite sure where "PR075" comes from, but all the tests named after it are far removed from the original change. LGTM on the change but the tests need to be collapsed.

mosskappa:
Thanks for the review. I collapsed this down to a focused unit test for openBlob in pkg/v1/layout/blob_test.go and removed the broader follow-on tests from this PR.

pkg/v1/layout/blob.go
```diff
@@ -23,15 +24,52 @@ import (
...
+	if err != nil {
+		return nil, err
+	}
+	if info.Mode()&os.ModeSymlink != 0 {
+		return nil, fmt.Errorf("layout blob %s is a symlink", h)
+	}
+	if !info.Mode().IsRegular() {
+		return nil, fmt.Errorf("layout blob %s is not a regular file", h)
+	}
+	f, err := os.Open(p)
```
Subserial, line 57: Instead of calling f.Close() on all returns, can you add `defer f.Close()` after this line?

Subserial:
LGTM but one small change before approval.

Subserial (approved):

loosebazooka:
This seems to have a negative effect on build systems like bazel, which use extensive symlinking. Any chance this failure can be made configurable?

mosskappa:
Thanks for flagging this. I think making this configurable as a follow-up could make sense for trusted local layouts / build-system caches, as long as the secure default remains rejecting symlink blobs.

The important boundary from this PR is that the default OCI layout reader should not follow blob-path symlinks, since that can make a layout reference bytes outside the layout root. An opt-in escape hatch with clear naming/documentation, e.g. for callers that fully trust the local layout producer and need Bazel-style symlink materialization, seems like the safer compatibility path than relaxing the default behavior.

lbcjbb:
The use of symbolic links also allows layers to be shared between different images, which drastically reduces the disk space/IO used for several thousand images.

anguslees:
> The important boundary from this PR is that the default OCI layout reader should not follow blob-path symlinks, since that can make a layout reference bytes outside the layout root.

.. It can only read bytes outside the layout root, if someone put a symlink _in_ the layout root - right?

I think you're gesturing towards a (security?) problem here that justified this change - but I'm not clear what the problem is/was.   Surely - following widespread Unix practice - we should just assume that whoever constructed an oci layout using symlinks knew what they were doing and we should just follow the symlinks?

Issue #2276, opened 2026-04-22
crane: hang on crane export


adrian-thurston:
### Describe the bug

Seeing a hang on a `crane export`. Suspecting commit 3502eacc.

Reverting to v0.21.5 clears the hang.

Sorry I cannot supply a reproducer as the image that crane hangs on is private.

gnix0:
> ### Describe the bug
> Seeing a hang on a `crane export`. Commit [3502eac](https://github.com/google/go-containerregistry/commit/3502eacc53e454817af87fa032284b845137d330). Suspecting:
> 
> [#2271](https://github.com/google/go-containerregistry/pull/2271)
> 
> Reverting to v0.21.5 clears the hang.
> 
> I cannot supply a reproducer as the image that crane hangs on is private.

Hey Adrian! Just saw this, thanks for pointing it out. Yeah, I think I see what’s happening.

`crane export` goes through `mutate.Extract`, and that opens layer readers in a loop. Those readers currently get closed with `defer`, so they stay open until the whole extraction finishes. After #2271 remote layer reads hold a job slot until the body is closed, so export can fill the available slots and then block when it tries to open the next layer.

I'll take a deeper look at it right now and send a fix asap with a regression test. Thanks again!

Mehmood-klariq:
We are also affected by this issue. In our case, the problem occurs with crane pull, which hangs while downloading/exporting the image. Downgrading from crane 0.21.6 to 0.21.5 resolved the issue immediately.

PR #2308, opened 2026-05-19
tarball: close layer readers during Write

### LINKED ISSUE #2309: crane pull hangs on multi-layer images (v0.21.6 regression, symmetric to #2276 — pkg/v1/tarball)
## Summary

In v0.21.6, `crane pull IMAGE TARBALL` hangs indefinitely on images with several layers. This is the same `pullLimiter` slot-leak deadlock that was fixed for `crane export` in #2277 (closes #2276) — but the symmetric leak in `pkg/v1/tarball/write.go` was not patched, so `crane pull` (default tarball format) still deadlocks.

Fix in #2308.

## Reproduction

```bash
# v0.21.5 — works
$ curl -sL https://github.com/google/go-containerregistry/releases/download/v0.21.5/go-containerregistry_Linux_x86_64.tar.gz | tar -xz crane
$ ./crane version
0.21.5
$ ./crane pull postgres:16 /tmp/pg.tar    # finishes in ~30s, ~160 MB

# v0.21.6 — hangs
$ curl -sL https://github.com/google/go-containerregistry/releases/download/v0.21.6/go-containerregistry_Linux_x86_64.tar.gz | tar -xz crane
$ ./crane version
0.21.6
$ timeout 120 ./crane pull postgres:16 /tmp/pg.tar
$ echo $?    # 124 (killed); tarball stalls around 37 MB
```

`postgres:16` has 14 layers. The hang also reproduces on other public many-layer images and was first hit in CI promoting a 15-layer image (OCI image index with attestation manifest) between two Artifact Registry projects — every pull-and-push run since v0.21.6 was released (2026-05-18) timed out at the runner limit.

## Root cause

#2271 (v0.21.6) introduced a `pullLimiter` in `pkg/v1/remote`. `remoteImageLayer.Compressed()` now returns a `limitedReadCloser` that holds a limiter token until `Close()` is called (`pkg/v1/remote/limiter.go`, `pkg/v1/remote/fetcher.go:295-341`).

In `pkg/v1/tarball/write.go:173-206` the layer-write loop is:

```go
for i, l := range layers {
    ...
    r, err := l.Compressed()                          // acquires a slot
    if err != nil {
        return ...
    }
    blobSize, err := l.Size()
    ...
    if err := writeTarEntry(tf, layerFiles[i], r, blobSize); err != nil {
        return ...
    }
    // r is never Closed — slot is never released
}
```

`writeTarEntry` takes `r io.Reader` and just does `io.Copy(tf, r)` (`pkg/v1/tarball/write.go:369-381`). Nothing closes `r`. With `defaultJobs = 4` (`pkg/v1/remote/options.go:106`), the loop fills all 4 slots in the first 4 iterations and the 5th `l.Compressed()` call blocks forever in `pullLimiter.acquire`.

#2277 fixed the analogous leak in `crane export` (`mutate.Extract`) and explicitly called out this deadlock pattern, but did not patch this call site.

## Workaround

Pin to `crane v0.21.5` until the fix lands.

## Environment

- crane v0.21.6 (released 2026-05-18)
- Reproduced on macOS arm64 and `ubuntu-latest` GitHub-hosted runners

nandbhat:
## Summary

`tarball.MultiRefWrite` (and `crane pull`, which uses it) opens each layer's `Compressed()` reader and `io.Copy`s it into the tar without closing the reader. After #2271, remote layer readers hold a pull-job slot until their response body is closed, so writing more than `defaultJobs` (4) layers fills all available slots and the next `l.Compressed()` blocks forever in `pullLimiter.acquire`.

This is the symmetric leak to the one #2277 fixed in `mutate.Extract` / `crane export`.

Fixes #2309.

## Reproduction (before this PR)

```bash
$ curl -sL https://github.com/google/go-containerregistry/releases/download/v0.21.6/go-containerregistry_$(uname -s)_$(uname -m).tar.gz | tar -xz crane
$ timeout 120 ./crane pull postgres:16 /tmp/pg.tar
$ echo $?    # 124, killed — tarball stalls at ~37 MB
```

`postgres:16` has 14 layers. v0.21.5 finishes the same pull in ~30s with a 160 MB tar.

## Fix

Extract the per-layer write into a `writeLayer` helper that uses `defer r.Close()` so the pull-limiter slot is released as soon as the layer is in the tar. Same shape as #2277's `extractLayer`.

## Test

Added `TestWriteClosesLayerBeforeOpeningNext`. It wraps a `random.Image` so each layer's `Compressed()` reader takes one of 4 slots (modeling `remote.WithJobs`), and writes an image with `cap(tokens)+1` layers via `tarball.MultiRefWrite`. Before this change the test fails with `layer reader opened before previous readers were closed`. After this change the test passes and the slot channel is fully drained.

## Tested

```
go test ./pkg/v1/tarball
go test ./pkg/v1/tarball ./pkg/v1/mutate ./pkg/crane
```

All pass.

## Related

- Same root cause as #2276 / #2277, but in the tarball write path rather than mutate extract.

### DIFF
--- a/pkg/v1/tarball/write.go
+++ b/pkg/v1/tarball/write.go
@@ -191,16 +191,7 @@ func writeImagesToTar(imageToTags map[v1.Image][]string, m []byte, size int64, w
 			}
 			seenLayerDigests[hex] = struct{}{}
 
-			r, err := l.Compressed()
-			if err != nil {
-				return sendProgressWriterReturn(pw, err)
-			}
-			blobSize, err := l.Size()
-			if err != nil {
-				return sendProgressWriterReturn(pw, err)
-			}
-
-			if err := writeTarEntry(tf, layerFiles[i], r, blobSize); err != nil {
+			if err := writeLayer(tf, layerFiles[i], l); err != nil {
 				return sendProgressWriterReturn(pw, err)
 			}
 		}
@@ -366,6 +357,24 @@ func dedupRefToImage(refToImage map[name.Reference]v1.Image) map[v1.Image][]stri
 }
 
 // writeTarEntry writes a file to the provided writer with a corresponding tar header
+// writeLayer streams a layer's compressed blob into the tar writer at name,
+// ensuring the layer reader is closed as soon as it has been written. Closing
+// the reader releases any pull-limiter slot held by remote-backed layers
+// (see remote.WithJobs / pkg/v1/remote.limiter), so leaving readers open here
+// could deadlock the write loop after defaultJobs layers.
+func writeLayer(tf *tar.Writer, name string, l v1.Layer) error {
+	r, err := l.Compressed()
+	if err != nil {
+		return err
+	}
+	defer r.Close()
+	blobSize, err := l.Size()
+	if err != nil {
+		return err
+	}
+	return writeTarEntry(tf, name, r, blobSize)
+}
+
 func writeTarEntry(tf *tar.Writer, path string, r io.Reader, size int64) error {
 	hdr := &tar.Header{
 		Mode:     0644,
--- a/pkg/v1/tarball/write.go
+++ b/pkg/v1/tarball/write.go
@@ -357,24 +357,6 @@ func dedupRefToImage(refToImage map[name.Reference]v1.Image) map[v1.Image][]stri
 }
 
 // writeTarEntry writes a file to the provided writer with a corresponding tar header
-// writeLayer streams a layer's compressed blob into the tar writer at name,
-// ensuring the layer reader is closed as soon as it has been written. Closing
-// the reader releases any pull-limiter slot held by remote-backed layers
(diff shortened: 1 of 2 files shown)


nandbhat:
@googlebot I signed it!

pkg/v1/tarball/write.go
```diff
@@ -366,6 +357,24 @@ func dedupRefToImage(refToImage map[name.Reference]v1.Image) map[v1.Image][]stri
...
+		return err
+	}
+	defer r.Close()
+	blobSize, err := l.Size()
+	if err != nil {
+		return err
+	}
+	return writeTarEntry(tf, name, r, blobSize)
+}
+
```
Subserial: You seem to have bisected the other function from its leading comment. Could you put it back?

nandbhat: Moved writeLayer after writeTarEntry so its leading comment is restored. Also rebased onto latest main.

Subserial (approved):

PR #2310, opened 2026-05-20
crane: preserve Cmd when --entrypoint is used without --cmd

### LINKED ISSUE #2041: `crane mutate` deletes Cmd if you try to override the `--entrypoint` without `--cmd`
### Describe the bug

Given a base image with Entrypoint + Cmd;

`crane mutate --entrypoint ...` without also specifying `--cmd` deletes the `Cmd`. You end up with an Entrypoint only.

### To Reproduce

As above

### Expected behavior

Expected it to preserve the original Cmd, and override the Entrypoint only.

Deleting Cmd is unexpected, Dockerfile syntax `ENTRYPOINT [...]` does not do that. Ultimately to use `crane mutate`, you have to look up the base image, get .Config.Cmd, and re-add it as --cmd if you are using --entrypoint alone. This is possible! But not expected behaviour

### Additional context

- Output of `crane version`: 0.19.1, but there are no relevant commits since then.

iahsanGill:
Fixes #2041

## Summary
- `crane mutate --entrypoint X` previously cleared `Cmd` even when `--cmd` wasn't passed, silently dropping the base image's command. Workaround was to read the base config and re-pass `--cmd` manually.
- Switch from `len(...) > 0` to `Flags().Changed(...)` so `--entrypoint` and `--cmd` are independently respected. Matches Dockerfile ENTRYPOINT/CMD semantics.

## Test plan
- [x] New `TestMutateEntrypointCmdPreservation` covers three cases: `--entrypoint` only preserves base Cmd, `--cmd` only preserves base Entrypoint, both flags override both.
- [x] Verified the bug-reproduction subtest fails on the unfixed code (`Cmd = [], want [old-cmd-arg]`).
- [x] `go test ./cmd/crane/... ./pkg/crane/...` and `go vet ./...` clean.

### DIFF
--- a/cmd/crane/cmd/mutate.go
+++ b/cmd/crane/cmd/mutate.go
@@ -101,14 +101,13 @@ func NewCmdMutate(options *[]crane.Option) *cobra.Command {
 				return err
 			}
 
-			// Set entrypoint.
-			if len(entrypoint) > 0 {
+			// Set entrypoint and/or cmd only when the user passed those flags,
+			// so overriding one doesn't silently wipe the other inherited from
+			// the base image. Matches Dockerfile ENTRYPOINT/CMD semantics.
+			if c.Flags().Changed("entrypoint") {
 				cfg.Config.Entrypoint = entrypoint
-				cfg.Config.Cmd = nil // This matches Docker's behavior.
 			}
-
-			// Set cmd.
-			if len(cmd) > 0 {
+			if c.Flags().Changed("cmd") {
 				cfg.Config.Cmd = cmd
 			}
 
--- a/cmd/crane/cmd/mutate_test.go
+++ b/cmd/crane/cmd/mutate_test.go
@@ -1,0 +1,130 @@
+// Copyright 2026 Google LLC All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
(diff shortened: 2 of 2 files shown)


Subserial (approved):

iahsanGill:
Heads up — CI surfaced a real design question I'd appreciate guidance on before fixing the failures.

**E2e failure root cause:** `.github/workflows/e2e.yaml` does `crane mutate --entrypoint=/app/crane,version` on top of `alpine` and then `docker run $img`. With the old behavior (Cmd wiped on `--entrypoint`), that runs `crane version`. With this PR's behavior, alpine's inherited `Cmd=["/bin/sh"]` survives and docker runs `crane version /bin/sh` → `unknown command "/bin/sh"`. So the e2e workflow itself depended on the behavior #2041 calls a bug.

**Design tension:** the existing `cfg.Config.Cmd = nil // This matches Docker's behavior.` matches `docker run --entrypoint X image` semantics (which clears CMD at runtime) and also matches BuildKit's Dockerfile ENTRYPOINT-resets-CMD rule per the [current Dockerfile reference](https://docs.docker.com/reference/dockerfile/#understand-how-cmd-and-entrypoint-interact). So the reporter's "Dockerfile doesn't do that" premise may not hold.

Three options I see, happy to take whichever you prefer:

1. **Close as won't-fix** — current behavior is intentional and matches Docker semantics; document that more clearly and close #2041.
2. **Make it opt-in** — add a `--preserve-cmd` flag, keep old behavior as default, e2e workflow unchanged.
3. **Proceed as-is** — accept the behavior change and update the e2e workflow to pass `--cmd=`.

Will hold off on further changes until I hear back.

PR #2317, opened 2026-05-23
pkg/name: avoid substring-based loopback HTTP fallback

### LINKED ISSUE #2313: pkg/name: unanchored loopback regex causes TLS downgrade for hosts containing "127.0.0.1" or "::1" substring
### Summary
     
  `pkg/name/registry.go` uses unanchored regular expressions to detect
  loopback addresses for the purpose of choosing HTTP vs HTTPS scheme.
  Because `regexp.QuoteMeta` only escapes metacharacters and does **not**
  add `^`/`$` anchors, the regex matches the literal substring
  `127.0.0.1` (or `::1`) anywhere in the hostname. Any registry hostname
  that embeds either substring is misclassified as loopback and
  contacted over plaintext HTTP instead of HTTPS.
  
  This was reported to the Google OSS-VRP and acknowledged as a valid
  security concern (TLS downgrade / potential data exposure) but
  classified as a product vulnerability under OT2/OT3 — Google
  suggested I file it here directly. Hence this issue.
  
  ### Affected code
  
  `pkg/name/registry.go` lines 30-34:
  
  ```go
  // Detect the loopback IP (127.0.0.1)
  var reLoopback     = regexp.MustCompile(regexp.QuoteMeta("127.0.0.1"))
  
  // Detect the loopback IPV6 (::1)
  var reipv6Loopback = regexp.MustCompile(regexp.QuoteMeta("::1"))
  ```
  
  Compare the sibling `reLocal` on the same file, which IS properly
  anchored — so the existing convention in this file is to anchor:
  
  ```go
  var reLocal = regexp.MustCompile(`.*\.localhost(?::\d{1,5})?$`)
  ```
  
  ### Reproducer
  
  ```go
  package main
  
  import (
        "fmt"
  
  )
  
  func main() {
        for _, h := range []string{
                "127.0.0.1",                 // intended loopback, should be http
                "127.0.0.1:5000",            // intended loopback, should be http
                "[::1]:5000",                // intended loopback, should be http
                "127.0.0.1.evil.com",        // NOT loopback, should be https
                "127.0.0.1.evil.com:5000",   // NOT loopback, should be https
                "my127.0.0.1registry.io",    // NOT loopback, should be https
                "evil.com.127.0.0.1.nip.io", // NOT loopback, should be https
                "x127.0.0.1x.com",           // NOT loopback, should be https
                "[2001:db8::1]:5000",        // global IPv6, should be https
        } {
                r, err := name.NewRegistry(h)
                if err != nil {
                        fmt.Printf("%q rejected: %v\n", h, err)
                        continue
                }
                fmt.Printf("%q -> %s\n", h, r.Scheme())
        }
  } 
  ```
  
  Output against `v0.21.6`:
  
  ```
  "127.0.0.1" -> http
  "127.0.0.1:5000" -> http
  "[::1]:5000" -> http
  "127.0.0.1.evil.com" -> http        <-- BUG (should be https)
  "127.0.0.1.evil.com:5000" -> http   <-- BUG
  "my127.0.0.1registry.io" -> http    <-- BUG
  "evil.com.127.0.0.1.nip.io" -> http <-- BUG
  "x127.0.0.1x.com" -> http           <-- BUG
  "[2001:db8::1]:5000" -> http        <-- BUG
  ```
  
  ### Versions confirmed vulnerable
  
  - v0.21.6 (latest released at time of report)
  - master @ commit `d807d512681c6b18391a8fa0ae0fb27c36ba5f4d` (2026-05-19)
  
  ### Impact
  
  Any consumer of `pkg/name` that uses `Registry.Scheme()` to construct
  outbound HTTP requests — which includes `crane`, `cosign`, `kpack`,
  `ko`, `skaffold`, and any tool that takes user-supplied image
  references. An attacker controlling a hostname that embeds the
  substring (trivial via wildcard DNS or owning any domain such as
  `127.0.0.1.attacker.tld`) can force the client to use plaintext HTTP,
  exposing bearer tokens / basic-auth credentials on the wire and
  enabling on-path tampering of manifest/layer responses.
  
  For IPv6 the issue is broader: any address whose zero-compression
  places `::1` anywhere in the textual form (`2001:db8::1`, `fe80::1`,
  `fec0::1`, etc.) is misclassified.
  
  ### Suggested fix
  
  Either anchor the regexes:
  
  ```go
  var reLoopback     = regexp.MustCompile(`^127\.0\.0\.1(?::\d{1,5})?$`)
  var reipv6Loopback = regexp.MustCompile(`^\[::1\](?::\d{1,5})?$`)
  ```
  
  Or, preferably, drop the regex and use structured parsing — this
  also fixes a related weakness in `isRFC1918()` which uses
  `strings.Split(r.Name(), ":")[0]` for port stripping and does not
  handle bracketed IPv6 hosts:
  
  ```go
  func (r Registry) isLoopback() bool {
      h := r.Name()
      // Strip port and IPv6 brackets, then parse as IP.
      if i := strings.LastIndex(h, ":"); i >= 0 && !strings.HasSuffix(h[:i], "]") {
          h = h[:i] 
      }
      h = strings.TrimSuffix(strings.TrimPrefix(h, "["), "]")
      if ip := net.ParseIP(h); ip != nil {
          return ip.IsLoopback()
      }
      return false
  }
  ```

Nick-Bug-Bounty-Hunter:
Fixes #2313

## Summary
- replace substring-based loopback regex checks with structured host parsing
- reuse the parsed host for RFC1918, localhost, and loopback detection
- add regressions for spoofed 127.0.0.1 hostnames and bracketed IPv6 cases

## Testing
- go test ./pkg/name -count=1
- go test ./pkg/... -count=1

### DIFF
--- a/pkg/name/registry.go
+++ b/pkg/name/registry.go
@@ -27,12 +27,6 @@ import (
 // Detect more complex forms of localhost references.
 var reLocal = regexp.MustCompile(`.*\.localhost(?::\d{1,5})?$`)
 
-// Detect the loopback IP (127.0.0.1)
-var reLoopback = regexp.MustCompile(regexp.QuoteMeta("127.0.0.1"))
-
-// Detect the loopback IPV6 (::1)
-var reipv6Loopback = regexp.MustCompile(regexp.QuoteMeta("::1"))
-
 // Registry stores a docker registry name in a structured form.
 type Registry struct {
 	insecure bool
@@ -69,9 +63,15 @@ func (r Registry) Scope(string) string {
 	return "registry:catalog:*"
 }
 
+func (r Registry) host() string {
+	if host, _, err := net.SplitHostPort(r.Name()); err == nil {
+		return host
+	}
+	return strings.TrimSuffix(strings.TrimPrefix(r.Name(), "["), "]")
+}
+
 func (r Registry) isRFC1918() bool {
-	ipStr := strings.Split(r.Name(), ":")[0]
-	ip := net.ParseIP(ipStr)
+	ip := net.ParseIP(r.host())
 	if ip == nil {
 		return false
 	}
@@ -84,6 +84,11 @@ func (r Registry) isRFC1918() bool {
 	return false
 }
 
+func (r Registry) isLoopback() bool {
+	ip := net.ParseIP(r.host())
+	return ip != nil && ip.IsLoopback()
+}
+
 // Scheme returns https scheme for all the endpoints except localhost or when explicitly defined.
 func (r Registry) Scheme() string {
 	if r.insecure {
@@ -92,16 +97,13 @@ func (r Registry) Scheme() string {
 	if r.isRFC1918() {
 		return "http"
 	}
-	if strings.HasPrefix(r.Name(), "localhost:") {
+	if r.host() == "localhost" {
 		return "http"
 	}
 	if reLocal.MatchString(r.Name()) {
 		return "http"
 	}
-	if reLoopback.MatchString(r.Name()) {
-		return "http"
-	}
-	if reipv6Loopback.MatchString(r.Name()) {
+	if r.isLoopback() {
 		return "http"
 	}
 	return "https"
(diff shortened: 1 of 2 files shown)


Subserial:
Closing as fixed in #2314

PR #2318, opened 2026-05-24
fix(crane): avoid creating export tar on pull failure

### LINKED ISSUE #2066: crane:
### Describe the bug

A tar file is created even though the image failed to be pulled.

### To Reproduce

```sh
# fails
crane export does-not-exist x.tar

# succeeds
file x.tar
```

### Expected behavior

When `crane export some-image some.tar` fails, then it should not create a `some.tar`.

### Additional context

```
❯ crane version
0.20.3
```

Haihan-Jiang:
Fixes #2066

## What changed
- Delay opening the destination tarball in `crane export` until after the source image has been loaded successfully.
- Add a command-level regression test using a local 404 registry to verify failed pulls do not create the destination tar file.

## Why
`crane export IMAGE out.tar` previously created/truncated `out.tar` before pulling `IMAGE`. If the pull failed, an empty or partial destination file could still be left behind.

## Validation
- `go test ./cmd/crane/cmd`
- `go test ./cmd/crane/...`
- `git diff --check`

### DIFF
--- a/cmd/crane/cmd/export.go
+++ b/cmd/crane/cmd/export.go
@@ -46,12 +46,6 @@ func NewCmdExport(options *[]crane.Option) *cobra.Command {
 				dst = args[1]
 			}
 
-			f, err := openFile(dst)
-			if err != nil {
-				return fmt.Errorf("failed to open %s: %w", dst, err)
-			}
-			defer f.Close()
-
 			var img v1.Image
 			if src == "-" {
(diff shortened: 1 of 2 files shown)


Haihan-Jiang:
This is ready for review from my side. Could someone take a look when convenient?

Subserial (approved):

Issue #2120, opened 2025-07-28
crane: registry needs to hold full layers in memory


malt3:
### Describe the bug

When using `crane registry serve` and pushing large layers with `PUT`, I observe excessive memory usage.
I traced this to the `PUT` handler function:
https://github.com/google/go-containerregistry/blob/59a4b85930392a30c39462519adc8a2026d47181/pkg/registry/blobs.go#L485

The in-flight blob is cached in full in a buffer until all chunks have been uploaded, and then `Put` is invoked after the transfer is complete.
This creates two problems:

- Excessive memory usage: with a typical CUDA base image, I can reliably produce OOM situations, since individual layers are gigabytes in size.
- The CLI side looks broken to the end-user: after every byte was sent via HTTP, the progress bar is at 100%, but the server still has to call `Put`. For a slow `BlobHandler` (one that has to upload data to a remote storage backend), this results in minutes of waiting where the progress bar is full and the user thinks that the upload may be stuck.

### To Reproduce

Terminal 1:

```
crane registry serve --address 127.0.0.1:1234
```

```
crane copy nvidia/cuda@sha256:f353ffca86e0cd93ab2470fe274ecf766519c24c37ed58cc2f91d915f7ebe53c localhost:1234/nvidia/cuda:latest
```

### Expected behavior

I want the memory usage to be reasonable, even for large blobs.

### Additional context

I'd like to provide a fix in the form of a custom `io.Reader` with an internal fixed-size buffer that allows a `BlobHandler` to read data in chunks.
This should make uploads faster, less memory intensive, and provide better feedback to users.
Is this an acceptable solution?

malt3:
I noticed that my proposed solution doesn't work in practice with the chunked upload (`POST`, `PATCH`, then `PUT`) from the [spec](https://specs.opencontainers.org/distribution-spec/?v=v1.0.1#pushing-a-blob-in-chunks), since the digest is only part of the final `PUT`.
To limit memory usage, we could add an option to use temporary files instead of a buffer.
If we want to truly solve this (and call `bph.Put` early), we would need to change the interface of `BlobHandler.Put` so that it doesn't know the digest in advance.
That's not a great solution either.

PR #2321, opened 2026-05-24
registry: spill chunked blob uploads to disk

Haihan-Jiang:
## Summary

Chunked registry uploads currently keep every PATCH byte in a per-upload []byte until the final PUT supplies the digest. Large layers can therefore consume memory proportional to the full layer size before BlobHandler.Put is called.

This PR replaces the in-memory upload slice with a small upload buffer that keeps the first 32 MiB in memory and spills additional bytes to a temporary file. The final PUT still verifies the full stream against the requested digest before calling BlobHandler.Put, and successful uploads remove the temporary file.

This addresses the memory pressure reported in #2120 without changing the BlobHandler.Put interface. The final backend Put still happens after the upload is finalized because the digest is only known on the final PUT for chunked uploads.

## Test plan

- go test ./pkg/registry
- go test ./pkg/registry ./pkg/v1/remote ./pkg/crane ./cmd/crane/cmd
- go test ./...
- git diff --check


Haihan-Jiang:
This is ready for review. The registry upload path now spills chunked blob uploads to disk, with coverage in pkg/registry. The visible checks are green.

PR #2322, opened 2026-05-24
name: treat localhost as an explicit registry

### LINKED ISSUE #2048: ggcr: registry on localhost is defaulted to docker.io - image not known when accessing image
Hi!

thanks for the great work done here. 

### Describe the bug

When i.e leveraging podman images are tagged with `localhost` as default registry [1]. Docker used docker.io. 
When calculating an images size [2] based on the locally tagged name, resolution will fail[3] because of defaulting back to docker.io. since localhost is not honnored as a relevant registry name.

I barely remember it, but i think containerd is defaulting to localhost as well (?)

As far as I can see, https://github.com/google/go-containerregistry/blob/main/pkg/name/repository.go#L81 could be the relevant line of code.

With docker the same behaviour works well, since docker uses internally docker.io as a registry name.

### To Reproduce
- Use Podman
- Build docker image [1]
- Run Test [2]

### Expected behavior

I would expect that a valid hostname, like localhost, is accepted as a registry

### Additional context

- Version of the module: github.com/google/go-containerregistry v0.20.2
```
❯ podman --version
podman version 5.3.2
```

[1]
```bash
❯ cat Dockerfile
FROM scratch
❯ docker build . -t testimage:mytag
STEP 1/1: FROM scratch
COMMIT testimage:mytag
--> 379ecbdbbcdc
Successfully tagged localhost/testimage:mytag
379ecbdbbcdcb68e330f76f5928b8aaf37f1a5f13f63d1684a572fa4c7655965
```

[2]
```go
func Test(t *testing.T) {
	refName := "localhost/testimage:mytag"
	reference, err := name.ParseReference(refName)
	assert.NoError(t, err)
	image, err := daemon.Image(reference)
	assert.NoError(t, err)
	_, err = image.Size()
	assert.NoError(t, err)
}
```
[3]
```
=== RUN   Test
    ...test.go:176: 
        	Error Trace:	...test.go:176
        	Error:      	Received unexpected error:
        	            	Error response from daemon: docker.io/library/testimage:mytag: image not known
        	Test:       	Test
--- FAIL: Test (0.01s)

FAIL
```

Haihan-Jiang:
## Summary

Fixes #2048.

Docker-style image reference parsing treats the first path component as a registry when it contains `.` or `:`, and also when it is exactly `localhost`. `NewRepository` only handled the `.` / `:` cases, so `localhost/testimage:mytag` was parsed as a Docker Hub repository path instead of using `localhost` as the registry. That caused daemon lookups for Podman-tagged images such as `localhost/testimage:mytag` to resolve as `index.docker.io/localhost/testimage:mytag` / `docker.io/library/testimage:mytag` instead of the local daemon tag.

This change recognizes `localhost` as an explicit registry component and also makes the bare `localhost` registry use the same `http` scheme fallback as `localhost:<port>`.

## Test plan

- go test ./pkg/name
- go test ./pkg/name ./pkg/v1/daemon ./pkg/v1/remote
- go test ./...
- git diff --check

### DIFF
--- a/pkg/name/registry.go
+++ b/pkg/name/registry.go
@@ -92,7 +92,7 @@ func (r Registry) Scheme() string {
 	if r.isRFC1918() {
 		return "http"
 	}
-	if strings.HasPrefix(r.Name(), "localhost:") {
+	if r.Name() == "localhost" || strings.HasPrefix(r.Name(), "localhost:") {
 		return "http"
 	}
 	if reLocal.MatchString(r.Name()) {
--- a/pkg/name/registry_test.go
+++ b/pkg/name/registry_test.go
@@ -222,6 +222,9 @@ func TestRegistryScheme(t *testing.T) {
 		domain: "my127.0.0.1registry.io",
 		scheme: "https",
 	}, {
+		domain: "localhost",
+		scheme: "http",
+	}, {
 		domain: "localhost:8080",
 		scheme: "http",
 	}, {
--- a/pkg/name/repository_test.go
+++ b/pkg/name/repository_test.go
@@ -27,6 +27,7 @@ var goodStrictValidationRepositoryNames = []string{
 	"example.text/foo/bar",
 	"mirror.gcr.io/ubuntu",
 	"index.docker.io/library/ubuntu",
+	"localhost/testimage",
 }
 
 var goodWeakValidationRepositoryNames = []string{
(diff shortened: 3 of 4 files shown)


Haihan-Jiang:
This is ready for review. The name parser now treats localhost as an explicit registry, with updated registry/repository/tag coverage. The visible checks are green.

PR #2323, opened 2026-05-24
Fix Extract hardlinks after target overwrite

### LINKED ISSUE #2002: `mutate.Extract` will export wrong content when have a hard link and later modify it. (under special conditions)
### Describe the bug

Make a file and create a hard link to it, if this file is not the first file in tar (In general, it is in dictionary order).
And then modify this file, the content from `crane export` of the hard link file will be wrong.

### To Reproduce

```Dockerfile
FROM busybox

RUN echo 123 -> t2
RUN ln t2 t1 && ln t2 t3
RUN echo 456 -> t1
```
```bash
# prepare dockerfile
docker build . -t xxx
docker push xxx
crane export xxx image.tar && tar xf image.tar
cat t1 t2 t3
```
or I have built a image and pushed to dockerhub,
just use it
```
crane export hunsh/ggcr-extract-hardlink image.tar && tar xf image.tar
cat t1 t2 t3
```

will get output
```
456 -
123 -
456 -
```

### Expected behavior
```
456 -
123 -
123 -
```

### Additional context

It's related to https://github.com/google/go-containerregistry/issues/977

- Output of `crane version`: 0.20.2

Haihan-Jiang:
Fixes #2002.

## What changed
- Track hardlink names that are skipped because a newer layer already overwrote that path.
- Rewrite older hardlinks that point at those overwritten names so they target the original older-layer path instead of the newer file.
- Add a regression test for the reported layer sequence: base file, hardlink aliases, then a later overwrite of one alias.

## Why
`mutate.Extract` flattens images by walking layers from newest to oldest. That makes whiteouts efficient, but it also means an older hardlink entry can point at a path that has already been replaced by a newer layer. In that case the exported tar can make an untouched hardlink name resolve to the newer file contents.

In the issue case, `t3` should keep pointing at the old `t2` content after `t1` is overwritten. Before this change, `t3` could still link to `t1` and inherit the newer `456` contents.

## Validation
- `go test ./pkg/v1/mutate -run TestExtractRewritesHardlinkTargetOverwrittenByNewerLayer`
- `go test ./pkg/v1/mutate`
- `PATH="$(go env GOPATH)/bin:$PATH" ./hack/presubmit.sh`

### DIFF
--- a/pkg/v1/mutate/mutate.go
+++ b/pkg/v1/mutate/mutate.go
@@ -267,6 +267,8 @@ func extract(img v1.Image, w io.Writer) error {
 	defer tarWriter.Close()
 
 	fileMap := map[string]bool{}
+	// Tracks old hardlink names that were overwritten by newer layers.
+	hardlinkRewrites := map[string]string{}
 
 	layers, err := img.Layers()
 	if err != nil {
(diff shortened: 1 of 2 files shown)


Haihan-Jiang:
This is ready for review. The extract path now handles hardlinks correctly after a target overwrite, and the visible checks are green.

PR #2326, opened 2026-05-27
Add default repository prefix option

### LINKED ISSUE #2145: Allow overriding the default repository prefix when parsing image references in `pkg/name`
## Description

The `pkg/name` package currently interprets shorthand image names like "ubuntu" as
"index.docker.io/library/ubuntu:latest", automatically defaulting missing fields to:

- Registry: index.docker.io
- Repository prefix: library
- Tag: latest

While convenient, this behavior cannot currently be customized. It would be useful to support environments where the default repository prefix differs from "library" (e.g., "default" or "base").

## Proposed solution

Add an optional argument or configuration option to let users specify the default repository prefix when parsing references, for example:

```
ref, err := name.ParseReference("ubuntu", name.WithDefaultRepo("default"))
// -> index.docker.io/default/ubuntu:latest

```

or via a package-level configuration:

`name.SetDefaultRepo("default")`

This would maintain backward compatibility (still defaulting to "library") while allowing users to adapt the behavior to custom registries or mirrors.

## Motivation

This feature would help:

- Integrate with registries that use different default namespaces.
- Improve flexibility when working in registry setups.
- Maintain consistency with existing configuration patterns in the package (e.g. WithDefaultRegistry).

## References

[pkg/name](https://pkg.go.dev/github.com/google/go-containerregistry/pkg/name) documentation

Similar defaulting logic in [name.go](https://github.com/google/go-containerregistry/blob/main/pkg/name/name.go)

Haihan-Jiang:
Adds `name.WithDefaultRepositoryPrefix` so callers can choose the default prefix used for single-component repository names.

The existing behavior stays the same unless the option is set:
- `ubuntu` still defaults to `index.docker.io/library/ubuntu:latest`
- `WithDefaultRegistry(...)` still defaults only the registry/tag unless a repository prefix is explicitly provided
- `WithDefaultRepositoryPrefix("base")` makes `ubuntu` parse as `index.docker.io/base/ubuntu:latest`

Fixes #2145.

Tests:
- go test ./pkg/name
- go test ./pkg/name ./pkg/v1/remote/transport
- git diff --check

### DIFF
--- a/pkg/name/doc.go
+++ b/pkg/name/doc.go
@@ -40,5 +40,6 @@
 // "library", the missing registry "index.docker.io", and the missing tag
 // "latest". To disable this defaulting, use the StrictValidation option. This
 // is useful e.g. to only allow image references that explicitly set a tag or
-// digest, so that you don't accidentally pull "latest".
+// digest, so that you don't accidentally pull "latest". Use
+// WithDefaultRepositoryPrefix to override the missing repository prefix.
 package name
--- a/pkg/name/options.go
+++ b/pkg/name/options.go
@@ -26,16 +26,19 @@ const (
 )
 
 type options struct {
-	strict          bool // weak by default
-	insecure        bool // secure by default
-	defaultRegistry string
-	defaultTag      string
+	strict                     bool // weak by default
+	insecure                   bool // secure by default
+	defaultRegistry            string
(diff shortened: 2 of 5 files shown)


Haihan-Jiang:
This is ready for review. The default repository prefix option is added, and the visible checks are green.

PR #2332, opened 2026-05-30
fix: prevent SSRF in google.List() pagination

tufstraka:
Prevents SSRF attacks via malicious pagination Link headers in registry responses.

When listing tags or repositories, the library follows `Link` header URLs for pagination. A malicious registry can return a Link header pointing to internal services (e.g., cloud metadata at `169.254.169.254`), causing the client to make unintended requests.

This PR validates pagination URLs before following them:
- Scheme must match the registry's scheme (prevents HTTP downgrade)
- Host must match the registry's host (prevents redirect to arbitrary servers)

Fixes `google.List()`, `remote.List()`, and `remote.Catalog()`.

### DIFF
--- a/pkg/v1/google/list_test.go
+++ b/pkg/v1/google/list_test.go
@@ -301,39 +301,147 @@ func makeResp(hdr string) *http.Response {
 		Header: http.Header{
 			"Link": []string{hdr},
 		},
+		Request: &http.Request{
+			URL: &url.URL{
+				Scheme: "https",
+				Host:   "example.com",
+			},
+		},
 	}
 }
 
 func TestGetNextPageURL(t *testing.T) {
+	repo, err := name.NewRepository("example.com/myrepo")
+	if err != nil {
+		t.Fatalf("failed to create repository: %v", err)
+	}
+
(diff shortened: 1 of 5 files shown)


pkg/v1/google/list_test.go
```diff
@@ -301,39 +301,147 @@ func makeResp(hdr string) *http.Response {
...
 	} {
-		u, err := getNextPageURL(makeResp(hdr))
+		u, err := getNextPageURL(makeResp(hdr), repo)
 		if err == nil && u != nil {
-			t.Errorf("Expected err, got %+v", u)
+			t.Errorf("Expected err or nil URL for %q, got %+v", hdr, u)
 		}
 	}
 
+	// "<>" resolves to current URL which is valid
```
Subserial: Can you provide a link to the Docker spec as to why this is valid? I don't see this in the HTTP API V2 document.

tufstraka: Hey, this is fixed. There's no spec basis for treating <> as valid pagination.

Subserial (approved):

PR #2347, opened 2026-06-16
internal/gzip: fix goroutine leak in ReadCloserLevel

### LINKED ISSUE #2342: internal/gzip: ReadCloserLevel goroutine leaks r on error, causing pullLimiter deadlock via layout.WriteImage
## Summary

`internal/gzip.ReadCloserLevel` starts a background goroutine that holds `r io.ReadCloser`. In several error paths that goroutine calls `pw.CloseWithError(err)` **without first calling `r.Close()`**. When `r` wraps a `limitedReadCloser` (i.e. a remote layer reader acquired via the `pullLimiter` introduced in #2271), the token slot is permanently leaked.

This is a distinct bug from the call-site leaks fixed by #2277 and #2308 — those fixes ensure callers close the reader; this bug means the **goroutine itself** fails to close its copy of `r` when the pipe reader is closed early by the caller.

## Relationship to existing issues

- #2276 / #2277: `mutate.Extract` / `crane export` — reader not closed at call site. Fixed.
- #2309 / #2308: `tarball.MultiRefWrite` / `crane pull` — reader not closed at call site. Fixed (merged, not yet released).
- **This issue**: `internal/gzip.ReadCloserLevel` — reader not closed inside the goroutine on error. Not yet fixed.

## Trigger

`layout/write.go`'s `writeBlob` exits early (without reading all data) when it detects the blob file already exists on disk:

```go
func (l Path) writeBlob(hash v1.Hash, size int64, rc io.ReadCloser, ...) error {
    defer rc.Close()   // ← closes the pipe reader even if not fully consumed
    ...
    if s, err := os.Stat(file); err == nil && !s.IsDir() && ... {
        return nil     // ← early exit
    }
    ...
}
```

When `rc` is the `pr` pipe reader returned by `ReadCloserLevel`, closing it early causes the goroutine's next `bw.Flush()` → `pw.Write()` to return `io.ErrClosedPipe`. The goroutine then hits this path:

```go
// internal/gzip/zip.go
if err := bw.Flush(); err != nil {
    return pw.CloseWithError(err)  // r.Close() is never called → token leaked
}
```

`limitedReadCloser.Close()` is never called, so the `pullLimiter` slot is permanently occupied.

## Why this deadlocks in practice

`layout.WriteImage` (used by kaniko to save a build stage) calls `img.Layers()` and spawns one goroutine per layer. After go-containerregistry v0.21.6's change to `mutate.image.Layers()` (which looks up layers by compressed digest), multiple manifest entries that share the same digest — e.g. empty-tar snapshots from metadata-only Dockerfile commands — are all backed by the **same layer object**. Every goroutine therefore tries to write the same blob.

The first goroutine succeeds and renames the blob file. Every subsequent goroutine hits `writeBlob`'s early-exit and leaks a token. With `defaultJobs = 4`, four leaked tokens fill the limiter channel and all remaining goroutines block forever in `pullLimiter.acquire`.

## Affected code

```go
// internal/gzip/zip.go — ReadCloserLevel goroutine
gw, err := gzip.NewWriterLevel(bw, level)
if err != nil {
    // r.Close() missing → leak
    return pw.CloseWithError(err)
}

if _, err := io.Copy(gw, r); err != nil {
    defer r.Close()   // ✓ this path is fine
    defer gw.Close()
    return pw.CloseWithError(err)
}

if err := gw.Close(); err != nil {
    // r.Close() missing → leak
    return pw.CloseWithError(err)
}

if err := bw.Flush(); err != nil {
    // r.Close() missing → leak  (most common trigger)
    return pw.CloseWithError(err)
}
```

## Fix

```diff
--- a/internal/gzip/zip.go
+++ b/internal/gzip/zip.go
@@ -56,6 +56,7 @@ func ReadCloserLevel(r io.ReadCloser, level int) io.ReadCloser {
 		gw, err := gzip.NewWriterLevel(bw, level)
 		if err != nil {
+			r.Close()
 			return pw.CloseWithError(err)
 		}
 
@@ -68,10 +68,12 @@ func ReadCloserLevel(r io.ReadCloser, level int) io.ReadCloser {
 		// Close gzip writer to Flush it and write gzip trailers.
 		if err := gw.Close(); err != nil {
+			r.Close()
 			return pw.CloseWithError(err)
 		}
 
 		// Flush bufio writer to ensure we write out everything.
 		if err := bw.Flush(); err != nil {
+			r.Close()
 			return pw.CloseWithError(err)
 		}
```

## Environment

- go-containerregistry v0.21.6
- Triggered via kaniko v1.27.6 (`osscontainertools/kaniko#775`) using `layout.WriteImage` on a cached stage

amarkdotdev:
## Problem

`ReadCloserLevel` starts a goroutine that holds `r io.ReadCloser`. In several error paths (`gw.Close()`, `bw.Flush()`), the goroutine calls `pw.CloseWithError(err)` without ever calling `r.Close()`. When `r` wraps a `limitedReadCloser` (pullLimiter from #2271), the token slot is permanently leaked, eventually deadlocking under `layout.WriteImage`.

## Fix

Move `defer r.Close()` and `defer pw.Close()` to the top of the goroutine (resolving the existing `TODO(go1.14)` comment). Add `defer gw.Close()` after creation. This ensures resources are released on every exit path.

## Testing

`go build ./internal/gzip/` and `go test ./internal/gzip/` pass.

Fixes #2342

### DIFF
--- a/internal/gzip/zip.go
+++ b/internal/gzip/zip.go
@@ -51,32 +51,30 @@ func ReadCloserLevel(r io.ReadCloser, level int) io.ReadCloser {
 
 	// Returns err so we can pw.CloseWithError(err)
 	go func() error {
-		// TODO(go1.14): Just defer {pw,gw,r}.Close like you'd expect.
-		// Context: https://golang.org/issue/24283
+		// Always close the source reader when the goroutine exits,
+		// regardless of which error path is taken. This prevents
+		// resource leaks (e.g. pullLimiter token slots held by
+		// limitedReadCloser wrappers around r).
+		defer r.Close()
+
 		gw, err := gzip.NewWriterLevel(bw, level)
 		if err != nil {
 			return pw.CloseWithError(err)
 		}
+		defer gw.Close()
+		defer pw.Close()
 
 		if _, err := io.Copy(gw, r); err != nil {
-			defer r.Close()
-			defer gw.Close()
 			return pw.CloseWithError(err)
 		}
 
-		// Close gzip writer to Flush it and write gzip trailers.
 		if err := gw.Close(); err != nil {
 			return pw.CloseWithError(err)
 		}
 
-		// Flush bufio writer to ensure we write out everything.
 		if err := bw.Flush(); err != nil {
 			return pw.CloseWithError(err)
 		}
-
-		// We don't really care if these fail.
-		defer pw.Close()
-		defer r.Close()
 
 		return nil
 	}()


internal/gzip/zip.go
```diff
@@ -51,33 +51,31 @@ func ReadCloserLevel(r io.ReadCloser, level int) io.ReadCloser {
...
+		// resource leaks (e.g. pullLimiter token slots held by
+		// limitedReadCloser wrappers around r).
+		defer r.Close()
+		defer pw.Close()
+
 		gw, err := gzip.NewWriterLevel(bw, level)
 		if err != nil {
 			return pw.CloseWithError(err)
 		}
+		defer gw.Close()
```
Subserial, line 64: These should close in order pw -> gw -> r. You should move `defer pw.Close()` after `defer gw.Close()`. to maintain this order.

amarkdotdev: Good catch — fixed. Moved `defer pw.Close()` after `defer gw.Close()` so the execution order is pw → gw → r.

Subserial (approved):

PR #2337, opened 2026-06-07
fix(transport): apply refreshed bearer token after cross-host redirect

64johnlee:
## Summary

When an `http.Client`-level redirect sends a request to a host that differs from `bt.registry`, the `sendRequest()` closure skips setting the `Authorization` header — its `matchesHost` guard returns false for cross-host requests. The previous code called `sendRequest()` again after `bt.refresh()`, so the freshly fetched token was never applied to the retry and the client received a second 401.

Fix: after a successful `bt.refresh()`, read the token from `bt.bearer` directly and set the `Authorization` header unconditionally before retrying, bypassing the `matchesHost` guard (which exists to prevent header forwarding on client-visible redirects, not to gate the post-challenge retry).

Closes #2333

## Changes

- `pkg/v1/remote/transport/bearer.go` — replace `return sendRequest()` after `bt.refresh()` with a direct `in.Header.Set("Authorization", ...)` + `bt.inner.RoundTrip(in)`
- `pkg/v1/remote/transport/bearer_test.go` — add `TestBearerTokenAppliedAfterCrossHostRedirect` regression test: sets up a registry server whose host differs from `bt.registry`, verifies that after a 401 challenge the retry succeeds with the fresh token applied

## Test plan

- [x] `go test ./pkg/v1/remote/transport/...` passes (all existing tests + new regression test)
- [x] New test `TestBearerTokenAppliedAfterCrossHostRedirect` reproduces the bug on the old code path and passes on the fix

Issue #2333, opened 2026-06-03
ggcr: Bearer token from redirected registry challenge is fetched but not applied to redirected registry host


amitz-jfrog:
### Describe the bug

When a registry request is redirected to another registry-compatible host, and that redirected host returns a valid `WWW-Authenticate: Bearer ...` challenge, `go-containerregistry` fetches a token from the redirected host's token realm but does not apply that token to the retried request.

This causes the retried request to fail with `401 Unauthorized`, even though the token endpoint returned `200 OK` and the same token works when applied manually.

### To Reproduce

Run `crane` (or the `go-containerregistry` library) against a primary registry where traffic is redirected by a network policy or registry gateway to a secondary registry requiring authentication.

**Observed HTTP Flow:**

```text
GET https://primary-registry.example.com/v2/
-> 307 Location: https://redirect-gateway.example.com/route?url=...

GET https://redirect-gateway.example.com/route?url=...
-> 307 Location: https://redirected-registry.example.com/v2/target-repo/

GET https://redirected-registry.example.com/v2/target-repo/
-> 401
   WWW-Authenticate: Bearer realm="https://redirected-registry.example.com/auth/token",
                      service="redirected-registry.example.com"

GET https://redirected-registry.example.com/auth/token?scope=repository%3Amy-namespace%2Fmy-image%3Apull&service=redirected-registry.example.com
-> 200 OK

GET https://redirected-registry.example.com/v2/target-repo/my-namespace/my-image/manifests/latest
-> 401 Unauthorized

```

### Expected behavior

After successfully fetching the bearer token from the redirected host's authorization server (`200 OK`), the library should apply that token to the subsequent retry request against the redirected host, resulting in a successful request fulfillment.

### Additional context

* **Version of the module:** `0.21.6`
* **Registry used:** Primary registry (e.g., Docker Hub) redirected via a gateway/network policy to a secondary registry

PR #2349, opened 2026-06-17
transport: do not re-attach bearer token after cross-host redirect

evilgensec:
## Summary

`bearerTransport.RoundTrip` re-attaches the registry bearer token on its post-401 retry without checking that the request is still talking to the registry we authenticated against.

PR #2337 replaced the post-`bt.refresh()` `return sendRequest()` with an unconditional `in.Header.Set("Authorization", ...)` + `bt.inner.RoundTrip(in)`, intentionally bypassing the `matchesHost` guard. That guard exists to keep the `Authorization` header from being forwarded across an `http.Client`-level redirect to a different host.

With the guard bypassed, the credential leaks across a cross-host redirect:

1. A pull is issued against a registry; the transport seeds a bearer token from the trusted realm.
2. The registry answers the manifest/blob GET with a `302` to a *different* host. `checkRedirectSSRF` (pkg/v1/remote/fetcher.go) allows it for any public host or DNS name (it only blocks private/link-local IP literals).
3. `RoundTrip` re-runs for the redirected host. The first `sendRequest()` correctly omits `Authorization` (host mismatch). The redirected host replies `401` with a Bearer challenge.
4. `RoundTrip` refreshes the token from the trusted realm and, after #2337, sets `Authorization: Bearer <token>` unconditionally and sends it to the redirected host.

A registry the client did not authenticate against (a malicious/compromised peer, or a MITM on insecure transport) thus receives the operator's registry token.

## Fix

Gate the re-attach on `matchesHost`: only apply the refreshed token when the request is still on the registry host we logged in to; otherwise fall back to `sendRequest()` (which omits the credential). This preserves the #2333 behavior for same-host 401 retries (e.g. token expiry mid-session) and closes the cross-host credential leak.

## Changes

- `pkg/v1/remote/transport/bearer.go` — guard the post-refresh re-attach with `matchesHost`.
- `pkg/v1/remote/transport/bearer_test.go` — `TestBearerTokenReappliedOnSameHostChallenge` (the #2333 same-host retry still works) and `TestBearerTokenNotLeakedOnCrossHostChallenge` (token is not sent to a cross-host challenger).

## Test plan

- [x] `go test ./pkg/v1/remote/...` passes (existing tests + both new tests).
- [x] Reverting the guard makes `TestBearerTokenNotLeakedOnCrossHostChallenge` fail (the credential is forwarded), confirming the test catches the regression.

### DIFF
--- a/pkg/v1/remote/transport/bearer.go
+++ b/pkg/v1/remote/transport/bearer.go
@@ -247,11 +247,18 @@ func (bt *bearerTransport) RoundTrip(in *http.Request) (*http.Response, error) {
 		if err = bt.refresh(in.Context()); err != nil {
 			return nil, err
 		}
-		// Apply the freshly fetched token directly rather than calling
-		// sendRequest(), which only sets Authorization when in.URL.Host
-		// matches bt.registry. When the request followed a cross-host
-		// redirect the hosts differ and sendRequest() would omit the
-		// header, causing an unnecessary second 401.
+		// Re-attach the freshly fetched token, but only when the request is
+		// still talking to the registry we authenticated against. matchesHost
+		// guards against forwarding the Authorization header across an
+		// http.Client-level redirect to a different host: a malicious or
+		// compromised registry can 302 the request to an attacker-controlled
+		// host, answer the follow-up with a Bearer challenge, and harvest the
+		// token if we re-attach it unconditionally. For a cross-host request
+		// fall back to sendRequest(), which omits the credential, rather than
+		// leaking it to a host we never logged in to.
+		if !matchesHost(bt.registry.RegistryStr(), in, bt.scheme) {
+			return sendRequest()
+		}
 		bt.mx.RLock()
 		tok := bt.bearer.RegistryToken
 		bt.mx.RUnlock()
--- a/pkg/v1/remote/transport/bearer_test.go
+++ b/pkg/v1/remote/transport/bearer_test.go
@@ -698,41 +698,36 @@ func TestValidateRealmURLSameHost(t *testing.T) {
 	}
 }
 
-// TestBearerTokenAppliedAfterCrossHostRedirect is a regression test for
-// https://github.com/google/go-containerregistry/issues/2333.
-//
-// When an HTTP redirect sends the request to a host that differs from
-// bt.registry, the sendRequest() closure omits the Authorization header
-// (its matchesHost guard returns false). Previously, the post-401 retry
-// also went through sendRequest(), so the freshly refreshed token was
-// never applied and the request failed with a second 401.
-//
-// The fix skips sendRequest() after a successful refresh and sets the
-// Authorization header directly before retrying.
-func TestBearerTokenAppliedAfterCrossHostRedirect(t *testing.T) {
+// TestBearerTokenReappliedOnSameHostChallenge is the regression test for
+// https://github.com/google/go-containerregistry/issues/2333: when the
+// registry we authenticated against answers with a 401 Bearer challenge
+// (e.g. token expiry mid-session), the freshly refreshed token must be
+// applied to the retry so it succeeds rather than failing with a second 401.
+func TestBearerTokenReappliedOnSameHostChallenge(t *testing.T) {
 	const (
 		staleToken = "stale-token"
 		freshToken = "fresh-token"
 	)
 
 	attempts := 0
-	// registryServer simulates a registry reachable after a cross-host redirect.
-	// Its host intentionally differs from bt.registry ("registry.example.com").
 	registryServer := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 		attempts++
 		if r.Header.Get("Authorization") == "Bearer "+freshToken {
 			w.WriteHeader(http.StatusOK)
 			return
 		}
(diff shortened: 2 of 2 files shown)


Subserial (approved):

evilgensec:
Hi @Subserial , thanks again for merging this.

A quick clarification for my records: I reported this cross-host bearer-token regression through the Google OSS VRP, and the response stated it had been "identified and fixed internally by the maintainers via PR [#2349](https://github.com/google/go-containerregistry/pull/2349)" before my report. Since [#2349](https://github.com/google/go-containerregistry/pull/2349) is the PR I opened (and you merged it about six minutes later), I wanted to check with you directly: was this regression already being tracked or fixed internally before I opened [#2349](https://github.com/google/go-containerregistry/pull/2349), or was this PR the fix that resolved it? I just want to represent the timeline accurately. Thanks either way, and thanks for the fast review.

amitz-jfrog:
@evilgensec Thanks for the contribution. 
Just to make sure, does this PR reverts the resolution of issue https://github.com/google/go-containerregistry/issues/2333?
What registry does it match to? My issue was involving cross-host redirect. But a specific case in which the same host generated the token and later did not receive it. 
Does it make sense?

evilgensec:
@amitz-jfrog you're right, it does. [#2349](https://github.com/google/go-containerregistry/pull/2349) re-gates the retry on matchesHost, so the cross-host case in #2333 no longer re-applies the token and regresses.

I couldn't keep #2337's unconditional re-attach: it re-applies bt.realm's token (the original registry's credential) to whatever host the redirect lands on, so a malicious or compromised registry can 302 to an attacker host, return a Bearer challenge, and harvest the token. That's what [#2349](https://github.com/google/go-containerregistry/pull/2349) closed.

The safe fix for your gateway case is to treat the redirected host's 401 as a fresh challenge: fetch a token from that host's own realm and apply it there, instead of forwarding the original registry's token cross-host.

amitz-jfrog:
@evilgensec I see, so how do you suggest opening the issue again in a way that will be understandable and not trigger a vulnerability?
My use-case is a constant redirect between host A to B.
So the flow is, as I understand it:
```
Docker calls A ->(redirect) B. Gets 401 from B, to B token endpoint
Docker calls B directly and gets the token.
Docker calls host A for the image since it now thinks it got the token. ->(redirect) B, does not attach the token received from B. 
Fails.

```

I would assume that there is no risk in attached the host B token when calling host B. 
Also, this use-case is working with other docker clients (when using docker-ce for example)

amitz-jfrog:
What do you think?

evilgensec:
@amitz-jfrog You're right, and that's the key distinction: attaching host B's token to host B is safe. The leak #2349 closed was #2337 attaching host A's token to B.

Why a simple "re-attach on same host" won't fix your flow: bearerTransport is built once from the original registry's challenge, so realm/service/creds stay pinned to A. On B's 401, refresh() re-mints a token from A's realm with A's creds (B's challenge realm/service are ignored, only scope is read). So the token in hand is always A's. Re-attaching it to B is the cross-host forward we removed.

The fix for #2333 that doesn't reintroduce the leak is a per-host token exchange: when the 401 comes from a host different from the one we authenticated to, treat it as a fresh auth context. Parse that host's realm/service/scope, resolve creds for that host from the keychain, exchange a token against its own realm, and attach it only to that host. A's credential never reaches B; B's token never leaves B. That's the per-host model docker-ce appears to use.

For reopening, frame it as a feature request, not a revert:

- Title: "support bearer auth for the redirected host on cross-host redirect"
- Include your A→B flow as the repro
- State the invariant: never forward the original registry's token to the redirect target. Only use a token minted by the target's own realm
- Cross-link #2333, #2337 (introduced the leak), #2349 (closed it)

amitz-jfrog:
@evilgensec thank you very much. 🙏 
Is this what you had in mind?

**Issue** https://github.com/google/go-containerregistry/issues/2359
**PR** https://github.com/google/go-containerregistry/pull/2360

amitz-jfrog:
@evilgensec following up on that, this PR reverted the other issue's fix.
I have created a more specific fix, but I am not sure how to get it reviewed.
Can you please assist?

evilgensec:
@amitz-jfrog Reviewed it. Correct approach, and it holds the [#2349](https://github.com/google/go-containerregistry/pull/2349) invariant: handleCrossHostChallenge exchanges with authn.Anonymous and fromChallenge still SSRF-validates the redirected host's realm, so the original registry's token never reaches host B. Restores [#2333](https://github.com/google/go-containerregistry/issues/2333) without reopening the leak.

Two tightenings:

- Move bt.refresh() into the same-host branch. On the cross-host path it mints an original-registry token that gets discarded, and pollutes shared bt.scopes with B's scope.

- Note in the description that the exchange is anonymous-only (public token endpoints); auth-gated endpoints 401 and fall through to no creds, keychain support is a follow-up. Pre-empts the obvious review question.

CI is green and the diff is small. Request review from @Subserial , who merged [#2349](https://github.com/google/go-containerregistry/pull/2349) and owns this path. Framing: "restores [#2333](https://github.com/google/go-containerregistry/issues/2333) while preserving the [#2349](https://github.com/google/go-containerregistry/pull/2349) no-cross-host-forward invariant."

PR #2360, opened 2026-06-30
transport: per-host bearer token exchange on cross-host redirect

### LINKED ISSUE #2359: transport: support bearer auth for the redirected host on cross-host redirect
### Background

Issue #2333 reported that when a registry request is permanently redirected (A→B), and host B issues a `401 WWW-Authenticate: Bearer` challenge with its own realm, `go-containerregistry` fetches a token from B's realm but fails to apply it to the retry against B.

PR #2337 fixed #2333 by unconditionally re-attaching the refreshed token. However, this introduced a credential-leak vulnerability: `bearerTransport` would forward the original registry's token (minted from A's realm) to the redirected host B. PR #2349 closed that vulnerability by re-gating the re-attach on \`matchesHost\`, which regressed #2333.

### Use Case

A client is configured to pull from registry A, but a network policy or registry gateway permanently redirects all traffic from A to B. Both registries are legitimate and trusted in this deployment.

**Observed HTTP Flow:**

```
GET https://A/v2/target-repo/manifests/latest
→ 307 Location: https://B/v2/target-repo/manifests/latest

GET https://B/v2/target-repo/manifests/latest
→ 401 WWW-Authenticate: Bearer realm="https://B/auth/token",service="B"

GET https://B/auth/token?scope=...&service=B
→ 200 OK  (token minted by B's own realm)

GET https://A/v2/target-repo/manifests/latest   ← library retries against A
→ 307 Location: https://B/...

GET https://B/v2/target-repo/manifests/latest   ← B's token is NOT attached
→ 401 Unauthorized
```

### Root Cause

`bearerTransport` is constructed once from the original registry's challenge (host A). Its `realm`, `service`, and credentials are pinned to A. On a cross-host 401 from B:

1. `refresh()` re-mints a token from A's realm with A's credentials (B's challenge realm/service are ignored beyond scope)
2. The `matchesHost` guard (restored by #2349) correctly prevents forwarding A's token to B
3. No mechanism exists to perform a **per-host** token exchange: fetch a token from B's own realm, with credentials for B, and apply it only to B

### Expected Behavior

When the redirected host B returns a `401 WWW-Authenticate: Bearer` challenge, the library should:

1. Parse B's `realm`, `service`, and `scope` from the challenge
2. Resolve credentials for B from the configured keychain
3. Exchange a token against B's own realm
4. Apply that token **only** to subsequent requests to B

A's credentials must never reach B. B's token must never leave B. This matches the per-host authentication model used by docker-ce and other OCI clients.

### Security Invariant

This feature request must preserve the invariant closed by #2349: the original registry's token (minted from A's realm using A's credentials) must never be forwarded to the redirected host.

### References

- #2333 — original bug report
- #2337 — introduced the credential-leak while fixing #2333  
- #2349 — closed the credential-leak, regressed #2333

amitzig:
## Problem

When a registry permanently redirects traffic to a different host (A→B) and host B issues its own `WWW-Authenticate: Bearer` challenge, the library silently retried without credentials after #2349. The fix from #2337 was correctly reverted (it leaked A's token to B), but no replacement was provided.

Closes #2359. See also #2333, #2337, #2349.

## Fix

`matchesHost` is now checked immediately on a `WWW-Authenticate` challenge, before any scope accumulation or token refresh. This branches into two fully separate paths:

- **Same-host** (request is still on the registry we authenticated against): unchanged behavior — accumulate the challenge's scope into `bt.scopes`, call `bt.refresh()`, reattach the refreshed token.
- **Cross-host** (request was redirected to a different host): `handleCrossHostChallenge()` is called instead, which:
  1. Parses the redirected host's own `WWW-Authenticate` Bearer challenge
  2. Validates the realm URL through the existing SSRF guard (`validateRealmURL`), relative to the redirected host
  3. Fetches a token from **that host's own realm** using `authn.Anonymous` — the original registry's credentials never leave their host
  4. Applies the token only to the retry against the redirected host, without touching `bt`'s shared state

Keeping the cross-host path from mutating `bt.scopes` / `bt.bearer` matters: without this, a cross-host challenge would accumulate the redirected host's scope into `bt.scopes` and mint (then discard) a token from the *original* registry's realm — polluting state used by subsequent same-host requests, for no benefit.

## Anonymous-only exchange (caveat)

The per-host token exchange in `handleCrossHostChallenge` uses `authn.Anonymous` only — it covers public token endpoints on the redirected host. If the redirected host's token endpoint itself requires credentials, the exchange 401s and the request falls through to a no-credentials retry (same as today's behavior, no regression). Resolving credentials for the redirected host via a keychain is a natural follow-up, tracked via the `TODO` left in `handleCrossHostChallenge`, and is intentionally left out of this PR to keep the fix minimal and easy to review.

## Security invariant

The credential-leak guard from #2349 is fully preserved: the original registry's token is never forwarded cross-host. The new path uses `authn.Anonymous` for the redirected host's token exchange.

## Testing

- `TestBearerTokenExchangedOnCrossHostChallenge` — new test covering the A→B redirect flow
- `TestBearerTokenReappliedOnSameHostChallenge` — same-host refresh still works
- `TestBearerTokenNotLeakedOnCrossHostChallenge` — credential leak still blocked
- `TestInsufficientScope` — fixed to construct `registry` from the actual test-server address instead of an unrelated hostname; it was incidentally exercising the cross-host branch and only passed due to the old unconditional-refresh ordering
- Full `./pkg/v1/remote/...` suite passes


amitz-jfrog:
@Subserial can this be considered and reviewed?

amitz-jfrog:
Any update on this?

amitz-jfrog:
@Haihan-Jiang  Hey, saw that you are an active contributor for this repo.
What is the process for getting this change looked at? 
Thanks

amitzig:
Restores #2333 while preserving the #2349 no-cross-host-forward invariant. Requesting review from @Subserial, who merged #2349 and owns this path.

amitz-jfrog:
@Subserial Hey Jared, 
Checking if you had the chance to look at this change?
Thanks

amitz-jfrog:
Hey,
Following up on this

Subserial (approved):

PR #2378, opened 2026-07-26
crane: allow empty label values

### LINKED ISSUE #2376: crane: allow empty label values
### Describe the bug

An empty label value is not accepted by the mutate command. But labels without values are allowed according to the spec.

https://github.com/opencontainers/image-spec/blob/main/annotations.md#rules
> While the value MUST be present, it MAY be an empty string.

### To Reproduce

`crane mutate --label my.key= my-image:tag`

produces the error

`Error: parsing label "my.key", value is empty`

### Expected behavior

I expect the image to be modified with a label set with a empty string as value.

### Additional context

Crane version e8813dd0a00e799459cae01d8a4659b9be2fd871

vigneshakaviki:
## Summary

- allow `crane mutate` to retain labels whose value is empty
- cover empty label values through an in-memory registry round trip

Fixes #2376

## Testing

- `./hack/presubmit.sh`
- `go test -race ./...`
- `go run github.com/golangci/golangci-lint/v2/cmd/golangci-lint@v2.11.0 run`
- `go run golang.org/x/tools/cmd/goimports@v0.48.0 -d cmd/crane/cmd/mutate.go cmd/crane/cmd/mutate_test.go`

### DIFF
--- a/.github/workflows/e2e.yaml
+++ b/.github/workflows/e2e.yaml
@@ -4,6 +4,9 @@ on:
   pull_request:
     branches: ['main']
 
+permissions:
+  contents: read
+
 env:
   GOTOOLCHAIN: local
 
--- a/.github/workflows/e2e.yaml
+++ b/.github/workflows/e2e.yaml
@@ -56,9 +56,14 @@ jobs:
                 --new_tag localhost:1338/append-test \
                 --new_layer crane.tar))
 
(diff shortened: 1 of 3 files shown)


Subserial:
The windows runner issues are transient, and we should address them in a different PR. Could you remove the changes from this PR?

PR #2382, opened 2026-07-29
tarball: use correct file extension for zstd/uncompressed

milas:
Currently, the tarball writer always uses `.tar.gz` as the layer extension.

Instead, look at the media type and set it based on that:
 * zstd -> `.tar.zst`
 * uncompressed -> `.tar`
 * gzip/fallback -> `.tar.gz`


Subserial (approved):

Subserial:
Seems like just a spellcheck error. I can accept the change after a small fix.

milas:
> Seems like just a spellcheck error.

🤦🏻 

Pushed a new commit with better speeling, if you'd prefer I rebase into one commit LMK!

Subserial (approved):

PR #2393, opened 2026-08-03
mutate: don't reject the rootfs root ("./") as an unsafe tar path

CoreyCook8:
Extract rejected any layer entry whose path cleaned to "." as an "unsafe tar path". Unlike ".." / "../…", a "." entry does not escape the extraction root -- it is the root -- so rejecting it serves no security purpose. It does, however, break real images: gcr.io/distroless base images (and everything built on them) ship a standalone "./" layer entry, so Extract, crane export, and crane flatten all fail on them with unsafe tar path ".".

Drop "." from unsafeArchivePath (keeping ".." and "../"), and skip the redundant root entry in extractLayer instead of writing it, mirroring how whiteout and duplicate entries are already skipped. Empty-name entries, which also clean to ".", are skipped the same way.

Steps to Reproduce:

`crane export --platform linux/amd64 gcr.io/distroless/static-debian12:latest /dev/null`


Subserial:
Fixed this in #2400.

PR #2406, opened 2026-08-12
remote: retry failed Puller and Pusher initialization

### LINKED ISSUE #2060: issue with caching "bad" fetchers in Puller/Pusher
See e.g. https://github.com/google/go-containerregistry/blob/c4dd792fa06c1f8b780ad90c8ab4f38b4eac05bd/pkg/v1/remote/puller.go#L65-L90 

(same issue is in the Pusher)

For the code like:

```
puller := remote.NewPuller() // cache the puller instance

// somewhere later

puller.Head(ctx, reference) 
```

In this flow, fetcher is created lazily on access, but it uses the `ctx` passed with the first request. If the fetcher fails to init itself because `ctx` got canceled, it will be cached forever in the failed state (due to `sync.Once`). While the next `.Head` request might come with non-canceled context, it will still fail with the first error.

Moreover, the `ctx` used to initialize the fetcher is same as `ctx` used for the `Head` request itself, so passing `context.Background` is not an option.

There is no interface to clean cached fetcher, or even skip the cache.

Is it by design?

The only workaround is to create `Puller`/`Pusher` on each request, which completely cancels the idea of caching/reusing fetchers.

iahsanGill:
A failed Puller or Pusher initialization is retained by sync.Once, so a transient context, authentication, or transport error poisons the reusable client.

Delete only the failed cached reader or writer with CompareAndDelete, allowing the next request to initialize it again without removing a concurrent replacement. This adds retry regressions for both paths.

Fixes #2060

Tests:
- ./hack/presubmit.sh
- go test -race ./pkg/v1/remote -count=1

### DIFF
--- a/pkg/v1/remote/puller.go
+++ b/pkg/v1/remote/puller.go
@@ -76,7 +76,11 @@ func (p *Puller) fetcher(ctx context.Context, target resource) (*fetcher, error)
 		o:      p.o,
 	})
 	rr := v.(*reader)
-	return rr.f, rr.init(ctx)
+	if err := rr.init(ctx); err != nil {
+		p.readers.CompareAndDelete(target, rr)
+		return nil, err
+	}
+	return rr.f, nil
 }
 
 // Head is like remote.Head, but avoids re-authenticating when possible.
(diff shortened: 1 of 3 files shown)


Subserial (approved):

Subserial:
Approved, but it looks like the change has some linter errors.

iahsanGill:
Updated the part which was failing.

PR #2408, opened 2026-08-14
validate: check artifact manifests structurally instead of as images

RogerHYang:
`crane validate` fails on any multi-arch image published by BuildKit with provenance/SBOM attestations enabled:

```
$ crane validate --fast --remote docker.io/arizephoenix/phoenix:version-20.2.0-nonroot
Error: validating children: failed to validate image Manifests[2](sha256:158c42…):
validating config: invalid ConfigFile.RootFS.Type: "" != "layers"
```

BuildKit attaches attestation manifests to the image index, and recent versions give them the OCI empty descriptor (`{}`) as their config. `validate.Index` validates every `OCIManifestSchema1`/`DockerManifestSchema2` child as a container image, so the attestation's config fails the `ConfigFile.RootFS.Type == "layers"` invariant and validation of an otherwise well-formed image reports failure.

Rather than skipping these children, this change validates them as what they are. A child is treated as an artifact manifest when its index descriptor or manifest declares an `artifactType`, or its config media type is not an image config (`!IsConfig()`). Artifact children get structural validation — manifest digest/size/content consistency, config blob existence and digest/size match, layer blob existence (`Fast`) or digest/size verification (full) — while image-only invariants (rootfs type, diff_ids, platform) are not applied. A registry copy with a missing or corrupt attestation blob still fails validation.

Also fixes two bounds bugs in `validateLayers`, found while testing with attestation-shaped images:

- the computed-layer slices are appended only for `IsLayer()` media types but indexed by overall layer position, so an image with a leading non-layer blob panics or compares the wrong layers;
- `cf.RootFS.DiffIDs[i]` panics with index out of range when the config declares fewer diff_ids than the image has layers (previously reachable via any image whose config is missing rootfs data — now reported as a validation error).

Testing the descriptor-level `artifactType` path surfaced that `mutate.AppendManifests` silently dropped `IndexAddendum.ArtifactType`; it now propagates like the other descriptor overrides.

Tests cover: an index with a BuildKit-style attestation child validates clean (full and `Fast`); a child detected only via its descriptor `artifactType` validates clean; a child with a tampered config blob fails on config digest; an image with layers but no diff_ids returns an error instead of panicking.

Verified against a real BuildKit-published image: `crane validate --fast --remote docker.io/arizephoenix/phoenix:version-20.2.0-nonroot@sha256:b20fc0…` now reports `PASS`.

### DIFF
--- a/pkg/v1/mutate/index.go
+++ b/pkg/v1/mutate/index.go
@@ -56,6 +56,9 @@ func computeDescriptor(ia IndexAddendum) (*v1.Descriptor, error) {
 	if ia.Data != nil {
 		desc.Data = ia.Data
 	}
+	if ia.ArtifactType != "" {
+		desc.ArtifactType = ia.ArtifactType
+	}
 
 	return desc, nil
 }
--- a/pkg/v1/validate/artifact.go
+++ b/pkg/v1/validate/artifact.go
@@ -1,0 +1,137 @@
+// Copyright 2026 Google LLC All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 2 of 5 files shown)


Subserial:
This work has been superceded by #2414.

PR #2414, opened 2026-08-18
validate: support index attestation manifests and empty configs

### LINKED ISSUE #2409: crane validate fails on indexes with BuildKit attestation manifests (empty OCI config)
### Describe the bug

`validate.Index` (and therefore `crane validate`) fails on any multi-arch image whose index carries BuildKit SBOM/provenance attestation manifests in the current format. Recent BuildKit gives attestation manifests the OCI empty descriptor (`application/vnd.oci.empty.v1+json`, content `{}`) as their config. `validateChildren` validates every `OCIManifestSchema1`/`DockerManifestSchema2` child as a container image, so these children fail image-only invariants and validation of a well-formed image reports failure.

### To Reproduce

BuildKit's own published image reproduces it:

```
$ go run github.com/google/go-containerregistry/cmd/crane@v0.20.3 validate --fast --remote docker.io/moby/buildkit:latest
...
failed to validate image Manifests[11](sha256:76db3f42ed8eebbbc0f8d00237a1fa26587ea3cff9536f6d41bc8dbe59753c52): validating config: invalid ConfigFile.RootFS.Type: "" != "layers"
failed to validate image platform[11](sha256:76db3f42ed8eebbbc0f8d00237a1fa26587ea3cff9536f6d41bc8dbe59753c52): config file missing platform fields
exit status 1
```

`Manifests[11]` is an attestation manifest (annotated `vnd.docker.reference.type=attestation-manifest`, platform `unknown/unknown`), not an image. Any image built with `docker buildx build --provenance=true --sbom=true` using a current BuildKit reproduces the same failure. Older BuildKit gave attestation manifests an image-style config with a fake rootfs, which happened to pass validation — the empty-config format is what broke it.

### Expected behavior

Validation of an image index should not apply image invariants (rootfs type, diff_ids, platform) to attestation/artifact children. They are still worth checking structurally — manifest digest/size consistency and config/layer blob existence — so corruption is still caught.

### Additional context

Two adjacent panics in `validateLayers`, reachable with attestation-shaped inputs (reproducible with `random.Image` + `mutate.ConfigFile(img, &v1.ConfigFile{})`):

- `cf.RootFS.DiffIDs[i]` panics with index out of range when the config declares fewer diff_ids than the image has layers;
- the computed-layer slices are appended only for `IsLayer()` media types but indexed by overall layer position, so an image with a leading non-layer blob (e.g. an in-toto layer) panics or compares the wrong layers.

Also, `mutate.AppendManifests` drops `IndexAddendum.ArtifactType` instead of propagating it like the other descriptor overrides, which makes artifact-typed index children awkward to construct programmatically.

Subserial:
Fixes #2409

Fix validate.Index and validate.Image failures on multi-architecture images carrying BuildKit attestation and artifact manifests.

This change also covers panics mentioned in the source bug.


lindsayismith:
Copilot has identified some cases that would no longer be caught.  Perhaps we add tests for these cases and make sure they do the right thing:

I'll load the PR understanding ability first to properly analyze this pull request, then examine the changes.Now I'll fetch the full PR data and then examine the changes in detail.Now let me get the commit information and review state to understand the PR better:Now let me examine the actual code changes more carefully by looking at the specific files:Now let me look at the index validation changes:Now let me look at the old version to see what was changed:Based on my analysis of this PR, I've identified **two significant breaking changes** that could cause regressions for existing users:

## Critical Breaking Changes

### 1. **Empty Config Files Now Validate Successfully (Previously Failed)**

**The change:** Lines 80-98 in `image.go` now add a conditional guard:
```go
if m.Config.MediaType.IsConfig() {
    // Parse and validate config file content
}
```

**Impact:** Previously, **any image with an empty or missing config would fail validation** (line 94: `if cf.RootFS.Type != "layers"`). Now, images with empty configs (media type `application/vnd.oci.empty.v1+json`) **will pass validation even though they have no RootFS information**.

**Who could break:** Code that relies on validation rejecting images with empty configs or that has workarounds to skip validation for such images may now validate successfully, potentially exposing downstream issues.

### 2. **Layer DiffID Validation Now Skipped for Non-Layer Blobs (Previously Checked)**

**The change:** Lines 168-244 in `image.go` fundamentally restructured layer validation:

**Before (main branch, lines 117-222):**
- Used `append()` to build slices, automatically skipping non-layer blobs
- Only validated real layers in the second loop (line 166: `if !mediaType.IsLayer() { continue }`)
- Compared layer DiffIDs using manifest index directly: `cf.RootFS.DiffIDs[i]` (line 207)

**After (PR, lines 119-244):**
- Pre-allocates slices with full layer count: `make([]v1.Hash, len(layers))`
- Stores **both** layer and non-layer blob data at their respective indices
- **Only compares DiffIDs for actual layers** (lines 226, 239-243)

**Critical issue:** The old code would panic if a manifest had a **leading non-layer blob** (e.g., in-toto attestation):
- Non-layer blob at index 0: skipped
- Real layer at index 1: validated against `cf.RootFS.DiffIDs[0]` → **index mismatch panic**

The new code **silently skips** this validation for non-layer blobs, which means:
- Images with non-layer blobs mixed into the layer array will no longer fail
- **But also:** There's no validation that non-layer blob digests match the manifest

## Risk Assessment

**Most likely regression:** A user has an image with a leading non-layer blob (BuildKit attestation, SBOMs) and was working around the validation failures. This PR will now make their images pass validation **even if the non-layer blob is corrupted** (wrong digest/size).

**Test coverage:** The new tests (`image_test.go`, `index_test.go`) cover the happy path (valid attestations) but don't test:
- Non-layer blobs with mismatched digests
- Non-layer blobs with mismatched sizes
- Attestation manifests with invalid inner structure

## Recommendation

This PR fixes a real bug (issue #2409) but loosens validation on non-layer blobs without adding comprehensive checks. Consider adding tests that verify non-layer blob structural integrity (digest/size consistency) is still caught, even if their internal content isn't validated like image configs.

Subserial:
The latest push adds negative tests to prevent regressions introduced by the new permissive checks.

tprussak (approved):

PR #2419, opened 2026-08-21
fix(build): unify new build flow into cloudbuild_v2.yaml

tprussak:
Merge the two new build files into a unified build which does both the golang and docker builds, to avoid needing to copy intermediate files between the two stages.


HassanJasim (approved):
LGTM

PR #2420, opened 2026-08-23
authn: do not overwrite AuthConfig.Auth when username and password are empty

### LINKED ISSUE #1864: ggcr: authn.AuthConfig JSON marshaling logic does not handle some edge cases correctly
### Describe the bug

Where there is nothing in the `Username` and `Password` fields and only the `Auth` field of `authn.AuthConfig` is filled, custom json marshalling fails to produce proper output.

### To Reproduce

```go
authConfig := authn.AuthConfig{Username: "", Password: "", Auth: "<any valid base64 auth string>"}
dockerCfgAuth, err := json.Marshal(authConfig)
```

Running the example above produces a JSON object with only the `auth` field as expected. However, the contents of this field will always be `Og==`, which is base64 for `:`.

This is due to the `encodeDockerConfigFieldAuth` function in `pkg/authn/auth.go` always overwriting contents of the `Auth` field with a concatenation of `username + ":" + password`, which leads to data corruption when there is only the auth string and no username and password is provided. 

### Expected behavior

Zero values of `authn.AuthConfig` are expected to marshal to `{}`, not `{"auth": "Og=="}` and values with only `Auth` fields filled are expected to either only produce JSON objects with this field and its contents or produce an object with both auth and username+password pair decoded from auth.

### Additional context

- Version of the module - v0.17.0
- Registry used (e.g., GCR, ECR, Quay) - irrelevant

nileshpatil6:
Fixes #1864.

### The bug

`AuthConfig.MarshalJSON` unconditionally recomputes the `auth` field:

```go
func (a AuthConfig) MarshalJSON() ([]byte, error) {
	shadow := (authConfig)(a)
	shadow.Auth = encodeDockerConfigFieldAuth(shadow.Username, shadow.Password)
	return json.Marshal(shadow)
}
```

When neither `Username` nor `Password` is set there is nothing to derive it from, so `auth` becomes the encoding of `":"`, which is `Og==`.

```go
AuthConfig{Auth: "dXNlcjpwYXNz"}    -> {"auth":"Og=="}
AuthConfig{RegistryToken: "bearer"} -> {"auth":"Og==","registrytoken":"bearer"}
AuthConfig{IdentityToken: "id"}     -> {"auth":"Og==","identitytoken":"id"}
```

The first case silently discards a valid credential: a config carrying only an already encoded `Auth` marshals to a useless one, and round-tripping it yields `Auth: "Og=="` with empty username and password. The second and third invent an `auth` field for configs that never had one.

This is the case reported in #1864, which was closed by the stale bot rather than resolved, with a follow-up noting it was still broken.

### The change

Derive `auth` only when a username or password is present, otherwise leave whatever the caller set:

```go
if shadow.Username != "" || shadow.Password != "" {
	shadow.Auth = encodeDockerConfigFieldAuth(shadow.Username, shadow.Password)
}
```

Configs carrying a username and password are unaffected, including the existing `auth field replaced` case where a stale `Auth` is deliberately overwritten.

### Testing

Two cases added to the existing `TestAuthConfigMarshalJSON` table: an auth-only config, and a token-only config. Without the change both fail:

```
--- FAIL: TestAuthConfigMarshalJSON/auth_field_preserved_when_username_and_password_are_empty
--- FAIL: TestAuthConfigMarshalJSON/no_auth_field_is_invented_for_a_token_only_config
```

With it, `TestAuthConfigMarshalJSON` and `TestAuthConfigUnmarshalJSON` pass in full. `TestPodmanConfig` fails identically before and after on my machine because it picks up a `REGISTRY_AUTH_FILE` from the environment; it is unrelated to this change.

### DIFF
--- a/pkg/authn/authn.go
+++ b/pkg/authn/authn.go
@@ -89,10 +89,6 @@ func (a *AuthConfig) UnmarshalJSON(data []byte) error {
 // MarshalJSON implements json.Marshaler
 func (a AuthConfig) MarshalJSON() ([]byte, error) {
 	shadow := (authConfig)(a)
-	// Only derive the auth field from the username and password when at least
-	// one of them is set. Otherwise an AuthConfig carrying only an already
-	// encoded Auth, or only a token, would have its auth field overwritten with
-	// the encoding of an empty username and password (":", i.e. "Og==").
 	if shadow.Username != "" || shadow.Password != "" {
 		shadow.Auth = encodeDockerConfigFieldAuth(shadow.Username, shadow.Password)
 	}
--- a/pkg/authn/authn.go
+++ b/pkg/authn/authn.go
@@ -89,7 +89,13 @@ func (a *AuthConfig) UnmarshalJSON(data []byte) error {
 // MarshalJSON implements json.Marshaler
 func (a AuthConfig) MarshalJSON() ([]byte, error) {
 	shadow := (authConfig)(a)
-	shadow.Auth = encodeDockerConfigFieldAuth(shadow.Username, shadow.Password)
+	// Only derive the auth field from the username and password when at least
+	// one of them is set. Otherwise an AuthConfig carrying only an already
+	// encoded Auth, or only a token, would have its auth field overwritten with
+	// the encoding of an empty username and password (":", i.e. "Og==").
+	if shadow.Username != "" || shadow.Password != "" {
+		shadow.Auth = encodeDockerConfigFieldAuth(shadow.Username, shadow.Password)
+	}
 	return json.Marshal(shadow)
 }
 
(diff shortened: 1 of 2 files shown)


Subserial (commented):
Looks good, but can you remove the comments? They are redundant after the fix, and the comments on the test are useless when the test case already has a name.

nileshpatil6:
Done, comments removed from both the fix and the test cases. Tests still pass.

Subserial (approved):

nileshpatil6:
The failing `e2e windows-latest` job looks like a runner problem rather than something in this change.

The job pushed and tagged the image fine, then `docker run` aborted with:

```
failed to connect to the docker API at npipe:////./pipe/docker_engine; check if the path is correct and if the daemon is running: open //./pipe/docker_engine: The system cannot find the file specified.
```

so the Docker daemon was not up on that runner. The same job passed on other branches the same day (run 33007300693), and this change only affects JSON marshalling in `pkg/authn`, which the e2e script does not exercise.

I do not have permission to re-run the job. Could you kick it off when you get a chance?

PR #2426, opened 2026-08-26
Reject non-canonical IP literals in all SSRF guards

### LINKED ISSUE #2284: transport: validateRealmURL SSRF bypass via 0.0.0.0 (unspecified address)
`validateRealmURL()` in `pkg/v1/remote/transport/bearer.go` checks `IsLoopback`, `IsPrivate`, `IsLinkLocalUnicast`, and `IsLinkLocalMulticast` but not `IsUnspecified`.

`0.0.0.0` and `::` pass all four checks, but on Linux and macOS they resolve to `127.0.0.1`. A malicious container registry can redirect auth requests to the victim's localhost by returning:

```
WWW-Authenticate: Bearer realm="https://0.0.0.0:9999/token",service="registry"
```

The client then sends Basic auth credentials to `localhost:9999` — credential theft.

Confirmed with a PoC:
- `net.ParseIP("0.0.0.0").IsLoopback()` = false
- `net.ParseIP("0.0.0.0").IsPrivate()` = false
- `net.ParseIP("0.0.0.0").IsUnspecified()` = true (not checked)
- `net.Dial("tcp", "0.0.0.0:PORT")` connects to 127.0.0.1

I'll submit a PR adding `ip.IsUnspecified()` to the existing check.

### LINKED ISSUE #2425: SSRF guards bypassed by non-canonical IP literals the dialer accepts but net.ParseIP rejects (all four guard sites)
## Summary

All four SSRF guards in the registry client decide "is this an IP literal" with `net.ParseIP(host)`, which accepts **only** canonical dotted-quad IPv4 and canonical IPv6. Go's dialer (hence `net/http`) accepts many more forms, so for those hosts the guards' private/loopback/link-local rejection never executes while the request is still delivered to the private address:

| Form | Example | Resolves to |
|---|---|---|
| 32-bit decimal | `2130706433` | `127.0.0.1` |
| decimal (metadata) | `2852039166` | `169.254.169.254` |
| hex integer | `0x7f000001` | `127.0.0.1` |
| partial dotted-quad | `127.1` | `127.0.0.1` |
| zero-padded octets | `127.000.000.001` | `127.0.0.1` |
| zone-qualified IPv6 | `[fe80::1%25eth0]` | link-local |

## Affected guards

- `validateRealmURL` — `pkg/v1/remote/transport/bearer.go:154` (WWW-Authenticate realm; the comment states it blocks "loopback, and link-local ranges including the cloud instance metadata service")
- `checkRedirectSSRF` — `pkg/v1/remote/fetcher.go:101` (3xx redirect targets during manifest/blob GET)
- `validateForeignURL` — `pkg/v1/remote/fetcher.go:383` (foreign-layer `urls[]` from a manifest)
- `nextLocation` — `pkg/v1/remote/write.go:196` (blob-upload `Location` header)

## Reproduction (real library, local fakes)

Evil registry answers `/v2/` with `WWW-Authenticate: Bearer realm="http://2130706433:PORT/token"`. The library fetches the token from that realm **sending the victim's Basic credentials** — observed `Authorization: Basic dmljdGltOlN1cGVyU2VzZWNyZXQxMjM` arriving at the collector. Control: the identical dotted-quad realm is correctly rejected (`realm host "127.0.0.1" is a private or link-local address`). Separately observed: realm host `2852039166` passes the guard and the library dials `169.254.169.254`.

Reproduction test module in the fix PR description.

## Why this is a bug, not intended

- The guards' comments claim exactly this coverage; these are IP literals, not the documented-out-of-scope DNS case.
- The same shape was fixed before: #2284 ("validateRealmURL SSRF bypass via 0.0.0.0") added `ip.IsUnspecified()` for another value `net.ParseIP` mishandles but the dialer connects to.

## Suggested fix

Normalize the host before the check: strip any IPv6 zone, then parse with `netip.ParseAddr`; additionally handle the legacy inet_aton forms (or resolve and re-check all resulting IPs). Apply uniformly in all four guards.

Reported via Google Bug Hunters (acknowledged; public disclosure invited).

shaggyinsomniac:
Fixes #2425.

## What

Adds `internal/ipaddr` — a parser for IP literals in **every form the Go dialer accepts**: canonical IPv4/IPv6, zone-qualified IPv6, IPv4-mapped IPv6, and the legacy inet_aton encodings (32-bit decimal `2130706433`, hexadecimal `0x7f000001`, partial dotted-quad `127.1`, zero-padded octets). Routes all four SSRF guards through it:

- `validateRealmURL` (`pkg/v1/remote/transport/bearer.go`)
- `checkRedirectSSRF` and `validateForeignURL` (`pkg/v1/remote/fetcher.go`)
- `nextLocation` (`pkg/v1/remote/write.go`)

## Why

`net.ParseIP` accepts only canonical forms, so the existing guards never fire for hosts the dialer happily connects to — e.g. a registry returning `WWW-Authenticate: Bearer realm="http://2130706433:PORT/token"` receives the victim's Basic credentials, and a realm of `2852039166` reaches the cloud metadata service. Same shape as the fixed #2284 (`0.0.0.0`). Details + reproduction in #2425.

## Verification

- `internal/ipaddr` unit tests cover every encoding (decimal/hex/octal/partial/padded/zone/mapped) plus negative cases.
- End-to-end before/after with the reproduction from #2425 (real library, local fakes): at HEAD, the victim's `Authorization` header is delivered to the attacker's realm endpoint via the integer form; with this change the same flow is rejected (`realm host ... is a private or link-local address`) and the canonical-form behavior is unchanged.
- `go test ./pkg/v1/remote/transport/ ./internal/ipaddr/` pass; `./pkg/v1/remote/` has two pre-existing local-environment failures (`docker-credential-desktop` not in PATH) identical with and without this change.

### DIFF
--- a/internal/ipaddr/ipaddr.go
+++ b/internal/ipaddr/ipaddr.go
@@ -1,0 +1,85 @@
+// Copyright 2026 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Package ipaddr parses IP literals in every form accepted by the Go dialer,
+// including non-canonical legacy inet_aton encodings that net.ParseIP and
+// netip.ParseAddr reject.
+package ipaddr
+
+import (
+	"net/netip"
+	"strconv"
+	"strings"
+)
+
+// IsPrivateOrLinkLocal reports whether host denotes a loopback, private,
+// link-local, or unspecified address. It accepts any IP-literal form the Go
+// dialer accepts — canonical dotted-quad IPv4 and IPv6, zone-qualified and
+// IPv4-mapped IPv6, and legacy inet_aton encodings (32-bit decimal
+// "2130706433", hexadecimal "0x7f000001", partial dotted-quad "127.1",
+// zero-padded octets) — so a guard based on it cannot be bypassed by
+// spelling an internal address in a non-canonical way. DNS names are not IP
+// literals and return false.
+func IsPrivateOrLinkLocal(host string) bool {
+	addr, ok := Parse(host)
+	if !ok {
+		return false
+	}
+	return addr.IsLoopback() || addr.IsLinkLocalUnicast() || addr.IsLinkLocalMulticast() ||
+		addr.IsPrivate() || addr.IsUnspecified()
+}
+
+// Parse parses an IP literal in any form the Go dialer accepts.
+func Parse(host string) (netip.Addr, bool) {
+	// The dialer ignores IPv6 zone identifiers, so guards must too.
+	host, _, _ = strings.Cut(host, "%")
+	if addr, err := netip.ParseAddr(host); err == nil {
+		// Treat IPv4-mapped IPv6 as the IPv4 address the dialer connects to.
+		return addr.WithZone("").Unmap(), true
+	}
+	return parseLegacyIPv4(host)
+}
+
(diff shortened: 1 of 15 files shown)


Subserial (approved):

Subserial:
Looks like it's just spacing in the boilerplate message causing test issues.

PR #2429, opened 2026-08-31
fix(mutate): make Time layer updates lazy

### LINKED ISSUE #1168: mutate.Time should be lazy
[`pkg/v1/mutate.Time`](https://pkg.go.dev/github.com/google/go-containerregistry/pkg/v1/mutate#Time) sets all timestamps in an image to the given time, typically to make the image more reproducible. For timestamps in layers, it calls [`layerTime`](https://github.com/google/go-containerregistry/blob/v0.6.0/pkg/v1/mutate/mutate.go#L376) which reads the layer contents and writes them to a buffered `tar.Writer`. This means that if the layer is huge, we may run out of memory buffering its contents.

We should consider making this a lazy transformation, so that layer mutations aren't made until they're read. We did a similar thing in [`pkg/v1/cache`](https://github.com/google/go-containerregistry/blob/v0.6.0/pkg/v1/cache/cache.go#L70), where layer contents aren't cached until they're read via `Compressed` or `Uncompressed`.

amarkdotdev:
## Summary
- Defer layer timestamp rewrites in mutate.Time until a layer is actually read
- Avoid upfront work for every layer when only some layers are consumed

Fixes #1168

## Test plan
- [x] go test ./pkg/v1/mutate/... -run TestTime


Subserial:
Can you recreate this PR on a more recent commit?

amarkdotdev:
rebased onto latest main

amarkdotdev:
Pushed a583be5: `layerTime` now returns only `v1.Layer` so golangci-lint `unparam` passes on the Lint check.

Subserial (approved):

PR #2435, opened 2026-09-01
fix(release): honor declared Go toolchain

### LINKED ISSUE #2430: crane: v0.22.0 release binaries built with go1.25.0 instead of the declared toolchain go1.26.6
### Describe the bug

The v0.22.0 release binaries are built with **go1.25.0**, while `go.mod` declares
`toolchain go1.26.6`. The previous release, v0.21.9, was built with go1.26.5.

Since `go 1.25.0` in `go.mod` is documented right above the directive as the
*minimum* supported version rather than the build version, the release binaries
now ship a Go standard library that is several releases behind what the module
declares.

### To Reproduce

```
$ curl -fsSL https://github.com/google/go-containerregistry/releases/download/v0.21.9/go-containerregistry_Linux_arm64.tar.gz | tar -xz crane
$ go version -m crane | head -1
crane: go1.26.5

$ curl -fsSL https://github.com/google/go-containerregistry/releases/download/v0.22.0/go-containerregistry_Linux_arm64.tar.gz | tar -xz crane
$ go version -m crane | head -1
crane: go1.25.0
```

Same result for `go-containerregistry_Linux_x86_64.tar.gz`.

### Expected behavior

Release binaries built with the toolchain the module declares (`go1.26.6` at
v0.22.0), or with an explicitly configured build version — as was the case up to
v0.21.9.

### Additional context

The change appears to come from #2416, which removed `.go-version` (contents:
`1.26.5`) and repointed the workflows at `go.mod`. In `.github/workflows/release.yml`:

```yaml
# v0.21.9
go-version-file: '.go-version'   # -> 1.26.5

# v0.22.0 and main
go-version-file: 'go.mod'        # -> resolves the `go` directive, 1.25.0
```

The `toolchain go1.26.6` directive does not compensate for this, because the same
workflow sets `GOTOOLCHAIN: local` (line 8), which disables automatic toolchain
switching. So the build pins itself to whatever `setup-go` installed, i.e. the
minimum version from the `go` directive.

`main` is currently in the same state (`go 1.25.0` / `toolchain go1.26.6` /
`go-version-file: 'go.mod'` / `GOTOOLCHAIN: local`), so the next release would be
affected the same way.

**Impact.** Scanning both binaries with Trivy 0.74.0 (`--severity HIGH,CRITICAL
--ignore-unfixed`) reports 8 findings against v0.21.9's stdlib and 22 against
v0.22.0's. The exact counts depend on the scanner's database, but the direction
is what matters: consumers who scan release artifacts in CI see a regression when
upgrading, which is what led us to look into this.

**Possible fixes** (in the maintainers' judgement, roughly in order of least
surprise):

1. Restore an explicit build-version file for CI (`.go-version`) and keep
   `go.mod`'s `go` directive as the minimum-version declaration it is documented
   to be.
2. Drop `GOTOOLCHAIN: local` from `release.yml` so the `toolchain` directive is
   honoured.
3. Keep `go-version-file: 'go.mod'` but pin the build version some other way,
   e.g. `go-version: '1.26.6'` in the release workflow.

- Output of `crane version`: `0.22.0`
- Registry used: n/a (build/release issue, not registry-specific)

rksharma-owg:
## Summary

- let the release workflow activate the exact Go toolchain declared in `go.mod`
- stop before GoReleaser if the active Go version does not match that declaration
- keep `go 1.25.0` as the module's minimum supported version

## Problem

The v0.22.0 release binaries were built with Go 1.25.0 even though the module declares `toolchain go1.26.6`. This leaves release artifacts on an older standard library than intended and makes the repository's documented release-toolchain setting ineffective.

## Root cause

#2416 replaced the dedicated `.go-version` file with the `toolchain` directive, but the release workflow retained `GOTOOLCHAIN: local`. `actions/setup-go` selects the `go` directive from `go.mod` (the minimum version), while `GOTOOLCHAIN: local` prevents the Go command from activating the newer declared toolchain.

## Fix

Set `GOTOOLCHAIN` to `auto` for the release job so the Go command honors the pinned toolchain directive. Before GoReleaser runs, compare `go env GOVERSION` with the toolchain parsed by `go mod edit -json` and fail with an actionable mismatch if they differ.

## Security considerations

This affects the trusted tag-release workflow and the provenance-bearing binaries it publishes. The change does not alter workflow permissions, token handling, or application runtime behavior. Automatic selection is still pinned to the exact toolchain in `go.mod`, and the new check prevents a mismatched builder from publishing. Existing v0.22.0 artifacts are not changed; the fix applies to future releases.

## Validation

- Published reproduction: `go version -m crane` from the v0.22.0 Linux x86_64 archive reports `go1.25.0`.
- Regression check before the fix: `expected=go1.26.6 actual=go1.26.5` with the workflow's `GOTOOLCHAIN=local` behavior.
- Regression check after the fix: `workflow_GOTOOLCHAIN=auto expected=go1.26.6 actual=go1.26.6`.
- Release-equivalent `CGO_ENABLED=0 go build -trimpath ./cmd/crane/main.go`; `go version -m` reports `go1.26.6`.
- `PATH="$(go env GOPATH)/bin:$PATH" ./hack/presubmit.sh` — passed.
- `go test -coverprofile=/tmp/gcr-coverage-2430.txt -covermode=atomic -race ./...` — passed.
- `go build ./... && go test -run=^$ ./...` — passed.
- `actionlint -ignore SC2086 .github/workflows/release.yml` — passed. (`SC2086` findings predate this change elsewhere in the workflow.)
- `git diff --check` — passed.

## Compatibility

No library API, CLI behavior, module minimum Go version, artifact format, or supported platform changes. Only the Go version used to build future release binaries changes to the already-declared release toolchain.

Fixes #2430

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -5,7 +5,7 @@ on:
     tags: ['*']
 
 env:
-  GOTOOLCHAIN: local
+  GOTOOLCHAIN: auto
 
 jobs:
   goreleaser:
@@ -25,6 +25,14 @@ jobs:
           go-version-file: 'go.mod'
           check-latest: true # no-op if full version is specified
           cache: false
+      - name: Verify release Go version
+        run: |
+          expected="$(go mod edit -json | jq -r .Toolchain)"
+          actual="$(go env GOVERSION)"
+          if [[ "${actual}" != "${expected}" ]]; then
+            echo "Release Go version mismatch: got ${actual}, want ${expected}" >&2
+            exit 1
+          fi
       - uses: goreleaser/goreleaser-action@f06c13b6b1a9625abc9e6e439d9c05a8f2190e94 # v7.2.3
         id: run-goreleaser
         with:


Subserial:
Ah, my bad. I didn't realize we needed auto set. Once you accept the CLA I'll be able to merge the PR.

Subserial (approved):

PR #2445, opened 2026-09-09
fix(build): escape PATH variable in cloudbuild_v2.yaml

tprussak:


cloudbuild_v2.yaml
```diff
@@ -67,7 +67,7 @@ steps:
     ARG TARGETARCH
     ARG TARGETVARIANT
     WORKDIR /
-    ENV PATH="/ko-app:\${PATH}"
+    ENV PATH="/ko-app:\$${PATH}"
```
Subserial, line 70: I believe you should remove the back-slash instead of adding a new dollar sign if your intent is to extend the PATH variable.

tprussak: This needs to be resolved by the Docker builder, but Cloud Build would consume it before it gets there if it was unescaped. Line 71 uses this same pattern in several places.

Subserial (approved):

Issue #1768, opened 2023-08-10
ggcr: name functionally discards tags when a reference has both tag and digest


mxmeinhold:
### Describe the bug

When `name.ParseReference` is passed a reference that has both tag and digest, such as `registry.example.com/repo:tag@sha256:b5bb9d8014a0f9b1d61e21e796d78dccdf1352f23cd32812f4850b878ae4944c`, it parses as a Digest.

This becomes noticeable when using `remote.Write`, which does not write the tag to the registry, as it uses the references `Identifier()` 
https://github.com/google/go-containerregistry/blob/a54d64203cffcbf94146e04069aae4a97f228ee2/pkg/v1/remote/write.go#L577

This behavior is well and good for pulling an image, where the digest is the most specific reference, but since `name.ParseReference` appears to discard the tag. I believe this occurs because `NewTag` splits on `:`, which would result in the sha256sum in the digest being read as the tag.
https://github.com/google/go-containerregistry/blob/a54d64203cffcbf94146e04069aae4a97f228ee2/pkg/name/tag.go#L77-L83

The `base` of our example reference is then considered as `registry.example.com/repo:tag`, which is parsed to a repo of `repo:tag`, which of course is not a valid repository name, so this reference can't be parsed with a tag
https://github.com/google/go-containerregistry/blob/a54d64203cffcbf94146e04069aae4a97f228ee2/pkg/name/repository.go#L78-L91

It looks like `NewDigest` actually does parse out the tag, but only uses this to get the repository https://github.com/google/go-containerregistry/blob/a54d64203cffcbf94146e04069aae4a97f228ee2/pkg/name/digest.go#L80-L83

This parsing is masked by `Digest.String()`, as it returns the unparsed reference https://github.com/google/go-containerregistry/blob/a54d64203cffcbf94146e04069aae4a97f228ee2/pkg/name/digest.go#L57-L60

### To Reproduce

```go
package refs

import (
	"testing"

	"github.com/google/go-containerregistry/pkg/name"
)

func TestGoContainerRegistryRef(t *testing.T) {

	t.Run("tag only", func(t *testing.T) {
		ref, err := name.ParseReference("registry.example.com/repo:tag")
		if err != nil {
			t.Error(err)
		}

		if ref.Identifier() != "tag" {
			t.Error("ref id:", ref.Identifier(), "expected:", "tag")
		}
	})

	t.Run("digest only", func(t *testing.T) {
[...]

PR #2451, opened 2026-09-15
Retain explicit tag when parsing digests

### LINKED ISSUE #2069: ggcr: Parsing image name with both tag and digest loses tag info
### Describe the bug

When parsing image in `repo.com/image:tag@sha256:f8ecdcd87d7d84b87e645074084dd7f57dd62c76e120bb21e5abde158755be56`format it gets parsed as a Digest type, which returns digest from the `ref.Identifier()` function, and doesn't provide easy access to the tag information

### To Reproduce
```go
ref, _ := name.ParseReference(repo.com/image:tag@sha256:f8ecdcd87d7d84b87e645074084dd7f57dd62c76e120bb21e5abde158755be56")
fmt.Println(ref.Identifier()) // sha256:f8ecdcd87d7d84b87e645074084dd7f57dd62c76e120bb21e5abde158755be56
// fmt.Println(ref.Tag()) - doesn't exist
```

### Expected behavior

I'd expect to have a `ref.Tag()` function in the `Reference` interface that can return image tag for both `Tag` and `Digest` structs.

### Additional context

- go-containerregistry v0.20.3

Tyagiquamar:
Fixes #2069 (see also #1768).

`NewDigest` parsed the `tag@digest` base through `NewTag` but discarded the tag, so `ParseReference("repo/image:tag@digest")` gave no way to recover `tag`. This stores an explicitly-present tag on `Digest` and exposes it via `TagStr()` (empty when the input carried no tag, so a bare `repo@digest` does not report a phantom `latest`). Hostname ports are not mistaken for tags, using the same disambiguation as `NewTag`.

Test: new `TestDigestTagStr` covers tag present / absent / `latest` / port cases plus the issue repro through `ParseReference`. Full `pkg/name` suite green, `go vet` clean.


Tyagiquamar:
Hi, just following up on this when you get a chance. The branch is up to date and checks are green. If it looks good from your side, it should be ready to merge. Happy to make any changes if needed. Thanks!

PR #2458, opened 2026-09-19
fix(crane): require both username and password for auth login

### LINKED ISSUE #4120: The cosign login command stores incorrect creds under /root/.docker/config.json
**Description**

 The cosign login command does not validate the credentials against the registry server and directly puts the credentials into the `/root/.docker/config.json`
~~~
$ cosign login quay.io -u username
logged in via /root/.docker/config.json
[root@vm-236-142 ~]# cat /root/.docker/config.json
{
	"auths": {
		"quay.io": {
			"auth": "dXNlcm5hbWUK"
		}
	}
}
~~~

The correct behaviour would be to ask for the password via prompt and once the creds are validated, the /root/.docker/config.json file should be updated. This leads to problems when signing the images.

**Version**

~~~
GitVersion: v2.4.3
GitCommit: 6a7abbf3ae7eb6949883a80c8f6007cc065d2dfb
GitTreeState: clean
BuildDate: 2025-02-19T19:34:52Z
GoVersion: go1.23.6
Compiler: gc
Platform: linux/amd64
~~~

ogulcanaydogan:
Fixes sigstore/cosign#4120 (cosign's \`login\` command is this same \`crane auth login\` code, imported directly via \`cranecmd.NewCmdAuthLogin\`).

## What this fixes

\`crane auth login\` (and by extension \`cosign login\`) only rejected the request when **both** \`--username\`/\`-u\` and \`--password\`/\`-p\` were empty:

```go
if opts.user == "" && opts.password == "" {
	return errors.New("username and password required")
}
```

So running e.g. \`crane login reg.example.com -u AzureDiamond\` (forgetting \`-p\`) silently succeeded and wrote a docker config entry with an empty password - no error, no prompt, nothing to indicate the stored credential is broken. The same happens with \`--password-stdin\` piped from an empty source.

Changed the check to \`||\` so either missing field is rejected up front, with a clearer error message pointing at \`--password-stdin\`.

## Testing

- Added \`cmd/crane/cmd/auth_test.go\` covering: username-only, password-only, and neither (all now error and store nothing), the happy path (both provided, credentials actually persisted and readable back via \`docker/cli/cli/config\`), and an empty \`--password-stdin\` source (now also errors).
- Verified the new tests fail against the pre-fix code and pass after the fix.
- \`./hack/presubmit.sh\`: all tests pass except two pre-existing, unrelated failures I confirmed are present without this change too - \`pkg/name/internal\`'s build-tag test harness (\`must_test.sh\`/\`must_test.go\`, unrelated package) and \`cmd/crane/rebase_test.sh\` (needs a local \`registry\` binary on \`PATH\`, an environment setup issue in my sandbox, not a code issue).


ogulcanaydogan:
I signed it!

